PR #2, opened 2016-01-29
Fix missing sender identity on blockwised message

sbernard31:
When request is sending by block, the sender identity is missing.
This PR fixes the problem.
I don't know the code enough, but maybe there is other missing identity setting.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -207,6 +207,7 @@ public class BlockwiseLayer extends AbstractLayer {
 					
 					// Assemble and deliver
 					Request assembled = new Request(request.getCode());
+					assembled.setSenderIdentity(request.getSenderIdentity());
 					assembleMessage(status, assembled);
 					
 					exchange.setRequest(assembled);


sophokles73:
Good catch, Simon :-)

However, I do not like the fact that we do this in multiple places. This suggests to me that we are doing it in the wrong place(s). There has to be a place where we can be sure all kinds of requests flow through and that should be the place where we set the sender identity, don't you agree?

jvermillard:
BTW we had this issue with leshan when the initial registration content is large and use blockwise transfert

sbernard31:
I set the `SenderIdentity` just after the call of `Request` constructor, so I'm not sure there is a way to factorize that.
Maybe @mkovatsc can give us a recommendation of this.

mkovatsc:
Where do you usually set this?
The problem is that the BlockwiseLayer is indeed a second place where a new Request or Response is created and configured to somewhat inject an assembeled Message into the stack.

One option would be to to always extend the initial message with the additional payload. Here, we have to make sure that no timeouts or MessageObservers are triggered falsely.

In general, the BlockwiseLayer also needs to make sure the SenderIdentity is the same for all blocks before forwarding the assembled message.

sbernard31:
> Where do you usually set this?

In [CoapEndpoint.InboxImpl.receiveMessage(RawData)](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java#L589).

> The problem is that the BlockwiseLayer is indeed a second place where a new Request or Response is created and configured to somewhat inject an assembeled Message into the stack.

So maybe, the setSenderIdentity duplication make senses here.

> In general, the BlockwiseLayer also needs to make sure the SenderIdentity is the same for all blocks before forwarding the assembled message.

For now the senderIdentity is only available on `Request`, but I believe this make sense for all `Message`
In this case we could use `BlockwiseStatus.status.first.getSenderIdentity()` to make this kind of check ?

fdur:
Just for info, this issue was seen on REGISTER and REg UPDATE message in block-wise mode

sophokles73:
> In general, the BlockwiseLayer also needs to make sure the SenderIdentity is the same for all blocks before forwarding the assembled message.

I suppose this check is currently missing altogether. If so, we should open a new issue for this.

The way I understand the design of the `CoapStack` we cannot be sure that the `Request` popping out of the top of the stack is the same instance as the one originally received from the network layer. Thus, we will probably have to live with the fact that we need to set the client identity in multiple places.

+1 for merging this PR.

Issue #5, opened 2016-02-02
issue with url-encoded path segments on server-side


h0ru5:
If you are unlucky enough to find yourself in the situation that you get non-ascii chars or even blanks in the path, its crucial to uri-decode the path segments. Quick fix:

replace line:
https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/server/ServerMessageDeliverer.java#L142

by something like:

``` java
String name = path.removeFirst();
String sanitizedName = java.net.URLDecoder.decode(name, "UTF-8");
current = current.getChild(sanitizedName);
```

mkovatsc:
I remember this issue at a different point, where it was actually the URI class to cause the problem. Not sure if this is the right place, since it is simply a .get() on a ConcurrentHashMap, which does not know about URI encoding.

I will look into this and try to fix it at the root...

For 2.x, we will need to have something better than Java's own URI class anyway to support alternative transports.

h0ru5:
problem is here for example: you have a uri path token like ``stränge name``, which will be plaintext in the hashmap, but uri-encoded to ``str%C3%A4nge+name``  or  ``str%c3%a4nge+name`` in the request (or even ``str%C3%A4nge%20name`` when e.g. sent from copper). 

Therefore they will not match. I tried a workaround by injecting encoded keys, but there are just too many combinations. (... so I resorted to using english names :grin:)

mkovatsc:
Yes, the problem is Java's URI class in Californium. We should do clean UTF-8 handling.

Copper has to clean what it gets from Firefox. I will have a look there in parallel when fixing Cf.

Luckily it is hard to inject stränge stringß in Erbium ;)
(which would be handled perfectly as UTF-8 bytes---I hope)

h0ru5:
Actually in the pointed code its just plain Strings, not the Java URI class.
The lines should fix the problem - just had not launched a PR yet

mkovatsc:
What I mean is that at this point the string should already be exactly what it is supposed to mean. It must be identical to the UTF-8-encoded bytes in the Uri-Option. If the Uri-Option already carried an encoded string, the client has to be fixed. For Cf, we have to make sure, we do not convert any URI option into an encoded string. The later, however, might happen when using the URI class to handle the full URI (e.g., sanity checks, normalizing, etc.).

I know of a similar problem when creating resources with a whitespace in the name.

sophokles73:
@mkovatsc,

you  mentioned that

[...]

PR #6, opened 2016-02-02
url-decode incoming path segments, addresses #5

h0ru5:


h0ru5:
it's ok for me to submit this patch under the mentioned license.

mkovatsc:
Please see the comment in #5.

Also, in order for us to be able to merge the pull request, you will need to sign an [Eclipse Contributor License Agreement](https://projects.eclipse.org/user/sign/cla) and to include a signed-off-by footer in your commit message(s) using the same e-mail address for which you signed the CLA (`git commit -s ...` with corresponding author e-mail address).

sophokles73:
AFAIC there is nothing to fix here. If we receive a request with options containing escaped characters then the client has either intentionally put these *escape characters* (e.g. '%E5`) into the value and thus wants them to be part of the value or has violated the decomposition rules laid out by [RFC 7252, Section 6.4](https://tools.ietf.org/html/rfc7252#section-6.4).

Issue #1, opened 2016-01-22
Request/Response matching does not comply with spec when using DTLS


sophokles73:
I would like to follow up on the discussion we started on the [mailing list](https://dev.eclipse.org/mhonarc/lists/cf-dev/msg00574.html) around this issue.

I am creating the issue in the top level repo because we are merging the *californium.core*, *californium.element-connector* and *californium.scandium* repos into this repo soon.

jvermillard:
For reference the ietf-core thread about CoAP RFC limitation on cross epoch
notifications
https://mailarchive.ietf.org/arch/search/?email_list=core&gbt=1&index=oJD7psrLwJj662gjbRku1hGp44c

--
Julien Vermillard

sophokles73:
Thanks for the link, @jvermillard :-)
My understanding is that you have proposed the same thing that we have also discussed on the Californium list, i.e. alternatively matching on session ID and cipher strength, right?
Did you get any further replies on the core list since you originally posted the issue there?

sophokles73:
fixed in 9a0235a0d78b41a1426361e007f8b30e0be819fd

PR #7, opened 2016-02-03
Use SessionListener to trigger sending of pending APPLICATION messages.

sophokles73:
Sending an APPLICATION message to a peer for which no DTLS session
exists yet triggers a new handshake with the peer. Previously it was the
responsibility of the `Handshaker` class to send the (pending) APPLICATION layer
message to the peer once the handshake has been completed.

Handshakers signal the establishment of a session by means of the
`SessionLifecycle.sessionEstablished()` callback method. In order to
improve separation of concerns pending APPLICATION layer messages are
now sent by the `DTLSConnector` on invocation of this callback method.

This PR is the first step towards fixing #1 in that it hands back control over the message sending to `DTLSConnector`. The approach to fixing #1 then is to include a callback handler in the `RawData` instance passed in to `DTLSConnector.send(RawData)` which will be called back with the *ID*, *epoch* and *cipher* of the `DTLSSession` used for sending the message to the peer.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -29,6 +29,8 @@
  *                                                    from network MTU
  *    Kai Hudalla (Bosch Software Innovations GmbH) - fix bug 483371
  *    Benjamin Cabe - fix typos in logger
+ *    Kai Hudalla (Bosch Software Innovations GmbH) - use SessionListener to trigger sending of pending
+ *                                                    APPLICATION messages
  ******************************************************************************/
 package org.eclipse.californium.scandium;
 
@@ -85,10 +87,12 @@ import org.eclipse.californium.scandium.dtls.InMemoryConnectionStore;
 import org.eclipse.californium.scandium.dtls.MaxFragmentLengthExtension;
 import org.eclipse.californium.scandium.dtls.ProtocolVersion;
 import org.eclipse.californium.scandium.dtls.Record;
+import org.eclipse.californium.scandium.dtls.RecordLayer;
 import org.eclipse.californium.scandium.dtls.ResumingClientHandshaker;
 import org.eclipse.californium.scandium.dtls.ResumingServerHandshaker;
 import org.eclipse.californium.scandium.dtls.ServerHandshaker;
 import org.eclipse.californium.scandium.dtls.SessionAdapter;
+import org.eclipse.californium.scandium.dtls.SessionListener;
 import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
 import org.eclipse.californium.scandium.util.ByteArrayUtils;
 
@@ -777,8 +781,7 @@ public class DTLSConnector implements Connector {
 	 */
 	private void processOngoingHandshakeMessage(final HandshakeMessage message, final Record record, final Connection connection) throws HandshakeException {
 		if (connection.hasOngoingHandshake()) {
-			DTLSFlight flight = connection.getOngoingHandshake().processMessage(record);
-			sendHandshakeFlight(flight, connection);
+			connection.getOngoingHandshake().processMessage(record);
 		} else {
 			LOGGER.log(
 				Level.FINE,
@@ -806,8 +809,9 @@ public class DTLSConnector implements Connector {
 			if (session == null) {
 				session = new DTLSSession(helloRequest.getPeer(), true);
 			}
-			Handshaker handshaker = new ClientHandshaker(null, session, connection, config, maximumTransmissionUnit);
-			sendHandshakeFlight(handshaker.getStartHandshakeMessage(), connection);
+			Handshaker handshaker = new ClientHandshaker(session, getRecordLayerForPeer(connection), connection,
+					config, maximumTransmissionUnit);
+			handshaker.startHandshake();
 		}
 	}
 
@@ -930,9 +934,9 @@ public class DTLSConnector implements Connector {
 		DTLSSession newSession = new DTLSSession(record.getPeerAddress(), false, record.getSequenceNumber());
 		// initialize handshaker based on CLIENT_HELLO (this accounts
 		// for the case that multiple cookie exchanges have taken place)
-		Handshaker handshaker = new ServerHandshaker(clientHello.getMessageSeq(),
-				newSession, peerConnection, config, maximumTransmissionUnit);
-		sendHandshakeFlight(handshaker.processMessage(record), peerConnection);
+		Handshaker handshaker = new ServerHandshaker(clientHello.getMessageSeq(), newSession,
+				getRecordLayerForPeer(peerConnection), peerConnection, config, maximumTransmissionUnit);
+		handshaker.processMessage(record);
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -338,8 +313,7 @@ protected DTLSFlight receivedHelloVerifyRequest(HelloVerifyRequest message) thro
 
 		DTLSFlight flight = new DTLSFlight(getSession());
 		flight.addMessage(wrapMessage(clientHello));
-
-		return flight;
+		recordLayer.sendFlight(flight);
 	}
```
sbernard31: We should exit the function after `recordLayer.sendFlight(flight);`

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -361,7 +363,7 @@ public final DTLSFlight processMessage(Record record) throws HandshakeException
 			try {
 				record.setSession(session);
 				DTLSMessage messageToProcess = inboundMessageBuffer.getNextMessage(record);
-				while (nextFlight == null && messageToProcess != null) {
+				while (messageToProcess != null) {
```
sbernard31: I see a possible different behavior.
In the case where we find a complete flight and there is still some messageToProcess, we will continue to loop even if we already send the flight. That was not the case before.
I suppose this is not really an issue ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
```diff
@@ -100,11 +104,9 @@ protected synchronized DTLSFlight doProcessMessage(DTLSMessage message) throws H
 				Level.FINER,
 				"Received server's [{0}] FINISHED message again, retransmitting last flight...",
 				message.getPeer());
-			return lastFlight;
+			recordLayer.sendFlight(lastFlight);
```
sbernard31: We should exit the function ?

sophokles73: Excellent!

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sophokles73: I am not sure if I understand correctly. You seem to suspect that there is the possibility of looping endlessly. Can you give an example of circumstances under which this might happen? I thought thought that we will always be able to break out of the loop simply because of the call to `doProcessMessage(Record)` which always *consumes* the message, thus removing it from the queue.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
sophokles73: We already do, don't we?

sbernard31: Yes, we do ... :astonished: : Sry, I meaned [here](https://github.com/eclipse/californium/pull/7/files#diff-5da9149ee37db8bcc492b03743f00dfeR224) in ServerHandshaker. 

(I review in eclipse and comment in github, this is confusing :stuck_out_tongue: )

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sbernard31: Forget this one, finally I think it's ok.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1289,13 +1315,21 @@ private void handleTimeout(DTLSFlight flight) {
 			LOGGER.log(Level.FINE, "Re-transmitting flight for [{0}], [{1}] retransmissions left",
 					new Object[]{flight.getPeerAddress(), max - flight.getTries() - 1});
 
-			flight.incrementTries();
-			sendFlight(flight);
+			try {
+				flight.incrementTries();
+				flight.setNewSequenceNumbers();
```
sbernard31: I 'm not sure we must increment sequence numbers on retransmission ?
Does we do that before ?

sbernard31: I take a deeper look and it seems we always did that before..
Now we do that only on retransmission.
And when I read the spec, I understand we must do that only the first time (as sequence number is for anti-replay)
So I'm a bit confused ... ^^
(This change is not directly linked to this issue ?)

sbernard31:
About the #1 issue, the discussion is very long and a lot of solution was exposed which one do you planned to implement ? because the last conclusion was to pass a context to `send` function.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: I did not change anything here (at least not intentionally ;-)). We have always used new (record) sequence numbers when re-transmitting a flight. From my point of view this also is in compliance with the DTLS spec.

Section 4.2.2 reads:
> When a (handshake) message is retransmitted, the same message_seq value is used. Note, however, that from the perspective of the DTLS record layer, the retransmission is a new record. This record will have a new DTLSPlaintext.sequence_number value.

This makes sense exactly because duplicate detection is done based on the record sequence number at the record layer. Imagine the following: 

1. Client sends ClientHello using record seq_no 1.
2. Server processes message, marks record seq_no 1 as "processed" and sends its flight.
3. Flight is lost during transmission.
4. Client re-transmits its ClientHello using the same record seq_no 1
5. Server's record layer discards record as duplicate and the handshake stalls...

sophokles73:
I have already started to work on this locally. The idea is to pass a `MessageCallback` as part of the `RawData` object passed into `Connector.send(RawData)`. The `DTLSConnector` then invokes `MessageCallback.onContextEstablished(CorrelationContext)` and passes in an instance of `DtlsCorrelationContext` which contains the DTLS session ID, epoch and cipher name.
When the `DTLSConnector` receives a message on an established session, it also creates a `DtlsCorreleationContext`object and includes it in the `RawData` object handed over to `RawDataChannel.receiveData(RawData)`.

Within Californium's `CoapStack` we *attach* the `DtlsCorrelationContext` provided by the `MessageCallback`to the ongoing `Exchange` and later, when the response is received and matched based on the token, the `Matcher` can also compare the values contained in the incoming `DtlsCorrelationContext` to the values contained in the context attached to the `Exchange`.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: Now `setNewSequenceNumbers()` is only called in `handleTimeout`. Before this was called on `sendFlight`. So It seems there is a different behavior.  Am I wrong ?

(Ok I reread the spec and I mixed `message_seq`  and `record.sequence_number`, I agree with you now)

sophokles73: Yes, but `DTLSConnector.sendFlight(DTLSFlight)` explicitly checked for `flight.getTries() > 0` and only in that case set new sequence numbers on the flight, i.e. this method explicitly checked whether the flight was a re-transmisson. However, we already have a method that is *only* executed for flights that are re-transmitted and that is `DTLSConnector.handleTimeout()` which is invoked by the `TimerTask` scheduled for the outbound flight ...

sophokles73:
Hi @sbernard31,

I have fixed `ServerHandshaker.doProcessMessage()`... 
any additional concerns regarding this PR or can I go ahead and merge it into master frmo your point of view?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: You're right this seems ok :)

sbernard31:
:+1:

sophokles73:
merged in 6bb3b59

PR #11, opened 2016-02-19
Do resume session on restart if needed.

sbernard31:
Currently we can not restart  a device with a different `inetAddress`. This just don't works.
The idea is to do a resume session in this case.

See more details on [this thread](https://dev.eclipse.org/mhonarc/lists/cf-dev/msg00497.html) on the mailing list.

(About the tests I didn't find a good way to check if we do a resume handshake or not, at least the test I added verify that we can do a start/stop/start)


sbernard31:
I pushed the new proposition without API break.
(previous one with comment available [here](https://github.com/eclipse/californium/commit/cfce97ef88c87b6ec486ed451923a191db238381))

sbernard31:
Should be ok now ?

PR #16, opened 2016-03-03
Add clustering support for observe

sbernard31:
The idea is to be able to handle observe notification in a cluster. (more detail available [here](https://github.com/eclipse/leshan/wiki/Cluster))

To resolve this issue, I introduce 2 new interfaces:
  1. `NotificationListener` : to be able to get notification from any Californium instance.
  1. `ObserveRequestStore`: to share the pending observation state through each instances.

The other "big" change is the `Exchange` instance lifetime for observation:
* Before, the exchange lived all along the observe relation lifetime.
* Now, we recreate an exchange for each notification. I mean a new `Exchange` instance is created from data stored in `ObserverRequestStore` each time we receive a new notification and will be completed/removed as soon as the notification is handled.

All of this is clearly a work in progress (e.g. notification over DTLS is not supported). The idea is to share my first step to know if I go in the right direction.
So feedback is welcome ;)

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -23,6 +23,7 @@ package org.eclipse.californium.core;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
@@ -35,12 +36,14 @@ import org.eclipse.californium.core.coap.CoAP;
 import org.eclipse.californium.core.coap.CoAP.Type;
 import org.eclipse.californium.core.coap.LinkFormat;
 import org.eclipse.californium.core.coap.MediaTypeRegistry;
+import org.eclipse.californium.core.coap.MessageObserver;
 import org.eclipse.californium.core.coap.MessageObserverAdapter;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.network.Endpoint;
 import org.eclipse.californium.core.network.EndpointManager;
 import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.observe.NotificationListener;
 
 /**
  * The Class CoapClient.
@@ -846,7 +849,14 @@ public class CoapClient {
 	private CoapObserveRelation observeAndWait(Request request, CoapHandler handler) {
 		Endpoint outEndpoint = getEffectiveEndpoint(request);
 		CoapObserveRelation relation = new CoapObserveRelation(request, outEndpoint);
-		request.addMessageObserver(new ObserveMessageObserverImpl(handler, relation));
+		// add message observer to get the response.
+		ObserveMessageObserverImpl messageObserver = new ObserveMessageObserverImpl(handler, relation);
+		request.addMessageObserver(messageObserver);
+		// add notification listener to all notification
+		NotificationListener notificationListener = new Adapter(messageObserver, request);
+		outEndpoint.addNotificationListener(notificationListener);
+		// relation should remove this listener when the request is cancelled
+		relation.setNotificationListener(notificationListener);
 		CoapResponse response = synchronous(request, outEndpoint);
 		if (response == null || !response.advanced().getOptions().hasObserve())
 			relation.setCanceled(true);
@@ -865,7 +875,14 @@ public class CoapClient {
 	private CoapObserveRelation observe(Request request, CoapHandler handler) {
 		Endpoint outEndpoint = getEffectiveEndpoint(request);
 		CoapObserveRelation relation = new CoapObserveRelation(request, outEndpoint);
-		request.addMessageObserver(new ObserveMessageObserverImpl(handler, relation));
+		// add message observer to get the response.
+		ObserveMessageObserverImpl messageObserver = new ObserveMessageObserverImpl(handler, relation);
+		request.addMessageObserver(messageObserver);
+		// add notification listener to all notification
+		NotificationListener notificationListener = new Adapter(messageObserver, request);
+		outEndpoint.addNotificationListener(notificationListener);
+		// relation should remove this listener when the request is cancelled
+		relation.setNotificationListener(notificationListener);
 		send(request, outEndpoint);
 		return relation;
 	}
@@ -924,7 +941,27 @@ public class CoapClient {
 			return EndpointManager.getEndpointManager().getDefaultEndpoint();
 		}
 	}
-	
+
+	/*
+	 * Adapt MessageObserver for a given request in NotificationListener
+	 */
+	private class Adapter implements NotificationListener {
+		private MessageObserver obs;
+		private Request req;
+
+		public Adapter(MessageObserver observer, Request request) {
+			obs = observer;
+			req = request;
+		}
+
+		@Override
+		public void onNotification(Request request, Response response) {
+			if (Arrays.equals(request.getToken(), req.getToken())) {
+				obs.onResponse(response);
+			}
+		}
+	}
+
 	/**
 	 * The MessageObserverImpl is called when a response arrives. It wraps the
 	 * response into a CoapResponse and lets the executor invoke the handler's
@@ -1031,7 +1068,7 @@ public class CoapClient {
 			super(handler);
 			this.relation = relation;
 		}
-		
+
 		/**
 		 * Checks if the specified response truly is a new notification and if,
 		 * invokes the handler's method or drops the notification otherwise.
@@ -1056,6 +1093,7 @@ public class CoapClient {
 		 * method.
 		 */
 		@Override protected void failed() {
+			// When relation is canceled remove the notification listener
 			relation.setCanceled(true);
 			super.failed();
 		}
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
@@ -27,6 +27,7 @@ import java.util.concurrent.TimeUnit;
 import org.eclipse.californium.core.coap.MessageObserver;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.network.Endpoint;
+import org.eclipse.californium.core.observe.NotificationListener;
 import org.eclipse.californium.core.observe.ObserveNotificationOrderer;
 
 /**
@@ -58,7 +59,9 @@ public class CoapObserveRelation {
 
 	/** The handle to re-register for Observe notifications */
 	private ScheduledFuture<?> reregistrationHandle = null;
-	
+
+	private NotificationListener notificationListener;
+
 	/**
 	 * Constructs a new CoapObserveRelation with the specified request.
 	 *
@@ -125,6 +128,7 @@ public class CoapObserveRelation {
 		
 		// cancel old ongoing request
 		request.cancel();
+		endpoint.cancelObservation(request.getToken());
 		setCanceled(true);
 	}
 	
@@ -134,6 +138,7 @@ public class CoapObserveRelation {
 	 */
 	public void reactiveCancel() {
 		request.cancel();
+		endpoint.cancelObservation(request.getToken());
 		setCanceled(true);
 	}
 	
@@ -165,9 +170,17 @@ public class CoapObserveRelation {
 		
 		if (this.canceled) {
 			setReregistrationHandle(null);
+
+			if (notificationListener != null) {
+				endpoint.removeNotificationListener(notificationListener);
+			}
 		}
 	}
 	
+	public void setNotificationListener(NotificationListener listener) {
+		notificationListener = listener;
+	}
+
 	/**
 	 * Sets the current notification.
 	 *
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -54,6 +54,9 @@ import org.eclipse.californium.core.network.stack.BlockwiseLayer;
 import org.eclipse.californium.core.network.stack.CoapStack;
 import org.eclipse.californium.core.network.stack.ObserveLayer;
 import org.eclipse.californium.core.network.stack.ReliabilityLayer;
+import org.eclipse.californium.core.observe.InMemoryObservationStore;
+import org.eclipse.californium.core.observe.NotificationListener;
+import org.eclipse.californium.core.observe.ObservationStore;
 import org.eclipse.californium.core.server.MessageDeliverer;
 import org.eclipse.californium.elements.Connector;
 import org.eclipse.californium.elements.CorrelationContext;
@@ -149,6 +152,9 @@ public class CoapEndpoint implements Endpoint {
 	/** The list of interceptors */
 	private List<MessageInterceptor> interceptors = new ArrayList<>(0);
 
+	/** The list of Notification listener (use for CoAP observer relations) */
+	private List<NotificationListener> notificationListeners = new ArrayList<NotificationListener>(0);
+
 	/**
 	 * Instantiates a new endpoint with an ephemeral port.
 	 */
@@ -214,9 +220,35 @@ public class CoapEndpoint implements Endpoint {
 	 * @param config the config
 	 */
 	public CoapEndpoint(Connector connector, NetworkConfig config) {
+		this(connector, config, null);
+	}
+
+	/**
+	 * Instantiates a new endpoint with the specified address and configuration
+	 * and observe request store (used to persist observation).
+	 *
+	 * @param connector the connector
+	 * @param config the config
+	 */
+	public CoapEndpoint(InetSocketAddress address, NetworkConfig config, ObservationStore store) {
+		this(createUDPConnector(address, config), config, store);
+	}
+
+	/**
+	 * Instantiates a new endpoint with the specified connector and configuration
+	 * and observe request store (used to persist observation).
+	 *
+	 * @param connector the connector
+	 * @param config the config
+	 */
+	public CoapEndpoint(Connector connector, NetworkConfig config, ObservationStore store) {
 		this.config = config;
 		this.connector = connector;
-		this.matcher = new Matcher(config);		
+		if (store == null) {
+			this.matcher = new Matcher(config, new NotificationDispatcher(), new InMemoryObservationStore());
+		} else {
+			this.matcher = new Matcher(config, new NotificationDispatcher(), store);
+		}
 		this.coapstack = new CoapStack(config, new OutboxImpl());
 		this.connector.setRawDataReceiver(new InboxImpl());
 	}
@@ -372,6 +404,28 @@ public class CoapEndpoint implements Endpoint {
 		this.matcher.setExecutor(executor);
 	}
 
+	/*
+	 * (non-Javadoc)
+	 * 
+	 * @see
+	 * org.eclipse.californium.core.network.Endpoint#addNotificationListener(org.eclipse.californium.core.observe.NotificationListener)
+	 */
+	@Override
+	public void addNotificationListener(NotificationListener lis) {
+		notificationListeners.add(lis);
+	}
+
+	/*
+	 * (non-Javadoc)
+	 * 
+	 * @see
+	 * org.eclipse.californium.core.network.Endpoint#removeNotificationListener(org.eclipse.californium.core.observe.NotificationListener)
+	 */
+	@Override
+	public void removeNotificationListener(NotificationListener lis) {
+		notificationListeners.remove(lis);
+	}
+
 	/* (non-Javadoc)
 	 * @see org.eclipse.californium.core.network.Endpoint#addObserver(org.eclipse.californium.core.network.EndpointObserver)
 	 */
@@ -484,6 +538,15 @@ public class CoapEndpoint implements Endpoint {
 	@Override
 	public NetworkConfig getConfig() {
 		return config;
+	}
+
+	private class NotificationDispatcher implements NotificationListener {
+		@Override
+		public void onNotification(Request request, Response response) {
+			for (NotificationListener notificationListener : new ArrayList<>(notificationListeners)) {
+				notificationListener.onNotification(request, response);
+			}
+		}
 	}
 
 	/**
@@ -736,6 +799,11 @@ public class CoapEndpoint implements Endpoint {
 				connector.send(Serializer.serialize(rst));
 			}
 		}
+	}
+
+	@Override
+	public void cancelObservation(byte[] token) {
+		matcher.cancelObserve(token);
 	}
 
 	/**
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/ExchangeObserver.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/ExchangeObserver.java
@@ -33,7 +33,7 @@ public interface ExchangeObserver {
 	 * @param exchange the exchange
 	 */
 	void completed(Exchange exchange);
-	
+
 	/**
 	 * Invoked when a new correlation context is set.
 	 * 
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
@@ -23,7 +23,9 @@
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
+import java.util.Arrays;
 import java.util.Iterator;
+import java.util.Map.Entry;
 import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
@@ -36,6 +38,7 @@ import java.util.logging.Logger;
 import org.eclipse.californium.core.coap.CoAP.Type;
 import org.eclipse.californium.core.coap.EmptyMessage;
 import org.eclipse.californium.core.coap.Message;
+import org.eclipse.californium.core.coap.MessageObserverAdapter;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.network.Exchange.KeyMID;
@@ -45,6 +48,10 @@ import org.eclipse.californium.core.network.Exchange.Origin;
 import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.core.network.deduplication.Deduplicator;
 import org.eclipse.californium.core.network.deduplication.DeduplicatorFactory;
+import org.eclipse.californium.core.observe.InMemoryObservationStore;
+import org.eclipse.californium.core.observe.NotificationListener;
+import org.eclipse.californium.core.observe.Observation;
+import org.eclipse.californium.core.observe.ObservationStore;
 import org.eclipse.californium.core.observe.ObserveRelation;
 import org.eclipse.californium.elements.CorrelationContext;
 import org.eclipse.californium.elements.DtlsCorrelationContext;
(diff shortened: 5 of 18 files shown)


sbernard31:
I agree this PR will be mainly focused on "Separate ObserveStore, which can then be persisted easily".
As I understand your comments, you seems ok with the main idea, right ?

I currently try to integrate this in Leshan. (without security)
So I will probably commit some minor change about the current API, I introduce.
Then when we will have something which seems to work,  *security* will be the next step.

sbernard31:
We integrate this in Leshan in this [PR #93](https://github.com/eclipse/leshan/pull/93).
The first elementary tests seems ok.

So I will continue. Before to start on security issue,  I need to find a way to integrate this works in the `CoapClient` high level API.

sbernard31:
I updated the high level API `CoapClient` to take into account the previous modification.
Next step fix the blockwise transfer issue.

sbernard31:
I fix the code and the tests.
We are now able to handle several observe notification using blockwise transfer (block 2) that why there is some changes in the tests.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/SynchronousNotificationListener.java
```diff
@@ -0,0 +1,105 @@
...
+		synchronized (lock) {
+			if (currentResponse != null)
+				r = currentResponse;
+			else {
+				lock.wait(timeoutInMs);
+				r = currentResponse;
+			}
+			currentResponse = null;
+		}
+		System.out.println("waitForResponse " + i + " " + r);
```
jvermillard: :point_up:

sbernard31:
I added correlation context support.
Feedbacks are welcome (not only about the correlation context support addition)

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -204,18 +208,33 @@ public CoapEndpoint(InetSocketAddress address, NetworkConfig config) {
...
+	 *
+	 * @param connector the connector
+	 * @param config the config
+	 */
+	public CoapEndpoint(Connector connector, NetworkConfig config, ObserveRequestStore store) {
 		this.config = config;
 		this.connector = connector;
-		this.matcher = new Matcher(config, notificationListener, store);
+		if (store == null)
+			store = new InMemoryObserveRequestStore();
```
sophokles73: Do not alter passed in arguments, use separate variable instead.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -174,6 +174,10 @@ public void sendRequest(Exchange exchange,final  Request request) {
 		if (request.getToken() == null) {
 			idByToken = createUnusedToken();
 			request.setToken(idByToken.token);
+			// if original request has no token set it too.
+			// this is the case where request and currentRequest is not the same
+			if (exchange.getRequest() != null && exchange.getRequest().getToken() == null)
```
sophokles73: Can you explain under which circumstances this happens?

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -182,11 +186,23 @@ public void sendRequest(Exchange exchange,final  Request request) {
 			}
 		}
 
-		if (!request.getOptions().hasBlock2() && request.getOptions().hasObserve() && request.getOptions().getObserve() == 0) {
+		// for observe request
+		if ( request.getOptions().hasObserve() && request.getOptions().getObserve() == 0 && !request.getOptions().hasBlock2()) {
```
sophokles73: Why do we check for the absence of the Block2 option? My understanding of the [Blockwise draft 19](https://tools.ietf.org/html/draft-ietf-core-block-19#section-2.4) is that an initial GET request MAY also include this option.

californium-core/src/main/java/org/eclipse/californium/core/network/ExchangeObserver.java
```diff
@@ -33,5 +33,12 @@
...
 	 */
 	public void completed(Exchange exchange);
-	
+
+	/**
+	 * Invoked when a new correlation context is set.
+	 * 
+	 * @param exchange the exchange
+	 */
+	public void contextEstablised(Exchange exchange);
```
sophokles73: TYPO: should be `contextEstablished()`, I guess....

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
```diff
@@ -20,27 +20,28 @@
...
+			map.put(new KeyToken(obs.getRequest().getToken()), obs);
 		}
 	}
 
 	@Override
-	public Request get(byte[] token) {
-		Request request = map.get(new KeyToken(token));
-		if (request != null) {
-			RawData serialize = Serializer.serialize(request, null);
+	public Observation get(byte[] token) {
```
sophokles73: Also add `get(KeyToken)` variant in order to prevent needless creation of duplicate KeyToken instances in `Matcher.receiveResponse()`.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -384,19 +385,20 @@ public Exchange receiveResponse(final Response response, final CorrelationContex
 		KeyToken idByToken = new KeyToken(response.getToken());
 
 		Exchange exchange = exchangesByToken.get(idByToken);
-		if (exchange == null && observeRequestStore != null ) {
-			final Request request = observeRequestStore.get(response.getToken());
-			if (request != null){
+		if (exchange == null && observationStore != null ) {
+			final Observation obs = observationStore.get(response.getToken());
```
sophokles73: try to re-use the `idByToken` instance ...

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -343,6 +366,50 @@ public Exchange receiveResponse(final Response response, final CorrelationContex
...
+
+					@Override
+					public void onCancel() {
+						notificationListener.onCancel(request);
+						observeRequestStore.remove(request);
+					}
+
+					@Override
+					public void onAcknowledgement() {
+						notificationListener.onCancel(request);
```
sophokles73: Shouldn't this delegate to `notificationListener.onAcknowledgement()`?

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -846,7 +849,15 @@ private Request ifNoneMatch(Request request) {
...
-		request.addMessageObserver(new ObserveMessageObserverImpl(handler, relation));
+		// add message observer to get the response.
+		ObserveMessageObserverImpl messageObserver = new ObserveMessageObserverImpl(handler, relation);
+		request.addMessageObserver(messageObserver);
+		// add notification listener to all notification
+		NotificationListener notificationListener = new Adapter(messageObserver, request);
+		outEndpoint.addNotificationListener(notificationListener);
+		// messageobserver should remove this listener when the request is cancelled
+		messageObserver.setNotificationListener(notificationListener);
+		relation.setNotificationListener(notificationListener);
```
sophokles73: I do not understand what these additions are good for. Can you explain?

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
sbernard31: even to manage optional parameters ?

californium-core/src/main/java/org/eclipse/californium/core/network/ExchangeObserver.java
sbernard31: You guess right ;)

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
sbernard31: oops :/

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
sbernard31: `request` parameter is the current request, while `exchange.getRequest` is the original request.
Since I add this [modification](https://github.com/eclipse/californium/pull/16/files#diff-20c9d7fed566d6ad9272e84175d7b9c2R459), we could encounter some NPE because `exchange.getRequest.getToken` is `null`.

This is mainly the case where you send a request with blockwise transfer (block1).
In this case a new request is created in the `blockwiselayer` and so when we are in the `matcher` and create a new token for this request we need to modify the original one too.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sbernard31: The idea is to keep the previous behavior for this class with the new implementation.
Instead of getting notification through the `MessageObserver`, we get in now throw a notificationListener.
But we need to remove this notification listener when the observation is cancelled.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
sophokles73: Well, I would. This is a public method and you have no control over of what clients will pass in and do with the object reference after your constructor has completed.
Personally, I would **always** stick to this principle. But since we have no general coding rule in place for this, I guess you are free to do it any way you like and I do not want to start a broad discussion about it. Just my personal opinion ...

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
sophokles73: But why would the original request have no token?

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sophokles73: I may not have fully understood the overall idea here but here's how I think it *should* work.
1. client code creates an instance of `CoapClient` and starts observing a resource using e.g. `CoapClient.observeAndWait(CoapHandler, int)`
1. `CoapClient` creates an `ObserveMessageObserverImpl` as already done in the original code and registers it with the original `Request` object.
1. the original request is put to ÒbservationStore`
1. when a notification comes in the original request is retrieved from the `ObservationStore` and the `MessageObserver.onResponse()` method is invoked (as part of `Request.setResponse()`)
1. the *global* `NotificationListener` registered on the endpoint is also notified about the response resceived in reply to the observe request.

I do not see any need for registering the additional observer specific `NotificationListener`adapter. In case the node on which the original observe request (including the handler object) has been issued crashes, there is no longer a handler instance to be notified anyways, or is there?
I think client code that is supposed to also work in case of a fail-over, must not use the *individual message observer* approach for observing resources but must instead register a notification listener at the endpoint level from the beginning.
In that regard I do not think that we are changing the behavior of `CoapClient` but instead only add support for fail-over scenarios by means of new (alternative) API.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
sbernard31: So, I will fix that :)

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
sbernard31: Because the API allow you to create request without token and the token is dynamically created in the matcher, but in this case the blockwise request is created in blockwiselayer before the original one go through the matcher.
```
 * +-----------------------+
 * | MessageDeliverer      |
 * +-------------A---------+
 *               |
 *             * A            
 * +-Endpoint--+-A---------+
 * |           v A         |
 * |           v A         |
 * | +---------v-+-------+ |
 * | | Stack Top         | |
 * | +-------------------+ |
 * | |  ObserveLayer     | |
 * | +-------------------+ |
 * | |  BlockwiseLayer   | |
 * | +-------------------+ |
 * | |  ReliabilityLayer | |
 * | +-------------------+ |
 * | | Stack Bottom      | |
 * | +--------+-+--------+ |
 * |          v A          |
 * |          v A          |
 * |       Matcher         |
 * |          v A          |
 * |    MessageInterceptor |  
 * |          v A          |
 * |          v A          |
 * | +--------v-+--------+ |
 * +-|    Connector      |-+
 *   +--------+-A--------+
 *            v A
 *            v A
 *         (Network)
```

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sbernard31: 4. only the first response is accessible through MessageObserver.onResponse()
5. only the next notification are available on NotificationListener.

[Here](https://github.com/eclipse/californium/pull/16#commitcomment-16503803) is the main explanation of this.

sophokles73: If I understand correctly you mean that only the first response will be delivered to the `MessageObserver` the client code registered with the original request and subsequent notifications will not be. But why do you think that would be necessarily so?
My understanding is that as long as the process the client code is part of has not crashed, we keep a reference to the original request in the ObservationStore and thus have a reference to the MessageObserver the client registered. So I do not see why we should not be able to deliver *all* incoming notifications to the observer ...

jvermillard: if you push all the notif to the MessageObserver you encourage user to write code not working in cluster mode. For example a device push some notif and then push the next one 2hours after. It will probably reach another node and if will be only available using the NotificationListener.

If you ask the user to directly use the NotificationListener for receiving notification he will write clusterisable code from day one.

sbernard31: You're right @jvermillard , but the high level API (`CoapClient`) is not really cluster friendly :/.

@sophokles73, I don't think this is a good idea to force ObservationStore implementation to keep in memory a reference to the original request with its `MessageObserver` . [Here](https://github.com/eclipse/leshan/blob/clustering/leshan-server-demo/src/main/java/org/eclipse/leshan/server/demo/cluster/RedisObservationRegistry.java#L330) an example of the leshan redis implementation. This seems to me

sophokles73: I agree completely with you, @jvermillard.
However, I still do not understand why we need the `Adapter` which, to my understanding, forwards the notifications matching the original request to the observer registered with the original request, which would get these notifications anyways (without the `Adapter`).

sophokles73: @sbernard31 I see your concern but I guess it is only valid because you are using Redis as the primary cache. I would rather use it as a secondary cache, i.e. I would keep the entries in memory and only get from Redis if I have a cache miss. When you put to the cache you can simply omit the MessageObserver from the serialization to Redis because it will be useless on another node anyways ...

sbernard31: @sophokles73  if you have a primary cache in memory how did you sync it when you cancel an observation which means with the current implementation just remove it from the store ? 
Even if this is possible to use redis as a secondary cache, should we force users to keep an in memory cache for all their implementation ? 

About, the adapter the idea is just to only get the notification from `coapEndpoint` which concerns the request for this `CoapClient` and let the ObserveMessageObserverImpl doing the job as before.

sophokles73: Well, the `RedisObservationStore` could itself implement the in-memory (first level) cache. When the observation is canceled, i.e. `RedisObservationStore.remove()` is invoked, you simply remove it from the map as well as from Redis :-)
This way we do not force anybody to do anything because the in-memory cache is part of the implementation anyways. If somebody else wants to implement `ObservationStore` in a different way they are still free to do so. I am actually following a very similar approach for [implementing DTLS session caching](https://github.com/eclipse/californium/tree/session_cache) ...

My concern with the adapter is that for every observation we register a new `NotificationListener` that needs to be invoked when a notification arrives. Thus, the number of listeners grows linearly with the number of connected (and observed) devices...

sbernard31: If we don't force anybody about in-memory cache. That means we can not consider that we always have a request with its `MessageObserver` in memory, right ? so you agree that we need to use NotificationListener in coapClient. (Or do I miss something ?)

About cleaning the first level cache, what about if the cancel is not done on instances which have this request  cached in memory ? (Not so important as we agree that this should not be mandatory)

About the adapter, you're right. I don't know how we can avoid this :'(... In all case I'm not sure the `CoapClient` API is well adapted when you want to handle a lot of clients, because you need to create a coapClient by [IP/port/resourceURI](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java#L95). That's why we don't use it in Leshan.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
sbernard31: not sure to understand ?

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
sbernard31: KeyToken is only use for map I don't want to change the observationStore API only for the inmemory implementation.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
sbernard31: You 're right, I should accept block2 option with num and m set to 0.

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
sophokles73: What do you mean by
> only for the inmemory implementation.

The `KeyToken` serves as the key not only in `ObservationStore` but also e.g. in `Matcher`'s exchange maps ...

sbernard31: `KeyToken` is an internal object used to allow a byte array as key in a map. I'm not sure this is a good idea to expose it the `observationStore` API just to avoid to create a keyToken object in `InMemoryObservationStore`.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sophokles73: So, maybe we should tackle it from another angle then and discuss the relevance of the `CoapClient` class in general. It looks like its usefulness is limited to simple use cases anyways so we could argue that we do not support fail-over for this type of client at all (it simply doesn't fit the programming model) and that we instead encourage developers to use an alternative API (based on NotificationListener) which we will then need to provide.

sbernard31: This is more or less what I propose, I just modify coapClient to keep the previsous behavoir. (But the API is not really adapted to the fail-over/multi instance use case)

The alternative API is:
org.eclipse.californium.core.network.CoapEndpoint.sendRequest(Request)
org.eclipse.californium.core.network.CoapEndpoint.addNotificationListener(NotificationListener)
org.eclipse.californium.core.network.CoapEndpoint.removeNotificationListener(NotificationListener)
org.eclipse.californium.core.network.CoapEndpoint.cancelObservation(byte[])

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
sophokles73: Oh yes, you are right, of course. I didn't realize it was an internal class ... forget about it :-)

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -855,8 +855,7 @@ private CoapObserveRelation observeAndWait(Request request, CoapHandler handler)
 		// add notification listener to all notification
 		NotificationListener notificationListener = new Adapter(messageObserver, request);
 		outEndpoint.addNotificationListener(notificationListener);
-		// messageobserver should remove this listener when the request is cancelled
-		messageObserver.setNotificationListener(notificationListener);
+		// relation should remove this listener when the request is cancelled
 		relation.setNotificationListener(notificationListener);
 		CoapResponse response = synchronous(request, outEndpoint);
```
sophokles73: Up to here this is the same code as in `observe()`.
Maybe we can refactor this code into a separate private method?

californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
```diff
@@ -171,7 +171,7 @@ protected void setCanceled(boolean canceled) {
 			setReregistrationHandle(null);
 		}
 
-		if (notificationListener != null) {
+		if (this.canceled && notificationListener != null) {
```
sophokles73: Move up two lines inside of the existing

    if (this.canceled) {
       setRegistrationhandle(null)
    }

block?

sbernard31: you're right.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sophokles73: Why didn't you say so in the first place? ;-)

However, I think the `NotificationListener` needs a lot more JavaDoc explaining when and how its methods are invoked. The existing doc often generically refers to *the message*. However, I think we need to explicitly explain what the context of the message and its meaning is.

We will also need to add JavaDoc to `CoapClient` indicating that it is not suitable for implementing code that can fail over to other nodes (and what should be used instead).

Do we also want to deprecate `CoapClient`?

sbernard31: About `coapClient` deprecation, I have no strong opinion on it. It seems to be used in some tests. I don't know which kind of use case it target. Maybe we should ask this on the mailing list and try to know if there is users attach to this API.
(We should get Matthias' opinion too)

About documentation, could we use your hyper-documentation-skills ? :D

sophokles73: My take is that the `CoapClient` has been incepted as an easy to use helper for accessing resources on a server. I think the observe stuff has been added later, clearly without a focus on high availability scenarios.

We shoudl definitely get Matthias opinion about it. Can you put something on the mailing list, @sbernard31?

I will give the documentation a try :-) I will simply also commit to this branch, ok?

sbernard31: ok :)

evgeniynickolaev: Excuse me, I did not catch the point on CoapClient deprecation.
Does it mean all the CoapClient API will be deprecated? What could be used instead?
I am using it in my functional tests to call the CoAP server resources.

sbernard31: I think we should use the mailing list for this discussion. But don't worry we just ask if this make sense to keep this API. If there is users attached to it, we will keep it. But as we said there is some limitation with fail-over/clustering use cases.

evgeniynickolaev: > I am using it [CoapClient API] in my functional tests to call the CoAP
server resources.
UPD: Oh, not only for tests... My CoAP server uses CoapClient to call
embedded devices back (sort of duplex connection).
I guess it would work for me if CoAP Client API was extracted in a separate
library (could it solve the clustering issues?)

mkovatsc:
Hi #16 :)
As already mentioned in #37, I have doubts about the idea of NotificationListeners.
I do not like the Adapters either (sorry :D) nor modifications to CoapClient. I think the problem is that something essential like the NotificationOrderer ended up here. I remember, there was a reason for this, but unfortunately no details (probably concurrency again, I guess).

My proposal would be to move the NotificationOrderer out of the CoapClient. Remember: CoapClient is supposed to be _one_ of multiple possible APIs. For the re-registration, I am not sure. I would expect the intention and initiative for re-registration to depend on the application, which is then reflected in the choice of API.

For this cluster issue, I would propose to addd an interface for the NotifiactionOrderer, so that either a local one ore the  coordinated cluster orderer can be plugged in (quite analogous to the ObservationStore).

mkovatsc:
However, it would also be interesting, what you have in mind for the actual notification handler. I guess it will be executed on one of the cluster nodes. Probably, it writes to a database. The point is, that the NotificationOrderer is applied just before simething like this happens. Thus, it was in the CoapClient because that has the notification handler in this simple API.

So, it would not make sense to have a coordinated NotificationOrderer, when the notifcation handler code again needs to do something coordinated...

Still, I do not get how multiple NotificationListeners are supposed to solve this issue.

sbernard31:
As I said on the mailing list, I think the `NotificationOrderer` question should be address after we agreed on the internal/advanced API and the hight level API modification.
I think both solution are possible, I mean keeping the Notification ordering at High level API or move it in the internal/advanced one. I didn't think a lot about this, but for cluster implementation I suppose we can store observe option value.

As Kai proposed, I integrated the `NotificationListener` interface from Kai's PR.

I encountered some instability in `MemoryLeakingHashMapTest` (I don't think this is related to the last modification)
It seems sometime the Maps was not empty. Each times this happened, I get this strange logs:
```
INFOS: /127.0.0.1:45 154 ==> req CON-GET    MID=13417, Token=879694724ec85a04, OptionSet={"Observe":0, "Uri-Port":47384, "Uri-Path":"piggy"}, no payload
INFOS: Successfully established observe relation between /127.0.0.1:45154#879694724ec85a04 and resource /piggy
INFOS: /127.0.0.1:45 154 <== res ACK-2.05   MID=13417, Token=879694724ec85a04, OptionSet={"Observe":0, "Content-Format":"text/plain", "Block2":"(szx=0/16, m=true, num=0)"}, "123456789.123456"
INFOS: /127.0.0.1:45 154 ==> req CON-GET    MID=13418, Token=150e9ac957c25d, OptionSet={"Uri-Port":47384, "Uri-Path":"piggy", "Block2":"(szx=0/16, m=false, num=1)"}, no payload
INFOS: /127.0.0.1:45 154 <== res NON-2.05   MID=65366, Token=879694724ec85a04, OptionSet={"Observe":1, "Content-Format":"text/plain", "Block2":"(szx=0/16, m=true, num=0)"}, "123456789.123456"
INFOS: /127.0.0.1:45 154 <== res ACK-2.05   MID=13418, Token=150e9ac957c25d, OptionSet={"Content-Format":"text/plain", "Block2":"(szx=0/16, m=true, num=1)"}, "789.123456789.12"
WARNING: Ongoing blockwise transfer reseted at num=1 by new notification: NON-2.05   MID=65366, Token=879694724ec85a04, OptionSet={"Observe":1, "Content-Format":"text/plain", "Block2":"(szx=0/16, m=true, num=0)"}, "123456789.123456"
WARNING: Possible MID reuse before lifetime end for token [150e9ac957c25d], expected MID 13 419 but received 13 418
INFOS: /127.0.0.1:45 154 ==> req CON-GET    MID=13419, Token=951796cd, OptionSet={"Uri-Port":47384, "Uri-Path":"piggy", "Block2":"(szx=0/16, m=false, num=1)"}, no payload
WARNING: Wrong block number. Expected 1 but received 1: ACK-2.05   MID=13418, Token=150e9ac957c25d, OptionSet={"Content-Format":"text/plain", "Block2":"(szx=0/16, m=true, num=1)"}, "789.123456789.12"
```
Set OBS_NOTIFICATION_INTERVAL to a bigger value hide this issue.

sbernard31:
I rebased on master (squash all the commit in one and fix conflicts issue)
I test it with Leshan cluster branch. It seems ok :).

sophokles73:
As discussed in our face-to-face in Toulouse, I have merged this PR to the 2.0.x branch in 87a9b1c

PR #17, opened 2016-03-04
Option Number Registry as an Enum type

mojadita:
Signed-off-by: Luis Colorado lcu@basen.net

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/OptionTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/OptionTest.java
@@ -143,9 +143,9 @@ public class OptionTest {
 		assertArrayEquals(option.getValue(), new byte[] {(byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF});
 		assertEquals(0xFFFFFFFFL, option.getLongValue());
 
-		option.setLongValue(0x9823749837239845L);
-		assertArrayEquals(option.getValue(), new byte[] {-104, 35, 116, -104, 55, 35, -104, 69});
-		assertEquals(0x9823749837239845L, option.getLongValue());
+		//option.setLongValue(0x9823749837239845L);
+		//assertArrayEquals(option.getValue(), new byte[] {-104, 35, 116, -104, 55, 35, -104, 69});
+		//assertEquals(0x9823749837239845L, option.getLongValue());
 
 		option.setLongValue(0xFFFFFFFFFFFFFFFFL);
 		assertArrayEquals(option.getValue(), new byte[] {(byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF,
@@ -211,25 +211,26 @@ public class OptionTest {
 	
 	@Test
 	public void testArbitraryOptions() {
+	    // arbitrary options not present in OptionNumberRegistry are not allowed anymore.
 		OptionSet options = new OptionSet();
 		options.addETag(new byte[] {1, 2, 3});
 		options.addLocationPath("abc");
-		options.addOption(new Option(7));
-		options.addOption(new Option(43));
-		options.addOption(new Option(33));
-		options.addOption(new Option(17));
+		options.addOption(new Option(OptionNumberRegistry.parse(7)));
+		//options.addOption(new Option(OptionNumberRegistry.parse(43)));
+		//options.addOption(new Option(OptionNumberRegistry.parse(33)));
+		options.addOption(new Option(OptionNumberRegistry.parse(17)));
 
 		// Check that options are in the set
 		Assert.assertTrue(options.hasOption(OptionNumberRegistry.ETAG));
 		Assert.assertTrue(options.hasOption(OptionNumberRegistry.LOCATION_PATH));
-		Assert.assertTrue(options.hasOption(7));
-		Assert.assertTrue(options.hasOption(17));
-		Assert.assertTrue(options.hasOption(33));
-		Assert.assertTrue(options.hasOption(43));
+		Assert.assertTrue(options.hasOption(OptionNumberRegistry.parse(7)));
+		Assert.assertTrue(options.hasOption(OptionNumberRegistry.parse(17)));
+		//Assert.assertTrue(options.hasOption(OptionNumberRegistry.parse(33)));
+		//Assert.assertTrue(options.hasOption(OptionNumberRegistry.parse(43)));
 		
 		// Check that others are not
-		Assert.assertFalse(options.hasOption(19));
-		Assert.assertFalse(options.hasOption(53));
+		//Assert.assertFalse(options.hasOption(OptionNumberRegistry.parse(19)));
+		//Assert.assertFalse(options.hasOption(OptionNumberRegistry.parse(53)));
 		
 		// Check that we can remove options
 		options.clearETags();
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/OptionTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/OptionTest.java
@@ -51,8 +51,13 @@ public class OptionTest {
 	}
 
     @Test
-    public void testEnumSize() {
+    public void testEnumSizeAndOrder() {
+        int last = 0;
         assertEquals(19, OptionNumberRegistry.values().length);
+        for (OptionNumberRegistry o: OptionNumberRegistry.values()) {
+            assertTrue(last < o.getProtocolValue());
+            last = o.getProtocolValue();
+        }
     }
 
     @Test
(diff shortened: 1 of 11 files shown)


mojadita:
I discovered this bug trying to integrate library on own code. As of reviewing more, I discovered more errors in the OptionNumberRegistry so I wrote several Unit tests to check all the options against RFC-7252 doc.

As you requested, I have pushed the branch to **californium** package.

mojadita:
Yes, but both methods were written without referring one to another... so that's the reason to put unit tests for both.

By the way, I plan to rewrite all these classes (Method, Options, MessageTypes, ResultCodes and Mediatypes as enums with added functionality, to add uniformity ---some of them are and others not, making it difficult to maintain some)

mojadita:
I have refactored completely OptionNumberRegistry to be an enumeration.  It has methods to map from enum to protocol numbers and backwards.  I have rewritten the serializer for integer and long options and the points where protocol numbers are used.

A consequence of this is that arbitrary option numbers are not permitted anymore.  I have commented the related tests in the unit tests section.

Thaks in advance for your consideration.

californium-core/src/test/java/org/eclipse/californium/core/test/OptionTest.java
```diff
@@ -143,9 +265,9 @@ public void testSetLongValue() {
 		assertArrayEquals(option.getValue(), new byte[] {(byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF});
 		assertEquals(0xFFFFFFFFL, option.getLongValue());
 
-		option.setLongValue(0x9823749837239845L);
-		assertArrayEquals(option.getValue(), new byte[] {-104, 35, 116, -104, 55, 35, -104, 69});
-		assertEquals(0x9823749837239845L, option.getLongValue());
+		//option.setLongValue(0x9823749837239845L);
+		//assertArrayEquals(option.getValue(), new byte[] {-104, 35, 116, -104, 55, 35, -104, 69});
+		//assertEquals(0x9823749837239845L, option.getLongValue());
```
mojadita: Sorry, these tests continue being valid... I commented those to see next assertions, but forgot to uncomment.  Expect another commit about this.

californium-core/src/test/java/org/eclipse/californium/core/test/ParserTest.java
```diff
@@ -70,11 +69,11 @@ public void testResponseParsing() {
...
-							.addLocationPath("/one/two/three/four/five/six/seven/eight/nine/ten")
-							.addOption(new Option(57453, "Arbitrary".hashCode()))
-							.addOption(new Option(19205, "Arbitrary1"))
-							.addOption(new Option(19205, "Arbitrary2"))
-							.addOption(new Option(19205, "Arbitrary3"));
+							.addLocationPath("/one/two/three/four/five/six/seven/eight/nine/ten");
+//							.addOption(new Option(57453, "Arbitrary".hashCode()))
+//							.addOption(new Option(19205, "Arbitrary1"))
+//							.addOption(new Option(19205, "Arbitrary2"))
+//							.addOption(new Option(19205, "Arbitrary3"));
```
mojadita: These options are also invalid.

mojadita:
Now the commented test is uncommented and tests pass all again

californium-core/src/test/java/org/eclipse/californium/core/test/OptionTest.java
```diff
@@ -211,25 +333,26 @@ public void testLocationOptions() {
...
-		options.addOption(new Option(43));
-		options.addOption(new Option(33));
-		options.addOption(new Option(17));
+		options.addOption(new Option(OptionNumberRegistry.parse(7)));
+		//options.addOption(new Option(OptionNumberRegistry.parse(43)));
+		//options.addOption(new Option(OptionNumberRegistry.parse(33)));
+		options.addOption(new Option(OptionNumberRegistry.parse(17)));
 
 		// Check that options are in the set
 		Assert.assertTrue(options.hasOption(OptionNumberRegistry.ETAG));
```
mojadita: assertions commented in this test are all invalid now, as they all refer to unexistent OptionNumberRegistry values.  If arbitrary option is to be used, either define them in OptionNumberRegistry (easier now with the refactoring) or a new patch must be added.  I think this is the best approach to the problem.

mojadita:
Ready for your consideration to be merged.

mkovatsc:
There is a bit too much going on in this PR to merge it quickly ;)
- The isCritical() fix and the unit tests are definitely good.
- The changes to the number--byte conversion I have to check unhurriedly at some point.
- The enum, however, brings some problems: we have to check if other projects break now with this change, check what happens when unknown options are received (in particular in regard to proxies), and check how people can introduce their own options without patching the core library.

mojadita:
- Well, the enum possible problems is something that I was expecting.  But it is something straightforward to correct as normally the interfaces where it appears have changed types and probably will complaint about the incorrect typing.  I used this fact to make all necessary corrections in Cf, but I understand your  point of view.  How much time do you think it will take to decide merging or not?  Either case, as using non-conformant options is possible by default, i've tried to maintain myself coherent and avoid IANA non approved diversions  (this happens also in the case of media types).

- The changes of the number---byte conversion have passed all the unit tests you made, plus some tests I have been doing until the fit.  The previous number conversion was definitely obscure and unnecessarily complex.  I expect you'll agree with me after you do the neccesary testing.

- Do you agree in merging the isCritical() fix and the unit tests related and merge on this point?  Could it be after https://github.com/mojadita/californium/commit/c123fb1743b32ddeb2506c46927303a50897daa0 commit and rebase the rest to a new branch?

mojadita:
@mkovatsc, Well, I have been looking around to the **OptionSet** class and have not chance but to claim that the design doesn't scale well with unbounded Option's.  If you maintain support for unregistered options, you need a place to store them, and as they are implemented as fields in the OptionSet class, OptionSet class might be modified to allow for storage of the new, unregistered option data.  My approach is to continue developing the OptionSet class to use a EnumMap to store the options efficiently (even so you don't have to sort the list before encoding, as it's mandated by te intrinsic order in the enum class, which is checked properly in the Unit Tests I prepared) and add support for lists of options, depending on a field in the OptionNumberRegistry class, so adding a new option implies only touching the OptionNumberRegistry enum, and not having to touch OptionSet and possibly Option classes.  As it is designed now, no possibility of option storage is supported, as the different implemented options are mapped to different OptionSet fields.  I don't fully understand your comment in reference of having to touch californium code to use unregistered options.   Believe me, but you don't have more interest in this working and allowing for easy integration of new experimental options.

sophokles73:
@mojadita,

I suggest that you split up this PR as proposed by Matthias. This PR should only deal with the fix for `isCritical()` as the PR's title also suggests. We also would like to merge the unit tests created by you for that matter.
Please move all other changes to separate PRs so that we can evaluate independently. However, I am with Matthias that we need to be careful not to break Californium's public API (we have released an *official* version 1.o, so we need to keep backwards compatible or make the changes on a 2.x branch).

mojadita:
@sophokles73, @mkovatsc, please, don't close this pull request, as with the merge it will be possible to rebase it from the master branch again.  I have corrected several bad coded pieces of code there, and it's in the middle of converting the full set of Options into a more efficient piece of code:
* Now the encoding process requires sorting on the OptionSet (via asSortedList() method call) for each message.
* Now the use of options already set requires decoding as these are already encoded to binary form.
* No possibility of removing list entries in several options encoded as lists (e.g. for Uri-Path option)

I plan to modify all of these without modifying the external api, so perhaps we'll agree on a good push on your work.  I have developed a full class of option types to avoid long switch's and having to edit several files to add a single option in OptionNumberRegistry (as it is done now)

sophokles73:
Hi @mojadita,

it's been a year now ... if you want to spend some more work on this, then _now_ is the time to do it :-)
Let me know how you would like to proceed ...

boaks:
@mojadita 
If you want to continue to work on this, please create an other PR.
The master has many changes since that.

mojadita:
Well, it has been a year since I worked on this... I was told then about other projects possibly affected by the interface change, and had no chance but to wait...  Even having passed and reworked some unit tests that failed to properly check te whole thing.

Since then, I have begun working on other things and I have lost the interest in contributing to Cf, as more than one year of waiting queue is enough to me... Thanks for the interest, feel free to do whatever you want with the branch.  And next time, take some time to allow to read the email, as I have been waiting for more than a year for a response, before closing the issue.  

Thanks anyway for the interest.

boaks:
> And next time, take some time to allow to read the email, as I have been waiting for more than a year for a response, before closing the issue.

Sorry, you have been pinged on 28. March (https://github.com/eclipse/californium/pull/17#issuecomment-289793107) and didn't answered.

PR #18, opened 2016-03-14
Range for import package: org.eclipse.californium.elements should be "[1.0, 1.2)".

tux2323:
This means the version 1.2.0 is not included and the actual version 1.1.0 is in this range.

Example version ranges:
[1.2.3,4.5.6) 1.2.3 ≤ x < 4.5.6
[1.2.3,4.5.6] 1.2.3 ≤ x ≤ 4.5.6

### DIFF
--- a/scandium-core/pom.xml
+++ b/scandium-core/pom.xml
@@ -92,6 +92,7 @@
 							org.eclipse.californium.scandium.util
 						</Private-Package>
 						<Import-Package>
+							org.eclipse.californium.elements;version="[1.0, 1.2)",
 							*
 						</Import-Package>
 						<Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
--- a/scandium-core/pom.xml
+++ b/scandium-core/pom.xml
@@ -92,7 +92,6 @@
 							org.eclipse.californium.scandium.util
 						</Private-Package>
 						<Import-Package>
-							org.eclipse.californium.elements;version="[1.0, 1.2)",
 							*
 						</Import-Package>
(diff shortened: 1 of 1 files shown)


sophokles73:
Hi @tux2323,

thanks for your PR. However, in order for us to be able to accept it you will need to sign an Eclipse CLA and sign off on your commit as indicated [here](https://dev.eclipse.org/eclipse-webhook/services/status_details.php?id=56e69206ccc11).

I also think that we should actually use `2.0.0` as the upper boundary since we expect to keep all minor versions backward compatible.

Regards,
Kai

tux2323:
Hi @sophokles73
than it's normal OSGi semantic version logic and the import statement in pom.xml can be removed.

sophokles73:
That's even better then :-) Why did you close the PR?

tux2323:
I'm not sure if you can handle the pull request some of the commits did not pass the Eclipse validation.

sophokles73:
No, we can't. I just wanted to know whether you are fed up with the complicated process (that would be a pity :-() or simply want to create a new PR after you have signed the CLA and signed-off on your commit (that would be great).

So, by all means, if the Eclipse process requirements haven't scared you off (which I hope), we would be glad to receive your change as a new PR.

Thanks again!

kartben:
@tux2323 you can force push amended commits with the proper signed-off by headers in your repo, that should automatically update this PR. BTW in the future you probably want to do PR from a branch of your fork rather than directly master, so as the PR does not "point" to what is essentialy the HEAD of your own repo.

tux2323:
thx for the git hints

PR #22, opened 2016-03-16
Range for import package ...californium.elements should be "[1.0, 2.0)"

tux2323:
details see eclipse#18

Signed-off-by: Baranowski Christian Martin (INST-IOT/ESW) <ChristianMartin.Baranowski@bosch-si.com>


sophokles73:
@tux2323,

thanks for the PR :-)

BTW I think you should not disclose the information about the org unit you work in at Bosch in your *signed-off-by* footer. A simple *Christian Baranowski* should be sufficient.

PR #25, opened 2016-03-29
Mistake in OptionNumberRegistry.isCritical

mojadita:
Signed-off-by: Luis Colorado <lcu@basen.net>

This is the old branch with only the patches for unit test of OptionNumberRegistry and the bug in the isCritical method.

Sorry for the looong delay.


sophokles73:
Merged in `master` (3f6151d) and `1.0.x` (20d4db3)

PR #37, opened 2016-04-20
Observe clustering (no changes to CoapClient)

sophokles73:
The main difference between this approach and the original `observe_clustering` branch is that no changes to CoapClient have been made and the NotificationListener is only used for delivering notifications.

@Simon: I have incorporated (and refactored) a lot of your testing code but I still think that this approach is a little more lightweight ... ;-)

I have also tidied up a lot of the classes I needed to touch so that the overall amoutn of changes is larger than would have been necessary. But I thought: if I am already on it ...

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
@@ -96,18 +96,18 @@ public class CoapServer implements ServerInterface {
 
 	/** The root resource. */
 	private final Resource root;
-	
+
 	/** The message deliverer. */
 	private MessageDeliverer deliverer;
-	
+
 	/** The list of endpoints the server connects to the network. */
 	private final List<Endpoint> endpoints;
-	
+
 	/** The executor of the server for its endpoints (can be null). */
 	private ScheduledExecutorService executor;
-	
+
 	private final NetworkConfig config;
-	
+
 	/**
 	 * Constructs a default server. The server starts after the method
 	 * {@link #start()} is called. If a server starts and has no specific ports
@@ -116,7 +116,7 @@ public class CoapServer implements ServerInterface {
 	public CoapServer() {
 		this(NetworkConfig.getStandard());
 	}
-	
+
 	/**
 	 * Constructs a server that listens to the specified port(s) after method
 	 * {@link #start()} is called.
@@ -126,7 +126,7 @@ public class CoapServer implements ServerInterface {
 	public CoapServer(final int... ports) {
 		this(NetworkConfig.getStandard(), ports);
 	}
-	
+
 	/**
 	 * Constructs a server with the specified configuration that listens to the
 	 * specified ports after method {@link #start()} is called.
@@ -136,23 +136,23 @@ public class CoapServer implements ServerInterface {
 	 * @param ports the ports to bind to
 	 */
 	public CoapServer(final NetworkConfig config, final int... ports) {
-		
+
 		// global configuration that is passed down (can be observed for changes)
 		if (config != null) {
 			this.config = config;
 		} else {
 			this.config = NetworkConfig.getStandard();
 		}
-		
+
 		// resources
 		this.root = createRoot();
 		this.deliverer = new ServerMessageDeliverer(root);
-		
+
 		CoapResource wellKnown = new CoapResource(".well-known");
 		wellKnown.setVisible(false);
 		wellKnown.add(new DiscoveryResource(root));
 		root.add(wellKnown);
-		
+
 		// endpoints
 		this.endpoints = new ArrayList<>();
 		// sets the central thread pool for the protocol stage over all endpoints
@@ -164,19 +164,19 @@ public class CoapServer implements ServerInterface {
 			addEndpoint(new CoapEndpoint(port, this.config));
 		}
 	}
-	
+
 	public void setExecutor(final ScheduledExecutorService executor) {
-		
+
 		if (this.executor != null) {
 			this.executor.shutdown();
 		}
-		
+
 		this.executor = executor;
 		for (Endpoint ep : endpoints) {
 			ep.setExecutor(executor);
 		}
 	}
-	
+
 	/**
 	 * Starts the server by starting all endpoints this server is assigned to.
 	 * Each endpoint binds to its port. If no endpoint is assigned to the
@@ -184,18 +184,18 @@ public class CoapServer implements ServerInterface {
 	 */
 	@Override
 	public void start() {
-		
+
 		LOGGER.info("Starting server");
-		
+
 		if (endpoints.isEmpty()) {
 			// servers should bind to the configured port (while clients should use an ephemeral port through the default endpoint)
 			int port = config.getInt(NetworkConfig.Keys.COAP_PORT);
 			LOGGER.log(Level.INFO, "No endpoints have been defined for server, setting up server endpoint on default port {0}", port);
 			addEndpoint(new CoapEndpoint(port, this.config));
 		}
-		
+
 		int started = 0;
-		for (Endpoint ep:endpoints) {
+		for (Endpoint ep : endpoints) {
 			try {
 				ep.start();
 				// only reached on success
@@ -204,7 +204,7 @@ public class CoapServer implements ServerInterface {
 				LOGGER.log(Level.SEVERE, "Cannot start server endpoint [" + ep.getAddress() + "]", e);
 			}
 		}
-		if (started==0) {
+		if (started == 0) {
 			throw new IllegalStateException("None of the server endpoints could be started");
 		}
 	}
@@ -249,7 +249,7 @@ public class CoapServer implements ServerInterface {
 			endpoint.setMessageDeliverer(deliverer);
 		}
 	}
-	
+
 	/**
 	 * Gets the message deliverer.
 	 *
@@ -274,7 +274,7 @@ public class CoapServer implements ServerInterface {
 		endpoint.setExecutor(executor);
 		endpoints.add(endpoint);
 	}
-	
+
 	/**
 	 * Gets the list of endpoints this server is connected to.
 	 *
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -53,6 +53,9 @@ import org.eclipse.californium.core.network.stack.BlockwiseLayer;
 import org.eclipse.californium.core.network.stack.CoapStack;
 import org.eclipse.californium.core.network.stack.ObserveLayer;
 import org.eclipse.californium.core.network.stack.ReliabilityLayer;
+import org.eclipse.californium.core.observe.InMemoryObservationStore;
+import org.eclipse.californium.core.observe.NotificationListener;
+import org.eclipse.californium.core.observe.ObservationStore;
 import org.eclipse.californium.core.server.MessageDeliverer;
 import org.eclipse.californium.elements.Connector;
 import org.eclipse.californium.elements.CorrelationContext;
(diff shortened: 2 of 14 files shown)


californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
```diff
@@ -135,72 +135,76 @@ public CoapServer(int... ports) {
...
-	
-	public void setExecutor(ScheduledExecutorService executor) {
-		
-		if (this.executor!=null) this.executor.shutdown();
-		
+
+	public void setExecutor(final ScheduledExecutorService executor) {
+
+		if (this.executor != null) {
+			this.executor.shutdown();
```
boaks: Why is the executor shutdown()? 
Is the executor intended to be "exclusive" for californium?
My understanding was, that I can use it to manage the threading of the callback and my own application logic.
If you shutdown(), you may shutdown more the expected.

boaks:
It's hard to go through all changes, because the formating- and final-cleanup "hides the real changes"

sbernard31:
> It's hard to go through all changes, because the formating- and final-cleanup "hides the real changes" 

+1

sophokles73:
@sbernard31 

+1 for @boaks comment or +1 for the changes ;-)?

What I can try to do is split up the change set into *formal* changes and *semantic* changes.
Would that help? I would then separately simply commit the formal corrections & improvements because they probably don't need much review or discussion.

sophokles73:
Ok, I have separated the code style changes (already on master) from the semantic changes (this PR).

sbernard31:
I didn't take so much time to look at the PR because I firstly want to know if your way could really handle the observe cancellation. I know that from the beginning  you want to keep the initial exchange in memory on the node we initiate the observation. And from my point of view, this can not work. (I will be glad if I'm wrong because this will really simplify the code)

I try to explain this point several time without success ^^, so I pushed a test which show the problem.

sophokles73:
@sbernard31,

I think I do not understand the scenario that the test is verifying the behavior for. My understanding is the following:
1. Client 1 establishes an observation and receives some notifications for a while.
1. Then client 1 crashes and the server fails over to client 2, i.e. now sends notifications to client 2 only.
1. Client 2 cancels the observation (we currently only remove the local state from the Matcher but do not also send the corresponding cancel message to the server, which we should, I guess).
1. Server sends another notification and receives an RST from client 2. Consequently, server cancels the observation and does not send notifications anymore.
1. When client 1 recovers (comes back online after a restart) its Matcher state is empty.

Why would client 1 receive any more notifications for the already cancelled observation?
Even if the server did send another notification, client 1 would neither find an exchange for the corresponding token in its hash maps nor in the shared ObservationStore because it has been removed from there as part of client 2 cancelling the observation.

Or am I missing something?

sbernard31:
The use case I talk about is the one @jvermillard  described before I start to work on this.
https://github.com/eclipse/leshan/wiki/Cluster

In this case, client1 and client  2 is up at the same time and can receive notification from the same server.
1. client1 initiate the observe.
2. client 1 receive notifications.
3. client 2 receive notificaitions.
4. observe is cancelled from client 2. (we remove observation from shared store)
5. client 1 receive notifications but the exchange is still in memory (and not in shared store)
==> we accept notifications instead of refuse it by sending a RST.

sophokles73:
Hmm, what I do not understand is why both client 1 & client 2 receive notifications for an observation created by client 1 only. Are the notifications multicast? In the cluster topology Julien documented he explicitly noted that incoming packets are routed based on source ip:port so they should always go to a single node, shouldn't they?

sbernard31:
Because of server is behind a NAT.

sophokles73:
Are you talking about the CoAP server, i.e. the component hosting the resource we are observing?

sbernard31:
Yes (It's confusing I confess :p)

sophokles73:
Then we seem to agree that client 1 & client 2 are (in the LWM2M scenario at least) two leshan nodes.
But why does the device (the CoAP server) being located behind a NAT firewall imply that the notifications are sent to both nodes?

sbernard31:
The IP of the device (CoAP server) could change as it was behind a NAT. So the load balancer which use IP/port for routing will probably not choose the same leshan node (CoAP Client).

sophokles73:
Ok, now I understand what you're talking about. However, this is clearly not a fail-over scenario but instead some kind of *load balancing* scenario from my point of view.
If we want to support that, I guess we will need to sync the local state with the shared state on every notification (regardless of which of the nodes we receive the notification on). In particular, we will need to share the most recent observe count, I guess.

However, if we actively cancel the observation on the server when user code invokes `Endpoint.cancelObservation()` then we probably can solve the problem described by you as well, can't we? In that case the CoAP server would no longer send any notifications (regardless of whether they end up at client 1 or client 2).

sbernard31:
This is the use case we want to support at Leshan side.

About notification orderer, I agree (I already mention it [here](https://github.com/eclipse/californium/pull/16#commitcomment-16505975)).

About proactive cancel, I don't think this solve the problem. You're right that if the server is well-coded, it should not send notification anymore. But if we are behind a NAT using reactive cancel seems to make more sense. The other issue is that using the proactive cancel will not clean the maps of client1.

sophokles73:
> About proactive cancel, I don't think this solve the problem. You're right that if the server is well-coded, it should not send notification anymore. But if we are behind a NAT using reactive cancel seems to make more sense. The other issue is that using the proactive cancel will not clean the maps of client1.

You're right, we cannot assume the proactive cancel will reach the CoAP Server (it might have gotten a new IP address already).

I had an idea yesterday evening, though. Let me play around a little with the code ...

sophokles73:
@boaks, @sbernard31 

I have pushed some changes that should solve the problem of detecting stale notifications ...

sbernard31:
This solve partially the problem. If Client 1 receives a notification, it checks if there is pending observation, if not we clean the map. (We should clean more than that if we choose this solution, see `Matcher.ExchangeObserverImpl.completed`).
But this also means that if client 1 never receives a notification the map will never be cleaned...
So yes, we can implement a periodic task which will do some cleaning ... but if we start to implements this kind of thing I'm not sure this will be finally less impacting than the solution I proprose which have the advantage to have the same behavior on all nodes.

sophokles73:
I see your point. However, 
> But this also means that if client 1 never receives a notification the map will never be cleaned...

has made me wonder whether the overall idea of *managing* observations at the `Endpoint` layer makes any sense at all.
@mkovatsc has also pointed out on the mailing list that in the original *CoAP spirit* the client would have to determine whether something is wrong with its observation and, when in doubt, simply re-register or do whatever it seems fit.
Translating this to our problem domain it seems that the `CoapClient` already is implemented in this spirit, i.e. it is a component *outside* of the `Endpoint`. The `Endpoint`'s only purpose is to forward notifications to client code like `CoapClient`. In case of LWM2M leshan would be in a similar role and it should therefore probably also be leshan's responsibility to *manage* observation state (which it already does at the LWM2M level anyways).
All of this seems to point into the direction of the approach taken by @sbernard31 in the `observe_clustering` branch, trying to reduce the state managed by `Endpoint` to an absolute minimum. I will have to think about this over the weekend, I guess :-)
I still don't like the `NotificationListenerAdapter` per observation but maybe I'll simply have to live with that ...

mkovatsc:
I am slowly getting up to speed here.

There should be no problem at the server side: It either receives a cancellation or it detects orphans through its intermittend CON notifications.

On the client side, we have this load balancer. Since notifications can span across a change of the server IP address, the notification might end up at a different client node:

With a central ObserveStore, the Endpoints can check if there is an ongoing Observe relation.
We still have a local entry in the Matcher state, so it easily works also without the central ObserveStore.

The actual problem is, when one of the client nodes cancels, there is no good way to get rid of all the local Matcher state. Due to the IP address changes the server might not receive proactive cancellations and due to the local state, all other client nodes still accept notifications. So you cannot get rid of Observe relations...

Please let me know if I missed something. (I will put thoughts on the architecture/solutions in a different comment.)

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -145,33 +159,75 @@ public synchronized void setExecutor(final ScheduledExecutorService executor) {
...
+	}
 
-		exchange.setObserver(exchangeObserver);
-		LOGGER.log(Level.FINE, "Tracking open request using {0}, {1}", new Object[]{idByMID, idByToken});
+	private void addToObservationStore(final Exchange exchange, final Request request) {
+		boolean isObserveRequest = request.getOptions().hasObserve() && request.getOptions().getObserve() == 0;
+		boolean containsEarlyNegotiationParams = request.getOptions().hasBlock2() && request.getOptions().getBlock2().getNum() == 0 && !request.getOptions().getBlock2().isM();
+		// for observe request.
+		// We ignore blockwise request, except when this is an early negotiation (num and M is set to 0)  
+		if (isObserveRequest || containsEarlyNegotiationParams) {
```
mkovatsc: Why does it go into the ObservationStore when it containsEarlyNegotiationParams?

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -465,8 +519,28 @@ public void sendEmptyMessage(final Exchange exchange, final EmptyMessage message
...
+	public void setMessageDeliverer(final MessageDeliverer deliverer) {
+		if (deliverer == null) {
+			throw new NullPointerException("Message deliverer must not be null");
+		} else {
+			coapstack.setDeliverer(new MessageDeliverer() {
+
+				@Override
+				public void deliverResponse(final Exchange exchange, final Response response) {
+					deliverer.deliverResponse(exchange, response);
+					if (response.isNotification()) {
```
mkovatsc: Why send it to both the local Response handler and the registered NotificationListener?

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -145,33 +159,75 @@ public synchronized void setExecutor(final ScheduledExecutorService executor) {
...
+		// request MID is from the local name space -- use blank address
+		final KeyMID idByMID = assertOutboundRequestHasMessageId(request);
+
+		// ensure Token is set
+		final KeyToken idByToken = assertOutboundRequestHasToken(exchange, request);
+
+		exchange.setObserver(exchangeObserver);
+		exchangesByMID.put(idByMID, exchange);
+		exchangesByToken.put(idByToken, exchange);
+		addToObservationStore(exchange, request);
```
mkovatsc: Why put it n both the ByToken hashmap and the ObservationStore?
In cluster mode, the some coordinated entitiy must make sure if the observe relation registration was successful and then order all incoming notifications and only accept the latest.
So why both paths?

mkovatsc: In cluster mode, we need this coordinated NotificationOrderer.
When not in cluster mode, the current NotificationOrderer should have the same interface as the cluster NotificationOrderer and be called at the same point.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -465,8 +519,28 @@ public void sendEmptyMessage(final Exchange exchange, final EmptyMessage message
...
+						for (NotificationListener listener : notificationListeners) {
```
mkovatsc: I do not understand why we should add multiple NotifiactionListeners...

mkovatsc: If this is related to Simon's "to be able to get notification from any Californium instance", should these listeners represent each other node in the cluster? And each listener has acommunication adapter to reach the other node (in different container, VM, machine, ...)?

I would have expected something analogous to the ClusterObservationStore that takes over a consistent, ordered reception of notifications. It would basically be a distributed NotificationOrderer that is coordinated over all custer nodes. (See other comment on how it should be quite similar to the current NotificationOrderer.)

boaks:
Is this PR still proposed? Or is it OK to close it?

sophokles73:
I think we should close it, maybe we can add another label `for reference` to it, because I think it might be worthwhile to take a look at the exchanged arguments at some time?

boaks:
"label and close" or "label and keep"?

sophokles73:
label and close :-)

PR #48, opened 2016-05-20
matches() method in LinkFormat, cf-core fixed

maradrian:
The matches method was returning true for any attribute that was
part of the querry instead of checking for all the attributes that
were in the querry.

Signed-off-by: Adrian Mardari mardari.adri@gmail.com

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
@@ -164,64 +164,71 @@ public class LinkFormat {
 	}
 	
 	public static boolean matches(Resource resource, List<String> queries) {
-		
-		if (resource==null) return false;
-		if (queries==null || queries.size()==0) return true;
-		
+
+		if (resource == null)
+			return false;
+		if (queries == null || queries.size() == 0)
+			return true;
+
 		ResourceAttributes attributes = resource.getAttributes();
-		String path = resource.getPath()+resource.getName();
-		
+		String path = resource.getPath() + resource.getName();
+
 		for (String s : queries) {
 			int delim = s.indexOf("=");
 			if (delim != -1) {
-				
+
 				// split name-value-pair
 				String attrName = s.substring(0, delim);
-				String expected = s.substring(delim+1);
-
+				String expected = s.substring(delim + 1);
 				if (attrName.equals(LinkFormat.LINK)) {
 					if (expected.endsWith("*")) {
-						return path.startsWith(expected.substring(0, expected.length()-1));
+						if (!path.startsWith(expected.substring(0, expected.length() - 1)))
+							return false;
 					} else {
-						return path.equals(expected);
+						if (!path.equals(expected))
+							return false;
 					}
 				} else if (attributes.containsAttribute(attrName)) {
 					// lookup attribute value
 					for (String actual : attributes.getAttributeValues(attrName)) {
-					
+
 						// get prefix length according to "*"
 						int prefixLength = expected.indexOf('*');
 						if (prefixLength >= 0 && prefixLength < actual.length()) {
-					
+
 							// reduce to prefixes
 							expected = expected.substring(0, prefixLength);
 							actual = actual.substring(0, prefixLength);
 						}
-						
+
 						// handle case like rt=[Type1 Type2]
-						if (actual.indexOf(" ") > -1) { // if contains white space
+						if (actual.indexOf(" ") > -1) { // if contains white
+														// space
 							String[] parts = actual.split(" ");
-							for (String part : parts) { // check each part for match
-								if (part.equals(expected)) {
-									return true;
+							for (String part : parts) { // check each part for
+														// match
+								if (!part.equals(expected)) {
+									return false;
 								}
 							}
 						}
-						
+
 						// compare strings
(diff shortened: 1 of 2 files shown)


sophokles73:
Hi @maradrian,

thanks for the PR :-) Could you also add a unit test illustrating (and of course verifying) the issue?

Kai

californium-core/src/test/java/org/eclipse/californium/core/test/LinkFormatTest.java
```diff
@@ -0,0 +1,334 @@
+package org.eclipse.californium.core.test;
```
sophokles73: Please add the eclipse license header (simply copy from other file and update the year and the organization you work for.

californium-core/src/test/java/org/eclipse/californium/core/test/LinkFormatTest.java
```diff
@@ -0,0 +1,334 @@
...
+		System.out.println("Result for request: /.well-known/core?if=sensor  with fixed matches() method:");
+		System.out.println(buffer.toString() + "\n");
+		Assert.assertEquals(
+				"</sensors/light>;if=\"sensor\";rt=\"light-lux\""
+						+ ",</sensors/temp>;bar=\"one two\";foo;if=\"sensor\";rt=\"temperature-c\",",
+				buffer.toString());
+	}
+
+	@Test
+	public void testMatchesOld() {
```
sophokles73: I think I have confused you when I said "test case illustrating the problem". I didn't mean that we need to have a test case for the old implementation. My expectation is that the `testMatchesFixed()` test case would fail when run against the _old_ revision of `LinkFormat`. However, for the sake of clarity and simplicity, you should remove the code based on the _old_ matching code... Sorry again for being unclear about my intentions.

californium-core/src/test/java/org/eclipse/californium/core/test/LinkFormatTest.java
```diff
@@ -0,0 +1,334 @@
...
+		request1.setURI("/.well-known/core?if=sensor");
+		request2 = new Request(Code.GET);
+		request2.setURI("/.well-known/core?if=sensor&rt=light-lux");
+		request3 = new Request(Code.GET);
+		request3.setURI("/.well-known/core?if=sensor&rt=angle");
+
+	}
+
+	@Test
+	public void testMatchesFixed() {
```
sophokles73: Can you find a more descriptive name indicating what is to be tested/asserted, e.g. something like `testMatchesCorrectlyRecognizesBlaBlaBla...`? Alternatively, please add a JavaDoc shortly describing what the test validates/asserts. This makes it much easier for others to a) understand what the purpose of the test is and b) refactor the code if necessary when introducing changes in other places.

sophokles73:
Hi @maradrian,

sorry for getting back to you so late but I have been buried in work on other OSS projects I am involved in. I have made some comments on the code and in addition to that it would be very helpful if you could rebase on current master and squash all your changes into a single commit once you are done with incorporating any outstanding fixes. That would make merging it much easier ...

Thanks for your patience and contribution :-)

californium-core/src/test/java/org/eclipse/californium/core/test/LinkFormatTest.java
```diff
@@ -0,0 +1,334 @@
...
+
+	/**
+	 * Matches method before fix
+	 * 
+	 * @param resource
+	 * @param queries
+	 * @return
+	 */
+	public static boolean matchesOld(Resource resource, List<String> queries) {
+
```
sophokles73: Again, I might have confused you but I didn't mean to reproduce the actual matcher implementation here in the test case. In particular not the old (erroneous) one...

californium-core/src/test/java/org/eclipse/californium/core/test/LinkFormatTest.java
```diff
@@ -0,0 +1,334 @@
...
+	/**
+	 * Fixed matches method
+	 * 
+	 * @param resource
+	 * @param queries
+	 * @return
+	 */
+	private static boolean matchesFixed(Resource resource, List<String> queries) {
+
```
sophokles73: use `LinkFormat.matches()` instead..

californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
```diff
@@ -164,64 +164,71 @@ public static StringBuilder serializeAttribute(String key, List<String> values)
...
-						return path.startsWith(expected.substring(0, expected.length()-1));
+						if (!path.startsWith(expected.substring(0, expected.length() - 1)))
+							return false;
 					} else {
-						return path.equals(expected);
+						if (!path.equals(expected))
+							return false;
 					}
 				} else if (attributes.containsAttribute(attrName)) {
 					// lookup attribute value
```
sophokles73: I think the logic for checking whether a resource has a particular matching attribute should be moved to class `ResourceAttributes` instead in order to increase cohesion and information encapsulation. I think we should have something along the lines of  `ResourceAttributes.hasMatchingAttribute(String attribute, String value)`...

sophokles73:
@maradrian,

are you still working on this? Otherwise I would like to close the PR ...

sophokles73:
No activity any more.

PR #61, opened 2016-06-23
Ok, this is an almost complete stab at implementing CoAP-over-TCP RFC…

joemag1:
Ok, this is an almost complete stab at implementing CoAP-over-TCP RFC. Things worth noting about this PR:

Made an attempt to keep client interface fully backwards compatible. This means users of CoapServer, CoapClient and Request/Response objects should not care whether protocol runs over TCP or UDP. The only downside is that it allows client ability to send ACKs and RSTs, which get silently dropped by the TCP stack. Given that you need to opt-in into using TCP stack, I think that's a reasonable trade off.

Now to the actual implementation. Introduced interfaces for Matcher, DataSerializer, DataParser and CoapStack and created split implementations UDP and TCP (if anybody has better ideas for names I'm all ears).

On the connector side, implemented Tcp client and server connectors. Note, that they are asymmetrical - which I guess is expected given the nature of TCP. I don't think this violates any promises of Californium, as one already needs to instantiate CoapServer and CoapClient separately.

Went with Netty implementation, just like in Simon's branch. But used later version of Netty as it gave support for connection pooling, and tuned some parameters based on end to end tests.

Head to ThroughputClient.java and ThroughputServer.java for examples of how to use TCP connectors. One pleasant side effect of moving to the new TCP spec is getting rid of 16 bit MIDs, which made a big impact on throughput. This was a major factor for us (AWS).

Lastly, no TLS implementation at this point. If this looks like a reasonable approach should be trivial to add that.

Signed-off-by: Joe Magerramov <joe.magerramov@gmail.com>


element-connector/src/main/java/org/eclipse/californium/elements/tcp/CloseOnErrorHandler.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Channel handler that closes connection if an idle event was raised.
+ */
+class CloseOnErrorHandler extends ChannelHandlerAdapter {
+
+    private final static Logger LOGGER = Logger.getLogger(UDPConnector.class.getName());
```
sophokles73: Sure this should log on `UDPConnector`'s logger?

element-connector/src/main/java/org/eclipse/californium/elements/tcp/CloseOnIdleHandler.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Channel handler that closes connection if an idle event was raised.
+ */
+class CloseOnIdleHandler extends ChannelDuplexHandler {
+
+    private final static Logger LOGGER = Logger.getLogger(UDPConnector.class.getName());
```
sophokles73: correct logger?

joemag1: Fixed.

sophokles73:
Hi Joe,

I am currently looking at your PR and it seems that you have broken the IP validation checks by using your amazon email address for the fixes you merged to the PR (originally signed-off with your gmail address).

sophokles73:
Joe,

I think this looks very good for a start :-) However, I haven't been able to execute the existing test cases with you modifications, i.e. a `mvn clean install` fails because of multiple tests in californium-core fail like this when sending an EmptyMessage:
```
51 SEVERE [CoapEndpoint$5]: Exception in protocol stage thread: No Token has been set, not even an empty byte[0] - (org.eclipse.californium.core.network.CoapEndpoint$5.java:776) run() in thread CoapEndpoint-0.0.0.0/0.0.0.0:0#1 at (2016-06-28 14:59:31)
java.lang.NullPointerException: No Token has been set, not even an empty byte[0]
	at org.eclipse.californium.core.network.serialization.UdpDataSerializer.serializeMessage(UdpDataSerializer.java:71)
	at org.eclipse.californium.core.network.serialization.UdpDataSerializer.serializeEmptyMessage(UdpDataSerializer.java:65)
	at org.eclipse.californium.core.network.serialization.Serializer.serialize(Serializer.java:121)
	at org.eclipse.californium.core.network.CoapEndpoint$OutboxImpl.sendEmptyMessage(CoapEndpoint.java:588)
	at org.eclipse.californium.core.network.stack.CoapUdpStack$StackBottomAdapter.sendEmptyMessage(CoapUdpStack.java:231)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendEmptyMessage(AbstractLayer.java:81)
	at org.eclipse.californium.core.network.stack.ReliabilityLayer.receiveResponse(ReliabilityLayer.java:259)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveResponse(AbstractLayer.java:101)
	at org.eclipse.californium.core.network.stack.CoapUdpStack.receiveResponse(CoapUdpStack.java:143)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl.receiveMessage(CoapEndpoint.java:707)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl.access$800(CoapEndpoint.java:607)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl$1.run(CoapEndpoint.java:621)
	at org.eclipse.californium.core.network.CoapEndpoint$5.run(CoapEndpoint.java:774)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

Any idea why that is?

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -216,8 +229,18 @@ public CoapEndpoint(final InetSocketAddress address, final NetworkConfig config)
...
-		this.matcher = new Matcher(config);		
-		this.coapstack = new CoapStack(config, new OutboxImpl());
+
+		if (connector.isTcp()) {
+			this.matcher = new TcpMatcher(config);
+			this.coapstack = new CoapTcpStack(config, new OutboxImpl());
+			this.serializer = new Serializer(new TcpDataSerializer());
+			this.parser = new TcpDataParser();
+		} else {
+			this.matcher = new TcpMatcher(config);
```
sophokles73: should be `new UdpMatcher(config)` ...

joemag1: Thanks Kai. I think you caught it - the tests likely failed because I fat
fingered "TcpMather" in CoapEndpoint.

I'll update and rerun.

sophokles73:
The wrong Matcher seems to be the cause for the failing tests. However, the NullPointExceptions seem to be thrown because you check for a token in all outgoing `Message`s and `EmptyMessage`s so far do not contain a token (not even a zero-length token) which seems to be in violation of the CoAP spec.

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
```diff
@@ -0,0 +1,180 @@
...
+    @Override
+    public void destroy() {
+        stop();
+    }
+
+    @Override
+    public void send(RawData msg) {
+        Channel channel = activeChannels.get(msg.getInetSocketAddress());
+        if (channel == null) {
+            // TODO: Is it worth allowing opening a new connection when in server mode?
```
sophokles73: Is it a problem to allow for opening an outbound connection? ;-)
I have to admit though that the obvious use cases (currently) should work with this kind of separation into `TcpServerConnector` and `TcpClientConnector`. In the DTLS case, however, the `DTLSConnector` is able to initiate a DTLS handshake no matter whether it is on the client or server side. 
The separation also seems to limit the user code to the role chosen by configuring either sevrer or client connector, doesn't it? The `CoapServer` allows for registering multiple `Endpoint`s. The one to use for communicating with a peer is currently chosen based on whether the remote URI is a non-secure or secure URI, i.e. we currently only differentiate between secure and non-secure endpoints. In order to be able to use UDP and TCP endpoints in parallel, we would need to introduce a way to pick the endpoint based on the URI scheme in general, i.e. a request with a URI with the `coap+tcp` scheme should be routed to a TCP endpoint whereas a simple `coaps` request should be routed to a UDP/DTLS endpoint.

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
```diff
@@ -0,0 +1,196 @@
...
+import io.netty.channel.pool.FixedChannelPool;
+import io.netty.channel.socket.nio.NioSocketChannel;
+import io.netty.handler.timeout.IdleStateEvent;
+import io.netty.handler.timeout.IdleStateHandler;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.GenericFutureListener;
+import org.eclipse.californium.elements.Connector;
+import org.eclipse.californium.elements.RawData;
+import org.eclipse.californium.elements.RawDataChannel;
+import org.eclipse.californium.elements.UDPConnector;
```
sophokles73: unused import

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
```diff
@@ -0,0 +1,196 @@
...
+import io.netty.channel.ChannelOption;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.pool.AbstractChannelPoolHandler;
+import io.netty.channel.pool.AbstractChannelPoolMap;
+import io.netty.channel.pool.ChannelPool;
+import io.netty.channel.pool.ChannelPoolMap;
+import io.netty.channel.pool.FixedChannelPool;
+import io.netty.channel.socket.nio.NioSocketChannel;
+import io.netty.handler.timeout.IdleStateEvent;
```
sophokles73: unused import

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
```diff
@@ -0,0 +1,180 @@
...
+import io.netty.channel.ChannelOption;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.socket.SocketChannel;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.netty.handler.timeout.IdleStateHandler;
+import org.eclipse.californium.elements.Connector;
+import org.eclipse.californium.elements.RawData;
+import org.eclipse.californium.elements.RawDataChannel;
+import org.eclipse.californium.elements.UDPConnector;
```
sophokles73: unused import

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/UdpDataSerializer.java
```diff
@@ -0,0 +1,142 @@
...
+	@Override
+	public byte[] serializeEmptyMessage(Message message) {
+		DatagramWriter writer = new DatagramWriter();
+		serializeMessage(writer, message, 0);
+		return writer.toByteArray();
+	}
+
+	private void serializeMessage(final DatagramWriter writer, final Message message, final int code) {
+		if (message.getToken() == null)
+			throw new NullPointerException("No Token has been set, not even an empty byte[0]");
```
sophokles73: We currently do not explicitly initialize the token in `Message`s to byte[0], thus this check accounts for many NullPointerException when running the tests. We should probably change this but until then you should better assume a zero-length token when `Message.getToken()` returns `null` ...

joemag1: The null token should be a violation of the spec, but I'l make
serializers/parsers more defensive in dealing with them.

On Server/Client split: we could combine server/client connectors into a
single implementation that opens outbound connections if needed. We'll need
to be careful to enforce the "responses can only go out over the same
connection that received the request" part of the spec, but I think that's
doable. There might be some other minor complications, but I don't see
anything major.

About the only downside I see is that simple client applications may not
desire to open a listening TCP socket (security considerations and such).
Like you say it's already the case with UDP/DTLS, but there we don't have a
choice.

One alternative is to keep TcpClientConnector as is (only opens outbound
connection), while TcpServerConnector will do both. CoapClient already
creates the right endpoint/connector based on URI scheme. What do you think?

sophokles73: Well, the null token itself is not a violation itself, I guess, since it still results in a zero length token being serialized on the wire (in current Californium). However, we might be violating the requirement to echo the request's token in ACK and RST messages ...

Server/Client split: I like the idea of having a "client only" and a "server that can do both" connector :-) But we can probably introduce that at a later point and for now stick with the existing client and server.

joemag1: Can you take another look. The tests are passing for me now - I think using
TCPMatcher instead of UDP was the root cause of all errors. TCPMatcher
wasn't setting token on empty messages (since it never expects to see
them). I still added one for good measure.

I haven't fixed signed-off footer yet. If PR looks good otherwise I can
repost a new PR with correct signed-by footers.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -137,6 +144,12 @@
 	/** The matcher which matches incoming responses, akcs and rsts an exchange */
 	private final Matcher matcher;
 
+	/** Serializer to convert messages to datagrams. */
+	private final Serializer serializer;
```
sophokles73: Can we make UdpSerializer and TcpSerializer static utility classes? Or do they need to keep some state?

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -120,7 +138,42 @@ private synchronized void createDefaultEndpoint() {
...
+		tcp_endpoint = new CoapEndpoint(connector, config);
+		try {
+			tcp_endpoint.start();
+			LOGGER.log(Level.INFO, "Created implicit tcp endpoint {0}", tcp_endpoint.getAddress());
+		} catch (IOException e) {
+			LOGGER.log(Level.SEVERE, "Could not create tcp endpoint", e);
+		}
+	}
+
+	public void setTcpEndpiont(Endpoint endpoint) {
```
sophokles73: typo: setTcpEndpoint

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -102,6 +108,18 @@ public Endpoint getDefaultEndpoint() {
 		return default_endpoint;
 	}
 
+	/**
+	 * Gets the default tcp endping for implicit use by client. By default, the tcp endpoint has single worker
+	 * thread, and uses default TCP settings.
+	 * Be careful to stop default tcp endpoing, as it stops all messages sent over it.
+     */
+	public Endpoint getTcpEndpoint() {
```
sophokles73: I guess this will need to be synchronized as well ...

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -120,7 +138,42 @@ private synchronized void createDefaultEndpoint() {
...
+		try {
+			tcp_endpoint.start();
+			LOGGER.log(Level.INFO, "Created implicit tcp endpoint {0}", tcp_endpoint.getAddress());
+		} catch (IOException e) {
+			LOGGER.log(Level.SEVERE, "Could not create tcp endpoint", e);
+		}
+	}
+
+	public void setTcpEndpiont(Endpoint endpoint) {
+		if (this.tcp_endpoint!=null) {
```
sophokles73: add spaces around comparator

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataParser.java
```diff
@@ -1,222 +1,114 @@
...
-			throw new MessageFormatException("Message contains illegal option delta/length: " + delta);
-		}
-	}
+public abstract class DataParser {
+
+    /** Generate RST for a specified input. */
+    public abstract EmptyMessage generateRst(RawData rawData);
+
+    /** Parses message based on the raw data. */
+    public Request parseRequest(RawData rawData, int code) {
```
sophokles73: shouldn't the code come from the RawData?
also, can you make the params and methods that are not to be overridden by subclasses final?

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/Serializer.java
```diff
@@ -34,7 +35,14 @@
  */
 public class Serializer {
```
sophokles73: I don't really see, why we would need the Serializer class at all, do you?

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/TcpDataSerializer.java
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	@Override
+	public byte[] serializeEmptyMessage(Message message) {
+		DatagramWriter writer = new DatagramWriter();
+		serializeMessage(writer, message, 0);
+		return writer.toByteArray();
+	}
+
+	private void serializeMessage(final DatagramWriter writer, final Message message, final int code) {
```
sophokles73: I think we should use Code as parm type instead of int for clarity.

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/TcpDataSerializer.java
```diff
@@ -0,0 +1,166 @@
...
+ * The DataSerialized serializes outgoing messages to byte arrays based on CoAP TCP/TLS spec:
+ * <a href="https://tools.ietf.org/html/draft-ietf-core-coap-tcp-tls"/>
+ */
+public final class TcpDataSerializer implements DataSerializer {
+
+	@Override
+	public byte[] serializeRequest(Request request) {
+		DatagramWriter writer = new DatagramWriter();
+		Code code = request.getCode();
+		serializeMessage(writer, request, code == null ? 0 : code.value);
```
sophokles73: I know that Request allows for instantiation with null Code but I have yet to see why that would be necessary ... do you have any idea?

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/TcpDataSerializer.java
```diff
@@ -0,0 +1,166 @@
...
+	@Override
+	public byte[] serializeEmptyMessage(Message message) {
+		DatagramWriter writer = new DatagramWriter();
+		serializeMessage(writer, message, 0);
+		return writer.toByteArray();
+	}
+
+	private void serializeMessage(final DatagramWriter writer, final Message message, final int code) {
+		if (message.getToken() == null)
+			throw new NullPointerException("No Token has been set, not even an empty byte[0]");
```
sophokles73: You have already taken care that the token is always set (to at least byte[0]), haven't you?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapTcpStack.java
```diff
@@ -0,0 +1,236 @@
...
+			exchange.setResponse(response);
+			super.sendResponse(exchange, response);
+		}
+		
+		@Override
+		public void receiveRequest(Exchange exchange, Request request) {
+			// if there is no BlockwiseLayer we still have to set it
+			if (exchange.getRequest() == null)
+				exchange.setRequest(request);
+			if (deliverer != null) {
```
sophokles73: if (!hasDeliverer()) ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapTcpStack.java
```diff
@@ -0,0 +1,236 @@
...
+			} else {
+				LOGGER.severe("Top of CoAP stack has no deliverer to deliver request");
+			}
+		}
+
+		@Override
+		public void receiveResponse(Exchange exchange, Response response) {
+			if (!response.getOptions().hasObserve())
+				exchange.setComplete();
+			if (deliverer != null) {
```
sophokles73: hasDeliverer() ?

element-connector/pom.xml
```diff
@@ -21,6 +21,11 @@
 			<artifactId>junit</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>io.netty</groupId>
+			<artifactId>netty-all</artifactId>
+			<version>4.0.28.Final</version>
```
sophokles73: Is there a particular requirement to use 4.0.28? We should tak care to use a version that has already undergone the Eclipse IP review process (which 4.0.28 has), but we can e.g. also use a more recent version like 4.1.0 or 4.0.36 ...
any preference?

element-connector/src/main/java/org/eclipse/californium/elements/Connector.java
```diff
@@ -91,4 +91,12 @@
...
 	 */
 	public InetSocketAddress getAddress();
+
+	/**
+	 * Returns true if this is a TCP connector. TCP connector is subject to a different spec (CoAP-over-TCP) and
+	 * will requir CoAP endpoint to instantiate a different matcher, serializer and coap stack.
+	 *
+	 * @return true if this is a TCP based connector.
+     */
+	public boolean isTcp();
```
sophokles73: I'd rather have a method checking support for a particular scheme (e.g. _coap_ or _coap+tcp_). WDYT?

element-connector/src/main/java/org/eclipse/californium/elements/tcp/CloseOnErrorHandler.java
```diff
@@ -0,0 +1,39 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Institute for Pervasive Computing, ETH Zurich and others.
```
sophokles73: In classes you created from scratch you may as well just indicate Amazon as the copyright holder (we at Bosch do, at least) ... but that is up to you, I guess.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
joemag1: No state, but using an instance allow only CoapEndpoint constructor to care about TPC vs UDP, as it instantiates the right serializer and parser. Otherwise every reference to serializer and parser needs to be wrapped by an if.

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
joemag1: Ack.

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
joemag1: Ack.

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataParser.java
joemag1: There's a bit of circular dependency between parsing the messages and knowing what type it is. Agreed, done this way is a bit counterintuitive.

I think the simpler way would be to not try and have common parse(Request|Response) methods on DataParser, and instead having (Udp|TCP)Parsers implement those directly, since they know how to extract code out of the message. A tiny bit of code duplication, but it's easier to understand.

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/Serializer.java
joemag1: No I don't see a good reason either - it seems to have some convenience wrappers, but nothing CoApEndpoint can't do itself. I was trying to minimize changes outside those needed to support TCP, but happy to nuke this class.

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/TcpDataSerializer.java
joemag1: I don't think you can do that, since request and response codes are different enums.

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/TcpDataSerializer.java
joemag1: I'm not even sure that null Code on request won't fail somewhere else. The only argument for keeping this check is for backwards compatible for existing client who may have gotten used to null request == GET. Another way is to have request constructor perform that conversion.

element-connector/pom.xml
joemag1: 4.0.28 is the minimum needed due to added support for connection pools and connection pool maps. Picking a later version sounds reasonable too.

element-connector/src/main/java/org/eclipse/californium/elements/Connector.java
joemag1: Yes, I was actually rather unhappy with isTcp() method. Your proposed approach on connector sounds good to me (and makes DTLS work better).

joemag1:
I'll create an update addressing comments in a little while. Sounds like this is getting close to being merge-worthy, what do you think about creating two fresh pull requests at this point (one for connector, the another one for californium-core)?

sophokles73:
That is a wonderful idea :-)
Thanks for your work on this, Joe. I really appreciate it!!

joemag1:
Ok, here are the split pull requests. Note that the tcp-core is branched off connector, so will contain both commits.  Main changes from this PR is switching Connector.isTcp to isProtocolSupported and refactoring DataParser and DataSerializer to hopefully simplify them even more. I did not undertake refactoring EndpointManager - wanted to keep changes small.

Connector: https://github.com/eclipse/californium/pull/66
Stack: https://github.com/eclipse/californium/pull/67

PR #66, opened 2016-07-05
Implementation of TCP connector for Californium per CoAP over TCP RFC.

joemag1:
Uses Netty 4.1 for non-blocking IO. Implementation splits into Server and Client TCP connector, with expectation that the client connector does not listen on a port. Still todo:

    1) Add TLS support to both client and server connector.
    2) Add call back from the TcpConnector when a connection is lost (per RFC all exchanges with that remote endpoint must be closed)
    3) Add ability for TcpServerConnector to open outbound connections. This is desireable when application is reusing the same endpoint between CoapClient and CoapServer

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
@@ -1,0 +1,187 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Amazon Web Services.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *    Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
+ ******************************************************************************/
+package org.eclipse.californium.elements.tcp;
+
+import io.netty.bootstrap.Bootstrap;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelOption;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.channel.pool.AbstractChannelPoolHandler;
+import io.netty.channel.pool.AbstractChannelPoolMap;
+import io.netty.channel.pool.ChannelPool;
+import io.netty.channel.pool.FixedChannelPool;
+import io.netty.channel.socket.nio.NioSocketChannel;
+import io.netty.handler.timeout.IdleStateHandler;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.GenericFutureListener;
+import org.eclipse.californium.elements.Connector;
+import org.eclipse.californium.elements.RawData;
+import org.eclipse.californium.elements.RawDataChannel;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * TCP client connection is used by CoapEndpoint when instantiated by the CoapClient. Per RFC the client can both
+ * send and receive messages, but cannot accept new incoming connections.
+ */
+public class TcpClientConnector implements Connector {
+
+    private final static Logger LOGGER = Logger.getLogger(TcpClientConnector.class.getName());
+
+    private final int numberOfThreads;
+    private final int connectionIdleTimeoutSeconds;
+    private final int connectTimeoutMillis;
+    private EventLoopGroup workerGroup;
+    private RawDataChannel rawDataChannel;
+    private AbstractChannelPoolMap<SocketAddress, ChannelPool> poolMap;
+
+    public TcpClientConnector(int numberOfThreads, int connectTimeoutMillis, int idleTimeout) {
+        this.numberOfThreads = numberOfThreads;
+        this.connectionIdleTimeoutSeconds = idleTimeout;
+        this.connectTimeoutMillis = connectTimeoutMillis;
+    }
+
+    @Override
+    public void start() throws IOException {
+        if (rawDataChannel == null) {
+            throw new IllegalStateException("Cannot start without message handler.");
+        }
+
+        if (workerGroup != null) {
+            throw new IllegalStateException("Connector already started");
+        }
+
+        workerGroup = new NioEventLoopGroup(numberOfThreads);
+        poolMap = new AbstractChannelPoolMap<SocketAddress, ChannelPool>() {
+            @Override
+            protected ChannelPool newPool(SocketAddress key) {
+                Bootstrap bootstrap = new Bootstrap()
+                        .group(workerGroup)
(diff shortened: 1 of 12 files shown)


element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
```diff
@@ -0,0 +1,187 @@
...
+                    LOGGER.log(Level.WARNING, "Unable to open connection to " + msg.getAddress(), future.cause());
+                }
+            }
+        });
+    }
+
+    @Override
+    public void setRawDataReceiver(RawDataChannel messageHandler) {
+        if (rawDataChannel != null) {
+            throw new IllegalStateException("Raw data channel already set.");
```
sophokles73: this is a problem only if the connector is already running (started), isn't it?

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
```diff
@@ -0,0 +1,170 @@
...
+            return;
+        }
+
+        channel.writeAndFlush(Unpooled.wrappedBuffer(msg.getBytes()));
+    }
+
+    @Override
+    public void setRawDataReceiver(RawDataChannel messageHandler) {
+        if (rawDataChannel != null) {
+            throw new IllegalStateException("RawDataChannel alrady set");
```
sophokles73: see above

sophokles73:
@joemag1,

this looks good :-)
However, I think we will need to put this into the 2.0.x branch because we are introducing non-backwards compatible  changes to the public `Connector` API (introducing the `isTcp()`/`isProtocolSupported()` method). Client code may provide its own `Connector` implementation and an implementation based on 1.x `Connector` will no longer work with our new `Connector` interface.

This is no problem per se because we actually expected TCP support to break the API. It only means that we cannot merge it to master (which reflects 1.1.x). We have two options:
1) you either rebase the PR on 2.0.x which will require some extra work because we had started to change the `Matcher` to support clustering.
2) we create a new branch from master (e.g. `tcp_support`) and I merge it later to the 2.0.x branch myself.

I think we should go for option 2 in order to get a first version into the code base that we can work with and play around with. We can then later migrate it to 2.0.x.

Another thing: the IP validation fails due to a missing CLA. This is probably because in either your Eclipse account or your GitHub account you have not added the AWS or gmail addresses you used to sign-off.

joemag1:
Yeah, that thought crossed my mind. What kind of timelines were you
thinking for 1.1 and/or 2.0?

We could in theory make a change to connector to make it backwards
compatible (use instanceof instead of isSchemeSupported in endpoint), which
could make it safe to merge into master. We could even use instanceof on
master, and proper isSchemeSupported on 2.0 branch.

Otherwise, I'd recommend going with option 1) - waiting on the merge will
likely make it more painful down the road.

sophokles73:
frankly, we haven't discussed any timelines for some time now but I would like to use this major contribution as a trigger for having one :-)
Do you have any plans/expectations regarding when you want to see this in a release?
If we were able to put this into 1.1 in a non-breaking way then we should  be able to roll out a 1.1 release within some weeks. The "paperwork" required for doing so shouldn't be too heavy since we haven't changed very much but mainly did bug-fixes.

sophokles73:
FYI: I started a discussion of the roadmap on the mailing list ...

joemag1:
We'd like to start integration testing by end of July, even if it's only a
"beta" release.

We'd likely want something stable before October.

sophokles73:
I see, you guys seem to have an ambitious schedule :-) Then let's do everything it takes to make it happen. The quickest way, I guess would then be to integrate into master trying to remain backwards-compatible. Creating a 1.1.0-M1 milestone can be done pretty quickly. Preparing a release takes some time, though ...
About the IP validation: that is something you'll need to fix ASAP because I cannot accept any code without a CLA in place ...

joemag1:
Thanks for working on this, appreciate your flexibility. I still need to
add TLS support, let me get going on that today or tomorrow.

Here's what we can do then: merge proper change (with interface changes
into 2.0). Merge modified one (backwards compatible into new 1.1 branch).

I signed CLA with joemag@amazon.com - not sure why it's still showing as
failed. Any ideas?

sophokles73:
You might want to make sure that you have added the email address you signed-off with to your GitHub account as well. I think the IP check is implemented as a GitHub webhook which uses GitHug to resolve the user data... Then, you might also want to check, whether you have added your GitHub username to your eclipse profile (https://dev.eclipse.org/site_login/myaccount.php#open_tab_accountsettings)

kartben:
> I signed CLA with joemag@amazon.com - not sure why it's still showing as
failed. Any ideas?

Any chance you hadn't sign the CLA yet, two days ago when you made and pushed the commit?

sophokles73:
In that case, Joe would need to re-create the PR _from scratch_, right?

kartben:
Commit-amend + push force on his branch would do the trick, I think.

sophokles73:
@joemag1,

my understanding is that this is now _the candidate_ for merging into master (1.1.x) and 2.0.x, or do you want to make any more changes before? I think adding TLS support should be done a second step (separate PR). Thanks again for your nice work :-)

joemag1:
Both 1.1 and 2.0 PRs are good to be merged. The only difference between the
two is that 1.1 uses a marker interface rather than extending Connector
interface.

I still have to add TLS support, but I expect that to be fairly
straightforward. I'll try to knock that out as soon as I find an hour or
two in the next few days.

joemag1:
This thread seems to have lost some steam :) what's the next step?

Btw, I added TLS support to the PR. Like promised, it's a fairly small
change.

sophokles73:
@joemag1,

sorry for the delay. I am currently attending a multi-day workshop so I didn't find the time to go on. I will merge the PRs tomorrow morning when I am back in the office.

joemag1:
No worries Kai. Just wasn't sure if you were waiting for me on something.

sophokles73:
Ok, while preparing the merge, I came across some issues that need to be resolved:
* in your code changes you are using spaces for indentation, we are using tabs for indentation, though (we are providing Eclipse code formatter settings in the parent folder that you can use to quickly *tidy up* your code)
* the `org.eclipse.californium.core.network.CoapEndpointTest` in the core module does not compile due to a missing implementation of `Connector.isSchemeSupported()`
* it would be nice if you could re-base your branch on master :-)

joemag1:
Which PR were you merging? The CoAP endpoint test looks correct on my
branch. I prepared two PRs, one with isSchemeSupported for merging into 2.0
branch, and one without for master/1.1.

I will switch to tabs, rebase and resubmit.

sophokles73:
> Which PR were you merging? The CoAP endpoint test looks correct on my
branch. I prepared two PRs, one with isSchemeSupported for merging into 2.0
branch, and one without for master/1.1.

Well, I was trying to merge *THIS* PR (#66) (the TCP Connector to be added to element-connector). Are there other PRs for that?

joemag1:
The original idea of having two PRs one for connector and one for core
turned out to be bad - as you noticed there is inter-dependence between the
two.

So you probably want to merge the "core" PR instead (
https://github.com/eclipse/californium/pull/67), which includes connector
commit in it.

That PR is backwards incompatible (because of connector API changes), so I
also prepared
https://github.com/eclipse/californium/pull/69 to be backwards compatible
for 1.x merge. The only change there is using marker interface instead of
new API.

Let me know if that makes sense. If it does I'll update #67 and #69 with
tabs and resubmit.

sophokles73:
It does make sense to me. However, you should then close *this* PR (#66) ...
So, please go ahead and make the necessary changes in #67 and #69 (don't forget to rebase ;-)) and I will do my best to merge it before the weekend :-)

joemag1:
Ok, re-ran modified files through formatter and rebased on top of master. I
had to recreate "backwards compatible" PR because I was having problems
rebasing it off mainline. So the two PRs are:

#67 <https://github.com/eclipse/californium/pull/67> - for merging into 2.0
branch
#71 <https://github.com/eclipse/californium/pull/71> - for merging into 1.1
branch

-Joe

PR #67, opened 2016-07-05
Changes to californium core to support TCP collector.

joemag1:
List of things that changed:

* Split Matcher into Udp and Tcp Matchers, with the latter not keeping track of exchanges by MID.
* Split Stack into Udp and Tcp stack, with the later not having reliability layer (but keeping blockwise, per TCP RFC recommendation)
* Changed serializers and parsers into instantiable classes to support parsing TCP and UDP payloads.
** Refactored DataParser and DataSerializer a bit to make both symmetric and only delegate parsing of header to individual implementations.
* Introduced Tcp Server and Client bandwidths tests that attempt to send large amount of requests and responses.
* Changes to EndpointManager to allow creating an endpoint with TCP connector whenever coap+tcp scheme is used.

With introduction of connector.isSchemeSupported() the whole endpoint management can likely be improved, but did not want to boil the ocean. Still todo:

* Call back from connector when TCP connection is closed.
* Support for coaps+tcp scheme.


joemag1:
Ok, I split the new pull request with minor tweaks to make it backwards compatible. Then this pull request can be merged into 2.0 feature branch. I still need to add TLS support - I'll try to get that in when I have a few spare hours this week.

sophokles73:
merged to 2.0.x in  b4764f1

PR #71, opened 2016-07-15
TCP RFC implementation (backwards compatible for merging into 1.1 branch)

joemag1:
Based on PR: #67

The only change is using marker interface + instance of instead of cleaner isSchemeSupported method on Connector interface.


sophokles73:
Merged to master in 40c0ca0

Issue #73, opened 2016-07-17
Small possibility of Exchange leak when receiving non-piggybacked responses


joemag1:
Repro steps:

1) Client sends a CON request
2) Server calls exchange.acknowledge(), but then never sends a response, either due to a bug, or because server crashed.
3) The exchange on the client side never times out.

This has two undersired effects:

If client used CoapClient.(put|get|delete|post) variant with CoapHandler, then neither onLoad() nor onError() methods are ever called. Furthermore, the offending exchange is never removed from Matcher's tracking maps.

Potential solution: have one of CoapStack layers add a task that will fire after EXCHANGE_LIFETIME and call Exchange.timeout() if Exchange is still incomplete.

joemag1:
This is a more generic problem of https://github.com/eclipse/californium/issues/72

It just so happen in TCP mode every request is hit by this issue because there are no ACKs in TCP land.

boaks:
@joemag1 

Is this issue still open?

boaks:
If this issue still occurs, don't hesitate to reopen this issue or create a new one.

Issue #72, opened 2016-07-15
CoAP over TCP leaks exchanges


joemag1:
With switching to CoAP over TCP we no longer close outgoing exchanges that never receive any response. This is because in the past reliability layer + deduplicator were performing this function, and TCP stack no longer uses those.

This has two impacts:

1) Clients using async invocation never get called back
2) TcpMatcher will slowly accumulate old exchanges in its map.

To solve, likely need to introduce a TCP version of reliability layer to close expired exchanges.

joemag1:
I will take this one and submit a patch in the next few days.

boaks:
@joemag1 

Is this issue still open?

boaks:
If this issue still occurs, don't hesitate to reopen this issue or create a new one.

PR #74, opened 2016-07-19
Fix Exchange leak if response is never received

joemag1:
Fixes issues: https://github.com/eclipse/californium/issues/73 and https://github.com/eclipse/californium/issues/72

Prior to this fix, if a response to the request (or in UDP stack response for an acknowledged request) never arrived we'd leak Exchange in the Matcher maps. We'd also never call back client's CoapHandler.onError() callback.

This change adds an explicit expiration to CoAP exchanges based on MAX_EXCHANGE_LIFECYCLE parameter.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -764,8 +764,13 @@ public class CoapClient {
 	private CoapResponse synchronous(Request request, Endpoint outEndpoint) {
 		try {
 			Response response = send(request, outEndpoint).waitForResponse(getTimeout());
-			if (response == null) return null;
-			else return new CoapResponse(response);
+			if (response == null) {
+				// Cancel request so appropriate clean up can happen.
+				request.cancel();
+				return null;
+			} else {
+				return new CoapResponse(response);
+			}
 		} catch (InterruptedException e) {
 			throw new RuntimeException(e);
 		}
--- a/californium-core/src/test/java/org/eclipse/californium/core/network/UdpMatcherTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/network/UdpMatcherTest.java
@@ -81,7 +81,7 @@ public class UdpMatcherTest {
 
 	@Test
 	public void testReceiveResponseAcceptsResponseFromDifferentEpochUsingLaxMatching() {
-		// GIVEN a request sent via a DTLS transport using a matcher set to lax matching 
+		// GIVEN a request sent via a DTLS transport using a matcher set to lax matching
 		UdpMatcher matcher = newMatcher(false);
 		Exchange exchange = sendRequest(matcher, new DtlsCorrelationContext(SESSION_ID, EPOCH, CIPHER));
 
@@ -129,7 +129,7 @@ public class UdpMatcherTest {
 
 	@Test
 	public void testReceiveResponseAcceptsResponseFromSameSessionEpochAndCipherUsingStrictMatching() {
-		// GIVEN a request sent via a DTLS transport 
+		// GIVEN a request sent via a DTLS transport
 		UdpMatcher matcher = newMatcher(true);
 		Exchange exchange = sendRequest(matcher, new DtlsCorrelationContext(SESSION_ID, EPOCH, CIPHER));
 
--- a/californium-core/src/test/java/org/eclipse/californium/core/network/stack/CoapStackTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/network/stack/CoapStackTest.java
@@ -1,0 +1,64 @@
+package org.eclipse.californium.core.network.stack;
+
+import org.eclipse.californium.category.Small;
+import org.eclipse.californium.core.coap.CoAP;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.network.Exchange;
+import org.eclipse.californium.core.network.Outbox;
+import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.mockito.ArgumentCaptor;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Executors;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.mock;
+
+@Category(Small.class) @RunWith(Parameterized.class)
+public class CoapStackTest {
+
+	private static final NetworkConfig CONFIG = NetworkConfig.createStandardWithoutFile();
+
+	private final CoapStack stack;
+	private final Outbox outbox;
+
+	public CoapStackTest(CoapStack stack, Outbox outbox) {
+		this.stack = stack;
+		this.stack.setExecutor(Executors.newSingleThreadScheduledExecutor());
+		this.outbox = outbox;
+	}
+
+	@Parameterized.Parameters public static List<Object[]> parameters() {
+		Outbox udpOutbox = mock(Outbox.class);
+		Outbox tcpOutbox = mock(Outbox.class);
+
+		List<Object[]> parameters = new ArrayList<>();
+		parameters.add(new Object[]{new CoapTcpStack(CONFIG, tcpOutbox), tcpOutbox});
+		parameters.add(new Object[]{new CoapUdpStack(CONFIG, udpOutbox), udpOutbox});
+		return parameters;
+	}
+
+	@Test public void cancelledMessageExpectExchangeComplete() {
+		Request request = new Request(CoAP.Code.GET);
+
+		ArgumentCaptor<Exchange> exchangeCaptor = ArgumentCaptor.forClass(Exchange.class);
+		doNothing().when(outbox).sendRequest(exchangeCaptor.capture(), eq(request));
+
+		stack.sendRequest(request);
+
+		// Capture exchange
+		Exchange exchange = exchangeCaptor.getValue();
+		assertFalse(exchange.isComplete());
+
+		request.setCanceled(true);
+		assertTrue(exchange.isComplete());
+	}
+}
--- a/californium-core/pom.xml
+++ b/californium-core/pom.xml
@@ -23,6 +23,12 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-all</artifactId>
+			<version>1.9.5</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>${project.groupId}</groupId>
 			<artifactId>element-connector</artifactId>
 			<type>jar</type>
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -246,7 +246,7 @@ public class UdpMatcher implements Matcher {
 		 * This request could be
 		 *  - Complete origin request => deliver with new exchange
 		 *  - One origin block        => deliver with ongoing exchange
-		 *  - Complete duplicate request or one duplicate block (because client got no ACK) 
+		 *  - Complete duplicate request or one duplicate block (because client got no ACK)
 		 *      =>
 		 * 		if ACK got lost => resend ACK
 		 * 		if ACK+response got lost => resend ACK+response
@@ -301,7 +301,7 @@ public class UdpMatcher implements Matcher {
 				return ongoing;
 
 			} else {
-				// We have no ongoing exchange for that request block. 
+				// We have no ongoing exchange for that request block.
 				/*
 				 * Note the difficulty of the following code: The first message
 				 * of a blockwise transfer might arrive twice due to a
@@ -495,7 +495,7 @@ public class UdpMatcher implements Matcher {
 
 		@Override public void completed(final Exchange exchange) {
 
-			/* 
+			/*
 			 * Logging in this method leads to significant performance loss.
 			 * Uncomment logging code only for debugging purposes.
 			 */
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapTcpStack.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapTcpStack.java
@@ -57,6 +57,8 @@ import java.util.logging.Logger;
  * | +---------v-+-------+ |
  * | | Stack Top         | |
  * | +-------------------+ |
+ * | | {@link ExchangeCleanupLayer}      | |
+ * * | +-------------------+ |
  * | | {@link ObserveLayer}      | |
  * | +-------------------+ |
  * | | {@link BlockwiseLayer}    | |
@@ -91,8 +93,13 @@ public class CoapTcpStack implements CoapStack {
 		this.top = new StackTopAdapter();
 		this.outbox = outbox;
 
-		this.layers = new Layer.TopDownBuilder().add(top).add(new ObserveLayer(config)).add(new BlockwiseLayer(config))
-				.add(bottom = new StackBottomAdapter()).create();
+		this.layers = new Layer.TopDownBuilder()
+				.add(top)
+				.add(new ExchangeCleanupLayer())
+				.add(new ObserveLayer(config))
+				.add(new BlockwiseLayer(config))
+				.add(bottom = new StackBottomAdapter())
+				.create();
 
 		// make sure the endpoint sets a MessageDeliverer
 	}
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
@@ -56,6 +56,8 @@ import java.util.logging.Logger;
  * | +---------v-+-------+ |
  * | | Stack Top         | |
  * | +-------------------+ |
+ * | | {@link ExchangeCleanupLayer} | |
+ * * | +-------------------+ |
  * | | {@link ObserveLayer}      | |
  * | +-------------------+ |
  * | | {@link BlockwiseLayer}    | |
@@ -100,8 +102,15 @@ public class CoapUdpStack implements CoapStack {
 			reliabilityLayer = new ReliabilityLayer(config);
 		}
 
-		this.layers = new Layer.TopDownBuilder().add(top).add(new ObserveLayer(config)).add(new BlockwiseLayer(config))
-				.add(reliabilityLayer).add(bottom = new StackBottomAdapter()).create();
+
+		this.layers = new Layer.TopDownBuilder()
+				.add(top)
+				.add(new ExchangeCleanupLayer())
+				.add(new ObserveLayer(config))
+				.add(new BlockwiseLayer(config))
+				.add(reliabilityLayer)
+				.add(bottom = new StackBottomAdapter())
+				.create();
 
 		// make sure the endpoint sets a MessageDeliverer
 	}
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ExchangeCleanupLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ExchangeCleanupLayer.java
@@ -1,0 +1,31 @@
+package org.eclipse.californium.core.network.stack;
+
+import org.eclipse.californium.core.coap.MessageObserverAdapter;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.network.Exchange;
+
+/**
+ * A layer that reacts to user cancelled outgoing requests, and completes exchange, which causes state clean up.
+ */
+public class ExchangeCleanupLayer extends AbstractLayer {
+
+	@Override public void sendRequest(Exchange exchange, Request request) {
+		request.addMessageObserver(new CancelledMessageObserver(exchange));
+		super.sendRequest(exchange, request);
+	}
+
+	private static class CancelledMessageObserver extends MessageObserverAdapter {
+
+		private final Exchange exchange;
+
+		CancelledMessageObserver(Exchange exchange) {
+			this.exchange = exchange;
+		}
+
+		@Override public void onCancel() {
+			if (!exchange.isComplete()) {
+				exchange.setComplete();
+			}
+		}
+	}
+}
--- a/californium-core/src/test/java/org/eclipse/californium/core/network/TcpMatcherTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/network/TcpMatcherTest.java
@@ -1,0 +1,74 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial creation
+ ******************************************************************************/
+package org.eclipse.californium.core.network;
+
+import org.eclipse.californium.category.Small;
+import org.eclipse.californium.core.coap.CoAP.ResponseCode;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.coap.Response;
+import org.eclipse.californium.core.network.Exchange.Origin;
+import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.elements.CorrelationContext;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+
+import static org.junit.Assert.assertSame;
+
+@Category(Small.class)
+public class TcpMatcherTest {
+
+	private static final InetSocketAddress dest = new InetSocketAddress(InetAddress.getLoopbackAddress(), 5684);
+	private static final InetSocketAddress source = new InetSocketAddress(InetAddress.getLoopbackAddress(), 12000);
+
+	@Test
+	public void testRequestMatchesResponse() {
+		TcpMatcher matcher = newMatcher(false);
+		Exchange exchange = sendRequest(matcher, null);
+
+		Exchange matched = matcher.receiveResponse(responseFor(exchange.getCurrentRequest()), null);
+		assertSame(exchange, matched);
+	}
+
+	private TcpMatcher newMatcher(boolean useStrictMatching) {
+		NetworkConfig config = NetworkConfig.createStandardWithoutFile();
+		config.setBoolean(NetworkConfig.Keys.USE_STRICT_RESPONSE_MATCHING, useStrictMatching);
+		return new TcpMatcher(config);
+	}
+
+	private Exchange sendRequest(final TcpMatcher matcher, final CorrelationContext ctx) {
+		Request request = Request.newGet();
+		request.setDestination(dest.getAddress());
+		request.setDestinationPort(dest.getPort());
+		Exchange exchange = new Exchange(request, Origin.LOCAL);
+		exchange.setRequest(request);
+		matcher.sendRequest(exchange, request);
+		exchange.setCorrelationContext(ctx);
+		return exchange;
+	}
+
+	private Response responseFor(final Request request) {
+		Response response = new Response(ResponseCode.CONTENT);
+		response.setMID(request.getMID());
+		response.setToken(request.getToken());
+		response.setBytes(new byte[]{});
+		response.setSource(source.getAddress());
+		response.setSourcePort(source.getPort());
+		return response;
+	}
+}


sbernard31:
I have doubt about this PR. Here is the UDP CoAP spec about that:
_"while the CoAP protocol itself
      does not make any specific demands here, there is an expectation
      that the response will come within a time frame that is reasonable
      from an application point of view.  As there is no underlying
      transport protocol that could be instructed to run a keep-alive
      mechanism, the requester may want to set up a timeout that is
      unrelated to CoAP's retransmission timers in case the server is
      destroyed or otherwise unable to send the response."_
[source](https://tools.ietf.org/html/rfc7252#section-5.2.2)
As I understand the spec,  this is to the application to handle this.  I suppose by defining a timeout and cancel the request on timed-out.
(For TCP I don't know)

joemag1:
Given the current Californium implementation, does requester even have the
ability to time out requests? If they are using CoapClient API they see
neither Request, nor Exchange.

Regardless client timeout, do you think cleaning up exchanges after
EXCHANGE_LIFETIME is a reasonable interpretation of the spec?

sophokles73:
I tend to agree with Simon that we generally should leave this responsibility with the application layer. However, coming back to Joe's point regarding a client's capability to actually do it: `CoapClient` supports this kind of *timing out* a request already by means of `CoapClient.setTimeout()`.
The value set there could be used by the asynchronous API as well and e.g. cancel the request proactively if no response has been received in time. If a client uses `CoapClient.advanced()` or  `Endpoint.sendRequest()` it is in full control of the Request anyways and can explicitly cancel it after any time it sees fit.

sbernard31:
@joemag1  You're right about CoapClient API. There is no easy way to do that ... The only way I see is using `advanced(Request)` method. (Personal feeling : I rather prefer endpoint/request API for production)

About EXCHANGE_LIFETIME:
  As I understand the spec, an exchange is correlated to message concept and not request/response concept :
[_"The CoAP messaging model is based on the exchange of messages over
   UDP between endpoints. "_](https://tools.ietf.org/html/rfc7252#section-2.1)
 So EXCHANGE_LIFETIME is about messages too (not linked to request/response correlation) :
[_" EXCHANGE_LIFETIME is the time from starting to send a Confirmable
      message to the time when an acknowledgement is no longer expected,
      i.e., message-layer information about the message exchange can be
      purged."_](https://tools.ietf.org/html/rfc7252#section-4.8.2)

 The confusion is probably because Californium use `Exchange` class to correlate request and response too.

sophokles73:
Thanks for the clarification, Simon :-)
I think the correct *time to wait* for a *response* (in particular a response being sent in a separate CON/NON) is highly application/use case dependent (as already pointed out by the spec). The application probably knows best what time frame is reasonable to expect the server being able to produce a result in.

joemag1:
So there are two things here: timing out the request (request.setTimeout)
and cleaning up the exchange (exchange.setTimeout or exchange.setComplete).
The latter is what causes Matcher to clean up internal state.

I think it's reasonable to say request timeout is application concern, and
there are already some mechanisms in the CoAPclient to handle that. I'm our
particular example, we wrap CoapHandler with a future, so we have full
control over timeout.

But I do think exchange cleanup is the framework concern, since it's
garbage collection, rather than interaction concern. So what do you think
if we only cleaned up the exchange after EXCHANGE_LIFETIME, but kept
request alone?

sbernard31:
What do you want to clean exactly ? because in Californium, I believe that `Exchange` is needed for request/response matching.

joemag1:
Matcher state - the exchange sticks around in Matcher maps until it's
complete.

sbernard31:
If you do that, you will lost response received after EXCHANGE_LIFETIME. For the reason exposed above, this is not a good idea.

sophokles73:
We need to be careful which branch we are talking about. On master the Exchange lives as long as there are responses expected to arrive. In particular it exists for the whole lifetime of an observation, i.e. from the time the request has been sent until the last notification is processed. The 2.0.x branch currently contains an experimental change from Simon were the Exchange only lives until the request has been ACKed (right, Simon?). For any further responses coming in, an Exchange object is recreated based on the incoming response and the state kept for established observations.
I guess we are now talking about what to do on master, right?

joemag1:
Ok, so what should we do then to address the leak? In UDP land at least,
there is ack timeout which cleans up exchange if ACK gets lost, so leak can
only happen with non piggybacked responses.

In tcp land, any lost response will cause a leak, regardless of what the
client does.

sbernard31:
> The 2.0.x branch currently contains an experimental change from Simon were the Exchange only lives until the request has been ACKed (right, Simon?).
Not exactly, the exchange lives until the reception of the first response.

sbernard31:
For UDP, you're right. But from my point of view, this is up to the application layer to cancel the request when response is no more expected. So state should be clean on `request.cancel()`.

For TCP, I don't know the spec as well, but I suppose this could be the same ?

sophokles73:
Yeah, we should take a look at the draft. But I think there are no separate ACKs there, or are there? So I guess the responsibility for canceling timed out requests also should/could lie with the application layer ...

joemag1:
Yeah, no ACKs in TCP - which is why it's more prone to leaks on timeouts.

Back to the problem at hand - I think like it's reasonable for the
framework to enforce maximum amount of time an exchange can exist based on
some configuration from the user (EXCHANGE_LIFETIME seemed like a good fit,
but can be a completely different parameter).

But ultimately, I don't have a strong opinion where and how the cleanup
gets triggered, I'm mostly interested in solving the underlying leak. So
how about this approach then:

CoapClient already has a "timeout" property which is controlled by the
user. It's enforced on synchronous requests, but isn't on asynchronous. We
make two changes to the CoapClient:

* Honor timeout value even asynchronous invocations, and
* Have Request.setTimeout() trigger Exchange.setTimedout() so proper
cleanup is performed.

Thoughts?

sophokles73:
>  Honor timeout value even asynchronous invocations, and

Yes, we should do that. However, synchronous send currently does not *cancel* the request either after the timeout occurs but instead simply returns `null`, i.e. we seem to leak `Exchange`s in this case as well already.

> Have Request.setTimeout() trigger Exchange.setTimedout() so proper
cleanup is performed

I do not think that this is helpful. `Request.setTimedOut()` is invoked by the reliability layer if the all retransmissions failed (i.e. if we didn't receive an ACK). It also explicitly invokes `Exchange.setTimedOut()`. However, this has nothing to do with the problem we are trying to deal with here IMHO. I have taken a look at `Message.cancel()` because I think that this is what the client should do if it no longer wants to wait for a response (not an ACK). However, `Message.cancel()` basically only invokes the `onCancel()` method on the message's registered `MessageObserver`s. So, what we need to make sure is that e.g. `Matcher` registers an observer on the outbound request which, when its `onCancel()` method is invoked, removes the corresponding `Exchange` from the maps.

joemag1:
I haven't seen message.cancel() - this seems reasonable.

So the flow would work like this then:

* CoapClient observes timeout (CoapClient.timeout property) and if exceeded
calls Message.cancel().
** Need to update CoapClient to do that for async requests as well.
* Matcher *also* performs clean up when outgoing request.onCancel() is
invoked

Is everybody cool with this approach? If yes, I'll update the PR.

sophokles73:
AFAIK this looks reasonable. Don't forget that the `Matcher` needs to register a `MessageObserver` for doing this and we need to be careful that we register the observer on the *correct* request(s), in particular with regard to blockwise transfers ...

joemag1:
Ok - updated. Unfortunately, there is no way to make async requests time out in the CoapClient, as there is no guarantee an executor is set.

Without having sync and async requests behave consistently, I'd rather not add cancellation logic only to sync requests.

What are your thoughts on having CoapClient always initialize and use an executor?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseStatus.java
```diff
@@ -126,6 +126,15 @@ public int getCurrentNum() {
...
 
 	/**
+	 * Returns true if this is the first block in the sequence.
+	 *
+	 * @return true if first block, false otherwise.
+	 */
+	public boolean isFirst() {
+		return currentNum == 0;
+	}
+
```
sophokles73: I like this :-) We are doing a lot of such comparisons outside of the objects that should know best about their internal state ...

sophokles73:
Hm, I share your concerns regarding inconsistent behavior between sync and async. I do not think that we should create a `CoapClient` owned `Executor` unconditionally but only create it if needed, i.e. if no `Executor` has been set and we want to send asynchronously. We can then invoke `CoapClient.useExecutor()` to create one and schedule a task for canceling the request after *timeout*.

`CoapClient` currently shuts down its `Executor` when `CoapClient.shutdown()` is invoked. I think this should only be done if the `Executor` has not been set via `setExecutor()`. Otherwise `CoapClient` shuts down the executor that is under control of the client code which is probably not the expected behavior.

joemag1:
FWIW, the inconsistent behavior is there today, sync requests will timeout,
while async won't. Neither will clean up the matcher state.

The concern with calling useExecutor from async paths that it's a behavior
change - CoapClient will now implicitly create an executor, where it
doesn't do that today. This could be an issue for users that create 1000s
of CoapClients.

About the only option I can think of is having CoapClient use Endpoint's
executor to schedule timeout task. The way I'd do that is similar to my
original PR, where there's an explicit timeout layer in CoapStack, but
instead of having it use EXCHANGE_LIFETIME to schedule timeout, have it use
an optional "timeout" property on Request.

What do you think?

sophokles73:
I agree with you that it would be more desirable to use the Endpoint's Executor for handling the timeout. I am not sure about passing in the timeout as a param of the request though, since this seems to be more of a global config param that is not really application specific. The amount of time a client wants to wait for a response is still under the control of the client, whereas the "TimeoutLayer" simply provides means to prevent leaking of exchanges. Thus, the timeout value for the TimeoutLayer should be much higher than what Clients would usually wait for a response.

joemag1:
I didn't fully catch the concern about global param. I was imagining the
timeout layer waits variable amount for each request, depending on
request.getTimeout(), which would be set by the client. If no response
arrived within that time, it calls request.cancel(), which causes the
client to timeout, and exchange to clean up.

Btw, while we continue this conversation, what do you think about merging
the current PR? The current PR should be pretty non controversial - it
cleans up exchange if the request is cancelled.

sophokles73:
Yes, that is what we originally had in mind. However, we could also use the `CancelledMessageObserver` as a *safeguard* against leaked `Exchange`s only. A client using `CoapClient` to send an async request might be responsible to explictly invoke `Request.cancel()` after an application specific period of time if it no longer wants to wait for a response. However, if it fails to do so, the `CancelledMessageObserver` kicks in after some extended period of time (configured globally) in order to prvent the leaking of the Exchange if the server does not send a response at all. That timeout value should then of course be much higher than MAX_EXCHANGE_WAIT. The worst thing that might happen is that a client sending an async request wants to wait longer than that global max wait time so the request gets cancelled before by the `CancelledMessageObserver` (which the client can also be notified about by means of the `MessageObserverImpl` implicitly registered for the client's `CoapHandler`).

BTW I have rebased the PR on master (builds successfully) and cherry picked the commit on 2.0.x where it doesn't pass all tests (yet)  :-(

joemag1:
I'm totally fine with the client tracking timeout, and then calling
message.cancel() to clean up. That's what we'd likely do. So with that in
mind, just having CancelledMessageObserver is good enough for our use-case.

On a side note, one risk with a long time out is that each time out future
sits in the executor queue until timeout expires. So the queue would grow
to requests_per_second * cleanup_timeout. Unlikely to be a problem for most
applications, but could be a concern for high TPS ones, or if cleanup
timeout is too large.

What kinds of errors are you seeing in 2.0.0?

sophokles73:
The problems on 2.0.x are related to the fact that in 2.0.x the `Exchange` no longer spans all messages exchanged between client and server (in particular not any notifications for observed resources received as additional `Response`s to the original `Request`). I am not sure whether removing the request from the maps in `CancelledMessageObserver` is the right thing to do. However, I also haven't really figured out what the right thing to do would be. I guess it would be something like *completing the exchange* and let the existing handlers and observers do the magic of removing the exchange from the maps ... I will need to look into this some more. This does not only affect 2.0.x but also master, I am afraid. Since I haven't written the original code for the `Matcher`, `Exchange` and the whole stack, my knowledge about these things is also limited. It would really help if @mkovatsc would join this discussion because he has been the original author and should probably know best, how we should approach the whole problem.
Sorry for that, Joe, but it seems that we are a little stuck on this. I will dig into this once I am back home from the business trip I am currently on. However, I am also on vacation during the next weeks so I will only find limited time working on this ....

joemag1:
Ok, thanks for update. I'm also happy to jump on hangouts and discuss
options.

sophokles73:
Hi Joe, 

I have given the whole thing some more thought throughout the last days.
As indicated in my previous comment, there is a fundamental difference in Exchange semantics between `master` and `2.0.x` branches. I think we should focus on the way we want to solve the problem in `master` first.

That said, I think the original idea of registering a `MessageObserver` for outbound requests in the `Matcher`s and then use the `onCancel()` callback to tidy up is not so bad after all. However, I think we should invoke the affected `Exchange`'s `setComplete()` method to trigger housekeeping within the `Matcher`s which register an `ExchangeObserver` on every `Exchange` for that purpose. 
Within our `MessageObserver.onCancel()` method we should then check whether it falls into one of the following categories:
1. CON, already acknowledged but no response received yet or
2. NON and no response received yet
If it does we can invoke `Exchange.setComplete()` and let the observer do its magic for removing the `Exchange` from the maps.
We could probably also use the `Matcher`'s `ScheduledExecutor' to schedule a task invoking `MessageObserver.onCancel()` after a certain amount of time (providing the *safeguard* mentioned in one of my earlier comments) ... however, this may also have the side effect of preventing garbage collection of those `Exchange`s that complete normally. So we would need to clear the reference from the task to the `Exchange` once it gets canceled explicitly or completes normally (similar to the way it is done with the re-transmissions in the `ReliabilityLayer`.
What do you think?

californium-core/src/test/java/org/eclipse/californium/core/network/UdpMatcherTest.java
```diff
@@ -62,6 +65,28 @@ public void testReceiveResponseAcceptsResponseWithoutCorrelationInformation() {
...
+		Exchange exchange = sendRequest(matcher, null);
+
+		exchange.getRequest().cancel();
+
+		Exchange matched = matcher.receiveResponse(responseFor(exchange.getCurrentRequest()), null);
+		assertNull(matched);
+	}
+
+	@Test
+	public void testCancelledRequestShouldRemoveFromMidMap() {
```
sophokles73: I don't think that this test is relevant for the kind of requests that cause the leaking:
* CON
* separately ACKed (i.e. already removed from the MID map because the ACK contained the original request's MID)
* no response received yet

californium-core/src/test/java/org/eclipse/californium/core/network/UdpMatcherTest.java
```diff
@@ -62,6 +65,28 @@ public void testReceiveResponseAcceptsResponseWithoutCorrelationInformation() {
 	}
 
 	@Test
+	public void testCancelledRequestShouldRemoveFromTokenMap() {
```
sophokles73: We should improve this test case to better cover the type of request we are targeting, i.e. we need to make sure that we have received a separate ACK before canceling the request, e.g.
````
...
Exchange matched = matcher.receiveEmptyMessage(emptyMessageFor(exchange.getCurrentRequest()));
assertNotNull(matched);
exchange.getRequest().cancel();
...
````

File: californium-core/src/test/java/org/eclipse/californium/core/network/UdpMatcherTest.java (hunk 2 above)
Comment (joemag1) [inline californium-core/src/test/java/org/eclipse/californium/core/network/UdpMatcherTest.java]: Any reason to not just blindly remove cancelled requests? I understand that the leak just happens when a response for acknowledged request is lost. But since we introduced a cleanup-on-cancel mechanism, is there any advantage to being selective on clean up? Seems safe to just clean up everything if the user cancelled the request. Also creates a consistent user experience, where cancelled requests invoke error callback, regardless of whether they received an ACK or not.

**Comment by joemag1** [comment]
I do like the idea of delegating to Exchange.setComplete to perform clean up. That was part of my original thinking as well. If we were going that route, instead of Matcher, I'd register Cancelled observer in Stack. Then cancelled observer just calls Exchange.setComplete(), whenever the user calls Request.cancel(), and matcher does all the cleanup, like it does today. This should work regardless of what matcher we use. 

I do, however, think it's best to cleanup indiscriminately, and not only ACKed CONs. My thinking is that if the user called cancel(), they have no interest in seeing a response. So might as well clean up. What do you think?

**Comment by sophokles73** [comment]
> I do like the idea of delegating to Exchange.setComplete to perform clean up. That was part of my original thinking as well. If we were going that route, instead of Matcher, I'd register Cancelled observer in Stack. Then cancelled observer just calls Exchange.setComplete(), whenever the user calls Request.cancel(), and matcher does all the cleanup, like it does today. This should work regardless of what matcher we use.

That sounds good :-)

> I do, however, think it's best to cleanup indiscriminately, and not only ACKed CONs. My thinking is that if the user called cancel(), they have no interest in seeing a response. So might as well clean up. What do you think?
 Show all checks

I can see your point. We should do two more things in order to *clean up* then, I guess:
1. Explicitly invoke `Request.cancel()` from within `CoapClient` once a synchronous request times out. So far, client code can invoke `cancel()` on the request it is waiting for (from a separate thread) to explicitly stop waiting for the timeout but we currently do not *clean up* when the request simply times out).
2. Schedule a *cancel* task on the stack's `Executor` that cancels pending requests (i.e. *complete* the corresponding `Exchange`) that we did not receive a response for after a certain time (should probably be a multiple of MAX_EXCHANGE_WAIT).

**Comment by joemag1** [comment]
>>1. Explicitly invoke Request.cancel()
This is my thinking as well.

>> 2. Schedule a *cancel* task on the stack's Executor that cancels pending
requests
What value of multiplier did you have in mind? The only concern would be
keeping exchanges locked up in executor's queue, preventing them from being
garbage collected until the multiplier elapsed.

For example, if multiplier was 10, an application doing 100 TPS would
constantly keep 10 * 247 * 100 = 247,000 exchanges in memory. So maybe it's
not that bad.

**Comment by sophokles73** [comment]
I was more thinking of a small multiplier, e.g. 2x or 3x. In order to prevent locking of Exchanges by the queued up clean-up tasks we might be able to implement something along the lines of how re-transmission tasks are managed, i.e. once the Exchange has completed because either the response has been received or the client code has invoked `cancel()` we could *clear* the reference that the clean-up task holds to the `Exchange` ... The task would still be in the executor's queue but it would not prevent the Exchange from being GCed. We would need to keep the `ScheduledFuture` for the clean-up task in the Exchange or its corresponding `ExchangeObserver`, I guess ...

**Comment by joemag1** [comment]
My initial thinking was that cancel() does not remove task from executor's
queue. I think you're saying that it will still help because it remove
Exchange reference from the task.

I'll take a look at ScheduledExecutor code - if that's indeed the case then
what you suggesting should work.

-Joe

**Comment by joemag1** [comment]
Let me do this as two parts. Here's the first part, that cleans up exchange on cancel() and and calls cancel() if synchronized request times out.

To also add time based expiration, I need to do a bit more digging. For example, Exchange only supports one observer, so either need to add support for observer list, or consider adding time-based cleanup (and cancellation of it if exchange completes normally) to the matcher.

**Comment by sophokles73** [comment]
> My initial thinking was that cancel() does not remove task from executor's
queue. I think you're saying that it will still help because it remove
Exchange reference from the task.

I think I was mistaken in assuming that we would be able to clear the reference via the `ScheduledFuture`. However, there does not seem to be a way to influence the scheduled task (`Runnable` instance) by means of e.g. invoking `ScheduledFuture.cancel()`. So, I guess we will need to stick to either simply let all scheduled clean-up tasks get executed at some time or prevent them from being executed by invoking their `cancel()` method if the exchange has completed before the timeout. The latter seems to require us to jump through some hoops in order to find the right place to keep the reference to the scheduled clean-up task (as you already mentioned).
Another idea would be to implement the MID & token maps as LRU caches which would automatically evict timed-out exchanges when new exchanges are being put to the maps. This might also be helpful in limiting the capacity of the maps in order to prevent `OutOfMemoryException`s in case of excessive message exchanges ...

File: californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java [hunk 4]
```diff
@@ -764,8 +764,14 @@ private CoapResponse synchronous(Request request) {
...
 		try {
 			Response response = send(request, outEndpoint).waitForResponse(getTimeout());
-			if (response == null) return null;
-			else return new CoapResponse(response);
+			if (response == null) {
+				// Cancel request so appropriate clean up can happen.
+				request.cancel();
+				return null;
+			}
+			else {
```
sophokles73: formatting, should be on same line as closing bracket

californium-core/src/test/java/org/eclipse/californium/core/network/stack/CoapStackTest.java
```diff
@@ -0,0 +1,73 @@
...
+		// Capture exchange
+		Exchange exchange = exchangeCaptor.getValue();
+		assertFalse(exchange.isComplete());
+
+		request.setCanceled(true);
+		assertTrue(exchange.isComplete());
+	}
+
+	/** Determinstic scheduler throws on isShutdown(), but we simply want it to return false. */
+	private static class NeverShutdownDeterministicScheduler extends DeterministicScheduler {
```
sophokles73: do we really need to introduce a dependency on jmock just for using `DeterministicScheduler`? We need to create a CQ for new dependencies so that the Eclipse Foundation can run IP checks on it. jMock 2.x has not yet been used by any Eclipse projects yet so we would need to run the full Monty ... so, if there is an alternative to this that does not require too much effort then we should avoid introducing the jMock dependency here ...

sophokles73:
> My initial thinking was that cancel() does not remove task from executor's
queue. I think you're saying that it will still help because it remove
Exchange reference from the task.

I guess I was mistaken in assuming that we would be able to somehow clear the clean-up task's reference to the `Exchange` in effect to `ScheduledFuture.cancel()` being invoked. But that doesn't seem to be the case. So I guess the strategy easiest to implement will be what you initially intended: simply let all scheduled tasks execute.
However, we could also implement the MID & token maps as LRU caches which could start to evict exchanges that have timed out once their max capacity has been reached and new exchanges are being added ... this would also allow us to introduce a max capacity preventing uncontrolled memory consumption.

joemag1:
Didn't know adding new dependencies was such a pain. I will remove jmock.
It's useful for testing things where you require passage of time, but can
figure a different way here.

So to summarize, the final solution will be the current "request cancel
triggers exchange completion", plus "all exchanges complete after 2x max
exchange lifetime"

Liu sounds like an intriguing idea - but if we ever go that route we need
to be careful not to evict active exchanges.

sophokles73:
> Didn't know adding new dependencies was such a pain. I will remove jmock.
It's useful for testing things where you require passage of time, but can
figure a different way here.

Well, I would gladly take on the burden to add a dependency if we would make a lot of use of it :-)

> So to summarize, the final solution will be the current "request cancel
triggers exchange completion", plus "all exchanges complete after 2x max
exchange lifetime"

Yes, and I think we should make the multiplier configurable.

> Liu sounds like an intriguing idea - but if we ever go that route we need
to be careful not to evict active exchanges.

Sure, but I would rather postpone that to another PR ...

joemag1:
Californium uses a lot of scheduled executors, so jmock can actually be a
pretty useful test dependency.

sophokles73:
Then we might be able to create another PR later illustrating its usefulness in some of the existing test cases? However, I would like to make some progress in this PR and merge it ASAP now that we have agreed on how we want to pull this off, ok?

joemag1:
Yeah, I'm fine with that.

sophokles73:
Great, then maybe you can make the last few changes, rebase on master and then we are good to go :-)

joemag1:
Haven't forgotten about this yet, been busy with work and family last few
days. I'll try update the PR sometime in the next few days. Meanwhile, I
was reading CoAP spec, and it made me question whether we should be using
any multiplier, or just purging Exchange after EXCHANGE_LIFETIME.

Relevant bits from the spec:
https://tools.ietf.org/html/rfc7252#section-4.8.2  Relevant paragraph (and
further down it even mentions that the same value MAY be used for NONs as
well):

EXCHANGE_LIFETIME is the time from starting to send a Confirmable
      message to the time when an acknowledgement is no longer expected,
      i.e., message-layer information about the message exchange can be
      purged.  EXCHANGE_LIFETIME includes a MAX_TRANSMIT_SPAN, a
      MAX_LATENCY forward, PROCESSING_DELAY, and a MAX_LATENCY for the

      way back.  Note that there is no need to consider
      MAX_TRANSMIT_WAIT if the configuration is chosen such that the
      last waiting period (ACK_TIMEOUT * (2 ** MAX_RETRANSMIT) or the
      difference between MAX_TRANSMIT_SPAN and MAX_TRANSMIT_WAIT) is
      less than MAX_LATENCY -- which is a likely choice, as MAX_LATENCY
      is a worst-case value unlikely to be met in the real world.  In
      this case, EXCHANGE_LIFETIME simplifies to:

         MAX_TRANSMIT_SPAN + (2 * MAX_LATENCY) + PROCESSING_DELAY

      or 247 seconds with the default transmission parameters


So what do you think about just using EXCHANGE_LIFETIME for
expiration, without any multipliers.

sophokles73:
Sounds good :-)

BTW No hurries needed, I have been on vacation for the last three weeks and will return to office next week only ...

sbernard31:
I re-read this thread and I'm a bit lost, so I will just summarize what I understand.

This should be up to the application layer to define the acceptable amount of time to wait for a response.
The **low level API** (endpoint/request ...) should provide a way to **cancel request**.
The **high level API** (coapEndpoint) could propose a **configurable timeout**.

If we choose to propose a default value for this timeout. I think this should be higher than EXCHANGE_LIFETIME. (see explanation [here](https://github.com/eclipse/californium/pull/74#issuecomment-234200263))

HTH

joemag1:
Almost. Here's the summary.

* CoapStack/CoapEndpoint should clean up exchange state when request is
cancelled
* Higher level APIs (CoAPclient or client themselves) cancel request when
it times out, or for any other reason
* CoapStack cleans up exchange from internal maps after EXCHANGE_LIFETIME.
This is based on interpretation of the spec:
https://tools.ietf.org/html/rfc7252#section-4.8.2

sbernard31:
>* CoapStack/CoapEndpoint should clean up exchange state when request is cancelled

Good for me.
>* Higher level APIs (CoAPclient or client themselves) cancel request when it times out, or for any other reason

Good for me. (Timeout should be configurable)
>* CoapStack cleans up exchange from internal maps after EXCHANGE_LIFETIME.

For reason explain [here](https://github.com/eclipse/californium/pull/74#issuecomment-234200263), I think we must not do that. The exchange lifetime is about Message not about Request/Response correlation. (and this will not works for observation)

joemag1:
>>For reason explain here, I think we must not do that. The exchange lifetime is about Message not about Request/Response correlation. (and this will not works for observation)

My reading of the spec made it sound like any state can be purged, but re-reading it again it specifies "_message-layer information about the message exchange can be purged_". I assume that means deduplicator and MID maps, but not token maps. 

Do you think it's still worth having exchange expiry based on some client configurable value? Or should we only cleanup on request.cancel() and leave that task to the client?

sbernard31:
Personally, I will leave that task to the client.

joemag1:
That's fine with me. Kai, what do you think?

sophokles73:
It seems I have been a little quick to agree with the `EXCHANGE_LIFETIME` value. The way I read the spec it defines the amount of time after which a sender of a CON can/should stop waiting for an ACK for the message it sent. However, in our case we want to purge information about exchanges where we already received an ACK but never received a response. This can only be the case where the server sent an empty ACK, i.e. not piggybacking the response, and has then *forgotten* to send the response in a separate CON or NON. So we should probably not use `EXCHANGE_LIFETIME` for the timeout but instead use some higher value.
However, unless we limit the overall capacity of the MID & token maps, I do not think that we should leave cleaning up to the client (code) simply because we cannot make sure that a client will always invoke `cancel()` on stale requests.

sbernard31:
>So we should probably not use EXCHANGE_LIFETIME for the timeout but instead use some higher value. However, unless we limit the overall capacity of the MID & token maps, I do not think that we should leave cleaning up to the client (code) simply because we cannot make sure that a client will always invoke cancel() on stale requests.

It's seems to me a bit difficult to define this "lifetime" especially for Observe request. (In master the californium CoAP Exchange for observe can live forever)

joemag1:
Can we just pipe this timeout value from the client? For normal
request/reply interactions a "reply-timeout" value is set on the Request by
CoapClient. A layer inside CoapStack waits for this timeout and calls
setTimeout() on Exchange/Request if the two are still incomplete.

For observe requests no timeout value is set, and instead CoapClient ties
cancellation of observe relationship (I need to see what is the mechanism
for that) to completion of Exchange.

Does that sound reasonable? If we still don't have an agreement, then at
the very least I'd like to merge the current PR (after I rebase it). In the
current PR coap stack at least performs a clean up on request cancellation,
which gives us an option to perform clean up ourselves.

sbernard31:
I still prefer limit complexity in the core API. Two way to do the same thing in that code (cancel+timeout) is maybe not necessary. So, I prefer to add timeout only at `CoAPClient` level.

But this is not a strong opinion.

sophokles73:
Having thought about all the arguments brought up in this thread (excellent discussion IMHO) I tend to agree with Simon that we should probably try to keep the implementation in the core (the CoapStack) as simple as possible and leave timeout handling to client code. This, BTW, is also what Matthias has always been arguing for if I am not mistaken :-)
In a way, we already have this pattern implemented by means of the `CoapClient` which can already been considered *client code* from my point of view. `CoapClient` already makes sure that observations are *refreshed* if an observed resource's value becomes *stale* so it seems reasonable to leave this up to other client code that doesn't use `CoapClient` as well.

I also agree with Joe that we should merge this PR in order to get the clean up on request cancellation for now so that clients can start managing (async) request timeout more properly ...

joemag1:
Btw, I do agree with Simon on performing timeout in CoAPClient. The only
reason we went down this path is because there was no straightforward way
to do that: Client does not have an access to a thread pool.

If somebody can think of a backwards compatible way to timeout at the
client, without creating a threadpool per client, the we can have our cake
and eat it too.

joemag1:
Rebased current "only clean up on cancell" PR on top of master.

sophokles73:
merged to master (14494826ead0548a14c0d1b254c4d97f0f052af9) and 2.0.x (7d32592bf2d56feaf42d9b33970950203311ce79)

PR #78, opened 2016-07-21
[fix #75] Server Treats NON-Message as CON-Message

### LINKED ISSUE #75: Server Treats NON-Message as CON-Message
I'm trying to send a non-confirmable message from the client as follows:

			Request request = new Request(Code.POST, Type.NON);
			request.setPayload("from CoAP client");
			CoapResponse response = client.advanced(request);

In this case, I'm not expecting any response back.  But, it seems the server always treats the request a confirmable message.

INFO: UDPConnector Receiver: org.eclipse.californium.elements.RawData@4a1f79c5
UDPConnector Receiver: org.eclipse.californium.elements.RawData@4a1f79c5

org.eclipse.californium.core.server.ServerMessageDeliverer.handleRequest(): request=[CON-POST   MID=48305, Token=76369a, OptionSet={"Uri-Path":"helloWorld"}, "from CoAP client"]

org.eclipse.californium.examples.HelloWorldServer$HelloWorldResource.handleRequest(): request=[CON-POST   MID=48305, Token=76369a, OptionSet={"Uri-Path":"helloWorld"}, "from CoAP client"]

request=[CON-POST   MID=48305, Token=76369a, OptionSet={"Uri-Path":"helloWorld"}, "from CoAP client"]
message from client = [from CoAP client];

As per output, it clearly shows the Server treats the NON-POST request as a CON-POST request.

evgeniynickolaev:
Request#type is now Type.CON by default.

CoapClient: Move setting the request type to the convenience methods.
Client code can use the advanced() methods for sending a request that
it has configured manually whereas the convenience methods can be used
in the established way.

Uses Request#uri (if not blank) in the advanced() methods instead of CoapClient#uri.

Signed-off-by: Eugene Nikolaev <eug.nikolaev@gmail.com>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -292,7 +292,7 @@ public class CoapClient {
 	}
 	
 	public Set<WebLink> discover(String query) {
-		Request discover = Request.newGet();
+		Request discover = newGet();
 		discover.setURI(uri); // for scheme and authority, but then remove path and query
 		discover.getOptions().clearUriPath().clearUriQuery().setUriPath("/.well-known/core");
 		if (query!=null) {
@@ -321,7 +321,7 @@ public class CoapClient {
 	 * @return the CoAP response
 	 */
 	public CoapResponse get() {
-		return synchronous(Request.newGet().setURI(uri));
+		return synchronous(newGet().setURI(uri));
 	}
 	
 	/**
@@ -332,7 +332,7 @@ public class CoapClient {
 	 * @return the CoAP response
 	 */
 	public CoapResponse get(int accept) {
-		return synchronous(accept(Request.newGet().setURI(uri), accept));
+		return synchronous(accept(newGet().setURI(uri), accept));
 	}
 	
 	// Asynchronous GET
@@ -344,7 +344,7 @@ public class CoapClient {
 	 * @param handler the Response handler
 	 */
 	public void get(CoapHandler handler) {
-		asynchronous(Request.newGet().setURI(uri), handler);
+		asynchronous(newGet().setURI(uri), handler);
 	}
 	
 	/**
@@ -355,7 +355,7 @@ public class CoapClient {
 	 * @param accept the Accept option
 	 */
 	public void get(CoapHandler handler, int accept) {
-		asynchronous(accept(Request.newGet().setURI(uri), accept), handler);
+		asynchronous(accept(newGet().setURI(uri), accept), handler);
 	}
 	
 	// Synchronous POST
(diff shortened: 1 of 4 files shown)


californium-core/src/test/java/org/eclipse/californium/core/test/ClientSynchronousTest.java
```diff
@@ -150,6 +153,34 @@ public void testSynchronousCall() throws Exception {
...
+		CoapClient client = new CoapClient(uri).useExecutor();
+
+		// Set NONs but expecting CONs as specified in request
+		client.useNONs();
+
+		Request request = new Request(Code.GET, Type.CON);
+
+		CoapResponse resp = client.advanced(request);
+
+		Assert.assertEquals(CONTENT_1, resp.getResponseText());
```
sophokles73: How does this assert that the request indeed contained a CON? The fact that we receive a response merely means that the request has been received and processed by the server ... We should at least assert that the response is of type ACK, i.e. the server piggybacked the response, shouldn't we?

evgeniynickolaev: Yes, thanks, sure we should...
I'll add the asserts but here is another one issue, I'll point to it. Need advice, please.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -930,7 +927,67 @@ protected Endpoint getEffectiveEndpoint(Request request) {
...
+	}
 	
+	/*
+	 * Assigns a CoapClient#uri if request has no uri.
+	 *
+	 * @param request the request
+	 */
+	private void assignClientUriIfEmpty(Request request) {
+		// request.getUri() is a computed getter and never returns null so checking options
+		if (request.getOptions().getUriHost() == null) {
```
evgeniynickolaev: This is too bad check and currently it does not work.
Not sure why it worked before...
How can I check that uri is set?

sophokles73: Maybe it's because you are testing this with "localhost" or an IP address as the hostname? `Request.setURI()` ignores those ...

evgeniynickolaev: @sophokles73, what do you think, is it OK to use getDestination instead of options host?

```
		if (request.getDestination() == null) {
			request.setURI(uri);
		}
```

According to the ```Request#setURI``` method, ```destination``` is the only one field which is being set anyway.
But on the other hand - ```setDestination(...)``` is used in many places...

sophokles73: At least the *destination* property seems to be always set when someone sets the URI on the request. So, yes, I think it would be a better choice for the check ...

evgeniynickolaev:
Closed due to new pull-request creation (apply review comments)

PR #81, opened 2016-08-09
Fixes to persist observation

ingoschaal:
Added some fixes to persist Observation in ObservationStore to a permanent store like Redis or MongoDB.

- In CorrelationContex added getKeys(), this allows to store the associated context.
- In UdpMatcher replaced the instantOf check of DtlsCorrelationContext. Check if in the given exchange the correlationContext with key "KEY_SESSION_ID" is available.
Otherwise we would have to persist something like the information about the instance of the correlationContext.
- Removed the cast to the DtlsCorrelationContext, because all information can be retrieved over the CorrelationContext, which is used in the isResponseRelatedToDtlsRequest and isResponseStrictlyRelatedToDtlsRequest method.
All operation can be done through the Interface.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -478,14 +478,13 @@ public class UdpMatcher implements Matcher {
 			// no correlation information available for request, thus any
 			// additional correlation information available in the response is ignored
 			return true;
-		} else if (exchange.getCorrelationContext() instanceof DtlsCorrelationContext) {
+		} else if (exchange.getCorrelationContext().get(DtlsCorrelationContext.KEY_SESSION_ID) != null) {
 			// original request has been sent via a DTLS protected transport
-			DtlsCorrelationContext exchangeDtlsContext = (DtlsCorrelationContext) exchange.getCorrelationContext();
 			// check if the response has been received in the same DTLS session
 			if (useStrictResponseMatching) {
-				return isResponseStrictlyRelatedToDtlsRequest(exchangeDtlsContext, responseContext);
+				return isResponseStrictlyRelatedToDtlsRequest(exchange.getCorrelationContext(), responseContext);
 			} else {
-				return isResponseRelatedToDtlsRequest(exchangeDtlsContext, responseContext);
+				return isResponseRelatedToDtlsRequest(exchange.getCorrelationContext(), responseContext);
 			}
 		} else {
 			// compare message context used for sending original request to context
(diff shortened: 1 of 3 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -478,14 +478,13 @@ private boolean isResponseRelatedToRequest(final Exchange exchange, final Correl
 			// no correlation information available for request, thus any
 			// additional correlation information available in the response is ignored
 			return true;
-		} else if (exchange.getCorrelationContext() instanceof DtlsCorrelationContext) {
+		} else if (exchange.getCorrelationContext().get(DtlsCorrelationContext.KEY_SESSION_ID) != null) {
```
sophokles73: Is there a particular reason why you would like to get rid of the *instanceof* check? You are still depending on the `DtlsCorrelationContext` class. The only reason I can see why you would want to do this is because you expect to find a DTLS correlation context in the exchange that is not a `DtlsCorrelationContext`. However, since Californium uses Element-Connector and Scandium, I wonder where such a correlation context should come from...

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
```diff
@@ -17,18 +17,27 @@
...
 	 * @param key the key to retrieve the value for.
-	 * @return the value or <code>null</code> if this context does not contain
-	 *         a value for the given key.
+	 * @return the value or <code>null</code> if this context does not contain a
+	 *         value for the given key.
 	 */
 	String get(String key);
+
+	/**
+	 * Sends the list of keys available in the context.
```
sophokles73: Should read: *Gets all keys use in this context.*

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
```diff
@@ -17,18 +17,27 @@
...
+	 *
+	 * @return list of keys set in the context.
```
sophokles73: Why are you returning a `Set` if the docs say *Gets a list of ...*?
Is the returned set ordered?

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
```diff
@@ -17,18 +17,27 @@
...
+	 */
+	Set<String> getKeys();
```
sophokles73: Is this method needed/used somewhere?

ingoschaal: If you implement ObservationStore and would like to use persitence like MongoDB or Redis (which has already begun), we need to persist the keys for CorrelationContext too. Therfore we ned the  getKeys() method in CorrelationContext.
Then we do not have to store the Object, which would be difficult. 
It is sufficient to know the key.

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
ingoschaal: I will of course change the comment. This is not the right comment.

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
ingoschaal: I will of course change the comment. This is not the right comment.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
ingoschaal: Yes, it is true about the first point of you.
We can not store the CorrelationContext Object (in MongoDB or Redis), we only will store the information about the keys (Therfore the getKeys() method in CorrelationContext). 
We can restore it to a MapBasedCorrelationContext, if its a DtlsCorrelationContext the KEY_SESSION_ID is available.
Besides, I am of the opinion, **instanceOf** is not best practice.
I do not quite understand the second point. If there may be no other CorrelationContext as DtlsCorrelationContext should we then not only use this?

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
sophokles73: I see. However, you will want to persist not only the keys but also the corresponding values, right?
Wouldn't a method like

````
Set<Map.Entry<String, String>> entrySet();
````

as defined in `Map` make more sense in that case? This would also not require to retrieve each value individually ...

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
sophokles73: Ok, I see your point. Makes sense to me :-)

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
ingoschaal: yes that would be possibility too. Thanks for this point.

sbernard31:
it sounds good to me.

ingoschaal:
can we bring this PR to a close?

sbernard31:
Integrated in master (commit 89651ab5097918cd7f057aa9ace794388f3a7172 and 5666e883d47b95575db481b5bffe07e11d973d17)

Issue #63, opened 2016-06-24
Californium re-uses unacknowledged MIDs in outbound requests


sophokles73:
Californium currently uses a simple counter for generating message IDs that is incremented with every message sent and wrapped at 2^16.

When client code uses Californium to send out requests to another CoAP server at a high rate, this may lead to re-use of message IDs for which Californium hasn't received an ACK or CON yet. While such a client's behavior seems questionable (flooding a constrained device with hundreds of requests per second doesn't really fit with the constrained device environment), Californium should make sure not to re-use IDs in this erroneous way. Also see #57 for a more detailed discussion of the issue.

magerramov:
Switching mid assignment to using a BitSet that tracks open exchanges, and then calling BitSet.nextUnset(currentMid) may do he trick here. The cost extra CPU.

PR #85, opened 2016-08-23
Introduce TokenProvider to prevent clashes of CoAP tokens

### LINKED ISSUE #83: Duplicated CoAP Tokens
In `Matcher` there is a method `createUnusedToken()` that should ensure that there are no duplicated CoAP tokens. However there are issues that lead to duplicated tokens anyways.

These issues differ on master branch and 2.0.x branch.

Master branch:

The createUnusedToken mechanism is not thread-safe. It gets checked if the token already exists but the new token gets added to the map later on, i.e. two threads may generate the same "unused" token.

2.0.x branch:

- The same concurrency issue as above,
- the AND condition in the while loop needs to be an OR condition.

I think on master branch we could fix this with `putIfAbsent()`, but on 2.0.x branch things get more complicated. Are there any suggestions how to fix this? Of course one fix would be to introduce some locks.

Remark (I am not a maths expert): One may assume that the probability for this clash is not that high, but it is (especially on 2.0.x branch). Caused by the variable length of the token we have a 1:8 probability to generate a one byte token. Inside these one byte tokens then we have some kind of birthday problem. In summary I get a clash on average with about 165 iterations with a test run.

Thanks
Daniel

DanielMaier-BSI:
Initial implementation to fix #83.

This PR introduces a `TokenProvider` which is responsible to provide unused token. The `TokenProvider` is plugable to give users the possibility to implement their own `TokenProvider`, e.g. a cluster ready one. The default `TokenProvider` is an in memory implementation that uses random tokens.

This PR does not modify the `TcpMatcher` because I am unsure which is the state of TCP implementation in 2.0.x branch (`TcpMatcher` does not use `ObservationStore`)

One problem is still with user provided tokens. `TokenProvider` may create same token as a user provided token, because `TokenProvider` is not aware of user provided tokens.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -25,6 +25,7 @@
  *                                      of Response(s) to Request (fix GitHub issue #1)
  *    Bosch Software Innovations GmbH - adapt message parsing error handling
  *    Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
+ *    Bosch Software Innovations GmbH - introduce TokenProvider
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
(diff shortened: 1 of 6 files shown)


sbernard31:
@sophokles73 What do you think about that ?

sophokles73:
I like the idea of using a strategy interface for generating tokens. I would also like to apply the same idea for creating message IDs which is currently also broken (see #63) :-( In order to remedy the situation I think we should start to scope both the generated tokens as well as the MIDs to the peer's endpoint address. I have already started to work on this but ran into problems on 2.0.x because of the (transparent) blockwise transfer support which makes keeping track of tokens and MIDs (and releasing them) a nightmare.

If you guys need this PR's functionality  urgently then by all means go with it. However, be advised that AFAIK the 2.0.x branch will undergo major changes before we create a release there. In particular, I would like to get rid of the blockwise layer before we create a release as already indicated on the mailing list.

sbernard31:
About the massive change you talked about, I'm interested to follow your work.  Could you share it even it's still in progress. :)

sophokles73:
I have pushed the `message_store` branch which basically is a fix for #63 but also introduces an  `MessageExchangeStore` which encapsulates all exchange maps and also takes care of generating and managing the MIDs and tokens so that assignment to messages should (hopefully) be thread safe ...

DanielMaier-BSI:
We don't need this PR urgently. We fixed the problem locally with locks inside UdpMatcher (https://github.com/bsinno/californium/tree/fixes_to_duplicate_tokens_inside_matcher_2.0.x) to not have too many local code changes. 

Nevertheless fixing this issue could be useful because leshan is using this branch and it is not usable with a few hundred observes. But I think it is not worth to introduce TokenProvider, because of all the upcoming refactorings. Perhaps we should also go with the quick lock based fix here as intermediate solution?

sbernard31:
IMHO, I think just fixing the random size could resolve a lot of problems at short term with minimal code modification.
Using a fixed 8 bytes should really limit collisions.

DanielMaier-BSI:
Mh, why not just use also the locks to have a solution that works in all cases and does not fail when you are unlucky? Or do you have any concerns regarding this (temporary) solution? I can create a PR for this.

sophokles73:
I have pushed [fix_#63_for_2_0_x](https://github.com/eclipse/californium/tree/fix_63_for_2_0_x) which introduces the `MessageExchangeStore` as the central point where message exchanges are managed and MIDs and tokens are created.

@DanielMaier-BSI I think we should add the `TokenProvider` to the `MessageExchangeStore` in a similar way that the `MessageIdProvider` is used there ...

sbernard31:
> Mh, why not just use also the locks to have a solution that works in all cases and does not fail when you are unlucky? Or do you have any concerns regarding this (temporary) solution? I can create a PR for this.

We talked about a short term fix. With 8 bytes = 2^64 combination = 1,844674407×10¹⁹. If you get twice time the same token pending the race description you described, I'm not sure "unlucky" was enough :p !

sophokles73:
@DanielMaier-BSI,

I have merged #98, the `MessageExchangeStore` is now the central place where MIDs and tokens are generated and managed.
Can you re-factor this PR and incorporate the `TokenProvider` into the `InMemoryMessageExchangeStore`?

DanielMaier-BSI:
OK. I put it onto my TODO list.

sophokles73:
Replaced by #113, merged in 52104bd8511f74935bfb11bce280200972cb4eed

PR #86, opened 2016-08-30
Store contextual information on the request

msangoi:
It can be needed to add some additional information to a request.
Observation is a typical use case: Californium is in charge of persisting the observation but it can be needed to persist the contextual information as well. It would not be necessary if the additional data is part of the request.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
@@ -30,6 +30,8 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.security.Principal;
+import java.util.Collections;
+import java.util.Map;
 
 import org.eclipse.californium.core.coap.CoAP.Code;
 import org.eclipse.californium.core.coap.CoAP.Type;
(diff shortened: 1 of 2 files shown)


sbernard31:
it sounds goods to me.
@sophokles73  Can I integrate this in 2.0.x branch ?

sophokles73:
Can you guys provide a concrete example of where this is useful/needed?

msangoi:
@sophokles73 A typical use case in Leshan is to persist the registrationId for a given observation request. If we can add this information to the Californium request context, there is no additional persistence to implement in the Leshan layer. 
Here is the example branch: https://github.com/eclipse/leshan/compare/observe-request-context

sophokles73:
I understand your motivation and my guess is that the expectation is that the user context can be retrieved via `Exchange.getRequest().getUserContext()` on every notification received, right?
Your code currently doesn't seem to achieve this, though, because the serializer doesn't consider the user context that you have added to the request (yet)...

msangoi:
Yes, I think this is what is done in this branch :)
In the Californium `InMemoryObservationStore`, the user context is set once the request is deserialized from the "wire format".
And in the Leshan branch, the user context is also serialized/deserialized in both implementations of the observation store (`InMemoryLwM2mObservationStore` and `RedisObservationStore`).

sophokles73:
Yes, you are right. I was a little confused by `InMemoryObservationStore` using serialization/deserialization to actually *clone* the original `Request`... an implementation using a real persistent store will probably need to find a way to persist the relevant information from the Request (including the user context values) by other means than serializing to the wire encoding.

So, AFAIK, you can go ahead and merge it to 2.0.x, @sbernard31

PR #88, opened 2016-09-09
fix stale observations in CoAPClient when CoAPServer cannot add an observation

balsmn:
As per CoAP Spec RFC 7641 Section [§2](https://tools.ietf.org/html/rfc7641#section-2) 

> If the server is unwilling or unable to add a new entry to the list of
   observers, then the request falls back to a normal GET request and
   the response does not include the Observe Option.

If the server sends a normal ACK for the observe request without the observe option, the [observation that was added to observation store in UdpMatcher#L204](https://github.com/eclipse/californium/blob/2.0.x/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java#L204) will never be removed leading to stale entries.

This PR aims to fix this problem.

Signed-off-by: Bala Azhagappan balasubramanian.azhagappan@bosch-si.com

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -322,12 +322,17 @@ public final class UdpMatcher extends BaseMatcher {
 
 					@Override
 					public void onResponse(Response response) {
+						// check whether the client has established the observe
+						// requested
 						if (!response.getOptions().hasObserve()) {
+							// Observe response received with no observe option
+							// set. It could be that the Client was not able to
+							// establish the observe. So remove the observe
+							// relation from observation store, which was stored
+							// earlier when the request was sent.
 							LOGGER.log(Level.FINE,
-									"Removed entry from observation store as observe response with token {0} didn't have observe option: ",
+									"Response to observe request with token {0} does not contain observe option, removing request from observation store",
 									idByToken);
-							// no observe option set. Client was not able to
-							// establish observe
 							observationStore.remove(request.getToken());
 						} else {
 							notificationListener.onNotification(request, response);
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -289,7 +289,7 @@ public final class UdpMatcher extends BaseMatcher {
 		 */
 
 		KeyMID idByMID = KeyMID.fromInboundMessage(response);
-		KeyToken idByToken = KeyToken.fromInboundMessage(response);
+		final KeyToken idByToken = KeyToken.fromInboundMessage(response);
 		LOGGER.log(Level.FINER, "received response {0}", response);
 		Exchange exchange = exchangeStore.get(idByToken);
 
@@ -321,8 +321,17 @@ public final class UdpMatcher extends BaseMatcher {
 					}
 
 					@Override
-					public void onResponse(final Response resp) {
-						notificationListener.onNotification(request, resp);
+					public void onResponse(Response response) {
+						if (!response.getOptions().hasObserve()) {
+							LOGGER.log(Level.FINE,
+									"Removed entry from observation store as observe response with token {0} didn't have observe option: ",
+									idByToken);
+							// no observe option set. Client was not able to
+							// establish observe
+							observationStore.remove(request.getToken());
+						} else {
+							notificationListener.onNotification(request, response);
+						}
 					}
 
 					@Override


sophokles73:
I think that you are right in assessing that we are currently prone to this problem. However, I think we should try to fix the problem by not putting the exchange to the `ObservationStore` nuless we have received the reponse to the original request that created the observation, i.e. we should not put the exchange to `ObservationStore` in `Matcher.sendRequest()` but instead in `Matcher.receiveResponse()`. If we receive the first response in reply to our original request, then we will be able to look up the exchange by token in the token map. At that point we can check if the response has the observe option and can then put it to the `ObserveStore`. For all subsequent responses received with that token, the lookup in the token map will fail and we will therefore restore the exchange from the ObservationStore.

What do you think?

sbernard31:
I'm pretty sure I did that the first time I implemented ObservationStore, then I changed it for the current less intuitive way. (I mean added the observation before response recept)

If I well remember this was to handle this scenario :

We have several instances of Californium CoAP client behind a NAT (The leshan cluster use case)
A CoAP client A send an observe request.
The CoAP server send a ACK.
Then the response is lost.
Later, the CoAP server send a notification and the response go through a different CoAPclient B.
As the observation is not in the store we will reject it.
(If the notification was sent to the client A, the notification would be accepted as the token is still present in memory in the Client A)

I don't now if this is a valid scenario to support :/

sophokles73:
To get the scenario described by Simon straight:

- The CoAP clients A & B are the nodes constituting e.g. a cluster of LWM2M servers, right?
- Consequently, the CoAP server in this scenario is a device that we want to observe resources on.
- In your scenario the CoAP server (device) responds with an empty ACK to client A's observe request, i.e. the response is not piggy-backed but is sent in a separate CON or NON later on. It is important to understand that this ACK will definitely be received by client A (otherwise client A will retransmit its request).
- If the response is sent in a NON that you are assuming that this NON is lost.
- If the response is sent in a CON then you are assuming that it is received by client B instead of A (for whatever reason)

The solution for both cases could be to also register the observation to the ObservationStore when we receive the ACK for the observe request, couldn't it?

sbernard31:
It could be. In this case, the fix proposed by @balsmn  still make sense, right ?

balsmn:
Any updates on this PR?

sophokles73:
@sbernard31,

I guess you're right. The `ACK` that client A receives from the server does not contain the*observe* option so the client does not know (yet) whether the server will actually accept the observe request. That can only  be determined from the `NON/CON` response sent later (which in your scenario is lost).
So I guess, Bala's PR makes sense in any case :-)

@balsmn, can you please rebase on current 2.0.x so that we can review and merge?

balsmn:
@sophokles73 I have rebased my changes on the current version of 2.0.x branch.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -321,8 +321,17 @@ public void onTimeout() {
 					}
 
 					@Override
-					public void onResponse(final Response resp) {
-						notificationListener.onNotification(request, resp);
+					public void onResponse(Response response) {
+						if (!response.getOptions().hasObserve()) {
```
sophokles73: I would like to see a comment here describing the circumstances under which this case will actually happen. I bet, if we do not document it here, we will all wonder why added this in two months time ;-)

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -321,8 +321,17 @@ public void onTimeout() {
...
+							LOGGER.log(Level.FINE,
+									"Removed entry from observation store as observe response with token {0} didn't have observe option: ",
```
sophokles73: may I suggest something like
> Response to observe request with token {0} does not contain observe option, removing request from observation store

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
balsmn: I had the comment in the line below. May be your right to add it before the if loop.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
balsmn: Not a problem

balsmn:
@sophokles73 I have added few more comments as requested.

sophokles73:
Thanks @balsmn :-)
Merged to 2.0.x in 10c7d81c31442cd18e305fbb7331fcfc76cc77c5

PR #90, opened 2016-09-16
#82 UdpMatcher shouldn't remove exchange in case of multicast requests

KMax:


sophokles73:
Hi @KMax,

thanks for the PR but I cannot really review the changes because you have completely ignored the formatting rules and thus all code is marked as changed :-(
Please use the code formatting rules contained in the top level folder. Also, you will need to sign an Eclipse Contributor Agreement first and sign-off your commits. You may want to take a look at the CONTRIBUTING.md file contained in the top level folder as well ...

Issue #92, opened 2016-09-20
Request for change in the CoapClient


Rajsmit:
Hi,

We want to create a Request with setContentFormat()  and some data in the header using  "coapRequest.getOptions().addOption()" before calling observe. but the current exposed observe API only accept  Response handler and Accept option.

Could you please provide a way to pass the customized request in the observe API or make to below two API public.

`private CoapObserveRelation observeAndWait(Request request, CoapHandler handler)`

`private CoapObserveRelation observe(Request request, CoapHandler handler)`

Regards,
Rajesh Kumar.

sophokles73:
You should be able to use the *advanced* methods of `CoapClient` for that:
````
Request req = Request.newGet().setURI("coap://server:5683/resourcename");
req.setObserve();
// now make all customizations, e.g. add options ...
req.getOptions().addOption(...);
...
CoapClient client = new CoapClient();
Response resp = client.advanced(req); // block while waiting for response
...
````

Rajsmit:
Hi Kai,

Thanks for your reply.
The Observe API provides the CoapObserveRelation as return parameter that we want to use to cancel the observer based on some APP conditions. when we use the advance API then we will not get the CoapObserveRelation and this will require more large effort from our side.

Thanks,
Rajesh Kumar.

sophokles73:
Oh, I see. That makes sense. I currently do not really see why we shouldn't make the more generic methods public. Would you like to create a PR for that?

Rajsmit:
Hi Kai,

We created PR ( https://github.com/eclipse/californium/pull/94) for making observe API public.
Please check and let us know if it is fine.

Thanks,
Rajesh Kumar.

sophokles73:
fixed in f6fa90d376749b651a6a6c8b4accb31bc6cd6eb6

PR #94, opened 2016-09-21
Change the visibilityof  observeAndWait() and observe() to public

prafulbhatnagar:
Change the visibility of  observeAndWait() and observe() to public as per
disucssion on the issue https://github.com/eclipse/californium/issues/92

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -848,7 +848,7 @@ public class CoapClient {
 	 * @param handler the Response handler
 	 * @return the CoAP observe relation
 	 */
-	private CoapObserveRelation observeAndWait(Request request, CoapHandler handler) {
+	public CoapObserveRelation observeAndWait(Request request, CoapHandler handler) {
 		Endpoint outEndpoint = getEffectiveEndpoint(request);
 		CoapObserveRelation relation = new CoapObserveRelation(request, outEndpoint);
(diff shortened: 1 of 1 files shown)


californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -848,7 +848,7 @@ private Request ifNoneMatch(Request request) {
 	 * @param handler the Response handler
 	 * @return the CoAP observe relation
 	 */
-	private CoapObserveRelation observeAndWait(Request request, CoapHandler handler) {
+	public CoapObserveRelation observeAndWait(Request request, CoapHandler handler) {
 		Endpoint outEndpoint = getEffectiveEndpoint(request);
```
sophokles73: Since this is now public, we will need to make sure that the observe option is set.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -867,7 +867,7 @@ private CoapObserveRelation observeAndWait(Request request, CoapHandler handler)
 	 * @param handler the Response handler
 	 * @return the CoAP observe relation
 	 */
-	private CoapObserveRelation observe(Request request, CoapHandler handler) {
+	public CoapObserveRelation observe(Request request, CoapHandler handler) {
 		Endpoint outEndpoint = getEffectiveEndpoint(request);
```
sophokles73: see above

PR #95, opened 2016-09-21
Exchange shouldn't be removed if it with a multicast request

KMax:
This PR is an attempt to fix the support of multicast requests. It's related to #82 issue.

All the existing tests pass, but I need a recommendation how to create a new test for the changes in this PR.

Signed-off-by: Maxim Kolchin kolchinmax@gmail.com

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -72,6 +72,7 @@ public class UdpMatcher implements Matcher {
 	/* Health status output */
 	private final Level healthStatusLevel;
 	private final int healthStatusInterval; // seconds
+    private final int multicastKeepAliveTimeout; //seconds
 
 	private boolean started;
 
@@ -105,6 +106,11 @@ public class UdpMatcher implements Matcher {
 
 		healthStatusLevel = Level.parse(config.getString(NetworkConfig.Keys.HEALTH_STATUS_PRINT_LEVEL));
 		healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL);
+
+        //As defined in https://tools.ietf.org/html/rfc7390#section-2.5
+        multicastKeepAliveTimeout = config.getInt(NetworkConfig.Keys.NON_LIFETIME)
+                + config.getInt(NetworkConfig.Keys.MAX_LATENCY)
+                + config.getInt(NetworkConfig.Keys.MAX_SERVER_RESPONSE_DELAY);
 	}
 
 	@Override public synchronized void start() {
@@ -154,10 +160,10 @@ public class UdpMatcher implements Matcher {
 			request.setMID(currendMID.getAndIncrement() % (1 << 16)); // wrap at 2^16
 		}
 		// request MID is from the local namespace -- use blank address
-		KeyMID idByMID = new KeyMID(request.getMID());
+		final KeyMID idByMID = new KeyMID(request.getMID());
 
 		// ensure Token is set
-		KeyToken idByToken;
+		final KeyToken idByToken;
 		if (request.getToken() == null) {
 			idByToken = createUnusedToken();
 			request.setToken(idByToken.token);
@@ -175,6 +181,16 @@ public class UdpMatcher implements Matcher {
 
 		exchangesByMID.put(idByMID, exchange);
 		exchangesByToken.put(idByToken, exchange);
+
+		if(request.isMulticast()) {
+			executor.schedule(new Runnable() {
+                @Override
+                public void run() {
+                    exchangesByMID.remove(idByMID);
+                    exchangesByToken.remove(idByToken);
+                }
+			}, multicastKeepAliveTimeout, TimeUnit.MILLISECONDS);
+		}
 	}
 
 	@Override public void sendResponse(Exchange exchange, Response response) {
@@ -365,7 +381,7 @@ public class UdpMatcher implements Matcher {
 			}
 			// ignore response
 			return null;
-		} else if (isResponseRelatedToMulticastRequest(exchange, responseContext)) {
+		} else if (isResponseRelatedToMulticastRequest(exchange)) {
 			if (response.getType() == Type.ACK && exchange.getCurrentRequest().getMID() != response.getMID()) {
 				// The token matches but not the MID.
 				LOGGER.log(Level.FINER,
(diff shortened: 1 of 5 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -508,11 +529,12 @@ private KeyToken createUnusedToken() {
 				KeyMID idByMID = new KeyMID(exchange.getCurrentRequest().getMID());
 				KeyToken idByToken = new KeyToken(exchange.getCurrentRequest().getToken());
 
-				exchangesByToken.remove(idByToken);
-
-				// in case an empty ACK was lost
-				exchangesByMID.remove(idByMID);
+				if (!exchange.getRequest().isMulticast() || isMulticastTimedOut(exchange)) {
+					exchangesByToken.remove(idByToken);
```
sophokles73: This basically means that an exchange for a multicast request could stay in the map forever if no responses arrive *after* the exchange has timed out, doesn't it?

sophokles73: Instead of using the reception of a response as the trigger to finally remove the exchange from the token map, I think we should instead schedule a task for cleaning up the multicast exchange after 

`NON_LIFETIME + MAX_LATENCY + MAX_SERVER_RESPONSE_DELAY`

as indicated by RFC 7390. This clean-up task would need to be scheduled when the original multicast request is sent out.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -415,6 +424,18 @@ private boolean isResponseRelatedToRequest(final Exchange exchange, final Correl
 		}
 	}
 
+	private boolean isResponseRelatedToMulticastRequest(final Exchange exchange, final CorrelationContext responseContext) {
```
sophokles73: Why do we need this separate method for comparing the correlation context? Since multicast always uses plain UDP (never DTLS) the correlation context will never contain a SESSION_ID and thus the comparison is basically the same as the one in `isResponseRelatedToRequest`, right?

KMax: @sophokles73 I didn't know that. Will remove that check in next commit.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
KMax: Should the cleanup task be scheduled in `sendRequest(Exchange, Request)` method of `UdpMatcher`?

Where to get the `MAX_LATENCY` and `MAX_SERVER_RESPONSE_DELAY` values? Looks like `NetworkConfig` doesn't have such keys.

KMax: Defined them in `NetworkConfig` and `NetworkConfigDefaults` in compliance with https://tools.ietf.org/html/rfc7390#section-2.5.

KMax:
I realised that directly removing an exchange from the maps in `UdpMatcher` I can break something. So I dived deeper in the californium code and refactored my changes to call timed out methods which later cause removal of the exchange from the maps.

Also I added a simple unit test to check that the exchange is not removed on the first response.

How could a test checking the removal of the exchange by timeout be implemented?

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -175,6 +181,20 @@ public UdpMatcher(final NetworkConfig config) {
 
 		exchangesByMID.put(idByMID, exchange);
 		exchangesByToken.put(idByToken, exchange);
+
+		if(request.isMulticast()) {
+			if(executor != null) {
```
sophokles73: no need to check for null. We have already checked for that in `UdpMatcher.start()`

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -105,6 +106,11 @@ public UdpMatcher(final NetworkConfig config) {
 
 		healthStatusLevel = Level.parse(config.getString(NetworkConfig.Keys.HEALTH_STATUS_PRINT_LEVEL));
 		healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL);
+
+        //As defined in https://tools.ietf.org/html/rfc7390#section-2.5
+        multicastKeepAliveTimeout = config.getInt(NetworkConfig.Keys.NON_LIFETIME)
```
sophokles73: may I propose to call this `multicastMaxLifetime` instead?

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -365,6 +385,15 @@ public UdpMatcher(final NetworkConfig config) {
 			}
 			// ignore response
 			return null;
+		} else if (isResponseRelatedToMulticastRequest(exchange)) {
```
sophokles73: I don't think that we need this separate block which mostly duplicates the code from the `else if (isResponseRelatedToRequest) {` block.
Instead, we could simply add another case to that other block checking for a multicast request:
````
} else if (isResponseRelatedToRequest(exchange, responseContext)) {
  Exchange prev = deduplicator.findPrevious(idByMID, exchange);
  Exchange prev = deduplicator.findPrevious(idByMID, exchange);
  if (prev != null) {
    LOGGER.log(Level.FINER, "Received duplicate response for open exchange: {0}", response);
    response.setDuplicate(true);
  } else if (exchange.getCurrentRequest().isMulticast()) {
    // keep exchange open because we might receive additional responses
    LOGGER.log(Level.FINER, "Received response for multicast request: {0}", response);
  } else {
    idByMID = new KeyMID(exchange.getCurrentRequest().getMID());
    exchangesByMID.remove(idByMID);
    LOGGER.log(Level.FINE, "Closed open request [{0}]", idByMID);
  }
  if (response.getType() == Type.ACK && exchange.getCurrentRequest().getMID() != response.getMID()) {
    LOGGER.log(Level.WARNING,
        "Possible MID reuse before lifetime end for token [{0}], expected MID {1} but received {2}",
        new Object[] { response.getTokenString(), exchange.getCurrentRequest().getMID(),
        response.getMID() });
  }
  return exchange;
}
````

WDYT?

KMax: > I don't think that we need this separate block

But then only the first response to a multicast request will fall in the block you're suggesting and the next responses will be marked as duplicates, because `prev` variable won't be `null`. Is that what supposed to be?

sophokles73: You're right. What about:
````
} else if (isResponseRelatedToRequest(exchange, responseContext)) {
  if (exchange.getCurrentRequest().isMulticast()) {
    // keep exchange open because we might receive additional responses
    LOGGER.log(Level.FINER, "Received response for multicast request: {0}", response);
  } else {
    Exchange prev = deduplicator.findPrevious(idByMID, exchange);
    if (prev != null) {
      LOGGER.log(Level.FINER, "Received duplicate response for open exchange: {0}", response);
      response.setDuplicate(true);
    } else {
      idByMID = new KeyMID(exchange.getCurrentRequest().getMID());
      exchangesByMID.remove(idByMID);
      LOGGER.log(Level.FINE, "Closed open request [{0}]", idByMID);
    }
  }
  ...
````
BTW I have merged #98 and #102 so it would be nice if you could re-base your PR on master in order to pick up the changes ...

boaks:
Can we close this PR? It seems nobody is working on it

PR #96, opened 2016-09-22
Add support for session fail over between nodes.

sophokles73:
Introduced a SessionCache strategy interface for storing and retrieving
state of an established DTLS session to and from e.g. a persistent
store.
The session information can be shared by multiple nodes and e.g. be used
to resume a session with a client in case the original node that the
client has established the session with, fails.


balsmn:
I tested this PR with a sample MongoDB based implementation. I was trying to get the following scenario working with this PR (a simulated fail-over)

_1. A DTLS client connects successfully to a LWM2M Server with a life time > 90 seconds
2. Restart LWM2M Server after the Session gets stored in session_cache.
3. DTLS client after 90 seconds successfully sends a register_update._

At first try, step 3 failed as the APPLICATION_DATA was not recognized by the `DTLSConnector` without an active session. Then with @sophokles73 help, I set the `DTLSConnector.setResumptionRequired()` on LWM2M client to force them do an abbreviated handshake. After setting it step 3 worked. :+1: 

Initially my expectation was that the failover is transparent to LWM2M Client. This could have been possible when the `ConnectionStore` was persistent which could have prevented the LWM2M Clients to do this abbreviated handshake. But I remember @sophokles73 saying that persisting `Connection` which in turn holds `Handshaker` is a complex task. May be that lead to the `SessionCache`.

The only question remaining is when does the LWM2M Client decide when to perform this abbreviated handshake.

sophokles73:
Thanks @balsmn for putting it to a test :-)
You mention the possibility of *transparently* failing over a connection (not a session) by means of sharing **all** connection state via `ConnectionStore`. While I agree that this may work in theory, I do not think that it would be a feasible approach because you would need to update the persisted state with **every** message received or sent in order to keep the record counters in sync. This would introduce a lot of latency into the processing/sending of messages which may be acceptable for some use cases but IMHO will not work as a general strategy.

I would like to merge this PR if there are no further objections.
@sbernard31, @mkovatsc any further concerns?

sbernard31:
I have quick a look on this. It seems ok for me.
(I see you preparing the implementation of the 5077 RFC)

PR #97, opened 2016-09-29
Enhance observer

prafulbhatnagar:
Changed the visibility of  observeAndWait(request,handler) and observe(request,handler) to public as per
discussion on the issue https://github.com/eclipse/californium/issues/92 .

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -849,7 +849,7 @@ public class CoapClient {
 	 * @param handler the Response handler
 	 * 
 	 * @return the CoAP observe relation
-	 * @throws IllegalStateException if the observe option is not set in the
+	 * @throws IllegalArgumentException if the observe option is not set in the
 	 *             request
 	 */
 	public CoapObserveRelation observeAndWait(Request request, CoapHandler handler) {
@@ -876,7 +876,7 @@ public class CoapClient {
 	 * @param handler the Response handler
 	 * 
 	 * @return the CoAP observe relation
-	 * @throws IllegalStateException if the observe option is not set in the
+	 * @throws IllegalArgumentException if the observe option is not set in the
 	 *             request
 	 */
 	public CoapObserveRelation observe(Request request, CoapHandler handler) {
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -19,7 +19,8 @@
  *    Kai Hudalla (Bosch Software Innovations GmbH) - use Logger's message formatting instead of
  *                                                    explicit String concatenation
  ******************************************************************************/
-package org.eclipse.californium.core;
+
+ package org.eclipse.californium.core;
 
 import java.io.IOException;
 import java.net.URI;
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -840,39 +840,55 @@ public class CoapClient {
 		return request;
 	}
 	
-	/*
+	/**
 	 * Sends the specified observe request and waits for the response whereupon
 	 * the specified handler is invoked when a notification arrives.
 	 *
 	 * @param request the request
+	 * 
 	 * @param handler the Response handler
+	 * 
 	 * @return the CoAP observe relation
+	 * @throws IllegalStateException if the observe option is not set in the
+	 *             request
 	 */
 	public CoapObserveRelation observeAndWait(Request request, CoapHandler handler) {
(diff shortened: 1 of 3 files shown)


prafulbhatnagar:
Made the email ID in github profile public.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -840,39 +840,55 @@ private Request ifNoneMatch(Request request) {
...
+			Endpoint outEndpoint = getEffectiveEndpoint(request);
+			CoapObserveRelation relation = new CoapObserveRelation(request, outEndpoint);
+			request.addMessageObserver(new ObserveMessageObserverImpl(handler, relation));
+			CoapResponse response = synchronous(request, outEndpoint);
+			if (response == null || !response.advanced().getOptions().hasObserve())
+				relation.setCanceled(true);
+			relation.setCurrent(response);
+			return relation;
+		} else {
+			throw new IllegalStateException("please make sure that the request has observe option set.");
```
sophokles73: since this is a problem with the provided parameter value (`request`), we should better throw an `IllegalArgumentException'

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -840,39 +840,55 @@ private Request ifNoneMatch(Request request) {
...
+	 */
+	public CoapObserveRelation observe(Request request, CoapHandler handler) {
+		if (request.getOptions().hasObserve()) {
+			Endpoint outEndpoint = getEffectiveEndpoint(request);
+			CoapObserveRelation relation = new CoapObserveRelation(request, outEndpoint);
+			request.addMessageObserver(new ObserveMessageObserverImpl(handler, relation));
+			send(request, outEndpoint);
+			return relation;
+		} else {
+			throw new IllegalStateException("please make sure that the request has observe option set.");
```
sophokles73: see above

californium-core/src/test/java/org/eclipse/californium/core/test/ClientAsynchronousTest.java
```diff
@@ -157,6 +157,46 @@ public void testAsyncObserveTriggersOnLoad() throws Exception {
 		Thread.sleep(50);
 		assertThat(receivedNotifications.get(), is(3));
 	}
+	
+	/**
+	 * Test case for CoapClient.observe(Request request, CoapHandler handler).
+	 * 
+	 * @author Praful Bhatnagar (prafulbhatnagar@gmail.com)
```
sophokles73: please remove the `@author` tag. We can always see who added which code by means of the commit log

californium-core/src/test/java/org/eclipse/californium/core/test/ObserveTest.java
```diff
@@ -205,6 +205,52 @@ public void onError() { }
 		
 	}
 	
+	/**
+	 * Test case for CoapClient.observeAndWait(Request request, CoapHandler handler).
+	 * @author Praful Bhatnagar (prafulbhatnagar@gmail.com)
+	 * @throws Exception
+	 */
+	@Test
+	public void testObserveAndWait() throws Exception {
```
sophokles73: I think we are already (implicitly) testing the existing behavior of `CoapClient.observeAndWait` by means of the `testObserve` test case, aren't we? In this test case we should probably only make sure that we detect an illegal argument, i.e. a `Request` without the *observe* option set.

californium-core/src/test/java/org/eclipse/californium/core/test/ClientAsynchronousTest.java
```diff
@@ -157,6 +157,46 @@ public void testAsyncObserveTriggersOnLoad() throws Exception {
...
+	 * @throws Exception
+	 */
+	@Test
+	public void testAsyncObserveWithRequestTriggersOnLoad() throws Exception {
```
sophokles73: see my comment for the other test method you added. I think we can refrain to simply make sure to detect a missing *observe* option in the passed-in `Request` ...

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
prafulbhatnagar: make sense, would update and re-submit.

sophokles73:
Not all commits in your PR have a signed-off-by footer. I suggest you rebase the branch on current master and squash all commits into a single one (don't forget the "signed-off-by" footer ;-))

californium-core/src/test/java/org/eclipse/californium/core/test/ObserveTest.java
prafulbhatnagar: Create a new pull request, please review https://github.com/eclipse/californium/pull/100 ..

prafulbhatnagar:
Created a new PR https://github.com/eclipse/californium/pull/100

PR #98, opened 2016-09-29
Fix #63 

### LINKED ISSUE #63: Californium re-uses unacknowledged MIDs in outbound requests

sophokles73:
This is supposed to fix the problem with re-using MIDs in exchanges with the same peer before EXCHANGE_LIFETIME has expired.
The PR introduces a `MessageExchangeStore` that encapsulates generation & tracking of MIDs per peer and the registration & management of `Exchange` objects in order to eliminate race conditions.

In a next step we should integrate the `TokenProvider` proposed in PR #85 to make sure that we always use a unique token as well.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
@@ -238,7 +238,7 @@ public class CoapServer implements ServerInterface {
 			running = false;
 		}
 	}
-	
+
 	/**
 	 * Destroys the server, i.e., unbinds from all ports and frees all system resources.
 	 */
@@ -248,6 +248,9 @@ public class CoapServer implements ServerInterface {
 		LOGGER.info("Destroying server");
 		// prevent new tasks from being submitted
 		executor.shutdown(); // cannot be started again
+		for (Endpoint ep : endpoints) {
+			ep.destroy();
+		}
 		try {
 			// wait for currently executing tasks to complete
 			if (!executor.awaitTermination(1, TimeUnit.SECONDS)) {
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
@@ -92,22 +92,24 @@ import org.eclipse.californium.core.server.resources.Resource;
 public class CoapServer implements ServerInterface {
 
 	/** The logger. */
-	private final static Logger LOGGER = Logger.getLogger(CoapServer.class.getCanonicalName());
+	private static final Logger LOGGER = Logger.getLogger(CoapServer.class.getName());
 
 	/** The root resource. */
 	private final Resource root;
-	
+
+	private final NetworkConfig config;
+
 	/** The message deliverer. */
 	private MessageDeliverer deliverer;
-	
+
 	/** The list of endpoints the server connects to the network. */
 	private final List<Endpoint> endpoints;
-	
+
 	/** The executor of the server for its endpoints (can be null). */
 	private ScheduledExecutorService executor;
-	
-	private final NetworkConfig config;
-	
+
+	private boolean running;
+
 	/**
 	 * Constructs a default server. The server starts after the method
 	 * {@link #start()} is called. If a server starts and has no specific ports
@@ -126,7 +128,7 @@ public class CoapServer implements ServerInterface {
 	public CoapServer(final int... ports) {
 		this(NetworkConfig.getStandard(), ports);
 	}
-	
+
 	/**
 	 * Constructs a server with the specified configuration that listens to the
 	 * specified ports after method {@link #start()} is called.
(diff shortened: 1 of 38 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/SweepDeduplicator.java
```diff
@@ -23,101 +23,133 @@
...
 	 * KeyMID has not yet arrived, this method returns null, indicating that
 	 * the message with the KeyMID is not a duplicate. In this case, the
 	 * exchange is added to the deduplicator.
 	 */
-	public Exchange findPrevious(KeyMID key, Exchange exchange) {
-		Exchange previous = incommingMessages.putIfAbsent(key, exchange);
-		return previous;
+	@Override
+	public Exchange findPrevious(final KeyMID key, final Exchange exchange) {
+		return incomingMessages.putIfAbsent(key, exchange);
```
sbernard31: map.putIfAbsent doesn't exist in java 7 ? I believe we are still targeting java 7 right (hudson is use jdk 1.7)? if this is the case `incomingMessages` should be a `ConcurrentHashMap` not a `Map`.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -0,0 +1,397 @@
...
+ * A {@code MessageExchangeStore} that manages all exchanges in local memory.
+ *
+ */
+public class InMemoryMessageExchangeStore implements MessageExchangeStore {
+
+	private static final Logger LOGGER = Logger.getLogger(InMemoryMessageExchangeStore.class.getName());
+	private static final int MAX_TOKEN_LENGTH = 8; // bytes
+	private final Map<KeyMID, Exchange> exchangesByMID = new ConcurrentHashMap<>(); // for all
+	private final Map<KeyToken, Exchange> exchangesByToken = new ConcurrentHashMap<>(); // for outgoing
+	private final Map<KeyUri, Exchange> ongoingExchanges = new ConcurrentHashMap<>();
```
sbernard31: same remark about java 7 and `Map.putIfAbsent()`

sbernard31:
Not directly linked to this PR, but I tried Leshan with this changes and I found an API break on `org.eclipse.californium.scandium.DTLSConnector.DTLSConnector(DtlsConnectorConfig, ResumptionSupportingConnectionStore)` which has now `package` visibility. We use this in [Leshan](https://github.com/eclipse/leshan/blob/master/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/LeshanClient.java#L102)

californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/SweepDeduplicator.java
sophokles73: Thanks for pointing this out, Simon :-) We *are* still targeting Java 7. I will change the type to `ConcurrentMap` ...

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
```diff
@@ -204,40 +216,67 @@ public void start() {
...
-		for (Endpoint ep:endpoints)
-			ep.destroy();
+	public synchronized void destroy() {
+
+		LOGGER.info("Destroying server");
+		// prevent new tasks from being submitted
 		executor.shutdown(); // cannot be started again
+		for (Endpoint ep : endpoints) {
+			ep.destroy();
+		}
```
sophokles73: Oops :-o, will remove ...

sophokles73:
Do you really need to provide an implementation of `ResumptionSupportingConnectionStore` or would it also be sufficient to provide your implementation of `SessionCache`? After all, this is the intended way of supporting session persistence ...

sophokles73:
I have rebased on 2.0.x and incorporated the changes suggested by @sbernard31

sbernard31:
In fact we don't need to persist session we just need to clear the connectionStore after a bootstrap.

sophokles73:
Then maybe we should add a `DTLSConnector.clearConnectionStore()` to do that?

sbernard31:
sounds good

sophokles73:
However, wasn't there also a need to be able to mark *all* connections in the store as required to resume the session?

sbernard31:
This makes sense too.

PR #99, opened 2016-09-29
Fix block1 bigger block size negotiation

sbernard31:
The RFC 7959 [§2.3](https://tools.ietf.org/html/rfc7959#section-2.3)   say:
_"Finally, the SZX block size given in a control Block1 Option
         indicates the largest block size preferred by the server for
         transfers toward the resource that is **the same or smaller than
         the one used in the initial exchange**; the client SHOULD use
         this block size or a smaller one in all further requests in the
         transfer sequence, even if that means changing the block size
         (and possibly scaling the block number accordingly) from now
         on."_

Currently when a server asked for a bigger size (something not really supported by the spec) we restart transfer from first block (0).
```
CON [MID=18708, T=28d2], PUT, /test, 1:0/1/128    ----->
<-----   ACK [MID=18708, T=28d2], 2.31, 1:0/1/256    
CON [MID=18709, T=28d2], PUT, /test, 1:0/1/256    ----->
    ...  ....
```
 I'm not sure this is the best way to do, the spec also say :
_"A server MAY also return a 4.08 error code for any (final or non-final) Block1
   transfer that is not in sequence; therefore, clients that do not have
   specific mechanisms to handle this case SHOULD always start with
   block zero and send the following blocks in order."_
and 
_"Still, the client SHOULD heed the preference indicated and, for all further
   blocks, use **the block size preferred by the server or a smaller one**."_

I suppose a better solution should be to keep the original size block when it is smaller than the preferred size block.
```
CON [MID=18708, T=28d2], PUT, /test, 1:0/1/128    ----->
<-----   ACK [MID=18708, T=28d2], 2.31, 1:0/1/256    
CON [MID=18709, T=28d2], PUT, /test, 1:1/1/128    ----->
<-----   ACK [MID=18709, T=28d2], 2.31, 1:1/1/256    
CON [MID=18710, T=28d2], PUT, /test, 1:2/0/128    ----->
<-----   ACK [MID=18710, T=28d2], 2.04, 1:2/0/256    
```

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
@@ -21,15 +21,10 @@
  ******************************************************************************/
 package org.eclipse.californium.core.test.lockstep;
 
-import static org.eclipse.californium.core.coap.CoAP.Code.GET;
-import static org.eclipse.californium.core.coap.CoAP.Code.POST;
-import static org.eclipse.californium.core.coap.CoAP.Code.PUT;
+import static org.eclipse.californium.core.coap.CoAP.Code.*;
+import static org.eclipse.californium.core.coap.CoAP.ResponseCode.*;
+import static org.eclipse.californium.core.coap.CoAP.Type.*;
 import static org.eclipse.californium.core.coap.OptionNumberRegistry.OBSERVE;
-import static org.eclipse.californium.core.coap.CoAP.ResponseCode.CHANGED;
-import static org.eclipse.californium.core.coap.CoAP.ResponseCode.CONTENT;
-import static org.eclipse.californium.core.coap.CoAP.ResponseCode.CONTINUE;
-import static org.eclipse.californium.core.coap.CoAP.Type.ACK;
-import static org.eclipse.californium.core.coap.CoAP.Type.CON;
 import static org.eclipse.californium.core.test.lockstep.IntegrationTestTools.*;
 import static org.junit.Assert.*;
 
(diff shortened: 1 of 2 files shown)


californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
```diff
@@ -336,6 +331,110 @@ public void testSimpleAtomicBlockwisePUT() throws Exception {
...
+		server.expectRequest(CON, PUT, path).storeBoth("A").block1(0, true, 128).payload(reqtPayload.substring(0, 128)).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("A").block1(0, true, 32).go();
+
+		server.expectRequest(CON, PUT, path).storeBoth("B").block1(4, true, 32).payload(reqtPayload.substring(128, 160)).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("B").block1(4, true, 32).go();
+
+		server.expectRequest(CON, PUT, path).storeBoth("C").block1(5, true, 32).payload(reqtPayload.substring(160, 192)).go();
+		server.sendResponse(ACK, CHANGED).loadBoth("C").block1(5, true, 32).go();
+
+		server.expectRequest(CON, PUT, path).storeBoth("D").block1(6, false, 32).go();
```
sophokles73: shouldn't the last PUT also contain the remaining bytes (193 - 200) as payload?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
```diff
@@ -336,6 +331,110 @@ public void testSimpleAtomicBlockwisePUT() throws Exception {
...
+
+		server.expectRequest(CON, PUT, path).storeBoth("A").block1(0, true, 128).payload(reqtPayload.substring(0, 128))
+				.go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("A").block1(0, true, 256).go();
+
+		server.expectRequest(CON, PUT, path).storeBoth("B").block1(1, true, 128)
+				.payload(reqtPayload.substring(128, 256)).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("B").block1(1, true, 256).go();
+
+		server.expectRequest(CON, PUT, path).storeBoth("C").block1(2, false, 128).go();
```
sophokles73: again, what about the remaining payload (257 - 300)?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
sbernard31: You're right I will do that.

sbernard31:
Integrated in master too ( commit 32ffb297258a8a9b09d0d9e1c53b3eb05c014724, 780f1858e0832e6c7e5ba557ba91e3fca63a394a )

PR #100, opened 2016-09-30
Changed the visibility of observeAndWait() and observe() to public

prafulbhatnagar:
Changed the visibility of observeAndWait(request,handler) and
observe(request,handler) to public as per discussion on the issue
https://github.com/eclipse/californium/issues/92 ..

Both the methods now checks that passed request has observer set on it,
if the observer is not set it raises IllegalArgumentException.

Test case also has been added for both methods.

Signed-off-by: Praful Bhatnagar <prafulbhatnagar@gmail.com>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -840,39 +840,55 @@ public class CoapClient {
 		return request;
 	}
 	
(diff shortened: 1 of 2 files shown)


sophokles73:
merged in f6fa90d376749b651a6a6c8b4accb31bc6cd6eb6

prafulbhatnagar:
Thanks..

PR #113, opened 2016-10-26
Merge PR#85 token provider to 2.0.x

ingoschaal:
This PR merge the  #85 "Introduce TokenProvider to prevent clashes of CoAP tokens" to the actual 2.0.x branch. 
The #85 is re-factor and the TokenProvider is incorporate  into the InMemoryMessageExchangeStore

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
@@ -69,15 +69,8 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 	 * 
 	 */
 	public InMemoryMessageExchangeStore(final NetworkConfig config) {
-		if (config == null) {
-			throw new NullPointerException("Configuration must not be null");
-		}
-
-		this.config = config;
-		this.tokenProvider = new InMemoryRandomTokenProvider(this.config);
+		this(config, new InMemoryRandomTokenProvider(config));
 		LOGGER.log(Level.CONFIG, "using default TokenProvider {0}", InMemoryRandomTokenProvider.class.getName());
-		LOGGER.log(Level.CONFIG, "using tokens of {0} bytes in length",
-				config.getInt(NetworkConfig.Keys.TOKEN_SIZE_LIMIT));
 	}
 
 	/**
@@ -97,8 +90,6 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 		}
 		this.tokenProvider = tokenProvider;
 		this.config = config;
-		LOGGER.log(Level.CONFIG, "using tokens of {0} bytes in length",
-				config.getInt(NetworkConfig.Keys.TOKEN_SIZE_LIMIT));
 	}
 
 	private void startStatusLogging() {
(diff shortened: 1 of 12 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -61,21 +61,28 @@
...
-			this.config = config;
-			this.tokenLength = config.getInt(NetworkConfig.Keys.TOKEN_SIZE_LIMIT, MAX_TOKEN_LENGTH);
-			LOGGER.log(Level.CONFIG, "using tokens of {0} bytes in length", tokenLength);
 		}
+		if (tokenProvider == null) {
+			throw new NullPointerException("TokenProvider must not be null");
+		}
+		this.tokenProvider = tokenProvider;
+		this.config = config;
+		this.tokenLength = config.getInt(NetworkConfig.Keys.TOKEN_SIZE_LIMIT, MAX_TOKEN_LENGTH);
```
sophokles73: Do we still need the token length here? It seems that the `TokenProvider` retrieves it from the config itself, doesn't it?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -61,21 +61,28 @@
...
 	/**
 	 * Creates a new store for configuration values.
 	 * 
 	 * @param config the configuration to use.
+	 * @param tokenProvider the TokenProvider which provides CoAP tokens that
+	 *            are guaranteed to be not in use.
+	 * 
 	 */
-	public InMemoryMessageExchangeStore(final NetworkConfig config) {
+	public InMemoryMessageExchangeStore(final NetworkConfig config, TokenProvider tokenProvider) {
```
sophokles73: I think we should still have a constructor that only takes a `NetworkConfig` as argument. We should also provide a setter for `TokenProvider` and during startup we can then check if a token provider is set and if not simply instantiate an `InMemoryRandomTokenProvider`. This way we do not need to change the test cases in so many places ....

ingoschaal:
@sophokles73  thanks for fast review

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
ingoschaal: Yes, good point. I will add and push it again.
Thanks

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -49,20 +49,36 @@
...
 	private TokenProvider tokenProvider;
 	private SecureRandom secureRandom;
+	
+	/**
+	 * Creates a new store for configuration values.
+	 * 
+	 * @param config the configuration to use.
+	 * 
+	 */
+	public InMemoryMessageExchangeStore(final NetworkConfig config) {
```
sophokles73: simply do a `this(config, null)` and handle the `null` case appropriately in the other constructor?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -81,8 +97,8 @@ public InMemoryMessageExchangeStore(final NetworkConfig config, TokenProvider to
 		}
 		this.tokenProvider = tokenProvider;
 		this.config = config;
-		this.tokenLength = config.getInt(NetworkConfig.Keys.TOKEN_SIZE_LIMIT, MAX_TOKEN_LENGTH);
-		LOGGER.log(Level.CONFIG, "using tokens of {0} bytes in length", tokenLength);
+		LOGGER.log(Level.CONFIG, "using tokens of {0} bytes in length",
```
sophokles73: No need to log this here. Since the `TokenProvider` is using this config param, it should be logged out in the `InMemoryRandomTokenProvider` instead ...

PR #116, opened 2016-11-03
fixed and handled the request Code null pointer exception

navoda:
When a CoAP Client is performing a Ping, the request sends a [**null** Code](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java#L279) which eventually caught by [**DataSerializer**](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataSerializer.java#L51) class and throws a _NullPointerException_.


sophokles73:
Hi @navoda,

I am not sure about the *intended* status of this PR. The commit lacks a valid signed-off footer. However, I also do not think that it addresses the root of the problem, which seems to be ( as @mettke pointed out in #146) that the corresponding `Request` does not contain a request code. Adding an *empty* request code to the `CoAP.Code` enumeration looks like a good idea to solve this problem ... do you want to give that one a try and create a new PR based on the proposed solution?

sophokles73:
Will be handled as part of #147 and #149

PR #126, opened 2016-11-07
Workaround for test, if DNS is not working.

boaks:
If tests are executed in an environment, where DNS is not working,
replace destination hostname "iot.eclipse.org" with the own "hostname".

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/coap/RequestTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/coap/RequestTest.java
@@ -19,6 +19,7 @@ import static org.hamcrest.CoreMatchers.*;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
 
@@ -67,8 +68,21 @@ public class RequestTest {
 
 	@Test
 	public void testSetURISetsUriHostOption() throws UnknownHostException {
-		Request req = Request.newGet().setURI("coap://iot.eclipse.org");
-		assertThat(req.getOptions().getUriHost(), is("iot.eclipse.org"));
+		String host = "iot.eclipse.org";
+		
+		try {
+			/* check, id DNS is working */
+			InetAddress.getByName(host);
+			/* DNS OK! */
+		}
+		catch(UnknownHostException ex) {
+			/* DNS failed, so use own host name */
+			host = InetAddress.getLocalHost().getHostName();
+		}
+
+		Request req = Request.newGet().setURI("coap://" + host);
+		assertThat(req.getOptions().getUriHost(), is(host));
+
 	}
 
 	@Test


californium-core/src/test/java/org/eclipse/californium/core/coap/RequestTest.java
```diff
@@ -67,8 +68,21 @@ public void testSetURISetsDestination() throws UnknownHostException {
 
 	@Test
 	public void testSetURISetsUriHostOption() throws UnknownHostException {
-		Request req = Request.newGet().setURI("coap://iot.eclipse.org");
-		assertThat(req.getOptions().getUriHost(), is("iot.eclipse.org"));
```
jvermillard: tried to use "coap://localhost" in the original code?

boaks: If I use  "localhost.localdomain" (at literally), I still get (on windows)

java.lang.IllegalArgumentException: cannot resolve host name: localhost.localdomain
	at org.eclipse.californium.core.coap.Request.setURI(Request.java:306)
	at org.eclipse.californium.core.coap.Request.setURI(Request.java:232)
	at org.eclipse.californium.core.coap.RequestTest.testSetURISetsUriHostOption2(RequestTest.java:89)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:50)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:459)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:675)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:192)

jvermillard: yes doesn't work on windows, now I wonder why we have such special case for "localhost" in the setURI code

boaks: maybe "localhost" and "127.0.0.1" are mapped trivial, "iot.eclipse.org" to its IP address not. 
So, with "127.0.0.1" as address, the "URI HOST" doesn't provide more information.
With "???.???.???.???" as address, the "URI HOST" provides more information.

sophokles73: That's a good questio, @jvermillard. The whole code around creating a `Request` from a URI could probably use an overhaul based on [Section 6.4 of the CoAP spec](https://tools.ietf.org/html/rfc7252#section-6.4) ...

jvermillard: As far I understand uri-host is used as 'host' for HTTP virtual host.
Using "localhost" as a virtual host sounds legit to me. I would remove the "localhost" case in setURI and change the test to use "localhost".
If you strictly apply the RFC I think it's incorrect to remove uri-host when it's value is "localhost"

sophokles73: Regarding the problem this issue is all about: the host option only needs to be set in a `Request` if it is *not* the same literal IP address as the Request's destination address. The way I read this is that the usual case is that the URI contains a literal IP address. If not, we simply put the hostname into the host option (no need to resolve it).

boaks: But then the "destination" will be undefined, if the hostname is not resolved.

jvermillard: I would remove the special case for 'localhost' and test with 'localhost' in place of 'iot.eclipse.org'

sophokles73:
In any case, it seems that this PR currently only addresses symptoms instead of the cause.
I agree with @jvermillard that we must not process `localhost` in a specific way. We should probably try to resolve the hostname and, if that fails, leave the destination address unspecified. When the request is sent later, we can throw an `IllegalStateException`. This would require client code to either make sure that DNS works or explicitly set the destination IP address before sending.

boaks:
`In any case, it seems that this PR currently only addresses symptoms instead of the cause.`

Sure, but the real "cause" is not in the californium code :-).

My intention was just: 
leave the implementation as it is, because the majority is even not aware of a environment with limited DNS (maybe, I'm the only one). 
Changing the implementation instead of the tests my introduce more trouble for that majority. 
Therefore I really think, work around a symptom, which only occurs very rarely, instead of changing the implemention, possible effect the majority, is the better way.
 
Or are there any other known issues according the "URI Host"?

sophokles73:
I guess Julien's and my point is that the current implementation regarding the handling of the URI option is not correct anyways and we should fix this.

boaks:
OK, so I close the PR 
(and just use it locally to fix my limited environment until the URI host handling is fixed).

PR #127, opened 2016-11-07
Bugfix notify over TCP.

boaks:
First correction according UDP processing.
Introducing NotificationListener and ObservationStore.
Please Note: the implementation is rather complex and requires the
attention of the implementors of the "clustered observe" feature.


sophokles73:
This currently looks like large scale code duplication between UDP and TCP stack :-( I do not think that this is the way to go ...

boaks:
Sure, but the "duplication pattern" is not limited to the "extended" ???Matcher :-).

boaks:
OK, I will just use a work branch until its stable enough and offers proper quality to be merged to 2.0.x

PR #128, opened 2016-11-08
Testfix for race condition at the CoapObserveRelation relation.

boaks:
If the CoapObserveRelation is set to late (latch.getCount() already
reached 1), the test may fail. Therefore add a check, if the relation is
already set before using the latch.

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/MemoryLeakingHashMapTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/MemoryLeakingHashMapTest.java
@@ -364,11 +364,21 @@ public class MemoryLeakingHashMapTest {
 			this.cancelProactively = cancelProactively;
 		}
 
-		public void setObserveRelation(CoapObserveRelation relation) {
+		public synchronized void setObserveRelation(CoapObserveRelation relation) {
 			this.relation = relation;
 		}
 
 		public void onLoad(CoapResponse response) {
+			CoapObserveRelation relation;
+			synchronized (this) {
+				relation = this.relation;
+			}
+			
+			if (null == relation) {
+				LOGGER.log(Level.INFO, "Client ignore notification {0}: [{1}]", new Object[]{counter++, response.getResponseText()});
+				return;
+			}
+			
 			latch.countDown();
 			LOGGER.log(Level.FINE, "Client received notification {0}: [{1}]", new Object[]{counter++, response.getResponseText()});
 


californium-core/src/test/java/org/eclipse/californium/core/test/MemoryLeakingHashMapTest.java
```diff
@@ -364,11 +364,21 @@ public CoapObserverAndCanceler(final CountDownLatch latch, final AtomicBoolean e
...
 		}
 
-		public void setObserveRelation(CoapObserveRelation relation) {
+		public synchronized void setObserveRelation(CoapObserveRelation relation) {
 			this.relation = relation;
 		}
 
 		public void onLoad(CoapResponse response) {
+			CoapObserveRelation relation;
+			synchronized (this) {
```
sophokles73: This will not prevent `onLoad()` being invoked before `setObserveRelation()`, in which case one notification will be *lost* and the test case will not work correctly anymore.
IMHO we need to add `synchronized` blocks in the test cases where `CoapObserverAndCanceler` is instantiated and `CoapObserverAndCanceler.setObserveRelation` is called, e.g.:

````
@Test
public void testObserveBlockwise() throws Exception {
  final String uri = uriOf(PIGGY);
  LOGGER.log(Level.FINE, "Test observe relation with blockwise notifications {0}", uri);

  // We need a long response text (>16) 
  currentResponseText = LONG_RESPONSE;

  CountDownLatch latch = new CountDownLatch(HOW_MANY_NOTIFICATION_WE_WAIT_FOR + 1);
  AtomicBoolean isOnErrorInvoked = new AtomicBoolean();

  CoapClient client = new CoapClient(uri);
  client.setEndpoint(clientEndpoint);
  CoapObserverAndCanceler handler = new CoapObserverAndCanceler(latch, isOnErrorInvoked, uri, true);
  synchronized (handler) {
    CoapObserveRelation rel = client.observe(handler);
    handler.setObserveRelation(rel);
  }

  // Wait until we have received all the notifications and canceled the relation
  assertTrue(
    "Client has not received all expected responses",
    latch.await(HOW_MANY_NOTIFICATION_WE_WAIT_FOR * OBS_NOTIFICATION_INTERVAL + 500, TimeUnit.MILLISECONDS));
    assertFalse(isOnErrorInvoked.get()); // should not happen
}
````

Then we make `CoapObserverAndCanceler.onLoad()` also `synchronized`. That way we prevent it from being invoked before the observe relation is set.

WDYT?

sophokles73: The latch is always initialized with a given number (of notifications to expect) in the corresponding test cases.
I guess the tests work properly as long as the time to wait for the latch to count down is sufficiently long, which sems to be the case with the *generously* added 500 ms and a notification interval of 50 ms ....

PR #129, opened 2016-11-08
Bugfix TCP schemes.

boaks:
Adjust schemes for received request according UPD or TCP usage.
Add common scheme dependent getDefaultEndpoint(String scheme) to EndpointManager. 
Adjust Request and CoapClient to use the scheme dependent default endpoint.
Fixing Request to consider TCP endpoints also (not only UDP).

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
@@ -22,16 +22,9 @@
  *    Kai Hudalla (Bosch Software Innovations GmbH) - use Logger's message formatting instead of
  *                                                    explicit String concatenation
  *    Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add getDefaultEndpoint(String scheme)
  ******************************************************************************/
 package org.eclipse.californium.core.network;
-
-import org.eclipse.californium.core.CoapServer;
-import org.eclipse.californium.core.coap.Request;
-import org.eclipse.californium.core.coap.Response;
-import org.eclipse.californium.core.network.config.NetworkConfig;
-import org.eclipse.californium.core.server.MessageDeliverer;
-import org.eclipse.californium.elements.tcp.TcpClientConnector;
-import org.eclipse.californium.elements.tcp.TlsClientConnector;
 
 import java.io.IOException;
 import java.net.InetAddress;
@@ -40,37 +33,57 @@ import java.net.SocketException;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.LinkedList;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import org.eclipse.californium.core.CoapServer;
+import org.eclipse.californium.core.coap.CoAP;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.coap.Response;
+import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.server.MessageDeliverer;
+import org.eclipse.californium.elements.tcp.TcpClientConnector;
 
 /**
  * A factory for {@link Endpoint}s that can be used by clients for sending
  * outbound CoAP requests.
  *
  * The EndpointManager contains the default endpoint for coap (on port 5683) and
- * coaps (CoAP over DTLS). When an application serves only as client but not
- * server it can just use the default endpoint to send requests. When the
+ * endpoint for other schemes (coaps, CoAP over DTLS, coap+tcp, COAP over TCP,
+ * coaps+tcp, COAPS over TCP,). When an application serves only as client but
+ * not server it can just use the default endpoint to send requests. When the
  * application sends a request by calling {@link Request#send()} the send method
  * sends itself over the default endpoint.
  * <p>
  * To make a server listen for requests on the default endpoint, call
- * <pre>{@code
- *  CoapServer server = new CoapServer();
- * }</pre>
+ * 
+ * <pre>
+ * 
+ * {
+ * 	CoapServer server = new CoapServer();
+ * }
+ * </pre>
+ * 
  * or more explicit
- * <pre>{@code
- *  Endpoint endpoint = EndpointManager.getEndpointManager().getDefaultEndpoint();
- *  CoapServer server = new CoapServer();
- *  server.addEndpoint(endpoint);
- * }</pre>
+ * 
+ * <pre>
+ * 
+ * {
+ * 	Endpoint endpoint = EndpointManager.getEndpointManager().getDefaultEndpoint();
+ * 	CoapServer server = new CoapServer();
+ * 	server.addEndpoint(endpoint);
+ * }
+ * </pre>
  */
 public class EndpointManager {
 
 	/** The logger */
-	private final static Logger LOGGER = Logger.getLogger(EndpointManager.class.getCanonicalName());
+	private static final Logger LOGGER = Logger.getLogger(EndpointManager.class.getCanonicalName());
 
 	/** The singleton manager instance */
-	private static EndpointManager manager = new EndpointManager();
+	private static final EndpointManager manager = new EndpointManager();
 
 	/**
 	 * Gets the singleton manager.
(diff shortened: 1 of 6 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -94,6 +95,36 @@ public static EndpointManager getEndpointManager() {
...
+	 * 
+	 * @param scheme scheme to select endpoint
+	 * @return the default endpoint for the provided scheme
+
+	 * @see #getDefaultEndpoint()
+	 * @see #getDefaultSecureEndpoint()
+	 * @see #getDefaultTcpEndpoint()
+	 * @see #getDefaultSecureTcpEndpoint()
+	 */
+	public Endpoint getDefaultEndpoint(String scheme) {
```
sophokles73: I'd rather keep the *default* endpoints in a `Map<String, Endpoint>` with the *scheme* as keys so that we can add more endpoints and do not need to do the string comparisons.

I also think that the *Default* should be dropped form all of `EndpointManager`'s methods. The whole purpose of the `EndpointManager` is to deal with the default endpoints, no need to explicitly mention this in the method names again. And we are already inconsistent there as well, e.g. `setSecureTcpEndpoint()` ...

We probably do not need the `getDefault???Endpoint()` methods any more, do we? We could keep a `getEndpoint()` method returning the standard coap endpoint but otherwise let client code use a `getEndpoint(String scheme)` to get an endpoint for the given scheme.

WDYT?

boaks: We need to check, how many test are using the "specialized" methods.
I tried such e replace in the "LeshanServer", but it affects a lot of tests.

I will check, how many changes would be required additionally.

boaks:
Refreshed.

I replaced 

```
if (!uri.startsWith("coap://") && !uri.startsWith("coaps://") && !uri.startsWith("coap+tcp://")
					&& !uri.startsWith("coaps+tcp://")) {

```

in Request with

```
if (!uri.contains("://")) {
```

I think, "http://" is handled by `uri.contains("://")` better, because it delegates the check to `setOptions(URI)`

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -48,29 +52,41 @@
...
  * sends itself over the default endpoint.
  * <p>
  * To make a server listen for requests on the default endpoint, call
- * <pre>{@code
- *  CoapServer server = new CoapServer();
- * }</pre>
+ * 
+ * <pre>
+ * {
+ * 	&#064;code
```
sophokles73: Do we need this within `<pre>`?

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -358,9 +258,12 @@ public void deliverRequest(Exchange exchange) {
 
 		@Override
 		public void deliverResponse(Exchange exchange, Response response) {
-			if (exchange == null) throw new NullPointerException();
-			if (exchange.getRequest() == null) throw new NullPointerException();
-			if (response == null) throw new NullPointerException();
+			if (exchange == null)
+				throw new NullPointerException("no CoAP exchange!");
```
sophokles73: Does it make sense to throw an exception here? It would probably be better to log the fact that there is no exchange and then ignore the response.

WDYT?

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
boaks: I don't need the `<pre>`. but its more a relict and just listed as difference caused be source formatting. 
Remove or replace by the old one?

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
boaks: In my opinion, the API requires a exchange and a response as precondition. 
Therefore the Exception signalling that something went wrong seems to be the right.
Or what do you have in mind, how should it be used without (accidentley?) ?

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
sophokles73: You need the `<pre>` in order to keep the line breaks in the generated HTML. I suggest removing the `{@code}` ...

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
sophokles73: True, the question is more: do we expect that this method will ever be invoked with null argument(s), given that it is basically an internal API only? I have no clue, though, whether this can actually happen ;-)

boaks: Only on misstaken implementation changes! Therefore I would vote for staying with the exceptions.

sophokles73: Ok

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -48,29 +52,38 @@
...
+ * 	CoapServer server = new CoapServer();
+ * 	server.addEndpoint(endpoint);
+ * }
+ * </pre>
  */
 public class EndpointManager {
 
 	/** The logger */
-	private final static Logger LOGGER = Logger.getLogger(EndpointManager.class.getCanonicalName());
+	private final static Logger LOGGER = Logger.getLogger(EndpointManager.class
```
sophokles73: this should be `private static final`

you seem to be using shorter lines than defined by our Eclipse code formatting rules. Are you using them (contained in project root folder)?

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -48,29 +52,38 @@
...
+			.getCanonicalName());
 
 	/** The singleton manager instance */
-	private static EndpointManager manager = new EndpointManager();
+	private final static EndpointManager manager = new EndpointManager();
```
sophokles73: same here

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
sophokles73: Can you also add curly braces around the statements while you're at it?

boaks:
```
Are you using the Eclipse code formatter settings we have included in the project root folder?
In particular you seem to use very short lines ...
```

Which one, `eclipse-cleanup-profile.xml` or `eclipse-formatter-profile.xml`?

I usually try not to "reformat" the code, even it was not proper formated before.
This makes it easier to follow up the changes. Ohterwise we should use 2 commits:
1. reformatting
2. changes

boaks:
I will change it on monday :-) !

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
boaks: Hm,  `eclipse-formatter-profile.xml` doesn't do this for me?
So it's intended to be manually?

sophokles73: Maybe it is missing from the formatter rules. Then we should add it :-)

boaks:
So, I "reformated" the EndpointManager, but left the other ony slightly touched classes unformated.

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -81,235 +92,112 @@ public static EndpointManager getEndpointManager() {
...
-	 * {@link ClientMessageDeliverer} rejects it. To receive requests, the
-	 * endpoint must be added to an instance of {@link CoapServer}. Be careful with
-	 * stopping or destroying the default endpoint as it affects all messages
-	 * that are supposed to be sent over it.
-	 *
-	 * @return the default endpoint
+	 * Get default endpoint for provided scheme.
+	 * 
+	 * @param uriScheme scheme to select endpoint
+	 * @return the default endpoint for the provided scheme
```
sophokles73: @throws IllegalArgumentException ?

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -81,235 +92,112 @@ public static EndpointManager getEndpointManager() {
...
+	 * 
+	 * @see #getDefaultEndpoint()
 	 */
-	public synchronized Endpoint getDefaultEndpoint() {
-		if (default_endpoint == null) {
-			createDefaultEndpoint();
+	public synchronized Endpoint getDefaultEndpoint(String uriScheme) throws IllegalArgumentException {
```
sophokles73: we do not declare `RuntimeException`s being thrown

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -81,235 +92,112 @@ public static EndpointManager getEndpointManager() {
...
+	 * the endpoint, if its not already started. Stop the old endpoint.
+	 * 
+	 * @param newEndpoint new default endpoint for the supported scheme.
+	 * @throws IllegalArgumentException, if the uri scheme is not supported or
+	 *             the endpoint is null.
 	 */
-	public synchronized void setTcpEndpoint(Endpoint endpoint) {
-		if (this.default_tcp_endpoint != null) {
-			this.default_tcp_endpoint.destroy();
+	public synchronized void setDefaultEndpoint(Endpoint newEndpoint) throws IllegalArgumentException {
```
sophokles73: we do not declare `RuntimeException`s being thrown

However, you should throw a `NullPointerException` instead of an `IllegalArgumentException` in case of the *newEndpoint* being `null` (and add a throws tag to the JavaDoc).

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -330,16 +218,24 @@ public synchronized void setDefaultSecureEndpoint(Endpoint endpoint) {
...
-			it.default_tcp_endpoint.clear();
+		for (Endpoint endpoint : it.endpoints.values()) {
+			endpoint.clear();
+		}
+	}
+
+	/**
+	 * Clear all endpoints. Stop them.
+	 */
+	public static void clearEndpoints() {
```
sophokles73: The naming of this method is a little confusing. If I am not mistaken, this method is about *removing*/*destroying* all default endpoints, not about clearing their internal state, right?

PR #130, opened 2016-11-09
Testfix irritating failure logging.

boaks:
I got that failure logging from the unit tests (once, not frequently):

Failed tests: 
  ObserveClientSideTest.testBlockwiseObserve:200 Expected empty message
but received
org.eclipse.californium.core.network.serialization.UdpDataParser@3a5ed7a6

That's irritating, because, sure, the parser isn't received.
Therefore replace the parser with the received message in the failure
logging. So maybe, next time we see what sporadically went wrong.

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
@@ -185,6 +185,10 @@ public class LockstepEndpoint {
 					Assert.assertEquals("Wrong MID:", mid, message.getMID());
 					print("Correct MID: "+mid);
 				}
+
+				public String toString() {
+					return "Expected MID: " +mid;
+				}
 			});
 			return this;
 		}
@@ -196,6 +200,11 @@ public class LockstepEndpoint {
 					Assert.assertEquals("Wrong MID:", expected, message.getMID());
 					print("Correct MID: "+expected);
 				}
+
+				public String toString() {
+					int expected = (Integer) storage.get(var);
+					return "Expected MID: " + expected;
+				}
 			});
 			return this;
 		}
@@ -206,6 +215,10 @@ public class LockstepEndpoint {
 					Assert.assertEquals("Wrong type:", type, message.getType());
 					print("Correct type: "+type);
 				}
+
+				public String toString() {
+					return "Expected type: " + type;
+				}
 			});
 			return this;
 		}
@@ -215,6 +228,9 @@ public class LockstepEndpoint {
 				public void check(Message message) {
 					org.junit.Assert.assertArrayEquals("Wrong token:", token, message.getToken());
 					print("Correct token: "+Utils.toHexString(token));
+				}
+				public String toString() {
+					return "Expected token: " + Utils.toHexString(token);
 				}
 			});
 			return this;
@@ -228,6 +244,10 @@ public class LockstepEndpoint {
 					Assert.assertEquals("Wrong payload length: ", expectedLength, actualLength);
 					Assert.assertEquals("Wrong payload:", payload, message.getPayloadString());
 					print("Correct payload ("+actualLength+" bytes):\n"+message.getPayloadString());
+				}
+
+				public String toString() {
+					return "Expected payload: '"+ payload + "'";
 				}
 			});
 			return this;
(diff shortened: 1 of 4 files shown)


boaks:
Smells, that even on localhost interface, the order has changed from sending (see log) to receiving
(see expected test).

boaks:
After analysing these failures, they are caused by message order exchanges (which are possible using UDP).

e.g.:
		server.expectRequest(CON, GET, path).storeBoth("F").block2(1, false, 16).go();
 
may be received in changed order.

boaks:
This may be fixed by replacing UDP and handle over the messages directly obey their order.
(If wanted, vote for this.)

boaks:
Or this may be fixed by adding a "unordered expectation", which handles the reordering.
(If wanted, vote for this.)

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
```diff
@@ -316,6 +384,42 @@ public void check(Message message) {
 			for (Expectation<Message> expectation:expectations)
 				expectation.check(message);
 		}
+		
+		@Override
+		public void go() throws Exception {
+			if (null != multi) {
+				add(multi);
```
sophokles73: couldn't this just be
````
if (null != multi) {
  multi.add(this);
  return;
}
````
and we drop the abstract `add(MultiMessageExpectation)` method?

File: californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java [hunk 2]
```diff
@@ -316,6 +384,42 @@ public void check(Message message) {
...
+			if (0 <= end && ',' == result.charAt(end)) {
+				result.setLength(end);
+			}
+			result.append("}");
+			return result.toString();
+		}
+
+		public abstract void go(Message msg) throws Exception;
+
+		public abstract void add(MultiMessageExpectation multi);
```
sophokles73: see comment above
my impression is that we can get rid of this method ...

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
```diff
@@ -572,17 +669,99 @@ public EmptyMessageExpectation(Type type, int mid) {
...
+		@Override
+		public void add(MultiMessageExpectation multi) {
+			multi.add(this);
+		}
+	}
+
+	public class MultiMessageExpectation implements Action {
+
+		private int counter;
+		private EmptyMessageExpectation emptyExpectation;
```
sophokles73: is there a particular reason why we only allow one message per type? Couldn't these be Lists instead?

boaks: My first impression was, that calling "MessageExpectation,check(Message)" is not free of side effects.
There is logging output and the "store" methods. So just iterate over all possible "Expectations" and call "check" to find one, which succeeds, seem for me to introduce some more noise.
Therefore I chose to use the tpyed approach to select the message check.
This will only work, as long as only ACKs and Responses are assumed to be send "at once" without further "step" interaction.

boaks:
Added some more "multi expectations".
Code view check to ensure, that no critical sequences of expectations is left.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
boaks: My impression was, that this is related to have a "untyped set of expectations".
Or am I wrong?
Staying with "typed expectations",  the methods build something very similar to the "visitor" pattern.
A alternative solution (still using the "typed expectations") would then check the instanceof the MessageExpectations.
Is that wanted?

sophokles73: Ok, now I see what you mean. I think you're right, we should keep it like this in order to prevent ugly instanceof checks ...

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
sophokles73: You are probably right in that the overall interaction becomes more and more undeterministic once we allow for even more messages to be exchanged before starting the check.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
boaks: Thanks!

PR #131, opened 2016-11-11
Use NONE as mid for TCP messages.

boaks:
Prevents accidently usage of MID.
(Alternatively introducing a TCP_MID -2 would distinguish even between empty and TCP.)
This is also very helpfull in analysing logs in mixed endpoints.


californium-core/src/main/java/org/eclipse/californium/core/network/serialization/TcpDataParser.java
```diff
@@ -44,6 +46,6 @@ public MessageHeader parseHeader(final DatagramReader reader) {
 		byte token[] = reader.readBytes(tokenLength);
 
 		// No MID/Type/VERSION in TCP message. Use defaults.
-		return new MessageHeader(CoAP.VERSION, CoAP.Type.CON, token, code, 0, 0);
+		return new MessageHeader(CoAP.VERSION, CoAP.Type.CON, token, code, Message.NONE, 0);
```
joemag1: What is Message.NONE - don't see it in this PR.

Issue #104, opened 2016-10-06
Missing Request/Response Session Check with DTLS Usage on CoAP Server Side


DanielMaier-BSI:
Californium implements in UdpMatcher.receiveResponse() discard of responses that do not belong to the same DTLS session as the request on CoAP-Client side.

This implements the requirement of CoAP specification:

“The following rules are added for matching a response to a request: 
The DTLS session MUST be the same, and the epoch MUST be the same.”

But there is no such such mechanism for this requirement on server side, i.e. a CoAP Server may send a response to a newly established DTLS session that comes from the same IP:Port as the request. To accomplish the RFC it is also mandatory to ensure that no response gets sent to a DTLS session that is not the same as the one of the request.

Example Use Case:
- There is a CoAP-Server using DTLS with server certificate but without required Client authentication.
- There is a CoAP-Client client1 with an established DTLS session that observes a resources of this server.
- A second CoAP-Client client2 initiates a DTLS session using the same IP:Port as client1, e.g. they are behind the same NAT and the NAT devices reuses port mappings after inactivity.
- Now client2 receives notifies (responses) that were intended for client1. Client2 can decrypt theses notifies because the CoAP-Server encrypts them for the newly created DTLS session. These  responses might leak sensitive information.

I tried this out with the given example. But this example is of course not comprehensive. I think this issue also exists when clients do client authentication as long as the client has e.g. a valid PSK.

sophokles73:
I think that Californium (currently) is exposed to this problem. However, in the LWM2M use case it is usually the CoAP server (the device = LWM2M client) that is behind the NAT, whereas the CoAP client (the LWM2M server) will usually keep the same IP:port.
I therefore think that there is no imminent problem for a LWM2M scenario where the LWM2M clients initiate the DTLS session.

DanielMaier-BSI:
I agree with you that this issue is not that critical for LWM2M use case. Only responses of Register, Udpate, deregister are affected.

(Even if someone gets control over LWM2M Server IP, the attacker needs to know the PSKs of clients (in case of using PSKs), Server certificate etc. to initiate new DTLS Session to LWM2M clients)

sbernard31:
I think this is a general issue. A scandium user is not able to know who we send a message to.
I mean the element connector abstraction just provide a IP/port address as destination.

Main DTLS implementation does not face this issue because, end users manipulate DTLS session.  In this case, you can control exactly what you do.

With scandium and element connector, the choice was to abstract this session/connection concept.
[...]

PR #132, opened 2016-11-11
Introduce conditional message sending for TCP based connections.

boaks:
Some messages must only be sent, if no new connection must be
established. Use CON/NON to determine, if conditional sending is used.
(Required e.g. for "cancel observe" or "response".)
(Cleanup  on losing the connection is right now not implemented.)

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -25,6 +25,8 @@
  *                                      of Response(s) to Request (fix GitHub issue #1)
  *    Bosch Software Innovations GmbH - adapt message parsing error handling
  *    Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - use CON/NON to setup conditional sending
+ *                                                    for messages over TCP
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -69,6 +71,7 @@ import org.eclipse.californium.core.server.MessageDeliverer;
 import org.eclipse.californium.elements.Connector;
 import org.eclipse.californium.elements.CorrelationContext;
 import org.eclipse.californium.elements.MessageCallback;
+import org.eclipse.californium.elements.MessageNotSentCallback;
 import org.eclipse.californium.elements.RawData;
 import org.eclipse.californium.elements.RawDataChannel;
 import org.eclipse.californium.elements.UDPConnector;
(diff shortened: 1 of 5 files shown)


joemag1:
Can you give a little more detail about what problem this PR is trying to solve. I can somewhat understand notification call back when sending response on closed connection in server mode. But I don't fully get why you would ever want to not open a connection in client mode.

boaks:
draft-ietf-core-coap-tcp-tls-05

    2.5. Message Transmission
    CoAP requests and responses are exchanged asynchronously over the
    TCP/TLS connection. A CoAP client can send multiple requests without 
    waiting for a response and the CoAP server can return responses in
    any order. Responses MUST be returned over the same connection as
    the originating request. Concurrent requests are differentiated by
    their Token, which is scoped locally to the connection.

But I get aware, that the CorrelationContext may be a better candidate todo so.
I have to investigate. This PR is therefore more to get a first experience using
coap+tcp including boserve/notify and leshan.

@kai: after a week working on that, I think it would be a very good idea, to have a branch for the tcp work. could your ceate one?

joemag1:
Can this part of the spec be met by closing all open exchanges when TCP
connection is terminated? That part is already in the spec.

On Tue, Nov 15, 2016 at 9:16 AM Achim Kraus <notifications@github.com>
wrote:

draft-ietf-core-coap-tcp-tls-05

2.5. Message Transmission
CoAP requests and responses are exchanged asynchronously over the
TCP/TLS connection. A CoAP client can send multiple requests without
waiting for a response and the CoAP server can return responses in
any order. Responses MUST be returned over the same connection as
the originating request. Concurrent requests are differentiated by
their Token, which is scoped locally to the connection.

But I get aware, that the CorrelationContext may be a better candidate todo
so.
I have to investigate. This PR is therefore more to get a first experience
using
coap+tcp including boserve/notify and leshan.

@kai <https://github.com/kai>: after a week working on that, I think it
would be a very good idea, to have a branch for the tcp work. could your
ceate one?

—
You are receiving this because you commented.


Reply to this email directly, view it on GitHub
<https://github.com/eclipse/californium/pull/132#issuecomment-260704957>,
or mute the thread

sophokles73:
Assuming that what @boaks is trying to achieve is the following:
Prevent a server from implicitly opening a new connection to a client when sending a response, e.g. a notification, to an already disconnected client (because of mobile network loss etc).

If this is the case then I agree with @joemag1 that we should try to invalidate/terminate all exchanges/observations affected by the connection loss (is there a way to detect such a connection loss without having to actively send out a packet?)

boaks:
`Can this part of the spec be met by closing all open exchanges when TCP connection is terminated? That part is already in the spec.`

Can we agree, that cite on the spec, we mention the section and copy the original text?
So, do you mean:
draft-ietf-core-coap-tcp-tls-05, Section 3.5 Closing the Connection,

```
All requests for which the CoAP client has not received a response
yet are cancelled when the connection is closed.

```
 
If so, I think we should raise an issue, that this should also be considered for the CoAP server.
And cancel "all request of a connection" may introduce race conditions, if responses are already convereded into RawData. So, yes, cancel all request when the connection is closed will improve the efficiency. But I see still the requirement for this "conditional send" (but it should be bound to the correlation context).

boaks:
https://github.com/core-wg/coap-tcp-tls/issues/80

joemag1:
Yes, there is a small chance of race. So maybe we need both - flag to
prevent new connections, and callback to cancel open exchanges.

On no new connections, what do you think about adding an explicit flag
(allow-new-connection) to rawmessage. That way, I feel the code will be
easier to reason.

On clearing exchanges: there is a netty call back on when a connection is
dropped. The challenge is getting that callback into CoAP stack, so that
the interface is still compatible with UDP.

boaks:
`On no new connections, what do you think about adding an explicit flag (allow-new-connection) to rawmessage. That way, I feel the code will be easier to reason.`

This was my first approach. But then I thought, a indicator interface would change less. 
And then I got aware, that using the CorrelationContext seems to be the superior idea.
It also helps on issue #104. 

`On clearing exchanges: there is a netty call back on when a connection is dropped. The challenge is getting that callback into CoAP stack, so that the interface is still compatible with UDP.`

Thanks for the netty infos.

sophokles73:
@boaks,

status of this PR? Needs rebasing at least, I guess ...

boaks:
Uups, sorry I forgot to close it. I think, the solution using the correlation context would be supiror. I'm still working on that.

PR #133, opened 2016-11-11
Reduce code duplications in CoapStack.

boaks:
Introduce BaseCoapStack with common functionality.
Add TcpAdaptionLayer and TcpObserverLayer.
Uses CON/NON for conditional sending over TCP. 
(based on PR#132)

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
@@ -18,24 +18,17 @@
  *    Kai Hudalla - logging
  *    Kai Hudalla (Bosch Software Innovations GmbH) - use Logger's message formatting instead of
  *                                                    explicit String concatenation
+ *    Achim Kraus (Bosch Software Innovations GmbH) - move common function to BaseCoapStack
  ******************************************************************************/
 package org.eclipse.californium.core.network.stack;
 
-import java.util.List;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import org.eclipse.californium.core.coap.EmptyMessage;
-import org.eclipse.californium.core.coap.Request;
-import org.eclipse.californium.core.coap.Response;
-import org.eclipse.californium.core.network.Exchange;
 import org.eclipse.californium.core.network.Outbox;
-import org.eclipse.californium.core.network.Exchange.Origin;
 import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.core.server.MessageDeliverer;
 import org.eclipse.californium.elements.Connector;
-
 
 /**
  * The CoAPStack builds up the stack of CoAP layers that process the CoAP
@@ -45,55 +38,48 @@ import org.eclipse.californium.elements.Connector;
  * The class <code>CoapStack</code> builds up the part between the Stack Top and
  * Bottom.
  * <hr><blockquote><pre>
- * +-----------------------+
- * | {@link MessageDeliverer}      |
- * +-------------A---------+
- *               A
- *             * A
- * +-----------+-A---------+
- * |       CoAPEndpoint    |
- * |           v A         |
- * |           v A         |
- * | +---------v-+-------+ |
- * | | Stack Top         | |
- * | +-------------------+ |
+ * +--------------------------+
+ * | {@link MessageDeliverer}         |
+ * +--------------A-----------+
+ *                A
+ *              * A
+ * +------------+-A-----------+
+ * |       CoAPEndpoint       |
+ * |            v A           |
+ * |            v A           |
+ * | +----------v-+---------+ |
+ * | | Stack Top            | |
+ * | +----------------------+ |
  * | | {@link ExchangeCleanupLayer} | |
- * * | +-------------------+ |
- * | | {@link ObserveLayer}      | |
- * | +-------------------+ |
- * | | {@link BlockwiseLayer}    | |
- * | +-------------------+ |
- * | | {@link ReliabilityLayer}  | |
- * | +-------------------+ |
- * | | Stack Bottom      | |
- * | +---------+-A-------+ |
- * |           v A         |
- * |         Matcher       |
- * |           v A         |
- * |       Interceptor     |
- * |           v A         |
- * +-----------v-A---------+
- *             v A 
- *             v A 
- * +-----------v-+---------+
- * | {@link Connector}             |
- * +-----------------------+
+ * | +----------------------+ |
+ * | | {@link ObserveLayer}         | |
+ * | +----------------------+ |
+ * | | {@link BlockwiseLayer}       | |
+ * | +----------------------+ |
+ * | | {@link ReliabilityLayer}     | |
+ * | +----------------------+ |
+ * | | Stack Bottom         | |
+ * | +----------+-A---------+ |
+ * |            v A           |
+ * |          Matcher         |
+ * |            v A           |
+ * |        Interceptor       |
+ * |            v A           |
+ * +------------v-A-----------+
+ *              v A 
+ *              v A 
+ * +------------v-+-----------+
+ * | {@link Connector}                |
+ * +--------------------------+
  * </pre></blockquote><hr>
  */
-public class CoapUdpStack implements CoapStack {
+public class CoapUdpStack extends BaseCoapStack {
 
 	/** The LOGGER. */
 	private final static Logger LOGGER = Logger.getLogger(CoapStack.class.getCanonicalName());
 
-	private List<Layer> layers;
-	private Outbox outbox;
-	private StackTopAdapter top;
-	private StackBottomAdapter bottom;
-	private MessageDeliverer deliverer;
-
 	public CoapUdpStack(NetworkConfig config, Outbox outbox) {
-		this.top = new StackTopAdapter();
-		this.outbox = outbox;
+		super(outbox);
 
 		ReliabilityLayer reliabilityLayer;
 		if (config.getBoolean(NetworkConfig.Keys.USE_CONGESTION_CONTROL) == true) {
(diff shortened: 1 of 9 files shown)


boaks:
CON/NON is not longer used for conditional sending over TCP.
(Therefore not logner based on PR#132)
The solution for conditional sending using the correlation context affects more and will be put in a separate PR.

boaks:
I think, that this cleanup is worth to be merged,or at least to be revisited.
(its the 1. part, the 2. related to the Matchers will come with the bugfix_tcp_notify after #129 was merged, because it depends on that).

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpAdaptionLayer.java
```diff
@@ -0,0 +1,80 @@
...
+
+	@Override
+	public void sendRequest(Exchange exchange, Request request) {
+		if (null == request.getType()) {
+			request.setType(Type.CON);
+		}
+		super.sendRequest(exchange, request);
+		// CoAP over TCP does not have acknowledgements. Everything is
+		// automatically acknowledged.
+		request.setAcknowledged(true);
```
sophokles73: This could make client code believe that the message has actually been received by the server by the time this property is set. Client code might even have registered a listener waiting for this property to become true. Since there is no acknowledgement in coap-over-tcp we might better want to always keep this property set to false instead to prevent this.

WDYT?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpAdaptionLayer.java
```diff
@@ -0,0 +1,80 @@
...
+
+	@Override
+	public void sendResponse(final Exchange exchange, Response response) {
+		if (null == response.getType()) {
+			response.setType(Type.CON);
+		}
+		super.sendResponse(exchange, response);
+		// CoAP over TCP does not have acknowledgements. Everything is
+		// automatically acknowledged.
+		response.setAcknowledged(true);
```
sophokles73: see sendRequest()

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpObserveLayer.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.logging.Logger;
+
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.coap.Response;
+import org.eclipse.californium.core.network.Exchange;
+import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.observe.ObserveRelation;
+
+public class TcpObserveLayer extends AbstractLayer {
```
sophokles73: It would probably be worth describing, why we need a *TCP* flavored `ObserveLayer`.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpObserveLayer.java
```diff
@@ -0,0 +1,70 @@
...
+			} else {
+				response.setLast(false);
+			}
+		} // else no observe was requested or the resource does not allow it
+		super.sendResponse(exchange, response);
+	}
+
+	@Override
+	public void receiveResponse(Exchange exchange, Response response) {
+		if (response.getOptions().hasObserve() && exchange.getRequest().isCanceled()) {
```
sophokles73: This is considered erroneous behavior of the server, right? According to A.3 of coap-over-tcp the client (this side) MUST actively cancel the observation by means of a GET request. This is what would have had to happen when the Request has been canceled, right?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpObserveLayer.java
boaks: Because the UDP `ObserveLayer` depends in `sendResponse` on 

> getType() == Type.NON

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpObserveLayer.java
boaks: > This is considered erroneous behavior of the server, right? 

It may also occur as "race condition". Until the CoAP client received the response for the observer canel request, there may be already transmitted notifies in the TCP connection.

Right now, there are is some stuff left, to provideda really good implemetation for observes over tcp. (also the conditional send stuff, or the connection lost).

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpAdaptionLayer.java
sophokles73: That's exactly my concern: that someone depends on it when in fact there isn't even a message type defined in coap-over-tcp and therefore makes false unjustified assumptions.
Wouldn't it be better to *let it crash* so that we at least see where these *false assumptions* are being made?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpObserveLayer.java
sophokles73: My point was more that this class would deserve a JavaDoc comment :-)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpAdaptionLayer.java
boaks: I'm just not sure, who is using that behavior.

Therefore, if configurable, someone, who complains, can enable it (even knowing, that it's more a fake and ust for temporary usage.)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpObserveLayer.java
boaks: OK

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpAdaptionLayer.java
boaks: @joemag1 
Can you comment on that?
After reflecting that, I'm wondering, if that auto-ack was introduced in early development state of the tcp implementation the suppress retransmission. If so, we can remove it.
Usually the ACK on receiving the response should be enough in my opinion.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
```diff
@@ -18,82 +18,74 @@
...
- * |           v A         |
- * |           v A         |
- * | +---------v-+-------+ |
- * | | Stack Top         | |
- * | +-------------------+ |
+ * <hr>
+ * <blockquote>
+ * 
+ * <pre>
+ * +----------------------------------+
```
sophokles73: You probably had good intentions when re-aligning the ASCII graphics. However, I think the result will not look like you expected it to do. The additional *whitespace* characters in lines containing a `{@link}` tag are there in order to make up for the `{@link }` characters which are not visible in the produced HTML ...

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ObserveLayer.java
```diff
@@ -160,45 +164,47 @@ public void receiveEmptyMessage(Exchange exchange, EmptyMessage message) {
...
-		
+
 		@Override
 		public void onAcknowledgement() {
 			synchronized (exchange) {
 				ObserveRelation relation = exchange.getRelation();
 				final Response next = relation.getNextControlNotification();
-				relation.setCurrentControlNotification(next); // next may be null
+				relation.setCurrentControlNotification(next); // next may be
+																// null
```
sophokles73: ???

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -50,15 +49,16 @@
...
 	private int ack_timeout;
 	private float ack_random_factor;
 	private float ack_timeout_scale;
 	private int max_retransmit;
 
 	/**
-	 * Constructs a new reliability layer.
-	 * Changes to the configuration are observed and automatically applied.
+	 * Constructs a new reliability layer. Changes to the configuration are
+	 * observed and automatically applied.
```
sophokles73: Actually, I think we should get rid of this, i.e. we should not observe the Configuration object for changes. IMHO the configuration should be static through the run-time of an Endpoint. If the config is supposed to be changed then the endpoint needs to be stopped and re-created with a new config object.

WDYT?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
boaks: OK, I revert that :-)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ObserveLayer.java
boaks: So, which line length is used in californium?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
boaks: On a first view, it's simply not working!
 
        NetworkConfig 
	public NetworkConfig set(String key, Object value) {
		if (key == null) {
			throw new NullPointerException("key must not be null");
		} else if (value == null) {
			throw new NullPointerException("value must not be null");
		} else {
			properties.put(key, String.valueOf(value));
			for (NetworkConfigObserver obs:observers) {
				obs.changed(key, value);
			}
			return this;
		}
	}

only calls the `changed(... Object)` method. 
The `NetworkConfigObserverAdapter `doesn't convert to calls of  the others. So `int`  and `float` are never called.

if they would have been called, I have no idea, how this was intended to be "thread safe".

So I think, remove it completly (but in an other PR).

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ObserveLayer.java
sophokles73: 120

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
sophokles73: +1 for removing it completely and +1 for doing it in a separate PR

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ObserveLayer.java
boaks: OK, and the maximum line widht for comments :-) 
(in my californium style its 80)

PR #135, opened 2016-11-11
Introduce message flavor to distinguish serialized message bytes.

boaks:
Message caches already serialized bytes. But, though the serialization
is depending on TCP or UDP flavor, this caching may fail. Introducing a
message flavor and check the flavor prior using the cached bytes fixes
possible resulting issues.


sophokles73:
What issues are you anticipating? Since we an `Endpoint` always either has a TCP or a UDP stack, IMHO there should not be any confusion regarding the serialization format of a message being referenced in any of the stacks, it is either TCP or UDP but there will not be both types being processed in one stack.
Or am I missing something?

boaks:
I still think, this solution, using the caching for Message.bytes, is better then always serialize/parse on every notify.

The current fix of #141 must be applied also to other usage (outside californium).

Issue #146, opened 2016-11-19
CoapClient.ping() is throwing a NullPointerException


mettke:
I'm trying to use the following Snipped to ping a Coap Device which is started on my Machine:
```java
String url = "coap://localhost:5683/";
CoapClient client = new CoapClient(url);
Log.debug(String.valueOf(client.ping()));
```
The Log always gives false no matter whether there is a device at the given address or not and the following Exception is thrown:
```
SEVERE: Exception in protocol stage thread: null
java.lang.NullPointerException
	at org.eclipse.californium.core.network.serialization.DataSerializer.serializeRequest(DataSerializer.java:49)
	at org.eclipse.californium.core.network.CoapEndpoint$OutboxImpl.sendRequest(CoapEndpoint.java:606)
	at org.eclipse.californium.core.network.stack.CoapUdpStack$StackBottomAdapter.sendRequest(CoapUdpStack.java:222)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendRequest(AbstractLayer.java:65)
	at org.eclipse.californium.core.network.stack.ReliabilityLayer.sendRequest(ReliabilityLayer.java:114)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendRequest(AbstractLayer.java:65)
	at org.eclipse.californium.core.network.stack.BlockwiseLayer.sendRequest(BlockwiseLayer.java:155)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendRequest(AbstractLayer.java:65)
	at org.eclipse.californium.core.network.stack.ObserveLayer.sendRequest(ObserveLayer.java:51)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendRequest(AbstractLayer.java:65)
	at org.eclipse.californium.core.network.stack.ExchangeCleanupLayer.sendRequest(ExchangeCleanupLayer.java:14)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendRequest(AbstractLayer.java:65)
	at org.eclipse.californium.core.network.stack.CoapUdpStack$StackTopAdapter.sendRequest(CoapUdpStack.java:181)
	at org.eclipse.californium.core.network.stack.CoapUdpStack$StackTopAdapter.sendRequest(CoapUdpStack.java:175)
	at org.eclipse.californium.core.network.stack.CoapUdpStack.sendRequest(CoapUdpStack.java:121)
[...]

PR #134, opened 2016-11-11
Bugfix tcp notify

boaks:
Collection of PRs to make notifies over TCP experimental working.
TODO: cleanup exchanges when TCP connection is closed or lost.

PR #147, opened 2016-11-21
Corrected NullPointerException for CoapClient#ping()

mettke:
* DataSerializer used request.getCode().value which lead to a NullPointerException using CoapClient.ping()
* QuickFix: Request#getRawCode returns 0 instead of a NullPointerException if Request#code is null

Fixes: eclipse/californium#146
Closes: eclipse/californium#116

Signed-off-by: Marc Mettke <marc@itmettke.de>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataSerializer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataSerializer.java
@@ -93,7 +93,7 @@ public abstract class DataSerializer {
 		byte[] body = serializeOptionsAndPayload(request);
 
 		MessageHeader header = new MessageHeader(CoAP.VERSION, request.getType(), request.getToken(),
-				request.getCode().value, request.getMID(), body.length);
+				request.getRawCode(), request.getMID(), body.length);
 		serializeHeader(writer, header);
 		writer.writeBytes(body);
 
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
@@ -146,7 +146,7 @@ public class Request extends Message {
 
 	@Override
 	public int getRawCode() {
-		return code.value;
+		return code == null ? 0 : code.value;
 	}
 
 	/**


californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataSerializer.java
```diff
@@ -93,7 +93,7 @@ public final RawData serializeRequest(final Request request, final MessageCallba
 		byte[] body = serializeOptionsAndPayload(request);
 
 		MessageHeader header = new MessageHeader(CoAP.VERSION, request.getType(), request.getToken(),
-				request.getCode().value, request.getMID(), body.length);
+				request.getRawCode(), request.getMID(), body.length);
```
sophokles73: This is a great change :-) IMHO it would now be sufficient to change `Request.getRawCode()` to return 0 if no `CoAP.Code` is set ... we wouldn't need to add the `EMPTY` value to the enum ...

WDYT?

mettke:
@sophokles73 
To be honest, I don't know if that's a good idea. 
I guess that was the approach of the one who implemented CoapClient.ping() this first time. After a while new Code was added and in the end it looked like it is currently and was throwing a NullPointerException.
IMHO it is never a good idea to use null for an Expected behaviour, 'cause you never know who is not expecting that it might be null. 
Even though I know that adding EMPTY as new enum value may add others Problems like the asynchron request which works using an Observer for Response but not for CoapClient we might not have thought off.

Anyway I'm happy to make the change if we decide to do it that way

sophokles73:
Yes, I guess you're right. Always having a code set in `Request` will make it easier to know what to expect for everybody. I wonder whether we should actually prevent the constructor(s) from accepting `null` codes...

mettke:
I've added a second commit which introduces a way to check whether Arguments are null. Maybe that goes in the direction of what you thought about.
I've also changed Code#valueOf. Nearly forgot about that...

boaks:
RFC7252, 4.3, Table 1, 
+----------+-----+-----+-----+-----+
| | CON | NON | ACK | RST |
+----------+-----+-----+-----+-----+
| Request | X | X | - | - |
| Response | X | X | X | - |
| Empty | * | - | X | X |
+----------+-----+-----+-----+-----+

If a ping s a empty message, why not try to add it there?

boaks:
I guess, the missbehaviour was introduced by the redesign of the serialization (tcp)?
And maybe, a junit test would help to ensure, that stays working.

mettke:
We could do that using Message instead of Request in the CoapClient. The Problem is that a User would be able to send a Ack or a Rst Message. 
```java
public CoapResponse advanced(Message request) {
	assignClientUriIfEmpty(request);
	return synchronous(request);
}
```

Another way would be to use a Message only in the ping command with something similar to this:
```java
public boolean ping(long timeout) {
	try {
		Message request = new Message(Code.EMPTY, Type.CON);
		request.setToken(new byte[0]);
		request.setURI(uri);
		request.send().waitForResponse(timeout);
		return request.isRejected();
	} catch (InterruptedException e) {
		// waiting was interrupted, which is fine
	}
	return false;
}
```
Both options would take a lot of work, 'cause there is no sending implemented into a Message. 'Till now they we're only used on the Server side and therefore that stuff is implemented in the Request.
I guess it would definitly be to much for my understanding of californium.

boaks:
I've checked old archives.

	public byte[] serializeRequest(Request request) {
		writer = new DatagramWriter();
		Code code = request.getCode();
		serializeMessage(request, code == null ? 0 : code.value);
		return writer.toByteArray();
	}

So, `null` in a Request was just `0` (EMPTY).

But, even if it's changing the UdpMatcher, I think the implementation closest to the RFC7252 would be extending the EmptyMessage. May be not right now. So for fixing, I would just propose to insert

     code == null ? 0 : code.value 

again.

sophokles73:
IMHO the right way to really address the problem at the root would be to extend `EmptyMessage` with e.g. a `newPing(InetSocketAddress)` factory method and then re-factor `CoapClient` to send the empty message instead of mis-using a request. This should not be too hard to do. I guess we would simply need to add some private methods for using `Endpoint.sendEmptyMessage()` to send out the empty ping message.

@mettke: I would really like to support you in doing this change if you are motivated to do it. However, I could also understand if you do not want to spend the (additional) time. In that case we could close the PR and I can make the change.

WDYT?

boaks:
As far as I checked it yesterday, the methods for sending EmptyMessages are "already" in
CoapEndpoint, CoapTcpStack, CoapUdpStack,TcpMatcher, UdpMatcher, but the implementations only consider ACK/RST.

And just to say, those are exactly the classes for PRs #133 and #134.

Therefore I would propose to first/temporary use to old "hack".

sophokles73:
@boaks,

I do not really see what your point is. Of course `Endpoint.sendEmptyMessage()` is already in place. I was talking about using this method to actually send out a ping as an `EmptyMessage` instead of using `Endpoint.sendRequest()` to send a ping as a an *empty* request.

boaks:
> I do not really see what your point is. 

Sure, but I think "all visited" methods must be checked to handle such an outgoing EmptyMessage, especially the matcher, which must do the matching/exchange stuff.

Currently only ACK/RST seems to be considered in the implementations for EmptyMessage.

sophokles73:
Sure, we need to make sure that the ping *exchange pattern* actually is handled correctly by the stack. After taking a first look I believe it will be...

boaks:
> ...

Easy? May be :-).

But please again, this affects the classes for PRs #133 and #134.
And thought the code deduplication (UDP/TCP) is done by moving common code, it will not be easy to follow up parallel changes. 

Therefore:
Old hack and the solution after finishing the code deduplication.

mettke:
Guys, what do you think of waiting till #133 and #134 are merged, before we continue this discussion? I mean it is not that urgent, is it?

The Alternativ would be to use:
```java
@Override
public int getRawCode() {
	return code == null ? 0 : code.value;
}

public final byte[] getByteArray(final Request request) {
	...
	MessageHeader header = new MessageHeader(CoAP.VERSION, request.getType(), request.getToken(),
			request.getRawCode(), request.getMID(), body.length);
	...
}
```
and a new Issue for adding an EmptyMessage to the EmptyMessages. 
Guess it would be better then adding the Hack directly to DataSerializer.java

sophokles73:
Sounds good to me. If you make these changes I will merge this PR.
I will also open a new issue for doing the *real* fix later.

mettke:
@sophokles73 Changes are made.
I'm happy to assist you at the upcoming issue. So if you want me too, just mention me.

PR #148, opened 2016-11-22
Add correlation context and principal to tcp connectors.

boaks:
Extract correlation context and principal.


element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
```diff
@@ -136,17 +148,22 @@ public RawData(byte[] data, InetSocketAddress address, Principal clientIdentity,
...
 	 * @param data the data that is to be sent or has been received
-	 * @param address the IP address and port the data is to be sent to or has been received from
-	 * @param clientIdentity the identity of the authenticated sender of the message
-	 *     (or <code>null</code> if sender is not authenticated)
-	 * @param correlationContext additional information regarding the context the message has been
-	 *      received in. The information contained will usually come from the transport layer, e.g.
-	 *      the ID of the DTLS session the message has been received in, and can be used to correlate
-	 *      this message with another (previously send) message.
-	 * @param multicast indicates whether the data represents a multicast message
+	 * @param address the IP address and port the data is to be sent to or has
```
boaks: What is the current "preferred" line length in californium?

element-connector/src/main/java/org/eclipse/californium/elements/tcp/NettyContextUtils.java
```diff
@@ -0,0 +1,171 @@
...
+	 */
+	public static CorrelationContext buildCorrelationContext(Channel channel) {
+		String id = channel.id().asShortText();
+		SslHandler handler = getSslHandler(channel);
+		if (null != handler) {
+			SSLEngine sslEngine = handler.engine();
+			SSLSession sslSession = sslEngine.getSession();
+			if (null != sslSession) {
+				byte[] sessionId = sslSession.getId();
+				if (null != sessionId && 0 < sessionId.length) {
```
boaks: Does someone has experience, if `SSLEngine.getHandshakeStatus()` works better to check, if the session is ready?

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
```diff
@@ -59,10 +67,11 @@ public TcpClientConnector(int numberOfThreads, int connectTimeoutMillis, int idl
 		this.numberOfThreads = numberOfThreads;
 		this.connectionIdleTimeoutSeconds = idleTimeout;
 		this.connectTimeoutMillis = connectTimeoutMillis;
-		this.listenUri = URI.create(String.format("%s://127.0.0.1:0", getSupportedScheme()));
+		this.listenUri = URI.create(String.format("%s://0.0.0.0:0", getSupportedScheme()));
```
boaks: replace the "localhost" with "any". "localhost" is usually only for loopback and not reachable for other host.

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
```diff
@@ -74,39 +83,52 @@ public TcpClientConnector(int numberOfThreads, int connectTimeoutMillis, int idl
...
+			@Override
+			public void operationComplete(Future<Channel> future) throws Exception {
 				if (future.isSuccess()) {
 					Channel channel = future.getNow();
+
+					CorrelationContext context = NettyContextUtils.buildCorrelationContext(channel);
 					try {
 						channel.writeAndFlush(Unpooled.wrappedBuffer(msg.getBytes()));
+						if (null == context) {
+							NettyContextUtils.futureCorrelationContext(channel, msg);
```
boaks: See comment above. With more netty knowledge we may change the order to first establis a TLS session and then send data, which would make this obsolete.

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
sophokles73: 120

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
sophokles73: I think that is the whole point since this is a client connector only that initiates connections to servers only but cannot itself serve connections initiated by others.

boaks: I think, the `listenURI` is not really used (except the scheme) ...

boaks: >  then we can leave it as it is :-)

Let's use the provided address. If this is "localhost", OK, if it's "any", also OK.
But always use "127.0.0.1" is in my opinion missleading.

boaks:
Pleas also wit on this PR. I will provide unit test.

boaks:
@sophokles73 

So now ready for discussion :-).
Tests are added. I switched to used the demo-certs.
Currently (still) only x.509 is supported for TLS.

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
```diff
@@ -59,10 +68,12 @@ public TcpClientConnector(int numberOfThreads, int connectTimeoutMillis, int idl
 		this.numberOfThreads = numberOfThreads;
 		this.connectionIdleTimeoutSeconds = idleTimeout;
 		this.connectTimeoutMillis = connectTimeoutMillis;
-		this.listenUri = URI.create(String.format("%s://127.0.0.1:0", getSupportedScheme()));
+		this.listenUri = URI.create(String.format("%s://%s:%d", getSupportedScheme(),
```
boaks: I really think, that "127.0.0.1" is missleading. Using that will result in "0.0.0.0".

element-connector/pom.xml
```diff
@@ -21,11 +22,23 @@
 			<artifactId>junit</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<!-- https://mvnrepository.com/artifact/org.hamcrest/hamcrest-library -->
+		<dependency>
+			<groupId>org.hamcrest</groupId>
+			<artifactId>hamcrest-library</artifactId>
+			<version>1.3</version>
```
sophokles73: I would like to add this dependency to the parent POM's managed dependencies and omit the version here. This library can then be used all other modules as well without the need to supply the version.

element-connector/pom.xml
```diff
@@ -21,11 +22,23 @@
...
+		</dependency>
 		<dependency>
 			<groupId>io.netty</groupId>
 			<artifactId>netty-all</artifactId>
 			<version>4.1.1.Final</version>
 		</dependency>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>demo-certs</artifactId>
+			<version>${project.version}</version>
```
sophokles73: see above, `demo-certs` is already used in other modules as well, so we should make it a managed dependency as well.

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
```diff
@@ -26,7 +26,8 @@
 import java.util.Arrays;
 
 /**
- * A container object for the data received or sent via a <code>Connector</code>.
+ * A container object for the data received or sent via a <code>Connector</code>
+ * .
```
sophokles73: I don't think that the trailing period should be on a new line ...

boaks:
Thats no problem.
The formatting changes are from applying the provided formatter.
(On other PR you ment, that if someone is on it, he should change the stuff tp apply the provided style.  :-)

sophokles73:
Yes, but what I meant was: fix the obvious formatting violations in the *area* of code you are making changes in, not: re-factor the whole class ;-)

element-connector/pom.xml
boaks: > I would like to add

So, you do it on your own,
or should I move it?

boaks: And I forgot to set the scope ;-(.

boaks: OK, I did it, you may check it, if I push ...

sophokles73: ???

boaks: I changed it localy, but I'm not ready with the other stuff.

boaks:
Should 

     @Override public

stay or could it be changed to

     @Override 
     public

if the class is changed?
Or only modified methods?
Or should we work with two commits?
1. reformat
2. changes

element-connector/src/main/java/org/eclipse/californium/elements/tcp/NettyContextUtils.java
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	/**
+	 * Byte array to hexadecimal string.
+	 * 
+	 * @param byteArray byte array to be converted to string
+	 * @param max maximum bytes to be converted.
+	 * @return hexadecimal string
+	 */
+	private static String toHexString(byte[] byteArray, int max) {
```
boaks: duplicate from scandium. Maybe we move it to californium core?

pom.xml
```diff
@@ -258,19 +271,15 @@
...
-					why this makes sense :-)
-					We can control whether we want to deploy to the Eclipse repo or Maven Central
-					by a combination of the version being a SNAPHOT or release version and property
-					skipStaging=true/false.
-					In any case we can take advantage of the plugin's "deferred deploy" feature which
-					makes sure that all artifacts of a multi-module project are deployed as a whole
-					at the end of the build process instead of deploying each module's artifacts
-					individually as part of building the module.
-				 -->
+				<!-- Use the Nexus Staging plugin as a full replacement for the standard
```
boaks: Hmm,  this looks just that a lot of stuff is not formatted according the formatter ;-(.
This make really a lot unnecessary work.

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
```diff
@@ -330,4 +330,16 @@ public boolean isSecure() {
 		return (correlationContext != null &&
 				correlationContext.get(DtlsCorrelationContext.KEY_SESSION_ID) != null);
 	}
+	
+	/**
+	 * Callback, when context gets available. Used on sending an message.
+	 * 
+	 * @param context established context to be forwarded to the callback.
+	 */
+	public void onContextEstablished(CorrelationContext context) {
```
sophokles73: we need either this method or `getMessageCallback()`, I guess.
Can you check whether `getMessageCallback()` is used for any other purpose than notifying it about the context?
If that is the case I would rather go with `onContextEstablished()` ...

element-connector/src/main/java/org/eclipse/californium/elements/TcpCorrelationContext.java
```diff
@@ -0,0 +1,50 @@
...
+ *                                      matching messages (fix GitHub issue #1)
+ ******************************************************************************/
+package org.eclipse.californium.elements;
+
+/**
+ * A correlation context that explicitly supports TCP specific properties.
+ */
+public class TcpCorrelationContext extends MapBasedCorrelationContext {
+
+	public static final String KEY_CONNECTION_ID = "CONNECTION_ID";
```
sophokles73: Where does this come from?

element-connector/src/main/java/org/eclipse/californium/elements/TlsCorrelationContext.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.eclipse.californium.elements.util.StringUtil;
+
+/**
+ * A correlation context that explicitly supports TLS specific properties.
+ * Currently not aware of renegotiation (API to acquire missing).
+ */
+public class TlsCorrelationContext extends TcpCorrelationContext {
+
+	public static final String KEY_SESSION_ID = "SESSION_ID";
```
sophokles73: Can't we reuse the keys defined in `DtlsCorrelationContext` somehow`

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TlsServerConnector.java
```diff
@@ -58,6 +59,7 @@ public TlsServerConnector(KeyManagerFactory keyManagerFactory, InetSocketAddress
 
 	@Override protected void onNewChannelCreated(Channel ch) {
 		SSLEngine sslEngine = sslContext.createSSLEngine();
+		sslEngine.setWantClientAuth(true);
```
sophokles73: For `DTLSConnector` this is configurable. I do not know whether you have given configuration of the TLS connector much thought yet ...

element-connector/src/test/java/org/eclipse/californium/elements/tcp/TcpConnectorTest.java
```diff
@@ -24,150 +21,154 @@
...
-                stream.write(buffer.array());
-            }
-
-            stream.write(1); // GET
-            stream.write(data);
-            stream.flush();
-            return new RawData(stream.toByteArray(), address);
-        }
-    }
+public class TcpConnectorTest extends BaseConnectorTest {
```
sophokles73: We try to avoid subclassing in test cases ...

scandium-core/pom.xml
```diff
@@ -10,11 +10,12 @@
 	<packaging>bundle</packaging>
 	<name>Scandium (Sc) Core</name>
 	<description>A library for exchanging data using DTLS over UDP</description>
-	
+
 	<dependencies>
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
+			<scope>test</scope>
```
sophokles73: no need to define the scope for managed deps if it does not differ ...

element-connector/src/main/java/org/eclipse/californium/elements/tcp/NettyContextUtils.java
boaks: OK, but for me this seems to extend the scope of this PR too much.
Separate PR?

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
boaks: It may be usefull, if the correlation conntext is not yet available, and the aquire it later has some overhead. I just want to reduce the always same "check & call".

element-connector/src/main/java/org/eclipse/californium/elements/TcpCorrelationContext.java
boaks: The idea of "id of (tcp) connection"?
Or the term "CONNECTION_ID"?
Or the value?

The idea was, that tcp connection, even with same endpoints, are potentially different.
The term was just selected by me.
The value, wich is used, is from netty, `channel.id().asShortText()`

element-connector/src/main/java/org/eclipse/californium/elements/tcp/NettyContextUtils.java
sophokles73: Yes, separate PR. This does not have much priority right now, I guess..

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TlsServerConnector.java
boaks: No, thoughts right now :-).

element-connector/src/main/java/org/eclipse/californium/elements/TcpCorrelationContext.java
sophokles73: I am asking because the term is also used by Thomas Fossati's draft for introducing a *CID* (connection ID) as part of a Hello Extension for DTLS to identify *security associations* independently from the transport. So, why do you want to propagate this Netty property up the layers?

element-connector/src/main/java/org/eclipse/californium/elements/TlsCorrelationContext.java
sophokles73: Maybe we do not need a TLS specific correlation context class anyways. Why can't we simply use `DtlsCorrelationContext` (and maybe rename it later to `TlsCorrelationContext`)? Is there something specific for TLS vs. DTLS?

element-connector/src/main/java/org/eclipse/californium/elements/TcpCorrelationContext.java
boaks: > So, why do you want to propagate this Netty property up the layers?

As I understood the intention of the correlation context, its mainly used to ensure, that a response is received with the same context as the request was sent.

So, though a TCP connection may be reopened, I think there is a requirement to ensure, that a response is received in the same and not in the reopend connection. The idea is, that netty overs that information. By now, its not tested.

element-connector/src/main/java/org/eclipse/californium/elements/TlsCorrelationContext.java
boaks: In TLS I can't access the "epoch" (at least, right now :-) ).
I think, having "special" classes for each type of connection, is useful. It makes it easier to distinguish.

element-connector/src/main/java/org/eclipse/californium/elements/TcpCorrelationContext.java
boaks: Tests are now added. See  correlationContextReconnectTimeout and correlationContextReconnectStopStart.

element-connector/src/main/java/org/eclipse/californium/elements/tcp/NettyContextUtils.java
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	/**
+	 * Report the correlation context related to the provided channel in future.
+	 * 
+	 * @param channel channel of correlation context
+	 * @param msg message for callback, when context gets established.
+	 */
+	public static void futureCorrelationContext(Channel channel, final RawData msg) {
+		if (null != msg.getMessageCallback()) {
```
boaks: Here is onel of the cases, where checking the availability of the callback is used for optimization.

boaks:
It seems for me to be impossible to make the changes in the unit test in a way, that the changes and the none changes are easy for review.

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TlsServerConnector.java
boaks: Just to add: the SSLEngine supports "want" and "need". So I thought, with "want", the upper layer can do his checks.

element-connector/src/main/java/org/eclipse/californium/elements/tcp/NettyContextUtils.java
boaks: Issue #168 created.

sophokles73:
I finally was able to create a [CQ for hamcrest-library](http://dev.eclipse.org/ipzilla/show_bug.cgi?id=12347)

element-connector/src/main/java/org/eclipse/californium/elements/TcpCorrelationContext.java
sophokles73: Is there a way to get at the session ID, epoch and cipher name used in the TLS session negotiated by Netty?

boaks: See NettyContextUtils. But the "epoch" seems to be not available. 
I'm currently working on the finalize of this work. I changed the comment

> /**
>  * A correlation context that explicitly supports TLS specific properties.
>  * Currently the context is not aware of renegotiation (API to acquire
>  * information is missing). According oracle, the renegotiate issues seems not
>  * be fixed, if your java is not to deprecated.
>  * 
>  * @see http://www.oracle.com/technetwork/java/javase/overview/tlsreadme2-176330.html
>  */
> public class TlsCorrelationContext extends TcpCorrelationContext {
> 

to indicate the information behind that.

boaks:
Currently the error cases for failed trusting x509 is missing.

Self-signed is currently not supported and not intended to be supported soon :-).

boaks:
@sophokles73 

I found some issues related to the "trust" of x509. I would like to work on that.
So, even if the CQ may be approved in the next days, please don't merge this PR!

sophokles73:
CQ has been approved :-)

boaks:
> CQ has been approved :-)

Good news! But still, please don't merge. I hope I can fix my issues today :-).

boaks:
Update:
After a couple of hours trying to test the "direct trust" (LWM2M TS, OMA-TS-LightweightM2M-V1_0-20161123-D, 7.1.3, "The LWM2M Client MUST either directly trust the server's X509 certificate or trust it indirectly by verifying it is correctly signed by a trusted CA.") a recognized the issue behind the SUN java x509 implementation:
If the client certificates are added to the keystore for the trust manager, these are interpreted as "issuer" and therefore sent as "DistinguishedName certificate_authorities" (RFC5246, 7.4.4). But the "direct client certificates" are no issuers and therefore the client the refuse the connection. 
The solution is to change the trust manager not to interpert these certificates as issuer.
I got for that and cleanup the stuff. May take some time, but I'm still on it.

boaks:
So, in the end, with unit tests and support for the "LWM2M x509 support" it's pretty large :-).

So, I will split it in several PRs but they may (will) denpend on each other ;-(.

boaks:
> So, I will split it in several PRs but they may (will) denpend on each other ;-(.

Next year ...

boaks:
Move TLS stuff to #294

PR #150, opened 2016-11-23
Bugfix received message order in ongoing blockwise with new

boaks:
notification.

It is not safe to assume, that the ACK arrives before the GET.
This testfix potentially changes the message flow:

Intended before testfix:
1.  => Notify
2.1 <= ACK
2.2 => BLOCK-Response(old)
2.3 <= GET
3.  => BLOCK-Response(new)

After testfix:
1.  => Notify
2.1 <= ACK
2.2 <= GET
3.  => BLOCK-Response(old)
4.  => BLOCK-Response(new)

On the first, it seems that this changes the order of the message flow.
But though 2.1 and 2.3 are undefined, the order before is undefined. So
it's not possible to realy ensure, that 2.2 is send before 2.3 is
received.


boaks:
@sophokles73 

Even if the test fails only very rarely, I think it's worth to be changed :-).

sophokles73:
This will be addressed by the fix for #155 as well ...

boaks:
OK

Issue #152, opened 2016-11-24
race condition


boaks:
https://hudson.eclipse.org/californium/job/1.1.x-nightly/5/testReport/junit/org.eclipse.californium.core.test/ObserveTest/testObserveClient/

This test has a race condition:

Main thread is calling `synchronous` request, which waits for a response on the result.
Additionaly it uses the `CoapHandler` callback for responses.

        Request

	public void setResponse(Response response) {
		this.response = response;

		// only for synchronous/blocking requests
		if (lock != null) {
			synchronized (lock) {
				lock.notifyAll();
			}
		}
		// else: we know that nobody is waiting on the lock

		for (MessageObserver handler : getMessageObservers()) {
			handler.onResponse(response);
		}
	}

This first `notifyAll` to return the reponse and then calls parallel the `CoapHandler`.
I would got for fixing the test and not the changing the behavior.
Any objections?

sophokles73:
I am not sure if I can follow. Where do you suspect the race condition?

boaks:
The failure `expected:<1> but was:<0>` at `assertEquals(1, notificationCounter); // only one notification received` indicates the race condition (or visibility problem).

		CoapClient client = new CoapClient(uriX);

		CoapObserveRelation rel = client.observeAndWait(new CoapHandler() {
			@Override
			public void onLoad(CoapResponse response) {
				System.out.println("Received Notification: "+response.advanced().getMID());
				++notificationCounter;
			}
			@Override
			public void onError() { }
		});

		rel.reactiveCancel();
		Thread.sleep(50); // wait for cancel to take effect

		for (int i=0; i < repeat; ++i) {
			resourceX.changed();
			Thread.sleep(50);
		}

		assertEquals(1, notificationCounter); // only one notification received


`++notificationCounter;` is called for the response asynchron to the execution of the thread. 
I alos add a check, if there is a reponse at all.

sophokles73:
Fixed on 2.0.x in da22dfbf4acc40f363f1a1e49003c9b73b2d5849
Fixed on master in d56c7e3fe8e1af4b1707500e257e11743a3a4054

PR #153, opened 2016-11-24
Testfix for testObserveClient.

boaks:
Fix race condition of response and onLoad().
Fix visibility of counters.
(Issue #152)

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/ObserveTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/ObserveTest.java
@@ -16,15 +16,20 @@
  *    Dominique Im Obersteg - parsers and initial implementation
  *    Daniel Pauli - parsers and initial implementation
  *    Kai Hudalla - logging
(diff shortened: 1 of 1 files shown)


californium-core/src/test/java/org/eclipse/californium/core/test/ObserveTest.java
```diff
@@ -100,8 +104,7 @@
 	private String uriX;
 	private String uriY;
 
-	private int notificationCounter = 0;
-	private int resetCounter = 0;
+	private AtomicInteger resetCounter = new AtomicInteger(0);
```
boaks: Its used in the inner class `ServerMessageInterceptor` and to move in to method level would require to change this also.

sophokles73: I know, but isn't that the right thing to do, now that we are already dealing with this class' synchronization issues? It would only mean moving the property to `ServerMessageInterceptor`. Not such a big deal, is it?

boaks:
@sophokles73 
Done!

PR #157, opened 2016-11-25
Server Name Indication Extension doesn't include server_name_list length field

kiemlicz:
Testing with mbedTLS reveals that mbed's client_hello sni extension cannot be decoded. 
This is due to missing server_name_list length field (encoding and decoding) in extension (as required for vectors).

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Response.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Response.java
@@ -133,4 +133,14 @@ public class Response extends Message {
 	public boolean isNotification() {
 		return getOptions().hasObserve();
 	}
+
+	/**
+	 * Checks whether this response has either a <em>block1</em> or
+	 * <em>block2</em> option.
+	 * 
+	 * @return {@code true} if this response has a block option.
+	 */
+	public boolean hasBlockOption() {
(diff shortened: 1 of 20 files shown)


sophokles73:
Thanks for the PR. However, you will need to sign an Eclipse CLA as described in the [contribution guide](https://github.com/eclipse/californium/blob/master/CONTRIBUTING.md).

Please also make sure to use tabs instead of spaces in order to prevent breaking indentation ...

kiemlicz:
I've just registered and signed ECA (like a 10 minutes ago) - sorry for inconvenience.
I believe I didn't change your indentation style

sophokles73:
Ok, I have checked your commits.
1. You did not include a `signed-off-by` footer in your commit messages. This is required by the Eclipse IP process. Please take a look at our contribution guide regarding how to add it to your commit message(s).
2. You are using blanks for indentation instead of tabs as we do in the Californium code.
3. It would be nice if you could squash your commits into a single one since they all concern the same issue and one alone doesn't fix the problem.

It would be great if you could change your PR accordingly so that we can accept it :-)
Let em know if you need any further assistance.

kiemlicz:
As usually my experiments went wrong...
Please excuse and close/reject this.

[This one is valid](https://github.com/eclipse/californium/pull/163)

PR #160, opened 2016-11-29
Bugfix set scheme from URI.

boaks:
Set field 'scheme' when calling 'setOptions(URI)'.
Introduce 'getDefaultPort(Strint uriScheme)'.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/CoAP.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/CoAP.java
@@ -17,13 +17,14 @@
  *    Daniel Pauli - parsers and initial implementation
  *    Kai Hudalla - logging
  *    Bosch Software Innovations GmbH - improve readability
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add getDefaultPort
+ *                                                    add CodeClass
  ******************************************************************************/
 package org.eclipse.californium.core.coap;
 
 import static org.eclipse.californium.core.coap.CoAP.MessageFormat.*;
 
 import java.nio.charset.Charset;
-
 
 /**
  * CoAP defines several constants.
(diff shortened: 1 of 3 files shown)


californium-core/src/main/java/org/eclipse/californium/core/coap/CoAP.java
```diff
@@ -135,6 +135,30 @@ public static boolean isSupportedScheme(final String uriScheme) {
 				CoAP.COAP_SECURE_URI_SCHEME.equalsIgnoreCase(uriScheme) ||
 				CoAP.COAP_SECURE_TCP_URI_SCHEME.equalsIgnoreCase(uriScheme);
 	}
+	
+	/**
+	 * Get default port for provided uri scheme.
+	 * 
+	 * @param uriScheme uri scheme for default port
+	 * @return default port, or 0, if no default port is defined for the provided uri scheme.
```
sophokles73: IMHO we should throw an `IllegalArgumentException` in case of an unsupported scheme (instead of returning 0 which needs to be checked by the client afterwards). Client code can use `isSupportedScheme()` before invoking this method then.

boaks:
I added also CodeClass to CoAP and some more unit tests.
Even if this is not really in the scope of the initial request, I think it should fit.
If not, let me know, then I would split it.

californium-core/src/main/java/org/eclipse/californium/core/coap/CoAP.java
sophokles73: Please update the JavaDoc accordingly ...

californium-core/src/test/java/org/eclipse/californium/core/test/CoapTest.java
```diff
@@ -16,43 +16,180 @@
...
+		assertFalse(CoAP.isSecureScheme("https:"));
+	}
+
+	@Test
+	public void testIsSupportedScheme() {
+		assertTrue(CoAP.isSupportedScheme(CoAP.COAP_SECURE_TCP_URI_SCHEME));
+		assertTrue(CoAP.isSupportedScheme(CoAP.COAP_SECURE_URI_SCHEME));
+		assertTrue(CoAP.isSupportedScheme(CoAP.COAP_TCP_URI_SCHEME));
+		assertTrue(CoAP.isSupportedScheme(CoAP.COAP_URI_SCHEME));
+		assertFalse(CoAP.isTcpScheme("https:"));
```
sophokles73: CoAP.isSupportedScheme() ?

boaks:
Uups, sorry, I have overseen that. Fixed.

californium-core/src/main/java/org/eclipse/californium/core/coap/CoAP.java
```diff
@@ -391,15 +470,15 @@ public String toString() {
 		}
 
 		public static boolean isSuccess(final ResponseCode code) {
-			return code.codeClass == 2;
+			return code.codeClass == CodeClass.SUCCESS_RESPONSE.value;
```
sophokles73: What if code == null? Shall we return `false` or throw a NPE?

californium-core/src/test/java/org/eclipse/californium/core/test/CoapTest.java
sophokles73: Do you want to fix this?

californium-core/src/main/java/org/eclipse/californium/core/coap/CoAP.java
boaks: NPE, I will add explizit ones.

boaks:
> Do you want to fix this?

Yes, until its supported :-).

boaks:
the request changes should be in.

sophokles73:
merged

PR #161, opened 2016-11-29
Bugfix adjust port after binding.

boaks:
Adjust port after binding (TCP server).

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
@@ -12,6 +12,7 @@
  * <p>
  * Contributors:
  * Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
+ * Achim Kraus (Bosch Software Innovations GmbH) - adjust port when bound.
  ******************************************************************************/
 package org.eclipse.californium.elements.tcp;
 
@@ -22,6 +23,7 @@ import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.channel.socket.nio.NioServerSocketChannel;
 import io.netty.handler.timeout.IdleStateHandler;
+
 import org.eclipse.californium.elements.Connector;
 import org.eclipse.californium.elements.RawData;
 import org.eclipse.californium.elements.RawDataChannel;
@@ -46,7 +48,6 @@ public class TcpServerConnector implements Connector {
 	private final String SUPPORTED_SCHEME = "coap+tcp";
 
 	private final int numberOfThreads;
-	private final InetSocketAddress localAddress;
 	private final int connectionIdleTimeoutSeconds;
 	private final ConcurrentMap<SocketAddress, Channel> activeChannels = new ConcurrentHashMap<>();
 
@@ -54,6 +55,7 @@ public class TcpServerConnector implements Connector {
 	private EventLoopGroup bossGroup;
 	private EventLoopGroup workerGroup;
 	private URI listenUri;
+	private InetSocketAddress localAddress;
 
 	public TcpServerConnector(InetSocketAddress localAddress, int numberOfThreads, int idleTimeout) {
 		this.numberOfThreads = numberOfThreads;
@@ -62,7 +64,8 @@ public class TcpServerConnector implements Connector {
 		this.listenUri = getListenUri(localAddress);
 	}
 
-	@Override public synchronized void start() throws IOException {
+	@Override
+	public synchronized void start() throws IOException {
 		if (rawDataChannel == null) {
 			throw new IllegalStateException("Cannot start without message handler.");
 		}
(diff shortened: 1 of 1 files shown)


boaks:
In my system netty sets the bind address to IPv6 in "any" is used, which causes some trouble. Therefore I just pick the port, if not already provided .

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
```diff
@@ -82,17 +97,33 @@ public TcpServerConnector(InetSocketAddress localAddress, int numberOfThreads, i
...
-
-			@Override
-			public void operationComplete(ChannelFuture future) throws Exception {
-				InetSocketAddress listenAddress = (InetSocketAddress) future.channel().localAddress();
-				listenUri = getListenUri(listenAddress);
+		ChannelFuture channelFuture = bootstrap.bind(localAddress).syncUninterruptibly();
+
+		if (0 == localAddress.getPort()) {
+			// get assigned port
+			channelFuture.addListener(new ChannelFutureListener() {
```
joemag1: Any value in listening asynchronously, since you already call syncUninterruptibly above?

boaks: Sure, the value is "no need to read the javadoc" :-).

I wass just miss lead by the introduction of adding the listener before.
Thanks for the hint, it makes the implementation much simpler!

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
```diff
@@ -82,17 +85,17 @@ public TcpServerConnector(InetSocketAddress localAddress, int numberOfThreads, i
...
-			public void operationComplete(ChannelFuture future) throws Exception {
-				InetSocketAddress listenAddress = (InetSocketAddress) future.channel().localAddress();
-				listenUri = getListenUri(listenAddress);
-			}
-		});
+		ChannelFuture channelFuture = bootstrap.bind(localAddress).syncUninterruptibly();
+
+		if (channelFuture.isSuccess() && 0 == localAddress.getPort()) {
+			InetSocketAddress listenAddress = (InetSocketAddress) channelFuture.channel().localAddress();
+			localAddress = new InetSocketAddress(localAddress.getAddress(), listenAddress.getPort());
```
sophokles73: wouldn't the following be sufficient:
```
localAddress = (InetSocketAddress) channelFuture.channel().localAddress();
listenUri = getListenUri(localAddress);
```

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
```diff
@@ -82,17 +85,17 @@ public TcpServerConnector(InetSocketAddress localAddress, int numberOfThreads, i
...
 		// Start the server.
-		bootstrap.bind(localAddress).syncUninterruptibly().addListener(new ChannelFutureListener() {
-
-			@Override
-			public void operationComplete(ChannelFuture future) throws Exception {
-				InetSocketAddress listenAddress = (InetSocketAddress) future.channel().localAddress();
-				listenUri = getListenUri(listenAddress);
-			}
-		});
+		ChannelFuture channelFuture = bootstrap.bind(localAddress).syncUninterruptibly();
```
sophokles73: I assume that `syncUninterruptibly()` means something like: "do it now and wait for the result", right?

element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
boaks: Using that results in:

     coap+tcp://0:0:0:0:0:0:0:0:5683

and using that as (leshan example):

        this.registrationEndpointUri = registrationEndpointUri;
        this.registrationEndpointAddress = new InetSocketAddress(registrationEndpointUri.getHost(), registrationEndpointUri.getPort());

results in:

    java.lang.IllegalArgumentException: hostname can't be null
	at java.net.InetSocketAddress.checkHost(InetSocketAddress.java:149) ~[?:1.8.0_51]
	at java.net.InetSocketAddress.<init>(InetSocketAddress.java:216) ~[?:1.8.0_51]

So it seems, IPv6 has some open issues. Therefore I just replaced the port.
I had a comment on that, but on the reworks that comment seems to be deleted.
Alternatively, IPv6 usage may be consequently supported. But this seems to be much out of this scope.

boaks: Analysing UDPConnector, I'm wondering, what the "exact" definition of "restart" with port 0 should be. Should it start on the previous port or could it alos chose a new one? UDPConnector preserves the 0 and would therefore potentially reopen a new port. Is that intended?

sophokles73: Oh, now I see that you are using `localAddress.getAddress()` instead of `listenAddress.getAddress()`.

The problem lies in the `getListenUri()` method. We need to distinguish between `Inet4Address` and `Inet6Address` because the URI constructor requires IPv6 literal addresses to be enclosed in square brackets:

```
private URI getListenUri(final InetSocketAddress listenAddress) {

  InetAddress address = listenAddress.getAddress();
  String addressString = address.getHostString();
  if (address instanceof Inet6Address) {
    addressString = String.format("[%s]", addressString);
  }
      return URI.create(String.format("%s://%s:%d", getSupportedScheme(), addressString, address.getPort()));
}
```

boaks: Hm, in this PR I would like to got for "the conservative solution" and just replace the port.
I'm afraid, that changing towards an IPv6 address would introduce more problems outsite.
Additionally we may add this InetSocketAddress to URI in a common util class.

sophokles73: Sounds good. Could you please open an issue for it so that we do not forget it?

sophokles73: > Analysing UDPConnector, I'm wondering, what the "exact" definition of "restart" with port 0 should be. Should it start on the previous port or could it alos chose a new one? UDPConnector preserves the 0 and would therefore potentially reopen a new port. Is that intended?

Is this related to this PR? If not and you suspect a problem with the current behavior, please open another issue.

boaks:
Created two issues for the left stuff (IPv6 and wanted behavior on restart with port 0).
I added also a comment to just replace the port.
So I hope the requested changes are done.

With that commit the PR #148 could also be simplified a bit. So if merged, I will adjust the other PR also.

PR #163, opened 2016-11-29
Adding list length field to the Server Name Indication extension (clean&new)

kiemlicz:
Testing with mbedTLS reveals that mbed's client_hello sni extension cannot be decoded.
This is due to missing server_name_list length field (encoding and decoding) in extension (as required for vectors).
Please close/reject https://github.com/eclipse/californium/pull/157 as I made a mistake while rebasing...

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerNameExtension.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerNameExtension.java
@@ -1,15 +1,15 @@
 /*******************************************************************************
  * Copyright (c) 2016 Bosch Software Innovations GmbH and others.
- *
+ * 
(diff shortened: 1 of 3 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerNameExtension.java
```diff
@@ -1,15 +1,15 @@
 /*******************************************************************************
  * Copyright (c) 2016 Bosch Software Innovations GmbH and others.
- * 
+ *
```
sophokles73: Can you please leave all comments intact? I assume you are using some code formatting rules that eliminate these whitespaces. I have checked the Californium code formatting rules but couldn't find such a setting that would be responsible for it. Are you using Eclipse at all?

Issue #166, opened 2016-12-02
NullPointerException when sending RST as response.


boaks:
Tcp/UdpMatcher / ExchangeObserverImpl / completed :

    java.lang.NullPointerException
	at org.eclipse.californium.core.network.UdpMatcher$ExchangeObserverImpl.completed(UdpMatcher.java:463)
	at org.eclipse.californium.core.network.Exchange.setComplete(Exchange.java:451)
	at org.eclipse.californium.core.network.UdpMatcher.sendEmptyMessage(UdpMatcher.java:164)
	at org.eclipse.californium.core.network.CoapEndpoint$OutboxImpl.sendEmptyMessage(CoapEndpoint.java:629)
	at org.eclipse.californium.core.network.stack.BaseCoapStack$StackBottomAdapter.sendEmptyMessage(BaseCoapStack.java:194)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendEmptyMessage(AbstractLayer.java:83)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendEmptyMessage(AbstractLayer.java:83)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendEmptyMessage(AbstractLayer.java:83)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendEmptyMessage(AbstractLayer.java:83)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendEmptyMessage(AbstractLayer.java:83)
	at org.eclipse.californium.core.network.stack.BaseCoapStack.sendEmptyMessage(BaseCoapStack.java:89)
	at org.eclipse.californium.core.network.CoapEndpoint.sendEmptyMessage(CoapEndpoint.java:521)
	at org.eclipse.californium.core.network.Exchange.sendReject(Exchange.java:196)
	at org.eclipse.californium.core.test.SmallServerClientTest$1.deliverRequest(SmallServerClientTest.java:124)

PR with fix will follow.

boaks:
See #155

Issue #155, opened 2016-11-25
overlapping notifications requiring blockwise transfer are not detected


sophokles73:
On the 2.0.x branch the `ObservationLayer` is not able to detect an already ongoing blockwise transfer for a notification when a new notification for the same token is received while the blockwise transfer is still in progress. This leads to the following problems:
* we no longer follow NSTART 1 because we run multiple blockwise transfers in parallel
* already running blockwise transfer(s) might mix up *old* with *new* blocks of the (changed) resource.
* since all blockwise transfers are conducted independently from each other, the blockwise layer delivers a separate *assembled* and potentially corrupted `Response` for each of them to the application layer when it should have delivered only one `Response` reflecting the most recent (completely transferred) state of the resource.

On master, this does not happen because an already running blockwise transfer is stopped and replaced with a new (blockwise) transfer for the newly received notification when such an *overlapping* notification is received.

boaks:
> we no longer follow NSTART 1 because we run multiple blockwise transfers in parallel

This question depends on the definition of NSTART. A blockwise transfer could be considered as a sequence of request/reponse pairs and NSTART 1 is only applied for each pair. Then also other "request/response" pairs would be possible, even if the complete blockwise is not completed.
Or the "complete blockwise transfer" is considered to be in NSTART 1, so no other request/response pairs are allowed.
I would prefer the pairwise definition. Otherwise the communication will be blocked for to long.

The NSTART 1 is mostly related to the resource restriction in the communication layer. Therefore blockwise is constructed in a way, where it is possible to handle the blockwise specific stuff outside of the communication layer. So I think NSTART 1 should be interpreted based on a single request/response pair and not the complete "blockwise transfer".

sophokles73:
I agree, but the issue is not whether the whole transfer is considered an exchange or not. The problem is that in the existing implementation there is no check making sure that there are no two exchanges (for a single block or the whole body) with the same endpoint are going on.

sbernard31:
Do we have a test for this bug ? 

The issue could not be fixed just by searching the exchange by URI instead of token in the [matcher](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java#L246) for block2 notification ? 
(I know we will add some specific blockwise behavior in the matcher :/)

sophokles73:
[...]

PR #167, opened 2016-12-02
Bugfix for NullPointerException on RST.

boaks:
Issue #166.
TcpMatcher is changed accordingly, even if RST is not used there.


sophokles73:
Will be resolved as part of the fix for #155 as well.

Issue #30, opened 2016-04-06
Randomly wrong responses in blockwise (option block2)


boaks:
(Im using leshan with californium 1.0.3)
During tests, I recognized, that sometimes the coap server resends already sent blocks.
(e.g. 23 is resend, 24 is requested).
Any Idea, what could cause this?

``````
- (org.eclipse.californium.core.network.interceptors.MessageTracer.java:64) sendResponse() in thread pool-4-thread-4 at (2016-04-06 17:58:21)
33 INFO [MessageTracer]: /192.168.10.106:12.345 ==> req CON-GET    MID=50932, Token=8ff6, OptionSet={"Uri-Path":["file","fw.zip"], "Block2":"(szx=3/128, m=false, num=22)"}, no payload - (org.eclipse.californium.core.network.interceptors.MessageTracer.java:78) receiveRequest() in thread pool-4-thread-3 at (2016-04-06 17:58:21)
33 INFO [MessageTracer]: /192.168.10.106:12.345 <== res ACK-2.05   MID=50932, Token=8ff6, OptionSet={"Block2":"(szx=3/128, m=true, num=22)"}, 
06 02 00 1d 07 02 00 57 07 02 00 fe 06 02 00 3b 07 02 00 70 07 02 00 89 07 02 00 9f 07 02 00 c3
07 02 00 dd 07 02 00 2e 08 02 00 47 08 02 00 8c 20 00 20 57 08 02 00 6e 08 02 00 ec 20 00 20 ec
25 00 20 9c 21 00 20 9c 20 00 20 28 26 00 20 10 27 00 00 80 08 02 00 e2 04 00 00 9b 08 02 00 30
b5 04 1c 85 b0 0d 1c 00 2a 04 d0 10 1c 24 49 24 4a ff f7 d7 fc 2b 78 15 2b 3e d1 22 4b 03 93 10
 - (org.eclipse.californium.core.network.interceptors.MessageTracer.java:64) sendResponse() in thread pool-4-thread-3 at (2016-04-06 17:58:21)
54 INFO [MessageTracer]: /192.168.10.106:12.345 ==> req CON-GET    MID=51188, Token=90f6, OptionSet={"Uri-Path":["file","fw.zip"], "Block2":"(szx=3/128, m=false, num=23)"}, no payload - (org.eclipse.californium.core.network.interceptors.MessageTracer.java:78) receiveRequest() in thread pool-4-thread-7 at (2016-04-06 17:58:21)
[...]

PR #170, opened 2016-12-06
[#155] Fix failure to detect overlapping blockwise notification transfers

sophokles73:
Refactored tracking of blockwise transfers so that the status is no
longer kept within the Exchange object but within BlockwiseLayer, thus
decoupling matcher from blockwise layer. Removed all knowledge about
(and special treatment of) blockwise transfer from Matchers.

Introduced support for using ETags for blockwise transfers. ETags
included in notifications are used in block2 transfers by clients.

Transparent blockwise transfer handling can now be disabled altogether
by setting MAX_RESOURCE_BODY_SIZE config property to 0.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
@@ -32,7 +32,6 @@ import org.eclipse.californium.core.coap.Message;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.network.stack.BlockwiseLayer;
-import org.eclipse.californium.core.network.stack.BlockwiseStatus;
 import org.eclipse.californium.core.network.stack.CoapStack;
 import org.eclipse.californium.core.observe.ObserveRelation;
 import org.eclipse.californium.core.server.resources.CoapExchange;
@@ -60,14 +59,27 @@ import org.eclipse.californium.elements.CorrelationContext;
  */
 public class Exchange {
 
+	private static final int MAX_OBSERVE_NO = (1 << 24) - 1;
+
 	/**
-	 * The origin of an exchange. If Cf receives a new request and creates a new
+	 * The origin of an exchange.
+	 * <p>
+	 * If Cf receives a new request and creates a new
 	 * exchange the origin is REMOTE since the request has been initiated from a
 	 * remote endpoint. If Cf creates a new request and sends it, the origin is
 	 * LOCAL.
 	 */
 	public enum Origin {
-		LOCAL, REMOTE;
+
+		/**
+		 * Indicates that a message exchange has been initiated locally.
+		 */
+		LOCAL,
+
+		/**
+		 * Indicates that a message exchange has been initiated remotely.
+		 */
+		REMOTE;
 	}
 
 	/** The endpoint that processes this exchange */
@@ -96,9 +108,6 @@ public class Exchange {
 	 */
 	private Request currentRequest; // Matching needs to know for what we expect a response
 
-	/** The status of the blockwise transfer. null in case of a normal transfer */
-	private BlockwiseStatus requestBlockStatus;
-
 	/**
 	 * The actual response that is supposed to be sent to the client. Layers
 	 * below the {@link BlockwiseLayer} should only work with the
@@ -109,9 +118,6 @@ public class Exchange {
 
 	/** The current block of the response that is being transferred. */
 	private Response currentResponse; // Matching needs to know when receiving duplicate
-
-	/** The status of the blockwise transfer. null in case of a normal transfer */
-	private BlockwiseStatus responseBlockStatus;
 
 	// indicates where the request of this exchange has been initiated.
 	// (as suggested by effective Java, item 40.)
@@ -129,12 +135,11 @@ public class Exchange {
 	// handle to cancel retransmission
 	private ScheduledFuture<?> retransmissionHandle = null;
 
-	// handle to extend blockwise status lifetime
-	private ScheduledFuture<?> blockCleanupHandle = null;
-
 	// If the request was sent with a block1 option the response has to send its
 	// first block piggy-backed with the Block1 option of the last request block
 	private BlockOption block1ToAck;
+
+	private Integer notificationNumber;
 
 	// The relation that the target resource has established with the source
 	private ObserveRelation relation;
@@ -262,25 +267,6 @@ public class Exchange {
 	}
 
 	/**
-	 * Returns the blockwise transfer status of the request or null if no one is
-	 * set.
-	 * 
-	 * @return the status of the blockwise transfer of the request
-	 */
-	public BlockwiseStatus getRequestBlockStatus() {
-		return requestBlockStatus;
-	}
-
-	/**
-	 * Sets the blockwise transfer status of the request.
-	 * 
-	 * @param requestBlockStatus the blockwise transfer status
-	 */
-	public void setRequestBlockStatus(BlockwiseStatus requestBlockStatus) {
-		this.requestBlockStatus = requestBlockStatus;
-	}
-
-	/**
 	 * Returns the response to the request or null if no response has arrived
 	 * yet. If there is an observe relation, the last received notification is
 	 * the response.
@@ -320,25 +306,6 @@ public class Exchange {
 	 */
 	public void setCurrentResponse(Response currentResponse) {
 		this.currentResponse = currentResponse;
-	}
-
-	/**
-	 * Returns the blockwise transfer status of the response or null if no one 
-	 * is set.
-	 * 
-	 * @return the status of the blockwise transfer of the response
-	 */
-	public BlockwiseStatus getResponseBlockStatus() {
-		return responseBlockStatus;
-	}
-
-	/**
-	 * Sets the blockwise transfer status of the response.
-	 * 
-	 * @param responseBlockStatus the blockwise transfer status
-	 */
-	public void setResponseBlockStatus(BlockwiseStatus responseBlockStatus) {
-		this.responseBlockStatus = responseBlockStatus;
 	}
 
 	/**
@@ -418,16 +385,33 @@ public class Exchange {
 		this.retransmissionHandle = retransmissionHandle;
 	}
 
-	public synchronized ScheduledFuture<?> getBlockCleanupHandle() {
-		return blockCleanupHandle;
+	/**
+	 * Sets the number of the notification this exchange is associated with.
+	 * <p>
+	 * This number can be used to match responses of a blockwise
+	 * transfer triggered by a notification.
+	 * 
+	 * @param notificationNo The observe number of the notification.
+	 * @throws IllegalArgumentException if the given number is &lt; 0 or &gt; 2^24 - 1.
+	 */
+	public synchronized void setNotificationNumber(final int notificationNo) {
+		if (notificationNo < 0 || notificationNo > MAX_OBSERVE_NO) {
+			throw new IllegalArgumentException("illegal observe number");
+		}
+		this.notificationNumber = notificationNo;
 	}
 
-	public synchronized void setBlockCleanupHandle(final ScheduledFuture<?> blockCleanupHandle) {
-		// avoid race condition of multiple block requests
-		if (this.blockCleanupHandle != null) {
-			this.blockCleanupHandle.cancel(false);
-		}
-		this.blockCleanupHandle = blockCleanupHandle;
+	/**
+	 * Gets the number of the notification this exchange is associated with.
+	 * <p>
+	 * This number can be used to match responses of a blockwise
+	 * transfer triggered by a notification.
+	 * 
+	 * @return The observe number of the notification or {@code null} if this
+	 *         exchange is not associated with a notification.
+	 */
+	public synchronized Integer getNotificationNumber() {
+		return notificationNumber;
 	}
 
 	public void setObserver(ExchangeObserver observer) {
@@ -741,80 +725,6 @@ public class Exchange {
 
 		public byte[] getToken() {
 			return Arrays.copyOf(token, token.length);
-		}
-	}
-
-	/**
-	 * A key based on a CoAP message's target URI that is scoped to an endpoint.
-	 * <p>
-	 * This class is used by the matcher to correlate requests by their target
-	 * URI (for observe relations).
-	 */
-	public static final class KeyUri {
-
-		private static final int MAX_PORT_NO = (1 << 16) - 1;
-		private final String uri;
-		private final byte[] address;
-		private final int port;
-		private final int hash;
-
-		/**
-		 * Creates a new key for a URI scoped to an endpoint address.
-		 * 
-		 * @param uri the URI.
-		 * @param address the endpoint's address.
-		 * @param port the endpoint's port.
-		 * @throws NullPointerException if uri or address is {@code null}
-		 * @throws IllegalArgumentException if port &lt; 0 or port &gt; 65535.
-		 */
-		public KeyUri(String uri, byte[] address, int port) {
-			if (uri == null) {
-				throw new NullPointerException("URI must not be null");
-			} else if (address == null) {
-				throw new NullPointerException("address must not be null");
-			} else if (port < 0 || port > MAX_PORT_NO) {
-				throw new IllegalArgumentException("port must be an unsigned 16 bit int");
-			} else {
-				this.uri = uri;
-				this.address = address;
-				this.port = port;
-				this.hash = (port * 31 + uri.hashCode()) * 31 + Arrays.hashCode(address);
-			}
-		}
-
-		@Override
-		public int hashCode() {
-			return hash;
-		}
-
-		@Override
-		public boolean equals(Object o) {
-			if (! (o instanceof KeyUri))
-				return false;
-			KeyUri key = (KeyUri) o;
-			return uri.equals(key.uri) && port == key.port && Arrays.equals(address, key.address);
-		}
-
-		@Override
-		public String toString() {
-			return new StringBuilder("KeyUri[").append(uri)
-					.append(", ").append(Utils.toHexString(address)).append(":").append(port)
-					.append("]").toString();
-		}
-
-		/**
-		 * Creates a new key for a request scoped to the request's source endpoint address.
-		 * 
-		 * @param request the request.
-		 * @return the key.
-		 * @throws NullPointerException if the request is {@code null}.
-		 */
-		public static KeyUri fromInboundRequest(final Request request) {
-			if (request == null) {
-				throw new NullPointerException("request must not be null");
-			} else {
-				return new KeyUri(request.getURI(), request.getSource().getAddress(), request.getSourcePort());
-			}
 		}
 	}
 }
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/TcpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/TcpMatcher.java
@@ -80,29 +80,6 @@ public final class TcpMatcher extends BaseMatcher {
 		// ensure Token is set
 		response.setToken(exchange.getCurrentRequest().getToken());
 
-		// Blockwise transfers are identified by URI and remote endpoint
-		if (response.getOptions().hasBlock2()) {
-			Request request = exchange.getCurrentRequest();
-			Exchange.KeyUri idByUri = new Exchange.KeyUri(request.getURI(), response.getDestination().getAddress(),
-					response.getDestinationPort());
-			// Observe notifications only send the first block, hence do not
-			// store them as ongoing
-			if (exchange.getResponseBlockStatus() != null && !response.getOptions().hasObserve()) {
-				// Remember ongoing blockwise GET requests
-				if (exchangeStore.registerBlockwiseExchange(idByUri, exchange) == null) {
-					LOGGER.log(Level.FINE, "Ongoing Block2 started late, storing {0} for {1}", new Object[] { idByUri,
-							request });
-				} else {
-					LOGGER.log(Level.FINE, "Ongoing Block2 continued, storing {0} for {1}", new Object[] { idByUri,
-							request });
-				}
-			} else {
-				LOGGER.log(Level.FINE, "Ongoing Block2 completed, cleaning up {0} for {1}", new Object[] { idByUri,
-						request });
-				exchangeStore.remove(idByUri);
-			}
-		}
-
 		// Only Observes keep the exchange active (CoAP server side)
 		if (response.isLast()) {
 			exchange.setComplete();
@@ -121,42 +98,10 @@ public final class TcpMatcher extends BaseMatcher {
 
 	@Override
 	public Exchange receiveRequest(Request request) {
-		/*
-		 * The differentiation between the case where there is a Block1 or
-		 * Block2 option and the case where there is none has the advantage that
-		 * all exchanges that do not need blockwise transfer have simpler and
-		 * faster code than exchanges with blockwise transfer.
-		 */
-		if (!request.getOptions().hasBlock1() && !request.getOptions().hasBlock2()) {
 
-			Exchange exchange = new Exchange(request, Exchange.Origin.REMOTE);
-			exchange.setObserver(exchangeObserver);
-			return exchange;
-		} else {
-			Exchange.KeyUri idByUri = new Exchange.KeyUri(request.getURI(), request.getSource().getAddress(),
-					request.getSourcePort());
-			LOGGER.log(Level.FINE, "Looking up ongoing exchange for {0}", idByUri);
-
-			Exchange ongoing = exchangeStore.get(idByUri);
-			if (ongoing != null) {
-				return ongoing;
-			} else {
-				// We have no ongoing exchange for that request block.
-				/*
-				 * Note the difficulty of the following code: The first message
-				 * of a blockwise transfer might arrive twice due to a
-				 * retransmission. The new Exchange must be inserted in both the
-				 * hash map 'ongoing' and the deduplicator. They must agree on
-				 * which exchange they store!
-				 */
-
-				Exchange exchange = new Exchange(request, Exchange.Origin.REMOTE);
-				LOGGER.log(Level.FINER, "New ongoing request, storing {0} for {1}", new Object[] { idByUri, request });
-				exchange.setObserver(exchangeObserver);
-				exchangeStore.registerBlockwiseExchange(idByUri, exchange);
-				return exchange;
-			} // if ongoing
-		} // if blockwise
+		Exchange exchange = new Exchange(request, Exchange.Origin.REMOTE);
+		exchange.setObserver(exchangeObserver);
+		return exchange;
 	}
 
 	@Override
@@ -189,7 +134,7 @@ public final class TcpMatcher extends BaseMatcher {
 		}
 	}
 
-	private boolean isResponseRelatedToRequest(final Exchange exchange, final CorrelationContext responseContext) {
+	private static boolean isResponseRelatedToRequest(final Exchange exchange, final CorrelationContext responseContext) {
 		return exchange.getCorrelationContext() == null || exchange.getCorrelationContext().equals(responseContext);
 	}
 
@@ -211,16 +156,7 @@ public final class TcpMatcher extends BaseMatcher {
 					exchangeStore.releaseToken(idByToken);
 				}
 			} else { // Origin.REMOTE
-				// this endpoint created the Exchange to respond to a request
-				Response response = exchange.getCurrentResponse();
-
-				Request request = exchange.getCurrentRequest();
-				if (request != null && (request.getOptions().hasBlock1() || response.getOptions().hasBlock2())) {
-					Exchange.KeyUri uriKey = new Exchange.KeyUri(request.getURI(), request.getSource().getAddress(),
-							request.getSourcePort());
-					LOGGER.log(Level.FINE, "Remote ongoing completed, cleaning up ", uriKey);
-					exchangeStore.remove(uriKey);
-				}
+				// nothing to do
 			}
 		}
 
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -33,7 +33,6 @@ import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.network.Exchange.KeyMID;
 import org.eclipse.californium.core.network.Exchange.KeyToken;
-import org.eclipse.californium.core.network.Exchange.KeyUri;
 import org.eclipse.californium.core.network.Exchange.Origin;
 import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.core.observe.NotificationListener;
@@ -106,27 +105,6 @@ public final class UdpMatcher extends BaseMatcher {
 			}
 		}
 
-		// Blockwise transfers are identified by URI and remote endpoint
-		if (response.getOptions().hasBlock2()) {
-			Request request = exchange.getCurrentRequest();
-			KeyUri idByUri = KeyUri.fromInboundRequest(request);
-			// Observe notifications only send the first block, hence do not store them as ongoing
-			if (exchange.getResponseBlockStatus() != null && !response.getOptions().hasObserve()) {
-				// Remember ongoing blockwise GET requests
-				if (exchangeStore.registerBlockwiseExchange(idByUri, exchange) == null) {
-					LOGGER.log(Level.FINE, "Ongoing Block2 started late, storing {0} for {1}",
-							new Object[]{idByUri, request});
-				} else {
-					LOGGER.log(Level.FINE, "Ongoing Block2 continued, storing {0} for {1}",
-							new Object[]{idByUri, request});
-				}
-			} else {
-				LOGGER.log(Level.FINE, "Ongoing Block2 completed, cleaning up {0} for {1}",
-						new Object[] { idByUri, request });
-				exchangeStore.remove(idByUri);
-			}
-		}
-
 		// Insert CON to match ACKs and RSTs to the exchange.
 		// Do not insert ACKs and RSTs.
 		if (response.getType() == Type.CON) {
@@ -141,7 +119,7 @@ public final class UdpMatcher extends BaseMatcher {
 				exchangeStore.registerOutboundResponse(exchange);
 			} else {
 				// we only need to assign an unused MID but we do not need to register
-				//the exchange under the MID since we do not expect/want a reply
+				// the exchange under the MID since we do not expect/want a reply
 				// that we would need to match it against
 				exchangeStore.assignMessageId(response);
 			}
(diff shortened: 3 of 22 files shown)


sbernard31:
No place for doubts, I have not masochistic tendencies ! :-p

sophokles73:
@sbernard31,

I suppose this means that you feel that the PR is too big to be reviewed, right?

sbernard31:
This means that I'm reviewing it but I'm suffering :)

sbernard31:
At first glance, I think this could have been done in several commit/PR.

1) Refactoring blockmode (moving blockstatus  from exchange to blocklayer)
2) fixing #155 issue
3) Making the transparent block mode optional
4) adding Etag support

For the time being, I like this idea to remove blockstatus from exchange to blocklayer.

sophokles73:
You are right, Simon. It could (should?) be split up into several steps. But I guess you know how it is if you start working on something that touches multiple areas of the code base ...
I can, though, try to re-factor the PR into several ones if that helps with reviewing and getting it merged more easily/quickly.

Moving the blockwise status to BlockwiseLayer also opens up the opportunity to move the BlockwiseLyer up in the stack one further step, I guess so that we get even closer to eventually moving it up into the application layer altogether. But that is something for the future :-)

sbernard31:
I know but I don't talk about not touching multiple areas of the code base.. I talk about limited the work to a precise goal. I know it's hard and I suppose I'm not perfect in this domain too. But if we want effective review,  I think we should be vigilant about that, especially for massive code change. As you say at eclipse con "one step at the time" :)

For this one, I don't advice you do re-split the PR. With the amount of code change this could be really error prone but I can not warranty that my review will be really efficient :)

sophokles73:
I am also prone to this fallacy of not *practicing what I preach* ;-)
I'd still very much appreciate any efforts for reviewing the PR, Simon. Your input is always welcome and important to me. I am pretty confident with this one, though, because I have spent a lot of time with the test cases as well.

boaks:
I try to understand, why Exchange still need currentRequest/currentResponse.

californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
```diff
@@ -129,13 +135,12 @@
...
 	private ScheduledFuture<?> retransmissionHandle = null;
 
-	// handle to extend blockwise status lifetime
-	private ScheduledFuture<?> blockCleanupHandle = null;
-
 	// If the request was sent with a block1 option the response has to send its
 	// first block piggy-backed with the Block1 option of the last request block
 	private BlockOption block1ToAck;
 
+	private Integer observeNo;
```
boaks: Why is this field named "observe..." but used only in "notification...()" methods?
Doc comment missing.

boaks:
I tested several firmware downloads successfully.
I will prepare for retest the #30.

californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
sophokles73: It's a private field thus, no Java Doc. The purpose of the field is described in the JavaDoc for the getter/setter, which, as you pointed out correctly, are not easy to identify as such ;-)
I should probably change this field's name to `notificationNumber` ...

boaks:
> 91.189236	CoAP	64	CON, MID:36079, GET, TKN:eb c6, End of Block #755, /files/fw.bin
> 91.189746	CoAP	180	ACK, MID:36079, 2.05 Content, TKN:e9 c6, Block #10 (text/plain)

OK, so, as long as the response has a matching MID, the client should verify the blocks ...
The PR would therefore also close #30.

californium-core/src/main/java/org/eclipse/californium/core/network/TcpMatcher.java
```diff
@@ -211,16 +156,7 @@ public void completed(final Exchange exchange) {
 					exchangeStore.releaseToken(idByToken);
```
boaks: if (!exchange.getCurrentRequest().getOptions().hasObserve()) {
				}

request with observe option 1 may be cleaned up on completion. rebase and use new Request.isObserve() ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -139,39 +159,38 @@ public BlockwiseLayer(final NetworkConfig config) {
...
+		exchange.setCurrentRequest(requestToSend);
+		lower().sendRequest(exchange, requestToSend);
 	}
 
-	private void startBlockwiseUpload(final Exchange exchange, final Request request) {
+	private Request startBlockwiseUpload(final Exchange exchange, final Request request) {
 
-		BlockwiseStatus status = findRequestBlockStatus(exchange, request);
+		final KeyUri key = getKey(exchange, request);
+		Block1BlockwiseStatus status = getOutboundBlock1Status(key, exchange, request);
```
sbernard31: This methods seems to be used to start a new blockwise Upload, so we should start with a fresh `BlockwiseStatus`
But in `getOutboundBlock1Status` we can return an already existing `Block1BlockwiseStatus` ?
I suspect this is the case where we try to do 2 blockwise for the same endpoint/same resource right ?

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -460,13 +380,6 @@ public void completed(final Exchange exchange) {
 					LOGGER.log(Level.FINER, "Exchange [{0}, {1}] completed", new Object[]{midKey, exchange.getOrigin()});
```
boaks: if (!exchange.getCurrentRequest().getOptions().hasObserve()) {
				}

Same as in TcpMatcher, request with observe option 1 may be cleaned up on completion. rebase and use new Request.isObserve() ?

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
```diff
@@ -76,7 +76,7 @@ public static void setDefaults(final NetworkConfig config) {
 		config.setInt(NetworkConfig.Keys.PREFERRED_BLOCK_SIZE, 512);
 		config.setInt(NetworkConfig.Keys.MAX_MESSAGE_SIZE, 1024);
 		config.setInt(NetworkConfig.Keys.MAX_RESOURCE_BODY_SIZE, DEFAULT_MAX_RESOURCE_BODY_SIZE);
-		config.setInt(NetworkConfig.Keys.BLOCKWISE_STATUS_LIFETIME, 10 * 60 * 1000); // ms
+		config.setInt(NetworkConfig.Keys.BLOCKWISE_STATUS_LIFETIME, DEFAULT_BLOCKWISE_STATUS_LIFETIME); // ms
```
boaks: Reason for changing the default from 10 minutes to 30 seconds?
A device restart or a lost DTLS session may exceed 30 seconds, so would go for a little more. 5 minutes?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -192,195 +211,256 @@ public void onAcknowledgement() {
...
-				error.getOptions().setBlock1(block1.getSzx(), block1.isM(), block1.getNum());
-				error.setPayload("Wrong block number");
-				exchange.setCurrentResponse(error);
+					// Remember block to acknowledge. TODO: We might make this a boolean flag in status.
+					exchange.setBlock1ToAck(block1); 
+
+					// Assemble and deliver
+					Request assembled = new Request(request.getCode());
+					assembled.setSenderIdentity(request.getSenderIdentity());
+					status.assembleMessage(assembled);
```
boaks: I'm not sure, if the MID and token from the first request should be used.
Even if it may only be used for logging, it may be confusing.
If it is used for the response, it would be wrong.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseStatus.java
```diff
@@ -20,89 +20,78 @@
...
 	 * It would be nice if we could get rid of this. Currently, the Cf client
 	 * needs it to mark a blockwise transferred notification as such. The
 	 * problem is, that the server includes the observe option only in the first
 	 * block of the notification and we still need to remember it, when the
 	 * last block arrives (block-14).
 	 */
-	/** The observe sequence number of this blockwise transfer */
-	private int observe = NO_OBSERVE;
+	protected int observe = NO_OBSERVE;
+	protected ByteBuffer buf;
```
boaks: `buf` is accessed sometimes after null check (getBufferSize) and sometimes without (addBlock and getBody).
Though a constructor without createing a buf exists, I think adding IllegalStateException would be a improvement.

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
sbernard31: i was a bit confuse when I saw this 30 seconds too.
Is it the whole time available to get or send the whole payload ? I mean all the blockwise request/response exchanges needed to send or receive the initial message payload ? Even 5 minutes seems to me really short. Just to get the first ACK of the first request this can take 247s (EXCHANGE_LIFETIME)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
```diff
@@ -0,0 +1,302 @@
...
+	 * @param response The CoAP response to be transferred blockwise.
+	 * @param preferredBlockSize The default size to use for individual blocks. If the exchange's request contains
+	 *                           an <em> early negotation</em> block2 option then the size indicated by that
+	 *                           option is used as the block size.
+	 * @return The tracker.
+	 */
+	static Block2BlockwiseStatus forOutboundResponse(final Exchange exchange, final Response response, final int preferredBlockSize) {
+		Block2BlockwiseStatus status = new Block2BlockwiseStatus(response.getOptions().getContentFormat());
+		status.response = response;
+		status.buf = ByteBuffer.allocate(response.getPayloadSize());
```
boaks: Why not using the othe constructor and let the base calss allocate?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
```diff
@@ -0,0 +1,302 @@
...
+	private static final Logger LOGGER = Logger.getLogger(Block2BlockwiseStatus.class.getName());
+
+	private Response response;
+	private byte[] etag;
+
+	/**
+	 * @param maxSize
+	 * @param contentFormat
+	 */
+	private Block2BlockwiseStatus(final int contentFormat) {
```
boaks: May be removed, if other constructor is used.

boaks: The currently caller `OutboundResponse` may better call the other constructor to use the allocate of the base class.
And the canditate for this `forRandomAccessRequest` requires the other 2 parameters.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	/**
+	 * Adds the payload of an incoming response to the buffer.
+	 * 
+	 * @param responseBlock The incoming response.
+	 * @param block2 The block2 option contained in the response.
+	 * @return {@code true} if the payload could be added.
+	 */
+	synchronized boolean addBlock(final Response responseBlock, final BlockOption block2) {
```
boaks: I'm not sure, if passing in a already accessed field as additional parameter "pays off".
In general, it MUST be the field, and the split in the parameter list seems to be a optimization.
(We could save a lot comments :-) ).

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
boaks: Uups, if simon is right, and it's the time from the first request/reponse, then we shold
1. keep the 10 minutes and 
2. create a issue to change that as timeout related to the last traffic.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
```diff
@@ -0,0 +1,145 @@
...
+ */
+final class Block1BlockwiseStatus extends BlockwiseStatus {
+
+	private Request request;
+
+	private Block1BlockwiseStatus(final int contentFormat) {
+		super(contentFormat, 0, 0);
+	}
+
+	private Block1BlockwiseStatus(final int contentFormat, final int bufferSize) {
```
boaks: Parameter swap to base class intended?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
```diff
@@ -0,0 +1,145 @@
...
+	 * 
+	 * @param exchange The message exchange the transfer is part of.
+	 * @param request The CoAP request containing the large body in its payload.
+	 * @param preferredBlockSize The size to use for individual blocks.
+	 * @return The tracker.
+	 */
+	static Block1BlockwiseStatus forOutboundRequest(final Exchange exchange, final Request request, final int preferredBlockSize) {
+		Block1BlockwiseStatus status = new Block1BlockwiseStatus(request.getOptions().getContentFormat());
+		status.request = request;
+		status.setCurrentSzx(BlockOption.size2Szx(preferredBlockSize));
```
boaks: Maybe consider adding this as parameter to the constructor.

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
sophokles73: This is the maximum time span *between* subsequent requests of the blockwise transfer. The timer is reset every time we receive a new request for an inbound block1 or block2 request.
I reduced this value drastically because I wasn't aware that it actually needs to be very high in order to support low-bandwidth, low-reliability transfers.
I have no problem with increasing it again. The Blockwise Transfer RFC mandates that clients do all required block transfers in rapid succession without further delays in between. I therefore assume that a value based on  EXCHANGE_LIFETIME could serve well.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
sophokles73: Yes, I was also thinking about the pros and cons of this. The biggest con IMHO is that we cannot be sure that the passed in block2 is really the block2 from the response (which it has to be for the method to produce the correct result).

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
boaks: > This is the maximum time span between subsequent requests

Thats good! 

> EXCHANGE_LIFETIME could serve well

Would be OK for me as default.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: I guess you're right. In a response to an inbound block1 upload we would need to use the token and MID from the last block1 request we received, right?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: The spec seems to say that we should just start a fresh blockwise sequence ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -397,35 +477,41 @@ public void sendResponse(final Exchange exchange, final Response response) {
...
-			if (response.getType()!=Type.ACK) {
-				LOGGER.finer("rejecting blockwise transfer for canceled Exchange");
-				EmptyMessage rst = EmptyMessage.newRST(response);
-				sendEmptyMessage(exchange, rst);
-				// Matcher sets exchange as complete when RST is sent
-			}
+		if (isTransparentBlockwiseHandlingEnabled()) {
 
-		} else if (!response.hasBlockOption()) {
+			if (exchange.getRequest().isCanceled()) {
```
sophokles73: You mean we should move this code block down into the `else` branch where we are dealing with responses that contain a block option?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: Yes, and the Matcher and ReliabilityLayer may even overwrite the wrong assembled values.
But the logs will be mixup, when we want to analyse such scenarios.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: It says so for multiple *inbound* block1 transfers from either the same or different clients.
However, in this case we are dealing with *outbound* block1 transfers, i.e. we are the client uploading a large body to a peer.

If we use leshan as an example, this would translate to: multiple threads or apps built on top of a leshan server issue block1 requests to the same resource on the same LWM2M client.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: I think, I begin to understand, that "non" is "none" (no blockwise response) and not NON (Non Confirmed).
So I agree, delete it from the blockwise layer.
There maybe something in the observer layer (NON/CON Notifies) and something in the ReliabilityLayer (CON).
But just sending RST to all none ACK seems to be strange to me.

sophokles73: I guess we need to deal with it at the Blockwise Layer because the response we received here will be a response to a request issued by the Blockwise Layer (not the application layer). But I guess we can safely move the block down to the `else` branch.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: I am proposing to not start the second request but instead return a 4.12 to the first request of the client that wants to start the second block1 upload. IMHO we should handle this on a *first come, first serve' basis. Why should the first client's request be cancelled if somebody else later on also wants to do an upload on the same resource? In a database you would also have the table or record locked by the first access which would prevent a second access to the same resource, wouldn't you?

californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
sophokles73: Done

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: I have changed it accordingly for an inbound block1 upload. For an inbound block2 download (where we retrieve multiple blocks transparently from a peer and deliver the assembled body to the app layer) I think using the MID and token of the *first* outbound request for the final (assembled) response seems right or doesn't it?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
```diff
@@ -0,0 +1,302 @@
...
+
+	/**
+	 * @param maxSize
+	 * @param contentFormat
+	 */
+	private Block2BlockwiseStatus(final int contentFormat) {
+		super(contentFormat, 0, 0);
+	}
+
+	private Block2BlockwiseStatus(final int contentFormat, final int bufferSize) {
```
sophokles73: No, I have fixed this :-)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
sophokles73: done

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseStatus.java
sophokles73: I have changed the super class constructors to always allocate a byte buffer (default size 0)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: For block2 I would agree to use MID/token from the first response, as long as no RST is sent for that response.
If we would be able to RST the response, we need to use the MID from the last.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: if (isTransparentBlockwiseHandlingEnabled()) {

			if (exchange.getRequest().isCanceled()) {
				// do not continue fetching blocks if canceled
				// reject (in particular for Block+Observe)
				if (response.getType() != Type.ACK) {
					LOGGER.finer("rejecting blockwise transfer for canceled Exchange");
					EmptyMessage rst = EmptyMessage.newRST(response);
					sendEmptyMessage(exchange, rst);
					// Matcher sets exchange as complete when RST is sent
				}

I think, the 
			if (exchange.getRequest().isCanceled()) {
				// do not continue fetching blocks if canceled

is at the right place here. But

				if (response.getType() != Type.ACK) {
					LOGGER.finer("rejecting blockwise transfer for canceled Exchange");
					EmptyMessage rst = EmptyMessage.newRST(response);
					sendEmptyMessage(exchange, rst);
					// Matcher sets exchange as complete when RST is sent
				}

is not only at the wrong place, it seems to be mixed up. I don't see, none ACK should be answered with RST.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -527,302 +620,328 @@ private void handleBlock2Response(final Exchange exchange, final Response respon
...
 
 			} else {
-				LOGGER.log(Level.FINER, "We have received all {0} blocks of the response. Assemble and deliver", responseStatus.getBlockCount());
-				Response assembled = new Response(response.getCode());
 
-				assembleMessage(responseStatus, assembled);
+				// ERROR, wrong block number (server error)
+				// TODO: This scenario is not specified in the draft.
+				// Canceling the request would interfere with Observe, so just ignore it
+				LOGGER.log(Level.WARNING,
```
sbernard31: we say we just ignore it, but we send a `RST` ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
sbernard31: I agree with @boaks.
Same "issue" in a lot of private functions in `BlockwiseLayer`.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -527,302 +620,328 @@ private void handleBlock2Response(final Exchange exchange, final Response respon
...
-				// copy options
-				block.setOptions(new OptionSet(request.getOptions()));
-				// make sure NOT to use Observe for block retrieval
-				block.getOptions().removeObserve();
+					status.setCurrentNum(nextNum);
 
-				block.getOptions().setBlock2(szx, m, num);
+					LOGGER.log(Level.FINER, "requesting next Block2 [num={0}]: {1}", new Object[]{ nextNum, block });
+					exchange.setCurrentRequest(block);
+					lower().sendRequest(exchange, block);
```
sbernard31: For observe case :  when we change the current request we could get a request with a new token value. I can not find where we clean the previous association `previousCurrentResquestToken` ==> `exchange`in the exchangeStore ?

sbernard31:
I think I'm done.
Maybe my remarks above are not directly linked to this PR, but I try to understand the californium blockwise implementation in a general way.

I feel the code will be more maintainable now, good job ;).

@sophokles73, do you think it will be possible to remove the currentRequest/currentResponse from exchange ? (not now, but in the future).  I feel this is maybe the last tricky part, but I'm not sure we can deal without it.

boaks:
>  I feel this is maybe the last tricky part, but I'm not sure we can deal without it.

I've also strange feelings about that trick ;-).

https://github.com/eclipse/californium/pull/170#issuecomment-265744902

sophokles73:
> do you think it will be possible to remove the currentRequest/currentResponse from exchange ? (not now, but in the future). I feel this is maybe the last tricky part, but I'm not sure we can deal without it.

I definitely hope so :-) However, I have not thought it through yet and therefore do not know what impact that might have ...

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: IMHO we should stop an ongoing outbound block1 or inbound block2 transfer if the original request got canceled by the application layer and clear the blockwise status accordingly.
I agree with @boaks that we do not need to send an RST here. If we receive a response (non-empty) in this method then it contains a block of the body or an error code and we can simply clear the block1 and/or block2 status and forget about the request (the application layer is not interested in the response anymore).
However, we will probably need to deal with empty responses we receive for block requests we have sent out subsequent to the initial request. We will need to e.g. mark the original request as rejected.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -527,302 +620,328 @@ private void handleBlock2Response(final Exchange exchange, final Response respon
...
+			// the corresponding Exchange and pass the response up the stack to the BlockwiseLayer.
+			// The problem we are facing is that the key used to look up the Block2BlockwiseStatus
+			// will always be the same if no ETag is contained in the notification (the key then only
+			// depends on the request's URI and source endpoint). So there is no way to distinguish a
+			// response to an outdated block2 request from a response that is part of the most recent
+			// block2 transfer.
+			// In order to be able to distinguish between a block of an old vs a new notification,
+			// the Exchange used to initially create the Block2BlockwiseStatus is
+			// marked with the notification's observe option (see Block2BlockwiseStatus.forInboundResponse()).
+			// We can then compare the observe option value from the Exchange that hasa been looked up
```
sophokles73: fixed

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -527,302 +620,328 @@ private void handleBlock2Response(final Exchange exchange, final Response respon
...
+			// for the incoming response with the value of the most recently received notification from the
+			// Block2BlockwiseStatus that has been looke up for the key.
```
sophokles73: fixed

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: I am not sure if I understand what you mean. Can you elaborate a little on what you suspect could be a problem here?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: I guess it means that we *do not process* the block returned by the peer. However, sending an RST doesn't make much sense here because the `ReliabiliytLayer` will already have ACKed the CON response. I will remove it ...

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
sbernard31: Making the code less readable to avoid the use of some getters, I'm not sure this make senses. All of this refactoring is to make the code more maintainable and readable. 

Not directly linked to this but [that][1] is pretty much how I see performance optimization : 

Did you measure the performance gain of this optimization ? 

[1]:https://stackoverflow.com/questions/6495030/java-how-expensive-is-a-method-call/6495096#6495096

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: Each time we send a request we add an entry in `exchangeStore` (token => exchange).
It seems that this entry is cleaned when `exchange` is completed.

With blockmode with notification we have a new token for each block, so we create a new (token => exchange) entry by block requested. But I can not see where we clean this entry. 

Not sure this is clearer ? (Not sure this is really clear in my head too :p)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: Yes, exactly. The whole point is not to *invalidate* the already ongoing transfer. You seem to think that this is not a good idea. May I ask why?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: This is done in the `handleBlock2Response()` method. More specifically, in the branch where the last block is handled:
```

				} else {

					// we have received the last block of the block2 transfer

					LOGGER.log(
							Level.FINER,
							"all {0} blocks have been retrieved, assembling response and delivering to application layer",
							status.getBlockCount());
					Response assembled = new Response(response.getCode());
					status.assembleMessage(assembled);

					// set overall transfer RTT
					assembled.setRTT(System.currentTimeMillis() - exchange.getTimestamp());

					if (status.isNotification()) {

						/*
						 * When retrieving the rest of a blockwise notification
						 * with a different token, the additional Matcher state
						 * must be cleaned up through the call below.
						 */
						if (!response.getOptions().hasObserve()) {
							// call the clean-up mechanism for the additional Matcher entry in exchangesByToken
							exchange.completeCurrentRequest();
						}
					}

					clearBlock2Status(key);
					LOGGER.log(Level.FINE, "assembled response: {0}", assembled);
					// Set the original request as current request so that
					// the Matcher can clean up its state based on the latest
					// ("current") request's MID and token
					exchange.setCurrentRequest(exchange.getRequest());
					// Set the assembled response as current response
					exchange.setResponse(assembled);
					upper().receiveResponse(exchange, assembled);
				}

```
Since we are using the same token for retrieving all blocks it is sufficient to clean up once the last block has been received.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: This was only a "theoretical scenario".  

> If we **would** be able to RST the response, we need to use the MID from the last.

**would** should be **should**.
The point was just, if someone has problems using RST on block2, then it may be interessting, that this solution doesn't considered that.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -192,195 +211,262 @@ public void onAcknowledgement() {
...
+					responseToSend.setType(response.getType());
+					responseToSend.setMID(response.getMID());
+					responseToSend.getOptions().setBlock2(requestBlock2);
+
+				}
+
+			} else if (requiresBlockwise(exchange, response, requestBlock2)) {
+
+				KeyUri key = getKey(exchange, response);
+				Block2BlockwiseStatus status = getOutboundBlock2Status(key, exchange, response);
```
sophokles73: When we receive multiple requests for the same resource coming from the same client (the scenario described above) then we will need to assume that the client wants to *start over* with the blockwise download. Based on what I wrote above, the only thing that would need to be changed is that we do not use `status.getNextResponseBlock()` (with no arg) but instead use `status.getNextResponseBlock(BlockOption)` with a newly instantiated BlockOption as the arg (indicating that we want to get the *first* response block). Explicitly creating a new instance (replacing a potentially already existing one) would also work but requires more code, I guess.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: I missed we used the same token, so this should be ok.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: Ok. It seems more intuitive to create new blockstatus for new block exchange. But If there is good reason to reuse the previous one, we should explain this in a comment.

sophokles73: I have added a corresponding comment.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: I have made some changes:
1. clear blockwise status for canceled requests
2. delegate handling of notification received for an already canceled observation to upper layer(s).

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sophokles73: Well, you would accept a response from an intermediary proxy instead of a server as well, wouldn't you? This seems to be the only remaining issue now before I can merge this PR :-)
Coming back to your proposal to cancel the first request: So far, *canceling* was something the  client code had to do deliberately, wasn't it? In any case, the outcome of the (first) client's request is not determined by the server as well if we follow your proposal, i.e. we also act *on behalf of the server* in this case by simply canceling the first client's request. IMHO it would be more informative for the client to receive an (error) response containing a code an payload so that it can better determine the reason why its request failed and react in a more appropriate way. WDYT?

sbernard31: (Just to be sure,  we talk about client and server CoAP)

> Well, you would accept a response from an intermediary proxy instead of a server as well, wouldn't you?

I don't see the problem here, I received a response from the server I connected to. (proxy or not)

> So far, canceling was something the client code had to do deliberately, wasn't it?

Yes, and we could assume that when the client send a request in block1 on a resource on which it already has a pending request, it want deliberately cancel the previous one.

This is exactly like when you put a new value in a map for an existing key. There is 2 implementation choices. You can refuse the addition or replace the old value by the new one.  In the second case, the map remove the old value for you but you can not say that you didn't do this deliberately.

> we also act on behalf of the server in this case by simply canceling the first client's request.

I don't think so. We just said before that cancelling request is be done at client initiative.

>IMHO it would be more informative for the client to receive an (error) response containing a code an payload so that it can better determine the reason why its request failed and react in a more appropriate way.

I'm still thinking this is not a good idea to make "as if" the server send a response.

sophokles73: > This is exactly like when you put a new value in a map for an existing key. There is 2 implementation choices. You can refuse the addition or replace the old value by the new one. In the second case, the map remove the old value for you but you can not say that you didn't do this deliberately.

I do not think that the analogy is correct here. When you put a value to a map you are talking about an atomic and consistent operation. Doing a blockwise transfer on the other hand consists of multiple PUT/POST requests. The new state is established only at the end of the whole transfer.
Therefore the analogy would more be like: you want to put a series of values for multiple keys to a map and only if you are able to put all values to the map will you get a new consistent state. In order to achieve this in a multi-threaded environment you would need to synchronize the access, i.e. as long as the one thread puts its values to the map the other thread has to wait. But you wouldn't interweave the individual puts to the map from the two threads....

sbernard31: My analogy just aims to explain why this could be considered as a deliberated action, nothing more. All of this depends on how you describe your API. This was just in response of :

>So far, canceling was something the client code had to do deliberately, wasn't it?

(I should be more cautious with analogies :/ )

sophokles73: But doesn't the argument still apply? The first client's intention is to PUT/POST data to a resource. Due to the size of the body this requires a blockwise transfer. Don't you think that the client's intention is to either upload the whole body or nothing at all? If a second PUT/POST (from another thread representing e.g. another client app on top of leshan) now triggers cancelation of the first block1 transfer, this could hardly be called *deliberately canceled by client one*, could it?
The *locking* analogy still applies IMHO ...

sbernard31: > Don't you think that the client's intention is to either upload the whole body or nothing at all? If a second PUT/POST (from another thread representing e.g. another client app on top of leshan) now triggers cancelation of the first block1 transfer, this could hardly be called deliberately canceled by client one, could it?

Probably not, but this is because you consider that there are 2 different "entities" (you call it _client app_) using the Californium client API  and sharing the same CoAP server resource without knowing each other.

FMPOV, There is only one user of the Californium client API, which should know what he is doing. In this case we could easily say that this is done deliberately.  If he decide to send a new request before it receive the response to the first one, this means that finally he doesn't need to send the previous request, only the last value make sense to be PUT/POST. 
If it want to support the use case you described, up to him to choose/implement if the first request should be cancelled or the second one rejected or maybe queued.

sophokles73: > Probably not, but this is because you consider that there are 2 different "entities" (you call it client app) using the Californium client API and sharing the same CoAP server resource without knowing each other.

Ok, now we're getting somewhere. So you are arguing that the Californium API is NOT (and should not be) intended to be used by multiple *independent* clients, right? We might have had this constraint already in Californium (even if not by intention but by failure of implementation) so I cannot really argue against it. However, this also seems to be a problem of the fact that we are doing *transparent blockwise handling* within the stack instead of in the client (where it would belong IMHO). So, we could say that this case of concurrent block1 uploads needs to be handled by the client(s) until we have refactored the BlockwiseLayer out of the stack and into a support library to be employed by `CoapClient` and `CoapResource`. Until then we could follow your approach. WDYT?

sbernard31: > So you are arguing that the Californium API is NOT (and should not be) intended to be used by multiple independent clients, right?

yes, It seems to me.

> Until then we could follow your approach. WDYT?

sounds good.

> However, this also seems to be a problem of the fact that we are doing transparent blockwise handling within the stack instead of in the client (where it would belong IMHO).

if the blockwise was handled out of the stack, each  stack clients would have its own blockwise state. So at client side nothing would avoid to have 2 block1 at the same time for the same resource/same server. But this would not be the case at server side. And when this would happen, the first request would be dropped. This would be near to cancelling approach.

sophokles73: Okay, I have just pushed the latest (last?) change based on our discussions, @sbernard31 :-)
I guess this concludes this PR and I would now like to merge it to 2.0.x. WDYT?

sbernard31: Good for me :)

PR #177, opened 2016-12-09
HP fortify issue Fix

Rajsmit:
Fixed some of the security issue

1. Code correctness :- Constructor invokes overridable function :-
CoapServer.java, BlockOption.java, Option.java, RemoteEndpoint.java,

2. Insecure Randomness :- MessageIdTracker.java, ReliabilityLayer.java

3. Redundent Null Check :- CoapServer.java, UdpMatcher.java

4. Object Model violation Just one equals() and HashCode() defined :-
BlockOption.java

5. Code Correctness :- Errorneous string Compare in Switch case.
Deduplicator.java

Signed-off-by: Rajesh Kumar <rajsmit@gmail.com>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
@@ -147,7 +147,7 @@ public class CoapServer implements ServerInterface {
 		}
 		
 		// resources
-		this.root = new RootResource();
+		this.root = createRoot();
 		this.deliverer = new ServerMessageDeliverer(root);
 		
 		CoapResource wellKnown = new CoapResource(".well-known");
@@ -157,17 +157,13 @@ public class CoapServer implements ServerInterface {
 		
 		// endpoints
 		this.endpoints = new ArrayList<>();
-		if(null != config) {
-			// sets the central thread pool for the protocol stage over all endpoints
-			this.executor = Executors.newScheduledThreadPool(//
-					config.getInt(NetworkConfig.Keys.PROTOCOL_STAGE_THREAD_COUNT), //
-					new Utils.NamedThreadFactory("CoapServer#")); //$NON-NLS-1$
-		}
-		if(null != this.config) {
-			// create endpoint for each port
-			for (int port : ports) {
-				addEndpoint(new CoapEndpoint(port, this.config));
-			}
+		// sets the central thread pool for the protocol stage over all endpoints
+		this.executor = Executors.newScheduledThreadPool(//
+				config.getInt(NetworkConfig.Keys.PROTOCOL_STAGE_THREAD_COUNT), //
+				new Utils.NamedThreadFactory("CoapServer#")); //$NON-NLS-1$
+		// create endpoint for each port
+		for (int port : ports) {
+			addEndpoint(new CoapEndpoint(port, this.config));
 		}
 	}
 
@@ -309,7 +305,7 @@ public class CoapServer implements ServerInterface {
 	 * @param endpoint the endpoint to add
 	 */
 	@Override
-	public final void addEndpoint(final Endpoint endpoint) {
+	public void addEndpoint(final Endpoint endpoint) {
 		endpoint.setMessageDeliverer(deliverer);
 		endpoint.setExecutor(executor);
 		endpoints.add(endpoint);
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
@@ -147,7 +147,7 @@ public class CoapServer implements ServerInterface {
 		}
 		
 		// resources
-		this.root = createRoot();
+		this.root = new RootResource();
 		this.deliverer = new ServerMessageDeliverer(root);
 		
 		CoapResource wellKnown = new CoapResource(".well-known");
@@ -157,13 +157,17 @@ public class CoapServer implements ServerInterface {
 		
 		// endpoints
 		this.endpoints = new ArrayList<>();
-		// sets the central thread pool for the protocol stage over all endpoints
-		this.executor = Executors.newScheduledThreadPool(//
-				config.getInt(NetworkConfig.Keys.PROTOCOL_STAGE_THREAD_COUNT), //
-				new Utils.NamedThreadFactory("CoapServer#")); //$NON-NLS-1$
-		// create endpoint for each port
-		for (int port : ports) {
-			addEndpoint(new CoapEndpoint(port, this.config));
+		if(null != config) {
+			// sets the central thread pool for the protocol stage over all endpoints
+			this.executor = Executors.newScheduledThreadPool(//
+					config.getInt(NetworkConfig.Keys.PROTOCOL_STAGE_THREAD_COUNT), //
+					new Utils.NamedThreadFactory("CoapServer#")); //$NON-NLS-1$
+		}
+		if(null != this.config) {
+			// create endpoint for each port
+			for (int port : ports) {
+				addEndpoint(new CoapEndpoint(port, this.config));
+			}
 		}
 	}
 
@@ -305,7 +309,7 @@ public class CoapServer implements ServerInterface {
 	 * @param endpoint the endpoint to add
 	 */
 	@Override
-	public void addEndpoint(final Endpoint endpoint) {
+	public final void addEndpoint(final Endpoint endpoint) {
 		endpoint.setMessageDeliverer(deliverer);
 		endpoint.setExecutor(executor);
 		endpoints.add(endpoint);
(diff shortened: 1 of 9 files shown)


californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
```diff
@@ -147,7 +147,7 @@ public CoapServer(final NetworkConfig config, final int... ports) {
 		}
 		
 		// resources
-		this.root = createRoot();
+		this.root = new RootResource();
```
sophokles73: This effectively removes the possibility to use another root resource by extending `CoapServer`.

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
```diff
@@ -157,13 +157,17 @@ public CoapServer(final NetworkConfig config, final int... ports) {
...
 		// endpoints
 		this.endpoints = new ArrayList<>();
-		// sets the central thread pool for the protocol stage over all endpoints
-		this.executor = Executors.newScheduledThreadPool(//
-				config.getInt(NetworkConfig.Keys.PROTOCOL_STAGE_THREAD_COUNT), //
-				new Utils.NamedThreadFactory("CoapServer#")); //$NON-NLS-1$
-		// create endpoint for each port
-		for (int port : ports) {
-			addEndpoint(new CoapEndpoint(port, this.config));
+		if(null != config) {
```
sophokles73: The solution to this problem is not checking for `null` but using `this.config` instead (it has been initialized to a non-null value in the code above).

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
```diff
@@ -305,7 +309,7 @@ public MessageDeliverer getMessageDeliverer() {
 	 * @param endpoint the endpoint to add
 	 */
 	@Override
-	public void addEndpoint(final Endpoint endpoint) {
+	public final void addEndpoint(final Endpoint endpoint) {
```
sophokles73: why do we need to make this method final and none of the others?

californium-core/src/main/java/org/eclipse/californium/core/coap/BlockOption.java
```diff
@@ -235,6 +235,14 @@ public boolean equals(Object o) {
 		return szx == block.szx && num == block.num && m == block.m;
 	}
 
+	@Override
+	public int hashCode() {
+		int result = szx;
+		result = 31 * result + (m ? 1 : 0);
+		result = 31 * result + num;
+		return result;
+	}
```
sophokles73: nice catch

californium-core/src/main/java/org/eclipse/californium/core/coap/Option.java
```diff
@@ -199,7 +199,7 @@ public long getLongValue() {
 	 *
 	 * @param value the new value
 	 */
-	public void setValue(byte[] value) {
+	public final void setValue(byte[] value) {
```
sophokles73: Why not make the whole class final?

californium-core/src/main/java/org/eclipse/californium/core/network/MessageIdTracker.java
```diff
@@ -57,7 +57,7 @@ public MessageIdTracker(final NetworkConfig config) {
 		exchangeLifetime = config.getLong(NetworkConfig.Keys.EXCHANGE_LIFETIME);
 		boolean useRandomFirstMID = config.getBoolean(NetworkConfig.Keys.USE_RANDOM_MID_START);
 		if (useRandomFirstMID) {
-			counter = new AtomicInteger(new Random().nextInt(1 << 10));
+			counter = new AtomicInteger(new SecureRandom().nextInt(1 << 10));
```
sophokles73: There is no need to use a cryptographically strong random in order to create the initial counter value. The counter will be incremented sequentially only afterwards ...

californium-core/src/main/java/org/eclipse/californium/core/observe/ObserveRelation.java
```diff
@@ -163,7 +163,7 @@ public Exchange getExchange() {
 	 *
 	 * @return the source address
 	 */
-	public InetSocketAddress getSource() {
+	public final InetSocketAddress getSource() {
```
sophokles73: Why do we need to make this method final but not the rest?

californium-core/src/main/java/org/eclipse/californium/core/server/resources/ConcurrentCoapResource.java
```diff
@@ -125,7 +125,7 @@ public ConcurrentCoapResource(String name, int threads) {
 	 * 
 	 * @param executor the executor service
 	 */
-	public void setExecutor(ExecutorService executor) {
+	public final void setExecutor(ExecutorService executor) {
```
sophokles73: same here, why only these two methods?

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
sophokles73: Maybe it is more about providing means for subclasses to provide a different *banner* for `GET /` requests? That could be easily achieved ...

californium-core/src/main/java/org/eclipse/californium/core/coap/BlockOption.java
sbernard31: I suppose the idea is to say that blockOption is now immutable ? 
- Should we still need setter (as there are private now)?
- Should we not define szx, num as final ?

californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/DeduplicatorFactory.java
```diff
@@ -69,14 +69,14 @@ public static synchronized void setDeduplicatorFactory(DeduplicatorFactory facto
 	public Deduplicator createDeduplicator(final NetworkConfig config) {
 
 		String type = config.getString(NetworkConfig.Keys.DEDUPLICATOR, NetworkConfig.Keys.NO_DEDUPLICATOR);
-		switch(type) {
-		case NetworkConfig.Keys.DEDUPLICATOR_MARK_AND_SWEEP:
+		if (type.equals(NetworkConfig.Keys.DEDUPLICATOR_MARK_AND_SWEEP)) {
```
sbernard31: Why replacing switch by if/else if ?
https://docs.oracle.com/javase/7/docs/technotes/guides/language/strings-switch.html

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
sbernard31: Why should we limit  custom behavior to  `GET /` request only ?

E.g. For  `Bootstrap delete` in LWM2M, you need a custom behavior on `DETELE /`  :)

sophokles73: I didn't want to limit the possibilities, I was just suspecting that people might not want the default response with the `CoAP RFC Californium` bla bla bla ;-)

Rajsmit: Right now it is very urgent for us to fix it. We can still live with it. For now we can park this until we find a better solution for this.

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
Rajsmit: I Agree, I will change it to use  this.config instead config and remove null check as this.config is non-null.

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
Rajsmit: This is same as point 1 so lets park it for now until we have more structured solution.

californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/DeduplicatorFactory.java
Rajsmit: I agree with you while using jdk7 and above it is not an issue.

californium-core/src/main/java/org/eclipse/californium/core/network/MessageIdTracker.java
Rajsmit: I Agree with you. I will remove it.

californium-core/src/main/java/org/eclipse/californium/core/coap/Option.java
Rajsmit: This is same as point 1 so lets park it for now until we have more structured solution.

californium-core/src/main/java/org/eclipse/californium/core/observe/ObserveRelation.java
Rajsmit: This is same as point 1 so lets park it for now until we have more structured solution.

sbernard31:
sounds good to me now :)

Issue #178, opened 2016-12-13
Infinite ping pong under TcpStack


joemag1:
If a TcpStack receives an empty message (which is possible, if for example the peer receives a badly formed message and calls CoapEndpoint.reject()) this will initiate an infinite ping pong. This happens because the peer receiving an empty message will itself send an empty message to the friend.

The offending line of code is in CoapEndpoint.reject(), where it checks if type is CON (under TCP we always hard code message type to CON, because we have to set it to something), it will send reject empty message. Which will trigger the same exact thing on the other peer.

The ping that initiates such ping pong is likely a failed message parsing, inside CoapEndpoint.receiveMessage() method.

boaks:
Hm, I didn't test it, but I thought, that both
TcpMatcher.receiveEmptyMessage and TcpAdaptionLayer.receiveEmptyMessage should discard the EmptyMessage on receiving. So doesn't this work?

boaks:
Ok, now I see, that's the InboxImpl!

joemag1:
This code path skips the stack and goes directly to the connector.

sophokles73:
fixed on master in fd9abfbdd3359533cfd50ed7bd93d3ccf08e6722
fixed on 2.0.x in 0f81d787068c940f89441037f3103c24694dee5e

PR #179, opened 2016-12-13
Fix for https://github.com/eclipse/californium/issues/178 (infinite ping pong)

joemag1:
Fixed a bug where empty message under TCP stack can cause an infinite ping pong between the client and the server. Added two layers of defense, one do not even allow empty messages to be deserialized by TCP stack: these are illegal. Secondly, added a protection by having endpoint ask CoapStack if it supports reject messages.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/RawPublicKeyIdentity.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/RawPublicKeyIdentity.java
@@ -125,7 +125,7 @@ public class RawPublicKeyIdentity implements Principal {
 	 */
 	@Override
 	public String toString() {
-		return new StringBuffer("RawPublicKey Identity [").append(niUri).append("]").toString();
+		return new StringBuilder("RawPublicKey Identity [").append(niUri).append("]").toString();
 	}
 
 	/**
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/util/Base64.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/util/Base64.java
@@ -967,7 +967,7 @@ public class Base64
 		// Else, don't compress. Better not to use streams at all then.
 		else {
 			boolean breakLines = (options & DO_BREAK_LINES) != 0;
-
+			boolean doPadding = (options & NO_PADDING) == 0;
 			//int    len43   = len * 4 / 3;
 			//byte[] outBuff = new byte[   ( len43 )                      // Main 4:3
 			//                           + ( (len % 3) > 0 ? 4 : 0 )      // Account for padding
@@ -975,7 +975,15 @@ public class Base64
 			// Try to determine more precisely how big the array needs to be.
 			// If we get it right, we don't have to do an array copy, and
 			// we save a bunch of memory.
-			int encLen = ( len / 3 ) * 4 + ( len % 3 > 0 ? 4 : 0 ); // Bytes needed for actual encoding
+			int encLen = ( len / 3 ) * 4; // number of full three byte sets encoded to corresponding 4 char set
+			int remainingBytes = len % 3;
+			if (doPadding) {
+				// encoding of remaining bytes will be padded to a full 4 byte char set
+				encLen += remainingBytes > 0 ? 4 : 0;
+			} else {
+				// remaining bytes will be encoded without padding
+				encLen += remainingBytes > 0 ? remainingBytes + 1 : 0;
+			}
 			if( breakLines ){
 				encLen += encLen / MAX_LINE_LENGTH; // Plus extra newline characters
 			}
--- a/scandium-core/src/test/java/org/eclipse/californium/scandium/auth/RawPublicKeyIdentityTest.java
+++ b/scandium-core/src/test/java/org/eclipse/californium/scandium/auth/RawPublicKeyIdentityTest.java
@@ -15,8 +15,11 @@
  ******************************************************************************/
 package org.eclipse.californium.scandium.auth;
 
-import static org.hamcrest.CoreMatchers.*;
-import static org.junit.Assert.*;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
 import java.security.GeneralSecurityException;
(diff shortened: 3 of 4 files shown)


joemag1:
And alternative way, might be to add validate(Message) method to coap stack, that gets invoked both on receiving and sending side, to decide if message should be dropped.  

It's a little less direct, but keeps the code more flexible for other future divergence between protocols.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -679,17 +679,20 @@ private void reject(final RawData raw, final MessageFormatException cause) {
 			rst.setToken(new byte[0]);
 			rst.setDestination(raw.getAddress());
 			rst.setDestinationPort(raw.getPort());
-			for (MessageInterceptor interceptor : interceptors) {
```
boaks: Removing the interceptors removes logging.

boaks:
From my understanding:
though the PING/PONG depends on the protocol flavor (TCP/UDP), we should try to move it to the Matcher and Stack and handle it there flavor aware.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -679,17 +679,20 @@ private void reject(final RawData raw, final MessageFormatException cause) {
...
-				interceptor.sendEmptyMessage(rst);
-			}
-			connector.send(serializer.serializeEmptyMessage(rst));
+
+			reject(rst);
```
boaks: May be to naive: 
but reject the RST, which was created to reject the original message, I'm not sure, if this is proper.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -679,17 +679,20 @@ private void reject(final RawData raw, final MessageFormatException cause) {
...
 				new Object[]{raw.getInetSocketAddress(), cause.getMessage()});
 		}
 
 		private void reject(final Message message) {
+			if (!coapstack.supportsRejects()) {
+				LOGGER.log(Level.FINE, "Rejects not supported under this stack.");
+				return;
+			}
+
 			EmptyMessage rst = EmptyMessage.newRST(message);
```
boaks: May be, I wrong, but on my first impression:

if the parameter message is a RST for a original message, the soruce/destination is already exchanged.

> EmptyMessage.newRST(message);

would exchange source/destination again.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
boaks: Hard to test: it's only executed, if a MessageFormatException occurs.
But I still believe, we would sent the double rejected message to our own, because we would also double exchange the source/destination.

boaks: ClientSynchronousTest

	@Test
	public void testSignal() throws Exception {
		/* ver. := 01, type:= CON (0), toklen := 0 | code := 7* | MID := 0x0102 */
		byte[] signal = new byte[] { 0x40, 0x07, 0x01, 0x02 };

		try (DatagramSocket socket = new DatagramSocket()) {
			DatagramPacket datagram = new DatagramPacket(signal, signal.length, InetAddress.getLoopbackAddress(), serverPort);
			DatagramPacket receive = new DatagramPacket(new byte[128], 128);
			socket.setSoTimeout(2000);
			socket.send(datagram);
			try {
				socket.receive(receive);
			} catch (SocketTimeoutException e) {
				Assert.fail("timeout instead of RST");
			}
		}
	}

with that test, and a temporary fix in 

	DataParser

	public final Message parseMessage(final RawData raw) {

		DatagramReader reader = new DatagramReader(raw.getBytes());
		MessageHeader header = parseHeader(reader);
		try {
			if (CoAP.isRequest(header.getCode())) {
				return parseMessage(reader, header, new Request(CoAP.Code.valueOf(header.getCode())));
			} else if (CoAP.isResponse(header.getCode())) {
				return parseMessage(reader, header, new Response(CoAP.ResponseCode.valueOf(header.getCode())));
			} else if (CoAP.isEmptyMessage(header.getCode())) {
				return parseMessage(reader, header, new EmptyMessage(header.getType()));
			}
		} catch (MessageFormatException e) {
			throw new MessageFormatException(e.getMessage(), header.getMID(), header.getCode(), CoAP.Type.CON == header.getType());
		}
		throw new MessageFormatException("illegal message code", header.getMID(), header.getCode(), CoAP.Type.CON == header.getType());
	}

the debugger shows the damage:

> reject(final RawData raw, final MessageFormatException cause)

creates a EmpytMessage only with destination and without source.

> private void reject(final Message message)

then creates in `EmptyMessage.newRST(message)` a RST for that 1. Level RST using 

	public static EmptyMessage newRST(Message message) {
		EmptyMessage rst = new EmptyMessage(Type.RST);
		rst.setDestination(message.getSource());
		rst.setDestinationPort(message.getSourcePort());
		rst.setMID(message.getMID());
		return rst;
	}

the empty source address as destination.

Result:
> Exception in network stage thread [UDP-Sender-localhost/127.0.0.1:0[0]]:
> java.net.BindException: Cannot assign requested address: Datagram send failed

(destination port 0)

joemag1: I don't think doing this in the Matcher will work. This particular code path bypasses Stack because it tries to manually send a RST for a badly formed message.

One option that I like a little more now, is to just send RST through the stack. I think that should have no side effect, because non of the layers do anything stateful with RST message (and in fact, may have positive side effect by clearing any state in the matcher). 

And that will solve the problem of infinite ping pong, because it TcpStack will correctly ignore RST message. Thoughts?

boaks: > I don't think doing this in the Matcher will work. 

> From my understanding:
> though the PING/PONG depends on the protocol flavor (TCP/UDP), we should try to move it to the Matcher and Stack and handle it there flavor aware.

I ment this very general: if there is something depending on the "flavor" (currently tcp/udp), then we should try to move it in the already flavor aware components (Matcher and Stack). Which one (or both), is then left to the implementor. 
I have also no problems with "if (TCP/UDP)" somewhere else to fix a bug, if we then but a reminder issue to solve it later without "if".

> One option that I like a little more now, is to just send RST through the stack. I think that should have no side effect, because non of the layers do anything stateful with RST message (and in fact, may have positive side effect by clearing any state in the matcher).
> 
> And that will solve the problem of infinite ping pong, because it TcpStack will correctly ignore RST message. Thoughts?
> 

I like this idea.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
boaks: OK.

But considering my "deep analysis" below, callng reject from reject will cause a message sent to port 0.

And there was a "idea",that in one case the RST could b canceled, in the other not.
(see "rst.isCanceled()"). To be frank, I don't know the background of that difference and may be we don't need it. But then I would prefer to have a explizit statementof changing this.
Otherwise, someone would habe it hard to see, if this was a intended or accidently change.

joemag1:
Abandoning this in favor of: https://github.com/eclipse/californium/pull/185

PR #182, opened 2016-12-15
Bugfix for MessageFormatException.

boaks:
Introduce new CoAPMessageFormatException to distinguish between general
errors and errors related to CoAP message.
Add tests.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataParser.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataParser.java
@@ -18,6 +18,8 @@
  * Kai Hudalla - logging
  * Bosch Software Innovations GmbH - turn into utility class with static methods only
  * Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
+ * Achim Kraus (Bosch Software Innovations GmbH) - add CoAP detail information 
+ *                                                 to MessageFormatException
  ******************************************************************************/
 package org.eclipse.californium.core.network.serialization;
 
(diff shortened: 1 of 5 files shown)


boaks:
This changes the processing for "missformed" messages in a way, that potentially more RST messages are sent. But I don't see the "clear intention", when RST should be sent or the message should just be ignored.

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataParser.java
```diff
@@ -53,17 +55,22 @@ public final Message parseMessage(final RawData raw) {
...
-			throw new MessageFormatException("illegal message code", header.getMID(), header.getCode(), CoAP.Type.CON == header.getType());
+		try {
+			if (CoAP.isRequest(header.getCode())) {
+				return parseMessage(reader, header, new Request(CoAP.Code.valueOf(header.getCode())));
+			} else if (CoAP.isResponse(header.getCode())) {
+				return parseMessage(reader, header, new Response(CoAP.ResponseCode.valueOf(header.getCode())));
+			} else if (CoAP.isEmptyMessage(header.getCode())) {
+				return parseMessage(reader, header, new EmptyMessage(header.getType()));
+			}
+		} catch (MessageFormatException e) {
```
sophokles73: Why do we need this try/catch block?

boaks: MessageFormatException are throw by util methods, which don't have the CoAP message context (MID, Code, CON/NON, e.g. CoAP.Code.valueOf). Trying to test the changes in "reject(final RawData raw, final MessageFormatException cause) " #179, I recognized that.
So catching and filling the context would allow to process such messages as CoAP message in `receiveMessage(final RawData raw)`. And as mentioned above, this increase the number of RST instead of ignored messages.

sophokles73:
I have a hard time understanding what this is good for.
Maybe @boaks can state the problem that this PR tries to address so that we can come to a conclusion whether we can close the PR or need to further pursue it.

boaks:
> 	public final Message parseMessage(final byte[] msg) {
> 
> 		DatagramReader reader = new DatagramReader(msg);
> 		MessageHeader header = parseHeader(reader); 
>             // after this line, it should be CoAP message with MID and a type => so RST may be sent 
> 		if (CoAP.isRequest(header.getCode())) {
>                      // if CoAP.Code.valueOf throws a general MessageFormatException,
>                      // it doesn't contain the already detected MID and type,
>                      //  => and so the message is ignored and no RST is sent!
> 			return parseMessage(reader, header, new Request(CoAP.Code.valueOf(header.getCode())));
> 		} else if (CoAP.isResponse(header.getCode())) {
> 			return parseMessage(reader, header, new Response(CoAP.ResponseCode.valueOf(header.getCode())));
> 		} else if (CoAP.isEmptyMessage(header.getCode())) {
> 			return parseMessage(reader, header, new EmptyMessage(header.getType()));
> 		} else {
>                      // so here we use the contructor, which may then trigger a RST 
> 			throw new MessageFormatException("illegal message code", header.getMID(), header.getCode(), CoAP.Type.CON == header.getType());
> 		}
> 	}
> 

Currently the behaviour of sending a RST for broken CoAP message is inhomogeneous. Even if MID and type is already detected, some "inner methods" don't have access to that information and therefore use the "general" MessageFormatException, which is ignored and not used to trigger a RST.
This PR takes those "general" MessageFormatException from "inner methods" and adds the already detected values of MID and type to potentially trigger a RST.

boaks:

sophokles73:
+1

PR #184, opened 2016-12-22
Bugfix use CloseOnErrorHandler.

boaks:
Use the CloseOnErrorHandler instead of the second CloseOnIdleHandler.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/tcp/TcpServerConnector.java
@@ -13,6 +13,7 @@
  * Contributors:
  * Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
  * Achim Kraus (Bosch Software Innovations GmbH) - adjust port when bound.
+ * Achim Kraus (Bosch Software Innovations GmbH) - use CloseOnErrorHandler.
  ******************************************************************************/
 package org.eclipse.californium.elements.tcp;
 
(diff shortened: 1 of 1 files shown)


sophokles73:
Is there any reason for this change? What problem is addressed?

boaks:
> Jan 09, 2017 10:39:47 AM io.netty.channel.DefaultChannelPipeline onUnhandledInboundException
> WARNUNG: An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
> 

The above message was reported frequently. so I investigated and found:

```
		protected void initChannel(SocketChannel ch) throws Exception {
			onNewChannelCreated(ch);

			// Handler order:
			// 0. Register/unregister new channel: all messages can only be sent over open connections.
			// 1. Generate Idle events
			// 2. Close idle channels.
			// 3. Stream-to-message decoder
			// 4. Hand-off decoded messages to CoAP stack
			// 5. Close connections on errors.
			ch.pipeline().addLast(new ChannelTracker());
			ch.pipeline().addLast(new IdleStateHandler(0, 0, connectionIdleTimeoutSeconds));
			ch.pipeline().addLast(new CloseOnIdleHandler());
			ch.pipeline().addLast(new DatagramFramer());
			ch.pipeline().addLast(new DispatchHandler(rawDataChannel));
			ch.pipeline().addLast(new CloseOnIdleHandler());
		}
```

So, the comment "5." (and the TcpClientConnector implementation) indicates for me, that the second addition of `new CloseOnIdleHandler()` is a typo and should be the `CloseOnErrorHandler`.

If there are still doubts, maybe the author can comment on this issue.

sophokles73:
Oh, I see. I totally oversaw the already existing addition of the idle handler...

PR #185, opened 2016-12-23
Fix for #178 (infinite ping-pong of empty messages)

joemag1:
where an empty message would get deserialized as CON (because there's no TYPE on the wire in TCP), and then make it to endpoint where it will get rejected, resulting in another empty message going back to the friend. To avoid this ping pong, send RST through the stack rather than having endpoint send it directly through the collector. All layers in the stack will ignore sendEmpty message, so from there it will make its way to connector, just like it does today. The exception is TCP stack, where TcpAdaptation layer will drop RST message.

Since we do not have Exchange, using null exchange to send empty message. Could also create a "fake" one instead. Open to opinion on this.

Signed-off-by: Joe Magerramov <joe.magerramov@gmail.com>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -584,8 +584,9 @@ public class CoapEndpoint implements Endpoint {
 			}
 
 			// MessageInterceptor might have canceled
-			if (!message.isCanceled())
+			if (!message.isCanceled()) {
 				connector.send(serializer.serializeEmptyMessage(message));
+			}
 		}
 
 		private void assertMessageHasDestinationAddress(final Message message) {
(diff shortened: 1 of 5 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -669,36 +673,18 @@ private void receiveMessage(final RawData raw) {
...
 
 		private void reject(final RawData raw, final MessageFormatException cause) {
-
 			// Generate RST, and send it if not cancelled by one of the interceptors
 			EmptyMessage rst = new EmptyMessage(Type.RST);
 			rst.setMID(cause.getMid());
-			rst.setToken(new byte[0]);
 			rst.setDestination(raw.getAddress());
 			rst.setDestinationPort(raw.getPort());
-			for (MessageInterceptor interceptor : interceptors) {
```
joemag1: Removed interceptor, because they are already applied in Outbox.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -669,36 +673,18 @@ private void receiveMessage(final RawData raw) {
...
-				"rejected malformed message from [{0}], reason: {1}",
-				new Object[]{raw.getInetSocketAddress(), cause.getMessage()});
+
+			coapstack.sendEmptyMessage(null, rst);
 		}
 
 		private void reject(final Message message) {
 			EmptyMessage rst = EmptyMessage.newRST(message);
-			// sending directly through connector, not stack, thus set token
-			rst.setToken(new byte[0]);
```
joemag1: Removed token because it's already added in Matcher.

PR #192, opened 2017-01-13
Add "cloneObservation" to Observation.

boaks:
Reduce code duplications and complexity for library user.

Removed serialization/deserialization.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
@@ -20,11 +20,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.eclipse.californium.core.Utils;
-import org.eclipse.californium.core.coap.Request;
-import org.eclipse.californium.core.network.serialization.DataParser;
-import org.eclipse.californium.core.network.serialization.DataSerializer;
-import org.eclipse.californium.core.network.serialization.UdpDataParser;
-import org.eclipse.californium.core.network.serialization.UdpDataSerializer;
 import org.eclipse.californium.elements.CorrelationContext;
 
 /**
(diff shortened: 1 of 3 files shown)


californium-core/src/main/java/org/eclipse/californium/core/observe/Observation.java
```diff
@@ -52,6 +52,20 @@ public Request getRequest() {
 	}
 
 	/**
+	 * @return a cloned observation with a shallow clone of request
+	 */
+	public Observation cloneObservation() {
```
sophokles73: why not implement `Cloneable` and overwrite `Object.clone()`?

boaks: Because of the **very specfic** "clone" of the Observation. A lot of information is not deep cloned and it was just done to get rid of the serialization/deserialization.
Mainly the InMemoryRegistrationStore in the leshan branch "improve_builders" remind me, that the current implementation seems to be fragile.

sophokles73: Ok, thinking about it I also came to the conclusion that it is probably not a good idea to make `Observation` cloneable in general since in a real implementation we would actually create the new instance by means of unmarshalling from persistent state.

sophokles73: However, for the same reason I do not think that it is a good idea to move the *cloning* code into `Observation`. The fact that we need to clone the `Observation` object by doing this *shallow* copying is actually a property of the `InMemoryObservationStore`. A DB based store would not need to do this because when reading from the DB a new object would be instantiated ...

boaks: May be we can move it to a util class, because the "InMemoryObservationStore" is not the only one using that shallow clone?

sophokles73: Yes, I thought so too. Maybe something like `ObservationSerializer`?

boaks: "..ShallowCloner" ?

boaks: Or do you want to stick with the serialization/deserialization?

boaks:
Changed to add ObservationUtil.shallowClone.
A unit test is also added.

PR #193, opened 2017-01-13
Add TCP correlation context.

boaks:
Used to distinguish different connections between the same endpoints.
The "localhost/127.0.0.1" mainly used for logging in the TCP client is
replaced by "any/0.0.0.0" because it reaches also targets on other
hosts, what would not be possible with "localhost".
The ConnectorTestUtil introduces common functions for TCP test. The
already available tests may be adpated to use this in a follow up PR.
This PR is the first part for correlation contexts for TCP based
connections. The larger part would be TLS and will follow, when this
get's merged.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
@@ -17,6 +17,7 @@
  *    Bosch Software Innovations GmbH - add support for correlation context to provide
  *                                      additional information to application layer for
  *                                      matching messages (fix GitHub issue #1)
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add onContextEstablished.
  ******************************************************************************/
 package org.eclipse.californium.elements;
 
@@ -293,8 +294,7 @@ public final class RawData {
 	/**
 	 * Gets the identity of the sender of the message.
 	 * 
-	 * This property is only meaningful for messages
-	 * received from a client.
+	 * This property is only meaningful for messages received from a client.
 	 * 
 	 * @return the identity or <code>null</code> if the
 	 *      sender has not been authenticated
@@ -304,7 +304,7 @@ public final class RawData {
 	}
 
 	/**
-	 * Gets additional information regarding the context this message has been 
+	 * Gets additional information regarding the context this message has been
 	 * received in.
 	 * 
 	 * @return the messageContext the correlation information or <code>null</code> if
@@ -330,4 +330,16 @@ public final class RawData {
 		return (correlationContext != null &&
 				correlationContext.get(DtlsCorrelationContext.KEY_SESSION_ID) != null);
 	}
+
+	/**
+	 * Callback, when context gets available. Used on sending an message.
+	 * 
+	 * @param context established context to be forwarded to the callback.
+	 */
+	public void onContextEstablished(CorrelationContext context) {
+		if (null != callback) {
+			callback.onContextEstablished(context);
+		}
+	}
+
 }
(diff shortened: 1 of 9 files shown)


element-connector/src/test/java/org/eclipse/californium/elements/tcp/TcpCorrelationTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+	@After
+	public void cleanup() {
+		for (Connector connector : cleanup) {
+			connector.stop();
+		}
+	}
+
+	@Test
+	public void correlationContext() throws Exception {
```
sophokles73: Can you please add a JavaDoc comment indicating what this test case is supposed to verify?
This makes it much easier to a) understand and b) review the code :-)

Same holds true for the other test cases as well...

element-connector/src/test/java/org/eclipse/californium/elements/tcp/TcpCorrelationTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+		cleanup.add(server);
+		cleanup.add(client);
+
+		Catcher serverCatcher = new Catcher();
+		Catcher clientCatcher = new Catcher();
+		server.setRawDataReceiver(serverCatcher);
+		client.setRawDataReceiver(clientCatcher);
+		server.start();
+		client.start();
```
sophokles73: It looks like the code for setting up the client and server is needed several times. Can you re-factor it into a separate method?

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
```diff
@@ -330,4 +330,16 @@ public boolean isSecure() {
 		return (correlationContext != null &&
 				correlationContext.get(DtlsCorrelationContext.KEY_SESSION_ID) != null);
 	}
+
+	/**
+	 * Callback, when context gets available. Used on sending an message.
+	 * 
+	 * @param context established context to be forwarded to the callback.
+	 */
+	public void onContextEstablished(CorrelationContext context) {
```
sophokles73: My understanding is that this method should be used instead of `getMessageCallback().onContextEstablished()`. I guess `DTLSConnector` uses the latter and should probably be changed to instead call this new method, shouldn't it?

element-connector/src/test/java/org/eclipse/californium/elements/tcp/TcpCorrelationTest.java
boaks: Currently this initializes "local variables" and soemtime it uses different parameters. Therefore I didn't move it to a seperate function to keep the test simple.

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
boaks: Yes, it could be changed now (or later, if it get's merged).

element-connector/src/test/java/org/eclipse/californium/elements/tcp/TcpCorrelationTest.java
```diff
@@ -0,0 +1,323 @@
...
+			connector.stop();
+		}
+	}
+
+	/**
+	 * Test, if the correlation context is determined proper.
+	 * Send a request and check, if the response has the same correlation context on the client side.
+	 * Send a second request and check, if this has the same correlation context on the client side.
+	 * Also check, if the server response is sent with the same context as the request was received.
+	 */
```
boaks: Even it seems to be more a set of tests, I think, it's covering a normal use case "request-response-request" and therefore I put it into one.

PR #196, opened 2017-01-23
basic correlation strategy for matching responses and request

boaks:
Replaces the "isResponseRelatedToRequest" in the Matchers with
"CorrelationStrategy.match()". Intended to be extended with a followup PR
for matching messages on sending to fix issue #104.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -26,6 +26,8 @@
  *    Bosch Software Innovations GmbH - adapt message parsing error handling
  *    Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
  *    Bosch Software Innovations GmbH - adjust request scheme for TCP
+ *    Achim Kraus (Bosch Software Innovations GmbH) - introduce CorrelationContextMatcher
+ *                                                    (fix GitHub issue #104)
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -69,6 +71,7 @@ import org.eclipse.californium.core.observe.ObservationStore;
 import org.eclipse.californium.core.server.MessageDeliverer;
 import org.eclipse.californium.elements.Connector;
 import org.eclipse.californium.elements.CorrelationContext;
+import org.eclipse.californium.elements.CorrelationContextMatcher;
 import org.eclipse.californium.elements.MessageCallback;
 import org.eclipse.californium.elements.RawData;
 import org.eclipse.californium.elements.RawDataChannel;
@@ -235,7 +238,7 @@ public class CoapEndpoint implements Endpoint {
 	 * @param config The configuration values to use.
 	 */
 	public CoapEndpoint(final InetSocketAddress address, final NetworkConfig config) {
-		this(createUDPConnector(address, config), config, null, null);
+		this(createUDPConnector(address, config), config, null, null, null);
 	}
 
 	/**
@@ -246,7 +249,7 @@ public class CoapEndpoint implements Endpoint {
 	 * @param exchangeStore The store to use for keeping track of message exchanges.
 	 */
 	public CoapEndpoint(final InetSocketAddress address, final NetworkConfig config, final MessageExchangeStore exchangeStore) {
-		this(createUDPConnector(address, config), config, null, exchangeStore);
+		this(createUDPConnector(address, config), config, null, exchangeStore, null);
 	}
 
 	/**
@@ -259,7 +262,7 @@ public class CoapEndpoint implements Endpoint {
 	 * @param config The configuration values to use.
 	 */
 	public CoapEndpoint(final Connector connector, final NetworkConfig config) {
-		this(connector, config, null, null);
+		this(connector, config, null, null, null);
 	}
 
 	/**
@@ -271,7 +274,7 @@ public class CoapEndpoint implements Endpoint {
 	 *              endpoint.
 	 */
 	public CoapEndpoint(final InetSocketAddress address, final NetworkConfig config, final ObservationStore store) {
-		this(createUDPConnector(address, config), config, store, null);
+		this(createUDPConnector(address, config), config, store, null, null);
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


element-connector/src/main/java/org/eclipse/californium/elements/CorrelationStrategy.java
```diff
@@ -0,0 +1,54 @@
...
+	 * Check, if data should be send.
+	 * 
+	 * @param dataContext correlation context of data to send. May be null, if
+	 *            no special correlation context is required for this data.
+	 * @param connectorContext current correlation context of connector. May be
+	 *            null, if not available or connection is not established.
+	 * @return true, if data should be sent by the connector, false, if data
+	 *         should be suppressed.
+	 */
+	boolean send(CorrelationContext dataContext, CorrelationContext connectorContext);
```
sbernard31: Do we really need a `send` method ? match is not enought ?

boaks: "send" is for the followup PR.
It's intended to check the correlation context just before "sending/encrypting" the message.
Currently both implementations are doing the "same checks", but I would like to allow different checkes for matching (mainly coap client side) and sending (mainly coap server side)

boaks:
> Why not removing the factory part and just pass the CorrelationStrategy in parameters ?

Because currently, it's "at least unclear" how someone is intended to create a "full working" Connector/CoAPEndpoint outside :-). 
See http://dev.eclipse.org/mhonarc/lists/leshan-dev/msg00681.html

But, yes, if it get's clear, how to create a working Connector/CoAPEndpoint, we can pass in the strategy and we could even make the field final.

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationStrategy.java
sophokles73: The `send` method is not used currently, so I suggest we *postpone* it to the subsequent PR.
I also would like to suggest some naming changes:
1. Rename `CorrelationStrategy` to `CorrelationContextMatcher` to better indicate what its purpose is.
2. IMHO `match` is overly generic and we can simply keep the name of the former `UdpMatcher.isResponseRelatedToRequest` method. It quite precisely describes the purpose of the method, doesn't it?
3. Once the `send` method is added in a later PR we also should come up with a more descriptive name for that as well.

californium-core/src/main/java/org/eclipse/californium/core/network/CorrelationStrategyFactory.java
```diff
@@ -0,0 +1,84 @@
...
+
+import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.elements.CorrelationStrategy;
+import org.eclipse.californium.elements.RelaxedCorrelationStrategy;
+import org.eclipse.californium.elements.StrictlyCorrelationStrategy;
+
+/**
+ * Factory for correlation strategy.
+ */
+public abstract class CorrelationStrategyFactory {
```
sophokles73: I am not a big fan of using such *factory* classes instead of passing in an instance of the interface to be created. However, I see your point, @boaks, regarding the somewhat *inflexible* process of creating and configuring a `CoapEndpoint` that can be used by leshan (this is addressed by the `improve_builder` branch already which provides means to set `Endpoint`s explicitly).
So, AFAIC it would be preferrable to allow for client code to pass in an instance of `CorrelationStrategy` to the `CoapEndpoint` constructor and, if the instance is `null` use this (private) factory for creating an instance based on the configuration property. However, I would rather not expose this factory as public API to be extended and used by client code.

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationStrategy.java
boaks: The intention is to decide, it the response should be processed for the request or the message should be sent based on the correlation context. Currently the implementations does this by "matching" the correlation context. But in general, I think "strategy" express more the intention of "decide" and not the current implementation of "match".
The renaming of method "match" to "isResponseRelatedToRequest" end remove of "send" will be contained in the next commit.

boaks:
Aside from the name "strategy" or "matcher" the other comment are taken into account.
So, please just give me feedback, if we shoudl take Kai's proposal and using a "Matcher" (emotion thumb down) name or if we should stay with "Strategy" (emotion thumb up).

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationStrategy.java
sbernard31: I rather prefer `match` because I suspect we will not only match Response with Request. But I will not fight for this.

sbernard31:
I voted for `matcher` but It's maybe too early to make polish on naming.

boaks:
@sbernard31 
Just to clarify, because there was also a comment from you on the method name "match".
There is a interface name "CorrelationStrategy" (current name) and Kai's proposal "CorrelationContextMatcher" (I think, it was that name). So the vote should be for the interface name.

The method name "match" I already replaced by the "isResponse...", 
because I think adding a second method for "send" makes it more flexible.

sbernard31:
I well understood. For now I see that class as a `CorrelationContextMatcher` and so it must `match` 2 `CorrelationContext`. But this could change as we have not  finished exploring this feature, that's why I'm not so attached to naming for now.

boaks:
So, three persons with three ideas :-).
I would like to continue to prepare the PR for the sending, so should we wait for some more feedback or which naming solution should be taken?

1. CorrelationStrategy.isResponseRelatedToRequest()
2. CorrelationContextMatcher.isResponseRelatedToRequest()
3. CorrelationContextMatcher.match()

sophokles73:
Since @boaks has already renamed the `match` method and we have a 2:1 preference for `CorrelationContextMatcher` we should be able to quickly proceed with this PR ...

boaks:
OK, so go for 2.

boaks:
Renaming is done.

element-connector/src/main/java/org/eclipse/californium/elements/StrictlyCorrelationContextMatcher.java
```diff
@@ -0,0 +1,74 @@
...
+package org.eclipse.californium.elements;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Strict correlation context matcher. Uses strictly matching for DTLS including
+ * the security epoch.
+ */
+public class StrictlyCorrelationContextMatcher implements CorrelationContextMatcher {
```
sophokles73: Should be `StrictCorrelationContextMatcher`

boaks:
done.

PR #200, opened 2017-01-24
Add TLS correlation context.

boaks:
Add TLS session information to correlation context.
Determine principal, if available. Allow to provide
client authentication mode for TLS server connector.
Replace use of cert.jks from element-connector with 
demo-certs.


boaks:
Still too big ...

Issue #156, opened 2016-11-25
RawData.isSecure() depends on DTLS


boaks:
RawData

	public boolean isSecure() {
		return (correlationContext != null &&
				correlationContext.get(DtlsCorrelationContext.KEY_SESSION_ID) != null);
	}

Remove DTLS dependeny by introducing a "isSecure()" method into CorrelationContext?

boaks:
If wanted, into PR #148 or new PR?

sophokles73:
There is no dependency on DTLS. There is just a dependency on a costant declared in class `DtlsCorrelationContext`. Since both classes are in the same package I do not see a problem here ...

boaks:
Sure, there is no dependency to scandium, but to DTLS!
If TLS (I hope its also secure :-) uses TlsCorrelationContext, this test is only true, because TlsCC uses the same key. But I would prefer to have a method ...

sophokles73:
The idea is to use the keys defined by `DtlsCorrelationContext`. Otherwise, a `Connector` may implement whatever type of `CorrelationContext` it wants. So, `TlsCorrelationContext` (if you want to create it), needs to use this key *by definition* instead of *by coincidence*.

sbernard31:
I continue [the discussion](https://github.com/eclipse/californium/pull/411#discussion_r138919812) here, because it's more appropriate. 

Looking more deeper to the code. It seems that `isSecure` is just used to set request scheme for incoming request on a specific endpoint.

Currently :
The code allow connector supporting several schemes. (coap, coaps, coap-tcp, ...)
CoapEndpoint allow to support only 2 schemes (secure and unsecure).
RawData.isSecure() depends on DTLS.

This seems not so clean to me.
1) I don't think that connector should express coap scheme, it is too coap dependent.
=> we should replace it by supported protocol  (UDP, DTLS,TCP,  TLS, ...). CoapEndpoint should deduce the supported scheme (coaps/coap/coap-tcp ...) from supported protocol.
2) I don't know if we want to keep connector/endpoint supporting several scheme/protocol. Currently our connector/endpoint support only one and I don't know how this could be different.
=> If we chose that connector/endpoint support only one protocol then we can remove the isSecure on RawData.
=> if we still want to support several protocol for connector/endpoint, I think this is up to the connector to say if this is a secure rawData or not, adding a boolean to `RawData.inbound`

boaks:
FMPOV: 
[...]

Issue #145, opened 2016-11-18
Request.userContext


boaks:
I'm pretty unsure, for what this is used. Currently leshan puts some data in.
Is intended to be related to the CorrelationContext? Or was it intended to be used unrelated to that context?

sophokles73:
I think this is used in the context of failing over observations between nodes.
`ObservationStore.add()` also persists the data from `Request.userContext`.

sbernard31:
UserContext is used to attach application data to a request. (a feature provided by californium for enduser) (see #86)
E.g we use it in Leshan to attach the registration id linked to an observe requests.

CorrelationContext contains data used to make correlation between CoAP messages. (needed for internal californium behavior)

HTH

boaks:
Thanks!

Issue #142, opened 2016-11-17
Matching of plain request and DTLS response


boaks:
private boolean isResponseRelatedToRequest(final Exchange exchange, final CorrelationContext responseContext) {
		if (exchange.getCorrelationContext() == null) {
			// no correlation information available for request, thus any
			// additional correlation information available in the response is ignored
			return true;
		}

Seems, that, if the request is sent in plain, its not checked, that the response is also plain.
May be this is not really important, but its also no big issue. 
So, does anbody depend on that behavior?

sophokles73:
The idea here was that the transport layer may indeed attach some context information in the response (even if the request was sent via UDP). However, in this case, we simply do not care.
Do you see any security issues here?

boaks:
My understanding was:
If a channel (connector) supports such a CorrelationContext, it should be used to ensure the request and response are mapped proper. I have already started an idea of introducing a "context matching strategy" to provide more flexibility her, but though there are too many changes required, I'm very delayed with that approach. 

> The idea here was that the transport layer may indeed attach some context information in the response (even if the request was sent via UDP). 

Any example?

>Do you see any security issues here?

Depending on the "token" definition (unique per directed connection/unique per peer for outgoing), there may be missinterpretations. So someone using DTLS relays on its secure mapping but would wrongely be delivered to a unsecure request.

sophokles73:
>My understanding was:
If a channel (connector) supports such a CorrelationContext, it should be used to ensure the request and response are mapped proper.

I agree if we replace `should` with `may` :-)

>Any example?

The transport layer might include information about the path MTU determined for the peer or whatever. The point is: not all information contained in the context is necessarily security related.

boaks:
Hm, so should we check, if the request has no context, if the response context is "secure"?

sophokles73:
If the request has no context then we only need to check whether the source of the response is the same as the destination of the request, right?

PR #202, opened 2017-01-25
Add correlation context matching when sending messages.

boaks:
TCP implementation is still dummy and will follow later.


boaks:
> Next step is to add a way to give a CorrelationContext when we use the CoapEndoint.send...() API ?

Your right. to make the context of the PR small, I just concentrated to the responses.
There was also sometime ago a "definition", that the correlation context is somehow internal to californium. Therefore I jsut want to show, how this could be used to ensure, that messages are sent to the right receiver.

sbernard31:
I agree with you  about the PR size. The scope you chose is fine for me :)
I thought about what could be the next step, but you're right,  let's focus on this PR for now.

boaks:
> Should we change the javadoc for the method RawData.getCorrelationContext ?

Right, I will do that :-).

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContextUtil.java
```diff
@@ -0,0 +1,60 @@
...
+			String value1 = context1.get(key);
+			String value2 = context2.get(key);
+			boolean match = (value1 == value2) || (null != value1 && value1.equals(value2));
+			if (!match && !warn) {
+				/* no warnings => fast return */
+				return false;
+			}
+			if (!match || info) {
+				/* logging differences with warnings and matches with level finest */
+				LOGGER.log(match ? Level.FINEST : Level.WARNING, "{0}, {1}: \"{2}\" != \"{3}\"", new Object[] { name, key,
```
boaks: Upps :-).

element-connector/src/main/java/org/eclipse/californium/elements/RelaxedCorrelationContextMatcher.java
```diff
@@ -36,29 +36,19 @@ public boolean isResponseRelatedToRequest(CorrelationContext requestContext, Cor
...
-					"(D)TLS session {0}, {1}",
-					new Object[] { requestedContext.get(DtlsCorrelationContext.KEY_SESSION_ID),
-							availableContext.get(DtlsCorrelationContext.KEY_SESSION_ID) });
-			LOGGER.log(
-					match ? Level.FINEST : Level.WARNING,
-					"(D)TLS cipher {0}, {1}",
-					new Object[] { requestedContext.get(DtlsCorrelationContext.KEY_CIPHER),
-							availableContext.get(DtlsCorrelationContext.KEY_CIPHER) });
-			return match;
+			return CorrelationContextUtil.match(getName(), KEYS, requestedContext, availableContext);
```
boaks: Good question. It's copied from the old UdpMatcher.

	private boolean isResponseRelatedToRequest(final Exchange exchange, final CorrelationContext responseContext) {
		if (exchange.getCorrelationContext() == null) {
			// no correlation information available for request, thus any
			// additional correlation information available in the response is ignored
			return true;
		} else if (exchange.getCorrelationContext().get(DtlsCorrelationContext.KEY_SESSION_ID) != null) {
			// original request has been sent via a DTLS protected transport
			// check if the response has been received in the same DTLS session
			if (useStrictResponseMatching) {
				return isResponseStrictlyRelatedToDtlsRequest(exchange.getCorrelationContext(), responseContext);
			} else {
				return isResponseRelatedToDtlsRequest(exchange.getCorrelationContext(), responseContext);
			}
		} else {
			// compare message context used for sending original request to context
			// the response has been received in
			return exchange.getCorrelationContext().equals(responseContext);
		}

There are some ideas an the correlation context (see https://github.com/eclipse/californium/issues/142), and, though I don't fully understand the usage of the correlation context, I didn't want to change too much.

element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java
```diff
@@ -281,13 +302,23 @@ private Sender(String name) {
...
+			/* check, if message should be sent with the "none correlation context" of UDP connector */
+			CorrelationContextMatcher correlationMatcher = getCorrelationContextMatcher();
+			if (null != correlationMatcher && !correlationMatcher.isToBeSent(raw.getCorrelationContext(), null)) {
+				if (LOGGER.isLoggable(Level.WARNING)) {
+					LOGGER.log(Level.WARNING, "UDPConnector ({0}) drops {1} bytes to {2}:{3}",
+							new Object[] { socket.getLocalSocketAddress(), datagram.getLength(), datagram.getAddress(),
+									datagram.getPort() });
+				}
+				return;
+			}
```
boaks: We drop it.
And yes, introducing something as "Message.isDropped()" in the future would be helpfull. At least for unit tests :-).
But for this, it would easyly exceed the number of line.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1176,6 +1190,9 @@ private void sendMessage(final RawData message) throws HandshakeException {
 
 		DTLSSession session = connection.getEstablishedSession();
 		if (session == null) {
+			if (!checkOutboundCorrelationContext(message, null)) {
```
boaks: Future Extension.
Without that, the response is sent but then dropped by the client (also without feedback).

boaks:
> Should we change the javadoc for the method RawData.getCorrelationContext ?

And the correction of the logging message in "CorrelationContextUtil.match" are done.

element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java
sbernard31: I agree, let's this for another PR.

element-connector/src/main/java/org/eclipse/californium/elements/RelaxedCorrelationContextMatcher.java
sbernard31: FMPOV, by default we should set a `strict` or `relaxed` CorrelationMatcher only for endpoint which use a  `DTLSConnector` and so remove this test about `KEY_SESSION_ID`  and `equals` statement.

boaks: I'm not sure. For me it seems to be more deterministic, if all connectors do their check. But in the end, I think the `CorrelationContext` (and its extendability) should be more specified. Currently having two point of extendability (`CorrelationContext` and `CorrelationContextMatcher`) may lead to issues.

boaks: I tried to sum-up the information about the correlation context and it's usage:

Information sum-up:

#142 
asymetric processing of requests and responses correlation context, potential extension of correlation context with information not intended to be used for "matching"

#145
UserContext and CorrelationContext, CorrelationContext is californium internal

#156 
Correlation Context is defined by keys (and their associated value) and a "secure" message uses the key "DtlsCorrelationContext.KEY_SESSION_ID".

Current usage:
Only DtlsCorrelationContext (and TcpCorrelationContext) is used. Therefore either no correlation context is provided, or a DtlsCorrelationContext. Therefore the "else ... equals" was not used.

So my conclusion:

Generally the first part of the matching:

```
        if (exchange.getCorrelationContext() == null) {
		// no correlation information available for request, thus any
		// additional correlation information available in the response is ignored
		return true;
	}
```
applies to "all" correlation contexts.

The second test, is currently base on DtlsCorrelationContext.KEY_SESSION_ID. I think, that matching should be defined by matching equal values for a set of keys. If null for both values of a key are also considered to be matching, the keys set may just be the super set over different correlation context types (e.g. TcpCorrelationContext.KEY_CONNECTION_ID could also be added. The result would be, that either the DTLS information is matched (provided by DtlsCorrelationContext) or the tcp information (provided by TcpCorrelationContext)). 

The third test 

```
        } else {
		// compare message context used for sending original request to context
		// the response has been received in
		return exchange.getCorrelationContext().equals(responseContext);
	}

```
should be removed

But I would call the "matcher" for all connectors, not only the "secure" ones.

boaks: If the above function should be implemented (half way already done in CorrelationContextUtil), the CorrelationContext may be extended by String[] getMatchingKeys().

The matcher may then reduce the key set by its requirements (e.g. Relaxed remove EPOCH).

sbernard31: > The third test should be removed

I agree about this.

> But I would call the "matcher" for all connectors, not only the "secure" ones.

I'm not sure to understand why. It seems to me that  `Relaxed` and `Strict` matcher are strongly linked to DTLS connector. So I don't see why we should use it for unsecure connector. For me this make sense to have a   `default` matcher for others connectors by default. But if you disagree, we can start with "matcher" for all connectors.

boaks: I guess the points are simply, that
- I would like to consider the TCP stuff (TCP_CONNECTION_ID) and, 
- when "send" would be extended to provide a correlation context, I would like to grant, that, if someone "outside" make a misstake and send data using a DTLS correlation context, but gets a plain connector, the message is dropped. 

I agree, that "californium internal", responses are sent via the connector the request comes in. and TCP is not in the focus. But I hope, the "send" extension will come soon :-).

boaks: I prepared a preview PR with the new "matching".

#208

boaks:
@sbernard31 

So the correlation  context  matcher is now related to the connector.
The "main" drawback seems for me, that the plain "coap:" correlation  context  matcher may not detect a DTLS correlation context.

element-connector/src/main/java/org/eclipse/californium/elements/StrictDtlsCorrelationContextMatcher.java
```diff
@@ -0,0 +1,31 @@
...
+ * Strict correlation context matcher. Uses strictly matching for DTLS including
+ * the security epoch.
+ */
+public class StrictDtlsCorrelationContextMatcher extends KeySetCorrelationContextMatcher {
+
+	private static final String KEYS[] = { DtlsCorrelationContext.KEY_SESSION_ID, DtlsCorrelationContext.KEY_EPOCH,
+			DtlsCorrelationContext.KEY_CIPHER };
+
+	public StrictDtlsCorrelationContextMatcher() {
+		super("relaxed correlation", KEYS);
```
sbernard31: bad copy/paste :)

californium-core/src/main/java/org/eclipse/californium/core/network/CorrelationContextMatcherFactory.java
```diff
@@ -13,30 +13,55 @@
...
+				 */
+			} else if (connector.isSchemeSupported(CoAP.COAP_TCP_URI_SCHEME)) {
+				/*
+				 * To be implemented in a future PR, in the meanwhile use
+				 * default dtls matcher as default for backwards compatibility
+				 */
+			}
+		}
+		return config.getBoolean(NetworkConfig.Keys.USE_STRICT_RESPONSE_MATCHING) ? new StrictDtlsCorrelationContextMatcher()
+				: new RelaxedDtlsCorrelationContextMatcher();
```
sbernard31: Why using DTLS correlation context matcher by default ?

element-connector/src/main/java/org/eclipse/californium/elements/StrictDtlsCorrelationContextMatcher.java
boaks: so true ;-(

californium-core/src/main/java/org/eclipse/californium/core/network/CorrelationContextMatcherFactory.java
boaks: For "backwards compatiblitiy". Mainly for some unit tests, which assume such dtls ... matcher. Reaching 8xx lines, I didn't wan't to change more with this PR.

boaks:
The typo should now be corrected.
If no objections, I would merge and go for extending the send for providing a correlation context.
I guess after that, we will, see if it works and fixes #104

sbernard31:
Good for me.

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
```diff
@@ -225,13 +227,15 @@ public static RawData inbound(byte[] data, InetSocketAddress address, Principal
...
 	 * @param data the data to send.
 	 * @param address the IP address and port the data is to be sent to.
+	 * @param correlationContext correlation context for sending data (may be <code>null</code>).
 	 * @param callback the handler to call when this message has been sent (may be <code>null</code>).
 	 * @param useMulticast indicates whether the data should be sent using a multicast message.
 	 * @return the raw data object containing the outbound message.
 	 * @throws NullPointerException if data or address is <code>null</code>.
 	 */
-	public static RawData outbound(byte[] data, InetSocketAddress address, MessageCallback callback, boolean useMulticast) {
+	public static RawData outbound(byte[] data, InetSocketAddress address, CorrelationContext correlationContext, MessageCallback callback, boolean useMulticast) {
```
sophokles73: The `callback` param is supposed to be used by the connector to set the `CorrelationContext` on an outbound request. How does that fit with providing the `CorrelationContext` in the `correlationContext` param?
Can you provide values for both params? If not, we should introduce another factory method accepting an existing correlation context instead of a callback, shouldn't we?

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
```diff
@@ -305,7 +309,7 @@ public Principal getSenderIdentity() {
 
 	/**
 	 * Gets additional information regarding the context this message has been
-	 * received in.
+	 * received in or should be sent out.
```
sophokles73: *received or should be sent in*

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContextMatcher.java
```diff
@@ -39,4 +42,15 @@
...
+	/**
+	 * Check, if message should be sent out using the current correlation
+	 * context of the connector.
+	 * 
+	 * @param messageContext correlation context of message
+	 * @param connectorContext correlation context of connector
+	 * @return true, if message should be sent, false, if message should not be
+	 *         sent.
+	 */
+	boolean isToBeSent(CorrelationContext messageContext, CorrelationContext connectorContext);
```
sophokles73: is there really a difference in matching inbound vs. outbound messages?
At least `KeySetCorrelationContextMatcher` doesn't seem to bother whether it's inbound or outbound ...

element-connector/src/main/java/org/eclipse/californium/elements/StrictDtlsCorrelationContextMatcher.java
```diff
@@ -0,0 +1,31 @@
...
+ *    Bosch Software Innovations GmbH - add flexible correlation context matching
+ *                                      (fix GitHub issue #104)
+ ******************************************************************************/
+package org.eclipse.californium.elements;
+
+/**
+ * Strict correlation context matcher. Uses strictly matching for DTLS including
+ * the security epoch.
+ */
+public class StrictDtlsCorrelationContextMatcher extends KeySetCorrelationContextMatcher {
```
sophokles73: Again, there doesn't seem to be a difference between matching context of inbound vs. outbound messages.
Do we need two methods in `CorrelationContextMatcher`?

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
boaks: The provided CorrelationContext may not contain all informations, which the connector can provide. E.g. someone may decide to put the pask indentity in and provide a matcher, which checks, if the dtls session actually used to encrypt the data matches that identity.
In that case, either the additional values must be added or a new correlation context must be handled back by the callback. I decided to use the already used one.

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
boaks: ok

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContextMatcher.java
boaks: Though it's extension point, the future will show, if it's used different.
One idea behind that, is the possiblity to offer a "backward compatible matcher", which only checks for received responses und ignores the correlation context on sending as it was until this PR.

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
sophokles73: Sticking to your example, I guess the PSK identity would come from a `CorrelationContext` established when the client initially authenticated and issued its first request. The server then sends a response to the client using the `CorrelationContext` from the ongoing `Exchange` the response is part of, right? I think this case could be handled by simply providing the `CorrelationContext` instance from the `Exchange` to `RawData.outbound()`, it should contain everything that's needed to make sure that the outbound response is indeed matching the request.
However, when the server later wants to send a request to the client itself, it needs to make sure that the request is sent to the client over the same DTLS session that was established when the client sent its initial request. Doesn't the server need to somehow *cache* or *associate* the then established `CorrelationContext` with the client identifier (e.g. the client's IP:port or whatever)?
Is this the responsibility of the application layer? How does an application provide the information to the stack which is required for matching the `CorrelationContext` to a client security association?

boaks: > The server then sends a response to the client using the CorrelationContext from the ongoing Exchange the response is part of, right?

Yes, that's done by this PR.

But let me short up the discussion:
I want to introduce an "extension point", means, I want to offer flexiblity outside of permanent patching californium. I sure, have not a "finished and very detailed idea" what exactly some will do with that. Therefore I would make too much assumptions and restrictions.

boaks:
@sophokles73 

After some comments back on yours, is it OK to have 
- two methods in the matcher
- in the outbound providing both correlation context and callback at the same time?

If so I woulde merge ...

sbernard31:
Ok for me.

sophokles73:
It is still unclear to me how the outbound matching is supposed to work/be used by application layer code.
However, we have been discussing this PR for ages now and I think we need to move on and see what we get out of it :-)
So, by all means, go ahead, Achim and merge it :-)

PR #203, opened 2017-01-26
Reduce error logging for security errors to comprehesive messages.

boaks:
Current security/encryption errors are logged with many "caused by
stacktraces". This reduces the message into a comprehensive format.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/tcp/CloseOnErrorHandler.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/tcp/CloseOnErrorHandler.java
@@ -12,28 +12,59 @@
  * <p>
  * Contributors:
  * Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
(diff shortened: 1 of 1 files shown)


boaks:
@joemag1 

Though this is only related to TLS/TCP, do you have any objections against reducing the logging of causes (of causes (of causes))?

sbernard31:
@boaks generally we used rebase instead of merge. The idea is to keep history more readable.

boaks:
@sbernard31 

Uups, sorry, I just pressed "merge pull request" without selecting the proper method "rebase and merge" ;-(. 

So, is it possible to correct this (I guess this would require "force push")? 
Or could we stay with the 2 merges (I did the same error with a other PR)?

sbernard31:
Personally, I'm ok for a force push. (it seems there is not so much works based on this)

boaks:
So, if no objections until monday, i would try to reset the 2.0 branch (hopefully it works).
And if the reset is possible, the I would reapply the PR with selecting rebase.

PR #204, opened 2017-01-26
Add client authentication mode to TLS server.

boaks:
Add new TlsServerConnector constructor with new enum parameter to define
the client authentication mode. Add unit tests.


boaks:
First check the proper usage of demo certs.

PR #205, opened 2017-01-27
Add more logging to analyse tests failing using demo-certs.

boaks:
My guess for the cause of the failure:
keystore contains client and server credentials. My test chose the server, maybe hudson takes the client.
I've implemented a SslContextUtil class to have a fine grande access and convertion possibilities. But though it's more then 500 lines, I tried to do it without.
  
### DIFF
--- a/element-connector/src/test/java/org/eclipse/californium/elements/tcp/TlsConnectorTest.java
+++ b/element-connector/src/test/java/org/eclipse/californium/elements/tcp/TlsConnectorTest.java
@@ -12,6 +12,7 @@
  * <p>
  * Contributors:
  * Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
+ * Achim Kraus (Bosch Software Innovations GmbH) - add more logging.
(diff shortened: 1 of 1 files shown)


sophokles73:
Is this a fix?

boaks:
No, it should help to gather the information to fix the issue.

sophokles73:
Then I suppose we should merge this PR once you found the reason for the problem and have amended this PR with the fix ...

boaks:
???

The error didn't occur on my system and the logs provided by hudson don't contain the required information to fix that issue. Is there a other possiblity to execute the test with more logging on hudson?

boaks:
Let me just add, that I think to provide more information in the logs is something which may also help in future analysis. Therefore I think, it's good to provide this also in the 2.0 branch. If it would just be temporary only to fix the current issue, your right, it should then not be a PR to 2.0.

sophokles73:
Sorry, I did not realize that the problem only occurs on the HIPP instance ... mea culpa.

PR #206, opened 2017-01-27
Bugfix: Change checkServerTrusted to relaxed DN test.

boaks:
Though keyStore.jks contains both, the server and the client keys, it seems to be platform dependend, which on is used. Therefore relax the check of the subject DN. PR #148 contains a SslContextUtil, which would allow to select the credentials and trusts. To use the demo-certs for more deterministic tests, I would prepare that Util in a separate PR. 


boaks:
> But I guess that this is another PR ...

A preview could be 

https://github.com/bsinno/californium/blob/9e5d3933966ddd1d3a03eadc7b26a6d50e0555f0/element-connector/src/main/java/org/eclipse/californium/elements/util/SslContextUtil.java

from PR #148 

I have laready started to isolate that into a seperate PR.

(There is also some "similar" functionallity in netty, but I don't think, we should get more dependent on netty.)

PR #208, opened 2017-02-01
Add alternative correlation matching

boaks:
For discus a alternative matching implementation in #202.
If agreed, I would squash this into #202

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
@@ -14,6 +14,7 @@
  *    Bosch Software Innovations GmbH - add support for correlation context to provide
  *                                      additional information to application layer for
  *                                      matching messages (fix GitHub issue #1)
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add getMatchingKeys
  ******************************************************************************/
 package org.eclipse.californium.elements;
 
@@ -27,6 +28,13 @@ import java.util.Set;
 public interface CorrelationContext {
 
 	/**
+	 * Gets the set of keys used to match a correlation contexts.
+	 * 
+	 * @return set of keys
+	 */
+	Set<String> getMatchingKeys();
+
+	/**
 	 * Gets a value from this context.
 	 * 
 	 * @param key the key to retrieve the value for.
@@ -36,7 +44,8 @@ public interface CorrelationContext {
 	String get(String key);
 
 	/**
-	 * Gets a Set of a Map.Entry which contains the key-value pair of the CorrelationContext.
+	 * Gets a Set of a Map.Entry which contains the key-value pair of the
+	 * CorrelationContext.
 	 *
 	 * @return A set of a map entry containing the key value pair.
 	 */
--- a/element-connector/src/main/java/org/eclipse/californium/elements/DtlsCorrelationContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/DtlsCorrelationContext.java
@@ -14,8 +14,11 @@
  *    Bosch Software Innovations GmbH - add support for correlation context to provide
  *                                      additional information to application layer for
  *                                      matching messages (fix GitHub issue #1)
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add getMatchingKeys
  ******************************************************************************/
 package org.eclipse.californium.elements;
+
+import java.util.Set;
 
 /**
  * A correlation context that explicitly supports DTLS specific properties.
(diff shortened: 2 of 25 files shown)


element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
```diff
@@ -27,6 +28,13 @@
 public interface CorrelationContext {
 
 	/**
+	 * Gets the set of keys used to match a correlation contexts.
+	 * 
+	 * @return set of keys
+	 */
+	Set<String> getMatchingKeys();
```
sbernard31: I would named it : `getKeys()` and I wouldn't talk about "matching" in the javadoc.

element-connector/src/main/java/org/eclipse/californium/elements/DtlsCorrelationContext.java
```diff
@@ -48,6 +53,14 @@ public DtlsCorrelationContext(String sessionId, String epoch, String cipher) {
 		}
 	}
 
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public Set<String> getMatchingKeys() {
+		return MATCHING_KEYS;
```
sbernard31: I would return all the Keys => `entries.keySet()`

element-connector/src/main/java/org/eclipse/californium/elements/MapBasedCorrelationContext.java
```diff
@@ -44,10 +50,18 @@ public final Object put(String key, String value) {
 	 * {@inheritDoc}
 	 */
 	@Override
+	public Set<String> getMatchingKeys() {
+		return EMPTY;
```
sbernard31: Why not : `entries.keySet()`

element-connector/src/main/java/org/eclipse/californium/elements/RelaxedCorrelationContextMatcher.java
```diff
@@ -47,10 +49,10 @@ private final boolean internalMatch(CorrelationContext requestedContext, Correla
 		} else if (null == availableContext) {
 			return false;
 		}
-		if (requestedContext.get(DtlsCorrelationContext.KEY_SESSION_ID) != null) {
-			return CorrelationContextUtil.match(getName(), KEYS, requestedContext, availableContext);
-		}
-		return requestedContext.equals(availableContext);
+		Set<String> keys = new CopyOnWriteArraySet<String>(requestedContext.getMatchingKeys());
+		keys.addAll(availableContext.getMatchingKeys());
```
sbernard31: I prefer to explicitly set the key used for matching. This is less error prone in case where somebody add new elements in the context.

element-connector/src/main/java/org/eclipse/californium/elements/DtlsCorrelationContext.java
boaks: The point seems for me, that, if new keys are added (e.g. MTU), they may be **not** considered for matching. And "explicit matching keys" may also detect missing entries in only one of the correlation contexts.

element-connector/src/main/java/org/eclipse/californium/elements/CorrelationContext.java
boaks: The idea was, that a correlation context has information, and a subset is used for matching.

boaks:
The proposal is based on my analysis:

https://github.com/eclipse/californium/pull/202#discussion_r98875008

Therefore I designed it according.
- Correlation Context may be extended or newly introdced.
- not all provided information is used for matching
- therefore, if a correlation context is created or extended, it should be determined there, what should be used for maching.

sbernard31:
As I explained above, I think this should be the `CorrelationContextMatcher` responsibility to define which keys must be used for matching.

My proposal respect : 
- _Correlation Context may be extended or newly introduced._
- _not all provided information is used for matching_

but not :
_if a correlation context is created or extended, it should be determined there, what should be used for maching._

Which part of your analysis make you think this is the right thing to do ?

boaks:
My intention was, that if some adds a key in a Correlation Context, there should be a hint, that a decision about the matching is required. A "matching key set" in the class would be that hint on my opinion. Otherwise a commet would be required as hint, but for me the set would be somehow more visible.

To have this in the matcher may introduce some "dependencies" there. Just have a look at the TCP correlation context, it introduces a TCP_CONNECTION_ID. Should the matcher then use a super set (DTLS + TCP) keys or should it be connector specific (makes it hard to define default matcher for all connector type), or should the matcher do it like the current and use some special keys to determine the specific key set?

sbernard31:
I think Matcher should be connector specific and so we don't need any hint.

> makes it hard to define default matcher for all connector type.

You're right. Maybe you can use `Connector.isSchemeSupported` or `Connector.getUri` to know the matcher to use.

boaks:
If we consider the correlation matcher to be related to the Connector's type, but also to be customized and passed into californium, the API will hardly "ensure" this "connector matching matcher" at compile time and must therefore assert that at runtime.
If that would be acceptable, I will prepare the changes.

sbernard31:
it seems acceptable for me.

boaks:
OK, then we're going to have:

- RelaxedDtlsCorrelationContextMatcher
- StrictdDtlsCorrelationContextMatcher
- (UdpCorrelationContextMatcher)
- TcpCorrelationContextMatcher
- (future: TlsCorrelationContextMatcher, though for TLS I currently can't get the epoch, there will be only one matcher).

Agreed?

sbernard31:
sounds good.

boaks:
Thanks for the input!
I will incorporate the input it into PR #202

PR #217, opened 2017-02-21
SslContextUtil for managing plain credentials and javax implementations.

boaks:
This utility converts between the informations used by the javax SSL
implementation and the plain credentials used by scandium. It offers
reading KeyStores, extracting credentials, and creating KeyManager and
TrustManager from the KeyStores or extracted credentials.

A folowup PR will use this for setup unit tests.


boaks:
Split into StringUtil (see #243) and a future PR for SslContextUtil

PR #227, opened 2017-03-03
To not drop FINISHED message when receiving it before CIPHERCHANGESPEC

sbernard31:
If FINISHED message is receive before CIPHER CHANGE SPEC, we will drop it because we have an ongoing handshake and the epoch does not match.

We should not drop it but give it to the handshaker which will queue it then processing it later.

This fixed is needed to move forward about access to psk store in an asynchronous way.

(No directly linked to this PR, but I would like to replace `LatchDecrementingDataHandler` by the RecordCollector I introduce in this PR, seems more readable to me)

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -760,9 +760,17 @@ public class DTLSConnector implements Connector {
 	 * @throws HandshakeException if the handshake message cannot be processed
 	 */
 	private void processHandshakeRecordWithConnection(final Record record, final Connection connection) throws HandshakeException {
-		if (connection.hasOngoingHandshake() && connection.getOngoingHandshake().getSession().getReadEpoch() == record.getEpoch()) {
-			// evaluate message in context of ongoing handshake
-			record.setSession(connection.getOngoingHandshake().getSession());
+		if (connection.hasOngoingHandshake()) {
+			if (connection.getOngoingHandshake().getSession().getReadEpoch() == record.getEpoch()) {
+				// evaluate message in context of ongoing handshake
+				record.setSession(connection.getOngoingHandshake().getSession());
+			} else if (record.getEpoch() > 0) {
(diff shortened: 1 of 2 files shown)


sophokles73:
In general, this looks good to me. I seems like a lot of effort required to test that the FINISHED message is forwarded to the handshake (*processing stage*), though. I was wondering whether we can make this kind of test much simpler if we build it on top of the changes introduced in PR #229. We could then simply create a mock `RecordProcessor` which would assert that all records are indeed forwarded to the processing stage, regardless of epoch. WDYT?

sbernard31:
I don't know I like the idea to check that the full handshake is ok.

sophokles73:
I do like that too in general but wouldn't we need to add tests for all possible permutations of *out-of-order* handshake messages as well? The unit tests for the `Handshaker` class already do test some of these so we might be able to add the missing ones there with a little less effort? Just an idea to save some code ... however, I have no problem with leaving it as is ...

sbernard31:
I'm not sure to understand, but I add the tests in DTLS connector as the bug was in DTLS connector. I mean adding the test in handshaker tests does not cover it.

sophokles73:
I know, the bug in `DTLSConnector` is just that it doesn't forward records from a future epoch to the processing stage. So the test might just be to ensure that it does. No need to make sure that the handshake itself succeeds. The tests for `Handshaker` then make sure that the records are actually processed in the correct order ...

sbernard31:
You're right the test I write is pretty high level and cover more than the fix I made.
You would like more low level tests which cover more bad order cases. I agree this could be valuable  but this is not really what I aim to achieve here. You're right the main thing I want to test is that records from a future epoch are forwarded to the processing stage. But even with the new RecordProcessor, I'm not sure I was able to do that easly. (Maybe when RecordProcessor will be extract from DTLSConnector).

Anyway, even if it's a bit complex, I try to make it readable.I think having one functional tests like this can make sense. (This is less sensitive to internal refactoring). (But this is not a strong opinion)

sophokles73:
Absolutely, please go ahead and merge the PR :-)

Issue #213, opened 2017-02-17
PSKStore latency impact


sbernard31:
I did some performance test with scandium.
And I face the same issue than Christian. (see [cf-dev](https://dev.eclipse.org/mhonarc/lists/cf-dev/msg01016.html))

Problems occurred when latency is added to PSK Store. (`getKey`)

Here problems I faced :
1) Access to keys through `PskStore` is done in a synchronous way.
2) We only have **1 thread which handle incoming messages** and this thread can be blocked to the `getKey` method. So when we wait for the key we stop to handle messages.
3) **DLTS retransmission** : We schedule a timeout for retransmission when we send a message. We cancel this scheduled retransmission before to send a new handshake message. In case where PSKstore take too much time (more than retransmission timeout) we can send retransmit message although we well received the next message (because we are blocked in `getKey` we don't sent the next message as fast enough and so don't canceled retransmission). In that case client will send again its message in response to of our retransmission and we will go for a new PSKStore access to answer... (Or how to shoot yourself in the foot)
4) **CoAP retransmission** : I did my tests using Californium at client side too. Currently when we send data through DTLSConnector and we don't have connection we initiate an handshake automatically. CoAP layer is not aware about that, it will schedule retransmission layer before data was really sent. So if handshake take too much time CoAP layer will try to resent data before the first handshake was finished and the CoAP message was sent. DTLSConnector will reintiate a full-handshake from scratch.

I just established a picture with this issue. We can create particular issues for each of this point to try to solve this general performance issue.

sophokles73:
There already is an open issue (#23) for item 4.

sbernard31:
Just a precision, this is not exactly the same issue.  Here we talk about retransmission and #23 talk about sending different application data. But it's a good thing to point this problem too as we should have both in mind to find a solution.

sbernard31:
#305 and #312  addressed the item 4.

boaks:
> In that case client will send again its message in response to of our retransmission and we will go for a new PSKStore access to answer... (Or how to shoot yourself in the foot)

I'm not sure. RFC6347 4.2.2

> DTLS implementations maintain (at least notionally) a
> next_receive_seq counter. This counter is initially set to zero.
> When a message is received, if its sequence number matches
> next_receive_seq, next_receive_seq is incremented and the message is
> processed. If the sequence number is less than next_receive_seq, the
> message MUST be discarded. 

[...]

PR #228, opened 2017-03-03
Delay retransmission when accessing to pskstore

sbernard31:
See #213 for more details.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -1539,6 +1539,22 @@ public class DTLSConnector implements Connector {
 			public void sendFlight(DTLSFlight flight) {
 				sendHandshakeFlight(flight, connection);
 			}
+
+			@Override
+			public void pauseRetransmission() {
+				DTLSFlight pendingFlight = connection.getPendingFlight();
+				if (pendingFlight != null) {
+					pendingFlight.cancelRetransmission();
+				}
+			}
+
+			@Override
+			public void resumeRetransmission() {
+				DTLSFlight pendingFlight = connection.getPendingFlight();
+				ScheduledFuture<?> f = timer.schedule(new RetransmitTask(pendingFlight), pendingFlight.getTimeout(),
+						TimeUnit.MILLISECONDS);
+				pendingFlight.setRetransmitTask(f);
+			}
 		};
 	}
 
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
@@ -471,6 +471,7 @@ public class ClientHandshaker extends Handshaker {
 			generateKeys(premasterSecret);
 			break;
 		case PSK:
+			recordLayer.pauseRetransmission();
 			String identity = pskStore.getIdentity(getPeerAddress());
 			if (identity == null) {
 				AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, session.getPeer());
@@ -483,6 +484,8 @@ public class ClientHandshaker extends Handshaker {
 				AlertMessage alert = new AlertMessage(AlertLevel.FATAL,	AlertDescription.HANDSHAKE_FAILURE, session.getPeer());
 				throw new HandshakeException("No preshared secret found for identity: " + identity, alert);
 			}
+			recordLayer.resumeRetransmission();
+
 			clientKeyExchange = new PSKClientKeyExchange(identity, session.getPeer());
 			LOGGER.log(Level.FINER, "Using PSK identity: {0}", identity);
 			premasterSecret = generatePremasterSecretFromPSK(psk);
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
@@ -205,6 +205,10 @@ public final class Connection implements SessionListener {
 		}
 	}
 
+	public DTLSFlight getPendingFlight() {
+		return pendingFlight;
+	}
+
 	/**
 	 * Gets the session containing the connection's <em>current</em> state.
 	 * 
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/RecordLayer.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/RecordLayer.java
@@ -39,4 +39,16 @@ public interface RecordLayer {
 	 *                  timespan to wait between re-transmissions. 
 	 */
 	void sendFlight(DTLSFlight flight);
+
+	/**
+	 * Pause any retransmission of the current pending flight.
+	 */
+	void pauseRetransmission();
+
+	/**
+	 * Resume previously paused retransmission. This means the retransmission
+	 * will be schedule again.
+	 */
+	void resumeRetransmission();
+
 }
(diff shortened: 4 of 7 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1539,6 +1539,22 @@ public void sendRecord(Record record) {
 			public void sendFlight(DTLSFlight flight) {
 				sendHandshakeFlight(flight, connection);
 			}
+
+			@Override
+			public void pauseRetransmission() {
```
boaks: `pause` and `resume` would make me beliefe, that the original retry timeout is continued. But in fact it's `cancel` an `restart`. Or did I oversee something?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -471,6 +471,7 @@ private void receivedServerHelloDone(ServerHelloDone message) throws HandshakeEx
 			generateKeys(premasterSecret);
 			break;
 		case PSK:
+			recordLayer.pauseRetransmission();
```
boaks: For the client side, if we reached this method, I think just `stop `the "own retransmission" would be enought. And we may move this more to the upper, say also for "expensive asymetric" cipher suites. Though very short later, the new flight will cancel the retransmission, I don't think it pays off to "resume/restart" the old retransmission.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -483,6 +484,8 @@ private void receivedServerHelloDone(ServerHelloDone message) throws HandshakeEx
 				AlertMessage alert = new AlertMessage(AlertLevel.FATAL,	AlertDescription.HANDSHAKE_FAILURE, session.getPeer());
 				throw new HandshakeException("No preshared secret found for identity: " + identity, alert);
 			}
+			recordLayer.resumeRetransmission();
```
boaks: canceled very short afterward when sending the new fligth. So omit it?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: Just to say: the functionality is OK for me. It's just that the names.

sbernard31: We cancel the current retransmission task, but not the whole retransmission, I mean we don't reset the flight tries counter and we don't reset the timeout value. That's why I choose `pause`/`resume` instead of `cancel`/`restart`. `restart` make me seem like we restart retransmission from the beginning  (flight tries = 0, timeout = 0).

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
sbernard31: You seems to say, that pause/resume is not necessary for client side ? why just for client side ?
You propose to just stop retransmission when message is received, right ?

I could be agree with that but this is no so easy :/
If we chose this solution,  where could be the right place to stop the retransmission ?
1) As soon as we received a new flight from the peer ?
[The spec](https://tools.ietf.org/html/rfc6347#section-4.1) seems to base retransmission on flight but we need to be able to detect duplicate, I'm not sure this really well match with our current implementation...
2) [As soon as we received a handshake message](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java#L842) ? 
This is not the right place, because message can be received in bad order or duplicate.
3) As soon as we receive the expected message (in Handshaker.doProcess() method) ?
Here we know we process the right message based on message sequence number, but there is no check about the nature of the message. I mean we can receive a message with the good sequence number but with bad type as [we doesn't store state](https://github.com/eclipse/californium/pull/309#issuecomment-304688301). 
More than that,  some time we don't expect one message but several messages (which could be sent in one flight or more), Should we ensure we received all messages before to stop retransmission ? or we just consider we can stop it when we receive the first message ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: Hm, so it's about the term "retransmission". 

I understood this as "single retransmission" triggered by the `RetransmitTask `scheduled in `scheduleRetransmission`. Based on that, there is no "retry counter" nor "timeout amplify".
My understanding was based on the usage of this term in DTLSFlight.

If "retransmission" is a potential sequence of retransmissions, your right, then the terms are proper. 

May be "pauseRetransmissions" and "resumeRetransmissions" would also be possible terms?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
boaks: > You seems to say, that pause/resume is not necessary for client side ? why just for client side ?

Oh no, I just got aware, that very short after the "resumeRetransmission" it's stopped again. But this is only true for the client side.

The other part, do also a "pause" for certificate bases handshake, was mainly influenced by issue #235 

So, if it's known, that there is potentially more time required to calculate then the timeout, we either pause or rely on working retransmission.

boaks: > And we may move this more to the upper, say also for "expensive asymetric" cipher suites. 

My idea, was to also pause for other potential expensive operations.
But the right places would require more research :-).

boaks:
My experience is, that for a lot of devices 1s (first retransmission timeout) is very fast, if crypo-functions are required. This PR could only solve the processing times of it's own side, but use "specific adjusted timeouts" for each flight may improve the performance. If we make such specific timeouts configurable, I'm not sure, if we require this PR.

Maybe we should also ask the IETF, if a TLS extension "relax timeouts" could be specified. This would enable clients to inform the server about it's capabilities and chose larger timeouts.

sbernard31:
>My experience is, that for a lot of devices 1s (first retransmission timeout) is very fast, if crypo-functions are required. This PR could only solve the processing times of it's own side, but use "specific adjusted timeouts" for each flight may improve the performance. If we make such specific timeouts configurable, I'm not sure, if we require this PR.

For me this is another issue. I think what your are talking about is a bit out of the spec. (It doesn't mean that this is not valuable). The issue is in-spec. I mean this does make sense to send a retransmission when we already received the expected message.
By the way, I don't think fixing the timeout issue solve this PR problem.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
sbernard31: >I would be afraid of stale handshake, because the retransmission may be stopped by a spoofed message (but I didn't go in to details, if this is really possible).

If you can spoof the message, you can send any messages and so completely stop the handshake by sending a malformed message or just sending an alert.

boaks: But I agree, that just stop the retransmission on an incoming handshake message would be a very simple solution for "longer processing times" and I don't see a easy argument against it. But not being an security expert, the unwanted side effects are out of my scope.

sbernard31:
I create a new PR #317  about stop retransmission instead of pause/resume (like discuss [here](https://github.com/eclipse/californium/pull/228#discussion_r119346208))
If this new PR is accepted, this will deprecate this one.

boaks:
Though #317 is merged, I think this one is to be closed. If I'm wrong, just reopen it again.

PR #229, opened 2017-03-06
Fix #216: Introduce queue for buffering inbound records

### LINKED ISSUE #216: Use multiple threads to process inbound records
This is particularly problematic because Scandium (currently) uses a single thread only for processing all incoming messages.

This is a follow-up issue of #213

sophokles73:
This is the initial step to fix #216.
The overall idea is to introduce an inbound Record buffer to `Connection` which is processed one-by-one using a thread pool based `Executor`. This way we achieve two things:
- make sure that all incoming records are processed in order
- allow multiple Handshake/Application records being processed in parallel

A next step could also be to re-factor the record processing logic into a separate class (i.e. puling it out of `DTLSConnector` which is already an ugly beast) implementing `RecordProcessor` ...

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
@@ -20,8 +20,13 @@ package org.eclipse.californium.scandium.dtls;
 import java.net.InetSocketAddress;
 import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertDescription;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertLevel;
 
 /**
  * Information about the DTLS connection to a peer.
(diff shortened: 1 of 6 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
```diff
@@ -283,7 +292,62 @@ public boolean addInbound(Record record) {
...
+	 * <li>invokes this method</li>
+	 * </ol>
+	 * <p>
+	 * This method uses an atomic boolean to make sure that only one record from this connection's buffer
+	 * is processed at a time, thus making sure that all records are processed in order of reception.
+	 * 
+	 * @param processor The object that should be used for processing the records.
+	 * @param executor The service to use for scheduling the processing tasks.
+	 */
+	public void processNextMessage(final RecordProcessor processor, final ExecutorService executor) {
```
sophokles73: sure, will do.

sophokles73:
Closed in favor of #233 so that we can collaborate on the code more easily ...

sbernard31:
Should I report comment on the #233 PR ?

sophokles73:
My understanding was that you wanted to do some changes, i.e. commits on the branch. That's the reason why I closed this one and pushed the branch to the Cf repo ... If you want to you can add the original comments from this PR there as well but I don't think that's necessary, is it?

sbernard31:
I wanted but I did not succeed :(. Let's restart the discussion on the new PR to avoid too much confusion.
I will add the original comments.

sophokles73:
That was the intention AFAIC :-)

PR #230, opened 2017-03-06
Remove nested synchonize section.

boaks:
Call synchronized section (getTracker() and tracker.getNextMessageId())
sequential.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
@@ -12,6 +12,9 @@
  * 
  * Contributors:
  *    Bosch Software Innovations - initial creation
+ *    Achim Kraus (Bosch Software Innovations GmbH) - remove nested synchronize.
+ *                                                    reduce blocking on 
+ *                                                    tracker.getNextMessageId()
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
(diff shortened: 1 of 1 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
```diff
@@ -51,17 +54,26 @@ public InMemoryMessageIdProvider(final NetworkConfig config) {
...
+	private synchronized MessageIdTracker getTracker(final InetSocketAddress destination) {
 		MessageIdTracker tracker = trackers.get(destination);
 		if (tracker == null) {
 			// create new tracker for destination lazily
 			tracker = new MessageIdTracker(config);
 			if (!trackers.put(destination, tracker)) {
-				// we have reached the maximum number of active peers
-				// TODO: throw an exception?
-				return -1;
+				return null;
```
sophokles73: since we're synchronizing this call we could also check capacity first, instead of instantiating the `MessageIdTracker` and throwing it away if the map is already full?

boaks: Sure.
I'm also thinking of implementing the MessageIdTracker "id page/epoch" oriented to shrink the timestamp arrays. So just save the Timestamp of the last used ID of a chunk (say 256 id per chunk). But not for now.

boaks:

sophokles73:
LGTM +1

Issue #211, opened 2017-02-08
Lost notification when first block2 notification is not complete.


sbernard31:
I faced a strange issue at client side when receiving 1 notification without blockmode during block2 transfer of the first notification on the same path. 

I will try to explain it but you'll see this is a bit tricky :p

1) Here is the failing scenario :
```
// client send a GET with observe on /test resource.
CON [MID=25457, T=f06bf010], GET, /test, (observe=0)    ----->

// server ACK the request and begin to send the response using block2
<-----   ACK [MID=25457], 0
<-----   CON [MID=111, T=f06bf010], 2.05, 2:0/1/1024, (observe=1)    

// client ACK and request the next block
ACK [MID=111], 0                          ----->
CON [MID=25458, T=ae04], GET, /test, 2:1/0/1024    ----->

// server finally send a new more recent notification but smaller without block2
<-----   CON [MID=222, T=f06bf010], 2.05, (observe=2)    

// we ack the response 
ACK [MID=222], 0                          ----->
// **** here at client side we get the response thanks to request.waitForResponse() : it works ****

// server send a new response and client acknowledge it
<-----   CON [MID=223, T=f06bf010], 2.05, (observe=3)  
ACK [MID=223], 0                          ----->
// **** But we never receive notification using Endpoint.addNotification() : failed :'( ****
```
2) I write a test about it [here](https://github.com/eclipse/californium/commit/fa4458c826c3e31a710014bb82c33d7213113242#diff-15e90779021001ad3493c536fd0165c0R315)

3) Here is a recap of how observe should works with notification:
We create a new exchange for new observe request then store it in Matcher maps. 
When we receive the response, we notify using `MessageObserver` API and we remove exchange from maps.
When a new response (notification) is received, [we create a new exchange from ObservationStore](https://github.com/eclipse/californium/blob/missing_notification/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java#L191) to handle it and we notify using `NotificationListener` API.

4) What happens in our use case :
The maps is cleaned in [ExchangeObserverImpl.completed(Exchange)](https://github.com/eclipse/californium/blob/missing_notification/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java#L294) using MID and Token of exchange.getCurrentRequest.
[...]

PR #231, opened 2017-03-06
Bugfix exchange management.

boaks:
Save remove of exchanges by token or MID.
For longterm exchange the MID or token may have been reused.
Don't use ExchangeObserver for notify Exchanges, not usefull there.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
@@ -12,6 +12,11 @@
  * 
  * Contributors:
  *    Bosch Software Innovations - initial creation
+ *    Achim Kraus (Bosch Software Innovations GmbH) - remove ExchangeObserver.
+ *                                                    Notify exchanges are not 
+ *                                                    stored within the matcher
+ *                                                    and therefore don't require
+ *                                                    a cleanup.
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -161,14 +166,12 @@ public abstract class BaseMatcher implements Matcher {
 	 * Special matching for notify reponses. Check, is a observe is stored in
 	 * {@link #observationStore} and if found, recreate a exchange.
 	 * 
-	 * @param exchangeObserver exchange observer for the new exchange
 	 * @param response notify response
 	 * @param responseContext correlation context of response
 	 * @return exchange, if a new one is create of the stored observe
 	 *         informations, null, otherwise.
 	 */
-	protected final Exchange matchNotifyResponse(final ExchangeObserver exchangeObserver, final Response response,
-			final CorrelationContext responseContext) {
+	protected final Exchange matchNotifyResponse(final Response response, final CorrelationContext responseContext) {
 
 		final Exchange.KeyToken idByToken = Exchange.KeyToken.fromInboundMessage(response);
 		Exchange exchange = null;
@@ -184,7 +187,6 @@ public abstract class BaseMatcher implements Matcher {
 			request.setDestinationPort(response.getSourcePort());
 			exchange = new Exchange(request, Origin.LOCAL, obs.getContext());
 			exchange.setRequest(request);
-			exchange.setObserver(exchangeObserver);
 			LOG.log(Level.FINER, "re-created exchange from original observe request: {0}", request);
 			request.addMessageObserver(new MessageObserverAdapter() {
 
@@ -224,7 +226,6 @@ public abstract class BaseMatcher implements Matcher {
 				public void onCancel() {
 					observationStore.remove(request.getToken());
 					exchangeStore.releaseToken(idByToken);
-
 				}
 			});
 		}
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
@@ -12,6 +12,13 @@
  * 
  * Contributors:
  *    Bosch Software Innovations - initial creation
+ *    Achim Kraus (Bosch Software Innovations GmbH) - remove only provided Exchange
+ *                                                    Observes and blockwise 
+ *                                                    exchanges may be used 
+ *                                                    longer, so that MID (observe)
+ *                                                    or token (blockwise) may
+ *                                                    be reused for an other
+ *                                                    exchange. 
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -253,26 +260,38 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 	}
 
 	@Override
-	public void remove(final KeyToken token) {
+	public void remove(final KeyToken token, final Exchange exchange) {
+		boolean removed;
+		int size;
 		synchronized (exchangesByToken) {
-			exchangesByToken.remove(token);
-			LOGGER.log(
-					Level.FINE,
-					"removing exchange for token {0}, remaining exchanges by tokens: {1}",
-					new Object[]{token, exchangesByToken.size()});
+			removed = exchangesByToken.remove(token, exchange);
+			size = exchangesByToken.size();
+		}
+		if (removed) {
+			LOGGER.log(Level.FINE, "removing exchange for token {0}, remaining exchanges by tokens: {1}",
+					new Object[] { token, size });
 		}
 	}
 
 	@Override
-	public Exchange remove(final KeyMID messageId) {
+	public Exchange remove(final KeyMID messageId, final Exchange exchange) {
+		Exchange removedExchange;
+		int size;
 		synchronized (messageIdProvider) {
-			Exchange removedExchange = exchangesByMID.remove(messageId);
-			LOGGER.log(
-					Level.FINE,
-					"removing exchange for MID {0}, remaining exchanges by MIDs: {1}",
-					new Object[]{messageId, exchangesByMID.size()});
-			return removedExchange;
+			if (null == exchange) {
+				removedExchange = exchangesByMID.remove(messageId);
+			} else if (exchangesByMID.remove(messageId, exchange)) {
+				removedExchange = exchange;
+			} else {
+				removedExchange = null;
+			}
+			size = exchangesByMID.size();
 		}
+		if (null != removedExchange) {
+			LOGGER.log(Level.FINE, "removing exchange for MID {0}, remaining exchanges by MIDs: {1}", new Object[] {
+					messageId, size });
+		}
+		return removedExchange;
 	}
 
 	@Override
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/MessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/MessageExchangeStore.java
@@ -9,7 +9,10 @@
  *    http://www.eclipse.org/legal/epl-v10.html
  * and the Eclipse Distribution License is available at
  *    http://www.eclipse.org/org/documents/edl-v10.html.
- * 
+ *
+ * Contributors:
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add Exchange to remove for
+ *                                                    save cleanup.
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -107,17 +110,21 @@ public interface MessageExchangeStore {
 	 * Removes the exchange registered under a given token.
 	 * 
 	 * @param token the token of the exchange to remove.
+	 * @param exchange Exchange to be removed, if registered with provided token.
 	 */
-	void remove(KeyToken token);
+	void remove(KeyToken token, Exchange exchange);
 
 	/**
 	 * Removes the exchange registered under a given message ID.
 	 * 
 	 * @param messageId the message ID to remove the exchange for.
-	 * @return the exchange that was registered under the given message ID or {@code null}
-	 * if no exchange was registered under the ID.
+	 * @param exchange Exchange to be removed. If {@code null}, the current
+	 *                 exchange with the MID is removed. If not {@code null},
+	 *                 only this exchange is removed, if it's registered with
+	 *                 the MID.
+	 * @return the removed exchange, or {@code null}, if no exchange was removed.
 	 */
-	Exchange remove(KeyMID messageId);
+	Exchange remove(KeyMID messageId, Exchange exchange);
 
 	/**
 	 * Gets the exchange registered under a given token.
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -23,9 +23,14 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - replace isResponseRelatedToRequest
  *                                                    with CorrelationContextMatcher
  *                                                    (fix GitHub issue #104)
+ *    Achim Kraus (Bosch Software Innovations GmbH) - remove obsolete ExchangeObserver
+ *                                                    from matchNotifyResponse. Don't
+ *                                                    remove exchange by MID on notify. 
+ *                                                    Add Exchange for save remove.
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -79,8 +84,8 @@ public final class UdpMatcher extends BaseMatcher {
 
 		exchange.setObserver(exchangeObserver);
 		exchangeStore.registerOutboundRequest(exchange);
-		// for observe request.
-		if (request.isObserve()) {
+		// for first observe request. Ignored for retries
+		if (request.isObserve() && 0 == exchange.getFailedTransmissionCount()) {
 			registerObserve(request);
 		}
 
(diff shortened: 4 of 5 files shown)


boaks:
Surprisingly this fixes #211 (at least the provided test there).

So, my analysis for #211, complete the ongoing blockwise exchange only cleans up the current request leaving the starting observe request in the token list. Adding the cleanup  for the starting request in the ExchangeObserver fixes that missbehaviour.

@sbernard31 
If it's possible, could you test, if this fixes not only the provided test but also the issue?

californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
```diff
@@ -184,7 +187,6 @@ protected final Exchange matchNotifyResponse(final ExchangeObserver exchangeObse
 			request.setDestinationPort(response.getSourcePort());
 			exchange = new Exchange(request, Origin.LOCAL, obs.getContext());
 			exchange.setRequest(request);
-			exchange.setObserver(exchangeObserver);
```
sbernard31: I'm not sure this is not usefull. Don't we need `exchangeObserver` for blockwise ?

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -310,11 +317,19 @@ public void completed(final Exchange exchange) {
...
-				exchangeStore.remove(idByToken);
-				if(!originRequest.getOptions().hasObserve()) {
+				exchangeStore.remove(idByToken, exchange);
+				Request request = exchange.getRequest();
+				if (request != originRequest && null != request &&  null != request.getToken()
+						&& !Arrays.equals(request.getToken(), originRequest.getToken())) {
+					// remove starting request also
+					originRequest = request;
+					idByToken = KeyToken.fromOutboundMessage(originRequest);
+					exchangeStore.remove(idByToken, exchange);
```
sbernard31: Why should we do that ?

californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
boaks: Hm, good question.
For other notifies, there is nothing to cleanup.

sbernard31:
>So, my analysis for #211, complete the ongoing blockwise exchange only cleans up the current request leaving the starting observe request in the token list. Adding the cleanup for the starting request in the ExchangeObserver fixes that missbehaviour.

Exactly that's why I propose this [fix](https://github.com/eclipse/californium/commit/8e83a1be9aeb8d4290e411113d0182ee23cbe224) when I opened the issue. This way we clean the original request. (I also clean the block2 status).

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
boaks: Why should we not cleanup the first request?
This is only valid for blockwise, where the first request has a different token.
That's the point where your #211 was broken.
when blockwise block2 finished regular, we do

> 	exchange.setCurrentRequest(exchange.getRequest());

But if the exchange is completed by a other response in between, we need to remove the first request also.

sbernard31: As it's a special blockwise behavior, should we not fix this in the blockwiselayer ?

boaks: Or in the ExchangeObserver as cleanup, which support all kinds of valid Exchange usage :-).

sbernard31: Ok but I believe that the idea about the blocklayer refactoring was to move as much as possible block2 behavior to this class. I also believe that we all would like to removing this confusing currentRequest/request. It seems a good strategy could be to isolate it before to remove it.

californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
boaks: The blockwise exchange only needs cleanup, if further blocks are read. But in that case, the implementations of Matcher.sendRequest() will set the ExchangeObserver at the point, where the Exchange is added to the matchers lists.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
boaks: Until that remove of the confusing currentRequest/request, it seems to be dangerous, not the cleanup for both request. By the way, thought the Matcher registers for all sent request, I think changing the request in the exchange is dangerous, because the cleanup may be not execute.
Currently this seems only working, because the token is not changed and the MID is cleaned-up by the ACK/Response.
So my view point is still: the Matcher stuff adds, then it should cleanup also. If Exchange is offering two requests, both had to be cleaned-up. Once, when one of the requests are removed, then this second clean up would be easy to remove.

sbernard31: I tried to understand better all of this reading the code, but I'm afraid I'm yet more lost.
Reading the javadoc of currentRequest :
```java
/**
* Returns the current request block. If a request is not being sent
* blockwise, the whole request counts as a single block and this method
* returns the same request as {@link #getRequest()}. Call getRequest() to
 * access the assembled request.
 * 
 * @return the current request block
*/
public Response getCurrentResponse()
```
and 
```java
/**
 * This method is only needed when the same {@link Exchange} instance uses
 * different tokens during its lifetime, e.g., when using a different token
 * for retrieving the rest of a blockwise notification (when not altered,
 * Californium reuses the same token for this).
 * See {@link BlockwiseLayer} for an example use case.
 */
public void completeCurrentRequest()
```
and knowing that blockwiselayer should be the only class which change the currentRequest (changing the orginal one by the current block request), I still feel that the cleaning should be initiated by it. 

But as I explained I'm not sure to see how all of this should work, I just want to let you know that  adding this cleaning step confusing me a bit more.

boaks: About the confusion I totally agree :-).

See also #232. There we search for observes in the exchangeStore, where we have already removed those Observe-Exchanges with the "onNotification" redesign :-).

I feel , that we now try to cleanup in a lot of places. But I don't feel, that we ensure not to forgett something. And removing the token or MID for a Exchange multipe times may be dangerous, if the token or MID is reused in the meantime. The most important of that PR is therefore the "save remove" of Exchanges. And checking for "leaks", I found the stuff wiht the two request an dadded the cleanup in the ExcahngeObserver, mainly because I don't feel, that it is easy to ensure, that the blockwise layer will correct the currentRequest before completion.

If the blockwise should do the "special request" cleanup, the blockwise layer should be informed about the completion. So one way to do that would introduce a special ExchangeObserver for (blockwise)Exchanges, set in the blockwise layer into the Exchange. But I don't think this would pay off.

boaks: @sbernard31 

I checked now the possibility to move the blockwise specific exchange stuff into the "blockwise layer".

Result:
Exchange only offers 1 ExchangeObserver, it could only be set not get
The UdpMatcher.ExchangeObserverImpl is private and could therefore not be extended.

So I don't see a way to move the stuff without making too complicated.

sbernard31: which blockwise stuff would you remove ?

boaks: > 				/* keep track of the starting request. Currently only used with blockwise transfer */
> 				Request request = exchange.getRequest();
> 				if (request != originRequest && null != request &&  null != request.getToken()
> 						&& !Arrays.equals(request.getToken(), originRequest.getToken())) {
> 					// remove starting request also
> 					originRequest = request;
> 					idByToken = KeyToken.fromOutboundMessage(originRequest);
> 					exchangeStore.remove(idByToken, exchange);
> 				}
> 

I thought you was complainig about the blockwise specific stuff in the general ExchangeObserver. Therefore I tried to check, if moving this is posssible. I think it woulb be too complicated to move these line into a special BlockwiseExchangeObserver.

sbernard31: The way this is currently done is by changing currentRequest/request on exchange and/or call setComplete() ou completeCurrentRequest() like [here](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java#L600).

That's what I meant by "cleaning should be initiated by blockwiseLayer".

boaks: The point is: relying on the replacing the `currentRequest` with the (origin)`Request` is very, **very fragile**!!! It's only done, it the blockwise is aware of being completed. If an other layer decided to do so, the leak is the result!

So, only in `ExchangeObserver.compeleted` we can ensure, that the request is relaesed, if it's not the `currentRequest`. I don't see, that we can do this reliable outside of that compeleted.

sbernard31: Ok, so I suppose "removing also request stuff" make sense only if you check that exchange.isComplete(). If you don't do that the `completeCurrentRequest` does make sense anymore ?

boaks: I don't get it!

`UdpMatcher.ExchangeObserverImpl.completed` is **ONLY** called, when the Exchange is completed! So why would you check `Exchange.isCompleted()`? You will not be called in `UdpMatcher.ExchangeObserverImpl.completed` if  `Exchange.isCompleted()` would return false.

sbernard31:
```java
	/**
	 * Call this method to trigger a clean-up in the Matcher through its
	 * ExchangeObserverImpl. Usually, it is called automatically when reaching
	 * the StackTopAdapter in the {@link CoapStack}, when timing out, when
	 * rejecting a response, or when sending the (last) response.
	 */
	public void setComplete() {
		this.complete = true;
		if (obs != null)
			obs.completed(this);
	}
	
	/**
	 * This method is only needed when the same {@link Exchange} instance uses
	 * different tokens during its lifetime, e.g., when using a different token
	 * for retrieving the rest of a blockwise notification (when not altered,
	 * Californium reuses the same token for this).
	 * See {@link BlockwiseLayer} for an example use case.
	 */
	public void completeCurrentRequest() {
		if (obs != null)
			obs.completed(this);
	}
 ```

Comment (boaks) [inline californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java]: So your right!!! I add the check :-).

(Its really, really confusing! Currenlty, if the `completeCurrentRequest()` is called, the `setComplete()` follows.)

Comment (sbernard31) [inline californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java]: I'm clearly confuse too. I'm currently searching if we still need this currentRequest.
I feel `currentRequest` concept should not be so usefull since @sophokles73 move blockstatus in blocklayer.

Comment (boaks) [inline californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java]: Get rid of the `currentRequest` would be the best solution.

But until that, I owuld prefer to have a workaround in the `completed` to ensure the cleanup.

Comment (sbernard31) [inline californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java]: I agree for the workaround too.

**Comment by boaks** [comment]
So, now, with `Exchange.isComplete()` test :-).

**Comment by boaks** [comment]

File: californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java [hunk 3]
```diff
@@ -239,26 +246,40 @@ public void registerOutboundRequestWithTokenOnly(final Exchange exchange) {
...
-			Exchange removedExchange = exchangesByMID.remove(messageId);
-			LOGGER.log(
-					Level.FINE,
-					"removing exchange for MID {0}, remaining exchanges by MIDs: {1}",
-					new Object[]{messageId, exchangesByMID.size()});
-			return removedExchange;
+			if (null == exchange) {
+				removedExchange = exchangesByMID.remove(messageId);
+			}
+			else if (exchangesByMID.remove(messageId, exchange)) {
```
sophokles73: put on same line as closing bracket

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -239,26 +246,40 @@ public void registerOutboundRequestWithTokenOnly(final Exchange exchange) {
...
+				removedExchange = exchange;
+			}
+			else {
```
sophokles73: put on same line as closing bracket

californium-core/src/main/java/org/eclipse/californium/core/network/MessageExchangeStore.java
```diff
@@ -101,17 +104,22 @@
...
 	 * @param messageId the message ID to remove the exchange for.
+	 * @param exchange Exchange to be removed. If {@code null}, the current
+	 *                 exchange with the MID is removed. If not {@code null},
+	 *                 only this exchange is removed, if it's registered with
+	 *                 the MID.
 	 * @return the exchange that was registered under the given message ID or {@code null}
 	 * if no exchange was registered under the ID.
 	 */
-	Exchange remove(KeyMID messageId);
+	Exchange remove(KeyMID messageId, Exchange exchange);
```
sophokles73: The docs for the returned value need to be updated with the additional constraint introduced by the *exchange* param, shouldn't it?

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -79,8 +84,8 @@ public void sendRequest(final Exchange exchange, final Request request) {
 
 		exchange.setObserver(exchangeObserver);
 		exchangeStore.registerOutboundRequest(exchange);
-		// for observe request.
-		if (request.isObserve()) {
+		// for first observe request. Ignored for retries
+		if (request.isObserve() && 0 == exchange.getFailedTransmissionCount()) {
```
sophokles73: :+1:

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -187,6 +192,7 @@ public Exchange receiveResponse(final Response response, final CorrelationContex
 		final KeyToken idByToken = KeyToken.fromInboundMessage(response);
 		LOGGER.log(Level.FINER, "received response {0}", response);
 		Exchange exchange = exchangeStore.get(idByToken);
+		boolean notify = false; // don't remove MID for notifies. May be already reused.
```
sophokles73: can we rename this to e.g. 'isNotification'

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -310,11 +317,23 @@ public void completed(final Exchange exchange) {
...
 					KeyMID idByMID = KeyMID.fromOutboundMessage(originRequest);
-					exchangeStore.remove(idByMID);
+					exchangeStore.remove(idByMID, exchange);
 				}
 				KeyToken idByToken = KeyToken.fromOutboundMessage(originRequest);
-				exchangeStore.remove(idByToken);
-				if(!originRequest.getOptions().hasObserve()) {
+				exchangeStore.remove(idByToken, exchange);
+				/* filter calls by completeCurrentRequest */
+				if (exchange.isComplete()) {
```
sophokles73: how could it not be completed? The `ExchangeObserver.completed()` is invoked only if the exchange is complete, isn't it?

sbernard31: https://github.com/eclipse/californium/pull/231#discussion_r105401490

boaks: I thought the same, but sbernard31 knew it better :-).

`Exchange.completeCurrentRequest()`

sophokles73:
In particular, all of this discussion is another excellent argument for completely removing the `BlockwiseLayer` from the stack and replacing it with helper classes for implementing clients and resources that want to support blockwise transfer :-)

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
sophokles73: Ok, understood (well, not really *understood* but I guess you guys figured it out ;-))

sbernard31: Nobody really understood  -_-!

boaks: But if we merge now, we don't see the mess "by default" :-)

sbernard31:
I'm not sure extracting the `BlockwiseLayer` without moving out observe is possible. I don't see how exchangeStore and TokenProvider will works with this.

If you really want to simplify the stack I suspect we need to be more radical : the stack should just handle messages(+ reliability) and let the upper layer matching request and response. Return to the spec meaning of exchange (CON + ACK)

boaks:
If tomorrow hudson is green again, I will merge this.

boaks:
hudson was green again :-)

PR #233, opened 2017-03-06
Fix #216: Introduce parallel processing of inbound records

### LINKED ISSUE #216: Use multiple threads to process inbound records

sophokles73:
This is the initial step to fix #216.
The overall idea is to introduce an inbound Record buffer to Connection which is processed one-by-one using a thread pool based Executor. This way we achieve two things:

make sure that all incoming records are processed in order
allow multiple Handshake/Application records being processed in parallel
A next step could also be to re-factor the record processing logic into a separate class (i.e. puling it out of DTLSConnector which is already an ugly beast) implementing RecordProcessor ...

Created from a branch on eclipse/californium so that @sbernard31 and other committers can collaborate on it. Replaces #229

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -108,6 +108,9 @@ import org.eclipse.californium.scandium.dtls.SessionTicket;
 import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
 import org.eclipse.californium.scandium.util.ByteArrayUtils;
 
+import eu.javaspecialists.tjsn.concurrency.stripedexecutor.StripedExecutorService;
+import eu.javaspecialists.tjsn.concurrency.stripedexecutor.StripedRunnable;
+
 
 /**
  * A {@link Connector} using <em>Datagram TLS</em> (DTLS) as specified in
@@ -131,6 +134,12 @@ public class DTLSConnector implements Connector {
 			+ 12 // DTLS message headers
 			+ 13 // DTLS record headers
 			+ MAX_CIPHERTEXT_EXPANSION;
+	/**
+	 * The default size of the striped executor's thread pool which is used for processing records.
+	 * <p>
+	 * The value of this property is 6 * <em>#(CPU cores)</em>.
+	 */
+	private static final int DEFAULT_EXECUTOR_THREAD_POOL_SIZE = 6 * Runtime.getRuntime().availableProcessors();
 
 	/** all the configuration options for the DTLS connector */ 
 	private final DtlsConnectorConfig config;
@@ -176,9 +185,10 @@ public class DTLSConnector implements Connector {
 	private CorrelationContextMatcher correlationContextMatcher;
 
 	private RawDataChannel messageHandler;
-
 	private ErrorHandler errorHandler;
 	private SessionListener sessionCacheSynchronization;
+	private StripedExecutorService executor;
+	private boolean hasInternalExecutor;
 
 	/**
 	 * Creates a DTLS connector from a given configuration object
@@ -194,14 +204,14 @@ public class DTLSConnector implements Connector {
 	/**
 	 * Creates a DTLS connector for a given set of configuration options.
 	 * 
-	 * @param configuration the configuration options
-	 * @param sessionCache an (optional) cache for <code>DTLSSession</code> objects that can be used for
+	 * @param configuration The configuration options.
+	 * @param sessionCache An (optional) cache for <code>DTLSSession</code> objects that can be used for
 	 *       persisting and/or sharing of session state among multiple instances of <code>DTLSConnector</code>.
 	 *       Whenever a handshake with a client is finished the negotiated session is put to this cache.
 	 *       Similarly, whenever a client wants to perform an abbreviated handshake based on an existing session
 	 *       the connection store will try to retrieve the session from this cache if it is
 	 *       not available from the connection store's in-memory (first-level) cache.
-	 * @throws NullPointerException if the configuration is <code>null</code>
+	 * @throws NullPointerException if the configuration is <code>null</code>.
 	 */
 	public DTLSConnector(final DtlsConnectorConfig configuration, final SessionCache sessionCache) {
 		this(configuration,
@@ -211,7 +221,15 @@ public class DTLSConnector implements Connector {
 						sessionCache));
 	}
 
+	/**
+	 * Creates a DTLS connector for a given set of configuration options.
+	 * 
+	 * @param configuration The configuration options.
+	 * @param connectionStore The registry to use for managing connections to peers.
+	 * @throws NullPointerException if any of the parameters is <code>null</code>.
+	 */
 	DTLSConnector(final DtlsConnectorConfig configuration, final ResumptionSupportingConnectionStore connectionStore) {
+
 		if (configuration == null) {
 			throw new NullPointerException("Configuration must not be null");
 		} else if (connectionStore == null) {
@@ -221,6 +239,31 @@ public class DTLSConnector implements Connector {
 			this.outboundMessages = new LinkedBlockingQueue<>(config.getOutboundMessageBufferSize());
 			this.connectionStore = connectionStore;
 			this.sessionCacheSynchronization = (SessionListener) this.connectionStore;
+		}
+	}
+
+	/**
+	 * Sets the executor to use for processing records.
+	 * <p>
+	 * If this property is not set before invoking the {@linkplain #start() start method},
+	 * a new {@link StripedExecutorService} is created with a thread pool of
+	 * {@linkplain #DEFAULT_EXECUTOR_THREAD_POOL_SIZE default size}.
+	 * 
+	 * This helps with performing multiple handshakes in parallel, in particular if the key exchange
+	 * requires a look up of identities, e.g. in a database or using a web service.
+	 * <p>
+	 * If this method is used to set an executor, the executor will <em>not</em> be shut down
+	 * by the {@linkplain #stop() stop method}.
+	 * 
+	 * @param executor The executor.
+	 * @throws IllegalStateException if his connector is already running.
+	 */
+	public final synchronized void setExecutor(StripedExecutorService executor) {
+
+		if (running.get()) {
+			throw new IllegalStateException("cannot set executor while connector is running");
+		} else {
+			this.executor = executor;
 		}
 	}
 
@@ -275,6 +318,11 @@ public class DTLSConnector implements Connector {
 		timer = Executors.newSingleThreadScheduledExecutor(
 				new DaemonThreadFactory("DTLS RetransmitTask-", NamedThreadFactory.SCANDIUM_THREAD_GROUP));
 
+		if (executor == null) {
+			// use a decently sized thread pool
+			executor = new StripedExecutorService(DEFAULT_EXECUTOR_THREAD_POOL_SIZE);
+			this.hasInternalExecutor = true;
+		}
 		socket = new DatagramSocket(null);
 		// make it easier to stop/start a server consecutively without delays
 		socket.setReuseAddress(true);
@@ -347,7 +395,7 @@ public class DTLSConnector implements Connector {
 	 * <p>
 	 * This method's execution time is proportional to the number of connections this connector maintains.
 	 */
-	public final synchronized void forceResumeAllSesions() {
+	public final synchronized void forceResumeAllSessions() {
 		connectionStore.markAllAsResumptionRequired();
 	}
 
@@ -367,8 +415,8 @@ public class DTLSConnector implements Connector {
 	 */
 	final synchronized void releaseSocket() {
 		running.set(false);
-		sender.interrupt();
 		outboundMessages.clear();
+		sender.interrupt();
 		if (socket != null) {
 			socket.close();
 			socket = null;
@@ -381,6 +429,10 @@ public class DTLSConnector implements Connector {
 		if (running.get()) {
 			LOGGER.log(Level.INFO, "Stopping DTLS connector on [{0}]", lastBindAddress);
 			timer.shutdownNow();
+			if (hasInternalExecutor) {
+				executor.shutdownNow();
+				hasInternalExecutor = false;
+			}
 			releaseSocket();
 		}
 	}
@@ -402,6 +454,7 @@ public class DTLSConnector implements Connector {
 	}
 
 	private void receiveNextDatagramFromNetwork() throws IOException {
+
 		byte[] buffer = new byte[inboundDatagramBufferSize];
 		DatagramPacket packet = new DatagramPacket(buffer, buffer.length);
 		synchronized (socket) {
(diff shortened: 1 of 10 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
```diff
@@ -283,7 +292,62 @@ public boolean addInbound(Record record) {
...
+	 * <li>invokes this method</li>
+	 * </ol>
+	 * <p>
+	 * This method uses an atomic boolean to make sure that only one record from this connection's buffer
+	 * is processed at a time, thus making sure that all records are processed in order of reception.
+	 * 
+	 * @param processor The object that should be used for processing the records.
+	 * @param executor The service to use for scheduling the processing tasks.
+	 */
+	public void processNextMessage(final RecordProcessor processor, final ExecutorService executor) {
```
sbernard31: There is a lot of synchronization tools (blocking queue + synchronized block+ atomic boolean)

I tried to find a better way :
- First time by removing synchronized block but I can not find any thread safe way to do.
- Then, I tried to just use synchronize block, we don't really need atomic boolean and concurrent queue, but I can not find a non concurrent queue implementation with a maximum size, so I'm just able to remove the atomic boolean...
- So finally, I think it's ok like that.

Another solution could be to use a [Striped Executor Services](http://www.javaspecialists.eu/archive/Issue206.html). That way, we lost the queue size limitation. (Maybe we don't really need it)

sbernard31:
I like the idea to enqueue Records then ensure that only 1 thread at the time can handle this queue.

I agree we should create a class responsible to do that. DTLSConnector should only handle messages without connection.

I have some doubts about adding the record queue on `Connection` object.
In a more general way, I have concerns about `Connection` and `Session` Object which is currently shared by several Threads (send worker, receiver worker, retransmission thread, new record handler thread pool).
Should we try to make immutable those 2 objects. This way we would be able to confidently share it between several threads. Do you have the same feeling ? (This is just a long term question)

Anyway, I think it's a really good first step :).

element-connector/src/main/java/org/eclipse/californium/elements/util/DaemonThreadFactory.java
```diff
@@ -0,0 +1,34 @@
...
+	 *            threads, created by this factory.
+	 * @param threadGroup the thread group or <code>null</code>
+	 */
+	public DaemonThreadFactory(final String threadPrefix, final ThreadGroup threadGroup) {
+		super(threadPrefix, threadGroup);
+	}
+
+	@Override
+	protected boolean createDaemonThreads() {
+		return false;
```
sophokles73: Absolutely. Thanks for spotting :-)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -286,6 +294,12 @@ public Thread newThread(Runnable r) {
 				return ret;
 			}
 		});
+		if (handshakerTaskExecutor == null) {
+			handshakerTaskExecutor = Executors.newFixedThreadPool(
+					Runtime.getRuntime().availableProcessors(),
+					new DaemonThreadFactory("DTLS Handshaker Task", SCANDIUM_THREAD_GROUP));
```
sophokles73: I guess because we do not want the JVM to have to wait until all handshaker tasks have finished before being able to shut down, or do we?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
sophokles73: I had the same problems finding a queue that requires less synchronization but enforcing a capacity limit. Maybe we can simply subclass an existing one and add capacity limitation ourselves ...

sophokles73:
> I have some doubts about adding the record queue on Connection object.
In a more general way, I have concerns about Connection and Session Object which is currently shared by several Threads (send worker, receiver worker, retransmission thread, new record handler thread pool).
Should we try to make immutable those 2 objects. This way we would be able to confidently share it between several threads. Do you have the same feeling ? (This is just a long term question)

I share your concerns in general. However, I do not see how we could make them immutable, e.g. the session object needs to have its counters and received records window being updated with every message received and sent. However, we might indeed be able to serialize access to these objects by scheduling tasks to a striped executor.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
```diff
@@ -283,7 +292,62 @@ public boolean addInbound(Record record) {
 	 * 
 	 * @return the next record or {@code null} if the buffer is empty.
 	 */
-	public Record getNextInboundRecord() {
-		return incoming.poll();
+	private Record getNextInboundRecord() {
+
+		synchronized(incoming) {
```
sophokles73: I guess it will do not harm then, right? If the monitor is already acquired, no waiting is needed or is it?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
sbernard31: Generally, when it's possible I prefer to re-use classes instead of customizing it. I suppose this is acceptable to start with a blockingQueue.

sbernard31:
I suppose we will see clearer about what is accessed concurrently when we will move RecordProcessor in another class. Concurrent access should be limited to DTLSConnector access as all tasks in Handshaker and RecordProcessor should be executed by only one thread for one connection,.

sophokles73:
I have incorporated the [StripedExecutorService](http://www.javaspecialists.eu/archive/Issue206.html) and have created a CQ for it. Once it gets approved, I will commit the change. The striped executor is part of the `DTLSConnector` and is currently used for inbound records only but I think we can (and should) also use it for outbound records as well.

However, I have removed the `RecordProcessor` interface for now simply because it caused way too much trouble trying to separate the record processing logic from the rest of the `DTLSConnector` (sigh) ...

Anyway, I think this is a good change and it should basically solve our original problem of not being able to perform multiple handshakes in parallel ...

sophokles73:
Any comments @sbernard31 @boaks @Christian-Schmid @DanielMaier-BSI ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -225,6 +237,27 @@ public DTLSConnector(final DtlsConnectorConfig configuration, final SessionCache
 	}
 
 	/**
+	 * Sets the executor to use for processing records.
+	 * <p>
+	 * If this property is not set before invoking the {@linkplain #start() start method},
+	 * a new {@link StripedExecutorService} is used with a thread pool of size #(CPU cores).
+	 * <p>
```
boaks: > * 6

Missing?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -419,40 +462,78 @@ private void receiveNextDatagramFromNetwork() throws IOException {
...
 				case CHANGE_CIPHER_SPEC:
-					processChangeCipherSpecRecord(record);
-					break;
-				case HANDSHAKE:
-					processHandshakeRecord(record);
+					executor.execute(new StripedRunnable() {
+
+						@Override
+						public Object getStripe() {
+							return record.getPeerAddress();
```
boaks: I'm not sure, if a "resume" from an other address creates problems.

boaks:
All in all, using the `StripedExecutorService` looks very good for me.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: nice catch :-)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: what kind of problem do you suspect to occur?

boaks: Assuming an "attack" (simultaneous messages from different addresses for the same DTLS session), the handshake may be executed in parallel from different executors.

sophokles73: Are you talking about an *intended* attack or an *accidental* attack? How would an attacker get the session ID, by guessing?

sbernard31:
Looks good for me. I ran some kind of "performances tests" that seems ok.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: I have no detailed analysis on that.
It's just a "very early idea" on what may go wrong (in that case, "executed in parallel").

sophokles73:
> I ran some kind of "performances tests" that seems ok.

Can you disclose some details/results?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: Anyway, this is a problem we are facing already, i.e. it is not introduced nor amplified by this PR, as far as I can see.

boaks: I don't want to stress this too much. It was just a very early thinking on what may go wrong.
(With a single thread, there was no "executed in parallel".)

sophokles73: That's true. Very good point. At least we were able to detect that an abbreviated handshake was already going on ...
Since we're synchronizing access to the `ConnectionStore` we should be able to still detect this situation and e.g. reject the concurrent resumption request.

sbernard31:
I just ran a LWM2M server with a PSKStore with 2s latency.
Then I run 100 LWM2M devices at the same time. (without coap retransmission because of 4) of #213)
And I mesure how many time to do a register. The tests stop after 5 minutes even if all registration is not done (success or failed).
The results :
 * With default value (24 threads)  : 100 success (max 10 813ms, min 8 997ms, avg 9 637ms)
 * With 100 threads : 100 success (max 4 549ms, min 3 500ms, avg 4 148ms)
 * With before this PR : 30 success (max 292 052ms, min 5 356ms, avg 267 104ms)

This is clearly better even if I expected just a bit more than 2s for the tests with 100 threads. I don't really understand this difference.

sbernard31:
A bit better when I removed some logs : 100 (max 3 063ms, min 2 599ms, avg 2 854ms)

sophokles73:
Context switching?

DanielMaier-BSI:
@sophokles73 you wrote that this is an initial step and the next step could be to extract the record processing. Do you still have in mind to implement the PSKStore lookups in an reactive way, as the next next step? Or do you think we can lessen the problem with enough threads to an acceptable behavior?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -275,6 +308,11 @@ private void start(final InetSocketAddress bindAddress) throws IOException {
 		timer = Executors.newSingleThreadScheduledExecutor(
 				new DaemonThreadFactory("DTLS RetransmitTask-", NamedThreadFactory.SCANDIUM_THREAD_GROUP));
 
+		if (executor == null) {
+			// use a thread pool with max 6 * #CPU cores
+			executor = new StripedExecutorService(Runtime.getRuntime().availableProcessors() * 6);
```
DanielMaier-BSI: just for curiosity: where does this metric "6 * cores" come from?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -225,6 +237,27 @@ public DTLSConnector(final DtlsConnectorConfig configuration, final SessionCache
...
+	 * than just the number of CPU cores. This helps with performing multiple handshakes
+	 * in parallel, in particular if the key exchange requires look up of identities, e.g.
+	 * in a database or using a web service.
+	 * <p>
+	 * If this method is used to set an executor, the executor will <em>not</em> be shut down
+	 * by the {@linkplain #stop() stop method}.
+	 * 
+	 * @param executor The executor.
+	 */
+	public void setExecutor(StripedExecutorService executor) {
```
DanielMaier-BSI: Should we prevent setting the executor while the connector is running? Otherwise we get concurrency issues with hasInternalExecutor flag. 

Is it even necessary to have this setter instead of setting it in the constructor? Then we could make the executor field even final. atm we are accessing the executor from different threads without appropriate locking or making it volatile.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: Just an *educated guess* ;-) Should be more than the number of cores because we still expect these threads to block so it should be a *pool* instead of just a *small set*.
Would you propose another default value?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: > Should we prevent setting the executor while the connector is running? Otherwise we get concurrency issues with hasInternalExecutor flag.

definitely, good point :-)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
DanielMaier-BSI: Hard to guess, because blocking time depends e.g. on the PSKStore implementation. I am not sure if we have blocking time when we use the default in memory PSK store.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: > Is it even necessary to have this setter instead of setting it in the constructor? Then we could make the executor field even final. atm we are accessing the executor from different threads without appropriate locking or making it volatile.

I do not want to bloat the constructor too much and given the fact that the connector needs to be started explicitly anyway ...
I was not aware of the fact that we are accessing the executor from multiple threads unsynchronized. Can you point out an example?

DanielMaier-BSI: In receiveNextDatagramFromNetwork() it gets accessed from receiver worker thread. In all other methods it gets accessed by the users thread. There are no race conditions or something like that, it is just about visibility.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: You were the one who required this in the first place because of your *remote DB PSK store*, right? ;-) My concern was more whether you think that the size is too small ...

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: So, do you suspect a problem or not?

DanielMaier-BSI: Yes, I do. I never saw a bug because of such things in the flesh. But you should not access a field from different threads without guarding the access with the same lock or use final fields or make the field volatile. This is also the case for read access. Tell me if you think I am not right.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
DanielMaier-BSI: My thoughts were:  When someone implements a custom PSK store the need to set a custom executor is fine, i.e. we will set a custom executor. So I thought the default configuration should be optimized for default in memory implementation.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
DanielMaier-BSI: Sorry for confusion. I disagree with my own comment. As we guarantee that executor field gets set before receiver worker thread gets started, we do not have any visibility issues with that. Start of a thread refreshes thread local memory.

I am not sure about some other fields, but I think we should postpone this discussion after refactoring of record processing.

sophokles73:
>  Do you still have in mind to implement the PSKStore lookups in an reactive way, as the next next step? Or do you think we can lessen the problem with enough threads to an acceptable behavior?

I think that we (you guys) will need to evaluate whether the striped executor with an adequately sized thread pool already solves your problem. If not, then you will either need to implement a lookup that doesn't take ages or we will need to think about how we could introduce a callback based implementation that does not rely on blocking threads.

For the time being, I will commit this PR so that you can start to evaluate. The numbers provided by @sbernard31 look promising to me, though.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sophokles73: > Assuming an "attack" (simultaneous messages from different addresses for the same DTLS session), the handshake may be executed in parallel from different executors.

Hmm, not sure if this is actually an *attack* or standard functionality of DTLS, i.e. if the client can pass the VERIFY_REQUEST *and* knows the session ID *and* knows the master secret then I guess the clients are all entitled to resume the session.

PR #236, opened 2017-03-08
Re-encrypt AEAD Record when record sequence number is incremented.

sbernard31:
All is in the title.
About test I'm afraid the same criticisms than #227 are available here. This test is really high level and this could be a good thing to add some tests about that in RecordTests, but it's not so easy to do. In particular initialize all the cryptographic stuff.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
@@ -36,7 +36,6 @@ import org.eclipse.californium.elements.util.DatagramReader;
 import org.eclipse.californium.elements.util.DatagramWriter;
 import org.eclipse.californium.scandium.dtls.cipher.CCMBlockCipher;
 import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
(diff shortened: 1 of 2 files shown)


sophokles73:
When I originally wrote the code in `Record` responsible for excluding re-encryption when using AEAD I was certain, that the record's sequence number was not part of the ciphertext being produced. It seems that I have missed that it is part of the `additional_data` being used for authenticating the ciphertext.
Great that you spotted this :-)
I think I will add some unit tests for `Record` to assert proper reencryption as well. Would you like me to merge this PR?

sbernard31:
> I think I will add some unit tests for Record to assert proper reencryption as well.

Great :)

> Would you like me to merge this PR?

I wanted a +1 review before merging. I believe you are ok with this fix, so  I can merge it or you can, I don't care :).
But I'm not sure to understand you want to add you tests in this PR or in another one ?

sophokles73:
Go head with merging. I will add unit test(s) later (no need for a PR, though ;-))

sbernard31:
Integrated in 2.0.x branch (commit ad4102e2174cb67d88b3cdaa50762aae681467d7 and 46e5c3355ce69efabe44014b10b682ffcff60b8a)

PR #237, opened 2017-03-09
Use localhost instead of any.

boaks:
Though the client1 endpoint address is used as destination, any doesn't
work on windows. Replaced by localhost.


sbernard31:
I faced the same kind of problem [here](https://github.com/eclipse/californium/pull/236/commits/58443525fd46f09655b6ac0422298eecff11f577#diff-f64728eeec115eb9a4036fa6b7f9c6a2R658).
So this sounds not totally specific to windows. I didn't really understand when this causes a problem.

Anyway, it's ok for me to use loopback address.

boaks:
As far as I understand it:
Listen on <any> is OK, but if we then use that <any> address as destination, it doesn't work.
So we either use <any> for the endpoint and <localhost>:<endpoint.port> as destination or <localhost> for both.
If you aggree, I would just merge this small fix.

boaks:

sbernard31:
+1 for me.

Issue #238, opened 2017-03-09
Introducing "in process" datagram messaging


boaks:
Hi all,

I would like to replace the DatagramSocketImpl used for testing to omit potential "package" loss using the real stack. It's not too much code, but I'm unsure, if this would require a new sub project.

Placing it in element-connector would mean, that this is available also for none testing stuff.
Placing it in element-connector test would mean, that this is not available outside the element-connector.
Placing it in a new subproject, I'm not sure if it's worth to generate such a new sub project.

So, any ideas where to place it?

sophokles73:
I wouldn't have a problem with putting it into `element-connector`. Others might find it useful to use in their experiments as well. It doesn't do any harm, does it?

boaks:
`DirectDatagramSocketImpl.initialize` must call `DatagramSocket.setDatagramSocketImplFactory()`, which is nonreversible until you restart the VM. But you have to call `DirectDatagramSocketImpl.initialize` otherwise it doesn't anything and so doesn't harm.

boaks:
After some investigations, I found, that exporting the test classes with

			<plugin>
			  <groupId>org.apache.maven.plugins</groupId>
			  <artifactId>maven-jar-plugin</artifactId>
			  <executions>
			    <execution>
			      <goals>
			         <goal>test-jar</goal>
			      </goals>
			    </execution>
		  	  </executions>
			</plugin>

and the using them with

            <dependency>
                <groupId>${project.groupId}</groupId>
                <artifactId>element-connector</artifactId>
                <version>${project.version}</version>
                <scope>test</scope>
                <classifier>tests</classifier>
            </dependency>

would do it without a extra subproject.

sophokles73:
@boaks 

can this be closed now?

boaks:
sure

PR #241, opened 2017-03-09
Introduce direct datagram socket

boaks:
Introduce DirectDatagramSocketImpl and add network handling in
TestTools. Use "in process" datagramm passing to omit potential "message loss"
using the UPD stack.
Second commit adjust most tests (except those, that need additional changes). 

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/rule/CoapNetworkRule.java
+++ b/californium-core/src/test/java/org/eclipse/californium/rule/CoapNetworkRule.java
@@ -1,0 +1,198 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations - initial implementation
+ ******************************************************************************/
+package org.eclipse.californium.rule;
+
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Logger;
+
+import org.eclipse.californium.core.coap.Message;
+import org.eclipse.californium.core.network.EndpointManager;
+import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.network.serialization.DataParser;
+import org.eclipse.californium.core.network.serialization.UdpDataParser;
+import org.eclipse.californium.elements.rule.NetworkRule;
+import org.eclipse.californium.elements.util.DatagramFormatter;
+
+/**
+ * CoAP network rules for junit test using datagram sockets.
+ * 
+ * Though {@link NetworkConfig} and {@link EndpointManager} depend on some
+ * internal state, this rule manages these states by setup and cleanup the
+ * values. Therefore it's intended, that test code uses the provided methods to
+ * access {@link NetworkConfig}, {@link #getStandardTestConfig()},
+ * {@link #createStandardTestConfig()}, and {@link #createTestConfig()}.
+ *
+ * The rule is intended to be mainly used as <code>&#64;ClassRule<code>
+ * 
+ * <pre>
+ * public class AbcNetworkTest {
+ *    &#64;ClassRule
+ *    public static CoapNetworkRule network = new CoapNetworkRule(Mode.DIRECT, Mode.NATIVE);
+ *    ...
+ *    &#64;BeforeClass
+ *    public static void init() {
+ *       network.getStandardTestConfig().setInt(NetworkConfig.Keys.MAX_MESSAGE_SIZE, 32);
+ *    ...
+ * </pre>
+ * 
+ * The {@link NetworkConfig} access methods then are valid from the
+ * <code>&#64;BeforeClass</code> until <code>&#64;AfterClass</code> method.
+ * 
+ * If used as
+ * <code>&#64;Rule<code>, the {@link NetworkConfig} access methods then are valid from the
+ * <code>&#64;Before</code> until <code>&#64;After</code> method.
+ * 
+ * For the DIRECT mode there are additional parameters available
+ * {@link #setMessageThreads(int)}, and {@link #setDelay(int)}.
+ * 
+ * In rare cases nested rules are allowed, but be careful when accessing the
+ * {@link NetworkConfig} to choose the active rule. The inner will overwrite the
+ * outer {@link NetworkConfig} and detach from that.
+
+ * <pre>
+ * public class AbcNetworkTest {
+ *    &#64;ClassRule
+ *    public static CoapNetworkRule network = new CoapNetworkRule(Mode.DIRECT, Mode.NATIVE);
+ *    &#64;Rule
+ *    public static CoapNetworkRule inner = new CoapNetworkRule(Mode.DIRECT, Mode.NATIVE);
+ *    ...
+ *    &#64;BeforeClass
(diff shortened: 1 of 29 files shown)


boaks:
Depending on #238 the DirectDatagramSocketImpl may be moved into an additional sub-project.

sophokles73:
It seems that I have totally misunderstood your intention with #238. I was under the impression that we would want to implement a `Connector` that could be wired up with another one to exchange `RawData` instances directly, i.e. without the need to go through the network stack. FMPOV your approach would work but goes way beyond what I think we need...

boaks:
My first variant was using such a a special Endpoint. But that introduced some "tes only dependencies" in CoapEndpoint and the "org.eclipse.californium.core.test.maninmiddle" stuff will require even a special DatagramSocket. Therefore I investigated in the DatagramSocketImpl. And it seems to fit perfect. It just needs a intialisation and after that ANY UDP traffic will use it. The only pain is, once set, it can't be removed.

boaks:
One small plus:
If something is already running on the machine and unfortunately using a UDP port, which is also "fix-used" in a test, this PR would allow to execute the tests, the other variant would fail (as the current implementation also does).

boaks:
Redesigned using rules and category.
I'm still convinced, that using the `DatagramSocket.setDatagramSocketImplFactory()` is the right approach. Any other would require at least, that a factory for DatagramSocket is used.

The limitations in the usage are unfortunatly from the design of that `setDatagramSocketImplFactory`, which is not intended to support changing the factory mroe then once. And the factory is not able to construct the OS DatagramSocketImpl. So it must be decided before running the test, if they should be executed with the OS DatagramSocketImpl or the Direct DatagramSocketImpl. For maven the introduced category could be used (even if right now, no test requires to be run with OS DatagramSocketImpl). If the test are run from eclipse, `-Dorg.eclipse.californium.junit.socketmode=???` (OS or DIRECT) must be used.

If this PR is accepted, I would check, if the element-connector and/or scandium test could be migrated also.

boaks:
[https://hudson.eclipse.org/californium/job/master/169/testReport/](url)

I'm looking forward to execute the test with the `DirectDatagramSocketImpl` :-).

boaks:
For the mission "green is always green", would it be possible to review this PR?

californium-core/src/test/java/org/eclipse/californium/rule/CoapNetworkRule.java
```diff
@@ -0,0 +1,144 @@
...
+
+	/**
+	 * Get standard network configuration for testing.
+	 * 
+	 * @return standard network configurations. Changes are visible to other
+	 *         usage of the standard configuration.
+	 * @see NetworkConfig#getStandard()
+	 */
+	public NetworkConfig getStandardTestConfig() {
+		return NetworkConfig.getStandard();
```
boaks: The intention is, that this `Rule` prepares the `NetworkConfig.getStandard()` within its scope according the setup of this `Rule` without using a file (see `createStandardTestConfig`). But yes, if used wrong, it may not work as intended. I think adjusting the documentation and make the implementation more safe against miss use should be possible.

californium-core/src/test/java/org/eclipse/californium/rule/CoapNetworkRule.java
```diff
@@ -0,0 +1,144 @@
...
+
+	/**
+	 * Create new network configuration for testing.
+	 * 
+	 * @return network configurations. Detached from the standard network
+	 *         configuration.
+	 */
+	public NetworkConfig createTestConfig() {
+		int threads = messageThreads.get();
+		NetworkConfig config = new NetworkConfig();
```
boaks: Looking on the test, I was not sure, if all test are using the `NetworkConfig` in a "proper intended way" or more "in random way". 
Therefore my intention is to move the access of `NetworkConfig` to `CoapNetworkRule` to put that configuration under the "scope control" of the rule and to make it easier, the search for "out of scope usage". 
So three methods are introduced with (in my opinion) proper documentation about how to use it and explicitly mentioning the side effect ("Changes are visible to other usage of the standard configuration").
So, if currently `createTestConfig` is not used, this is not really "by design", it's more just as its used now. I just want to make sure, that nobody uses `NetworkConfig` directly.

boaks:
> protected DatagramSocket(DatagramSocketImpl impl)

has `protected` scope, so I'm not sure, if this is intended to be used in general for providing a special impl. Using that requires the introduction of an additional `DatagramSocket` factory (or `Connector` factory), which then must be consequently used. That's possible but come with more changes and would not have the "general" effect, which I would prefer.

In my opinion, it should be possible, that:
- each test/testsuite defines, if it's possible to run on OS or DIRECT stack (most would run on both, but running on the OS stack will come with the potential errors). (covered by the introduced `Rule`)
- decide, which implementation is used when running the test (covered by the property "org.eclipse.californium.junit.socketmode")
- select the tests to run depending on OS/DIRECT, so that the tests are not unintended executed multiple times. (covered by the introduced `Category`)
This PR covers these requirements.

I'm not sure, if it's also a requirement, that a test "actively selects" the stack implementation (OS or DIRECT). After using the DIRECT for a couple of days (and with the "LoopbackAddress" for tests in mind) I don't see the advantage/need of that "active select". So I would prefer to use that factory and try, if we/someone has trouble with that. It's very easy to disable the DIRECT feature (and by default it's currently not activated for running the test locally within Eclipse) so on my opinion the risc is very low.

sophokles73:
Oh, you're right. I hadn't noticed the `protected`.

element-connector/src/test/java/org/eclipse/californium/elements/category/OsDatagramSocketImpl.java
```diff
@@ -0,0 +1,26 @@
...
+package org.eclipse.californium.elements.category;
+
+/**
+ * A marker interface for test cases that requires the OS DatagamSocketImpl for
+ * real network operations.
+ * 
+ * Be careful when design such test, so that dropping and reordering of messages
+ * doesn't break your test!
+ */
+public interface OsDatagramSocketImpl {
```
sophokles73: Can we rename this to e.g. `NativeDatagramSocketOnly` or `RequireNativeDatagramSocket`?

boaks: Sure, we can use another name.
I would prefer a name aligned with the naming in `NetworkRule.Mode`.
So if "Native" should be used in the `Category`, we should use it in `Mode` also.

sophokles73: Then let's do that, if you don't mind :-)

sophokles73: Are you updating this? I think we could merge it then :-)

boaks: Yes, I'm working on this. I hope the example snippet will be finished soon.

boaks:
Renaming and docu is done.
Should/could I squash , rebase and merge?

element-connector/src/test/java/org/eclipse/californium/elements/category/NativeDatagramSocketImplRequired.java
```diff
@@ -0,0 +1,26 @@
...
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations GmbH - Initial implementation
+ ******************************************************************************/
+package org.eclipse.californium.elements.category;
+
+/**
+ * A marker interface for test cases that requires the OS DatagamSocketImpl for
```
boaks: OK, I correct the OS usage to native and squash.

boaks:
hudson explore works, so this PR would not break the tests :-).

I would like to rebase/merge this tomorrow, any objections?

sophokles73:
Are you merging this to master and 1.0.x as well?

boaks:
That's my plan (otherwise "green is always green" would not come true).

But after PR #254 :-)

PR #243, opened 2017-03-10
Add StringUtil.

boaks:
Currently provides only "hex2CharArray", which will then be used by
SslContextUtil. Intended to be extended in the future.


sophokles73:
@boaks 

FMPOV there is nothing to be reviewed here. I would be ok if you just committed such small changes without the hassle of creating a PR for it ...

PR #245, opened 2017-03-22
Fix broken test depending on MTU.

boaks:
The "testReceivingMessagesInBadOrderDuringHandshake" depends on the MTU,
because it expects record and not flights. Running the test on windows
set the MTU to default of 1280 and so the second flight is split into
two records which causes the test to fail.
This fix will therefore wait for the flight (using a timeout) and so
overcome the MTU dependency.


boaks:
#246 

An alternative approach to fix this windows localhost MTU issue may be to adjust the MTU detection and so prevent scandium from split the flight.

sophokles73:
LGTM

Waiting for the flight to arrive to me seems the more robust approach instead of *hoping*  that the MTU will be large enough that all messages of the flight fit into one datagram.

Issue #226, opened 2017-02-27
NPE on request.cancel()


sbernard31:
I faced some NPE, when I cancel a request.
```
java.lang.NullPointerException: token bytes must not be null
	at org.eclipse.californium.core.network.Exchange$KeyToken.<init>(Exchange.java:637)
	at org.eclipse.californium.core.network.Exchange$KeyToken.fromOutboundMessage(Exchange.java:670)
	at org.eclipse.californium.core.network.UdpMatcher$ExchangeObserverImpl.completed(UdpMatcher.java:315)
	at org.eclipse.californium.core.network.Exchange.setComplete(Exchange.java:435)
	at org.eclipse.californium.core.network.stack.ExchangeCleanupLayer$CancelledMessageObserver.onCancel(ExchangeCleanupLayer.java:45)
	at org.eclipse.californium.core.coap.Message.setCanceled(Message.java:579)
	at org.eclipse.californium.core.coap.Request.setCanceled(Request.java:616)
	at org.eclipse.californium.core.coap.Message.cancel(Message.java:658)
```
I'm not sure but I suppose this could happen when request is canceled before token was assigned.

See the stack
```
 * +--------------------------+
 * | {@link MessageDeliverer}         |
 * +--------------A-----------+
 *                A
 *              * A
 * +------------+-A-----------+
 * |       CoAPEndpoint       |
 * |            v A           |
 * |            v A           |
 * | +----------v-+---------+ |
 * | | Stack Top            | |
 * | +----------------------+ |
 * | | {@link ExchangeCleanupLayer} | |
 * | +----------------------+ |
 * | | {@link ObserveLayer}         | |
 * | +----------------------+ |
 * | | {@link BlockwiseLayer}       | |
 * | +----------------------+ |
 * | | {@link ReliabilityLayer}     | |
 * | +----------------------+ |
 * | | Stack Bottom         | |
 * | +----------+-A---------+ |
 * |            v A           |
 * |          Matcher         |
 * |            v A           |
 * |        Interceptor       |
 * |            v A           |
 * +------------v-A-----------+
 *              v A 
 *              v A 
 * +------------v-+-----------+
 * | {@link Connector}                |
 * +--------------------------+
```
The `ExchangeCleanupLayer` add listener on `cancel` (which will call `exchange.complete`).
The `matcher` add listener on `exchange.complete` then attribute a token to the request.

More than the NPE, I suspect there is some scenario where exchangeStore will not be cleaned and token not release.

[...]

PR #247, opened 2017-03-22
[#226] Do not register ExchangeObserver on unsent requests.

sophokles73:
Made sure that the Matchers register their ExchangeObservers only on
outbound Exchanges that have been registered with the
MessageExchangeStore successfully.

When canceling a request that has not been registered successfully, e.g.
because there are no MIDs left to be used, there is no ExchangeObserver
to be invoked at all.

WDYT, @sbernard31 ?


boaks:
These are some changes on the same classes as #231.

So I would like to rebase/merge 231 first. Or do you think 231 is not the right way?
Just if 231 should not be rebased/merged, could you please extend your changes with
the "save remove" from 231 (add the `Exchange` to the remove methods)?

sophokles73:
I have rebased this PR on master so that it now includes @boaks changes from #231

boaks:
I think, we discussed enough about the Exchange complexity and the far future strategies to overcome that. So LGTM!

Issue #248, opened 2017-03-22
Provide proper OSGi meta-data for all modules


fipro78:
I noticed that Californium is inconsistent in providing OSGi meta-data.
Some bundles provide proper OSGi meta-data, some not. IMHO all provided bundles should provide proper OSGi meta-data.

californium-core-2.0.0-M3 – no OSGi meta-data
californium-proxy-2.0.0-M3 – no OSGi meta-data
element-connector-2.0.0-M3 –incomplete OSGi meta-data provided
scandium-2.0.0-M3 – OSGi meta-data provided

sophokles73:
@fipro78 

is this fixed by your PR? If so, could you please close the issue?

fipro78:
for 2.0.0 it can be closed. How to proceed with the other branches? master and 1.0.x

sophokles73:
It would be great if you could provide PRs for them as well, of course :-)

boaks:
Could you check, if this issues could be closed after mergin your PRs for 1.0 and master branch?

fipro78:
After merging the three PRs for the different branches I think this can be closed now. Further work on the OSGi bundle should be another issue and PR.

PR #249, opened 2017-03-22
Provide proper OSGi meta-data for all modules #248

fipro78:
Signed-off-by: Dirk Fauth <dirk.fauth@googlemail.com>

### DIFF
--- a/californium-core/pom.xml
+++ b/californium-core/pom.xml
@@ -61,7 +61,7 @@
 						<Import-Package>
 							*
 						</Import-Package>
-						<Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+						<Bundle-SymbolicName>${project.groupId}.core</Bundle-SymbolicName>
 					</instructions>
 				</configuration>
 			</plugin>
--- a/californium-core/pom.xml
+++ b/californium-core/pom.xml
@@ -11,7 +11,7 @@
 		<version>2.0.0-SNAPSHOT</version>
 	</parent>
 	<artifactId>californium-core</artifactId>
-	<packaging>jar</packaging>
+	<packaging>bundle</packaging>
 
 	<name>Californium (Cf) Core</name>
 	<description>CoAP library module</description>
@@ -36,6 +36,35 @@
 
 	<build>
 		<plugins>
+			<plugin>
+				<groupId>org.apache.felix</groupId>
+				<artifactId>maven-bundle-plugin</artifactId>
+				<extensions>true</extensions>
+				<configuration>
+					<instructions>
+						<Export-Package>
+							org.eclipse.californium.core,
+							org.eclipse.californium.core.coap,
+							org.eclipse.californium.core.network,
+							org.eclipse.californium.core.network.config,
+							org.eclipse.californium.core.network.interceptors,
+							org.eclipse.californium.core.network.stack,
+							org.eclipse.californium.core.observe,
+							org.eclipse.californium.core.server,
+							org.eclipse.californium.core.server.resources
+						</Export-Package>
+						<Private-Package>
+							org.eclipse.californium.core.network.deduplication,
+							org.eclipse.californium.core.network.serialization,
+							org.eclipse.californium.core.network.stack.*
+						</Private-Package>
+						<Import-Package>
+							*
+						</Import-Package>
+						<Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+					</instructions>
+				</configuration>
+			</plugin>
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId>
--- a/californium-proxy/pom.xml
+++ b/californium-proxy/pom.xml
@@ -11,7 +11,7 @@
 		<version>2.0.0-SNAPSHOT</version>
 	</parent>
 	<artifactId>californium-proxy</artifactId>
-	<packaging>jar</packaging>
+	<packaging>bundle</packaging>
 
 	<name>Californium (Cf) Proxy</name>
 	<description>Cross-proxy module</description>
@@ -46,4 +46,26 @@
 			<version>15.0</version>
 		</dependency>
 	</dependencies>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.felix</groupId>
+				<artifactId>maven-bundle-plugin</artifactId>
+				<extensions>true</extensions>
+				<configuration>
+					<instructions>
+						<Export-Package>
+							org.eclipse.californium.proxy,
+							org.eclipse.californium.proxy.resources
+						</Export-Package>
+						<Import-Package>
+							*
+						</Import-Package>
+						<Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+					</instructions>
+				</configuration>
+			</plugin>
+		</plugins>
+	</build>
 </project>
--- a/californium-proxy/pom.xml
+++ b/californium-proxy/pom.xml
@@ -16,6 +16,12 @@
 	<name>Californium (Cf) Proxy</name>
 	<description>Cross-proxy module</description>
 
+	<properties>
+		<httpcomponents.version>4.2.5</httpcomponents.version>
+		<httpcomponents.version.lowerbound>4.2</httpcomponents.version.lowerbound>
+		<httpcomponents.version.upperbound>5</httpcomponents.version.upperbound>
+	</properties>
+
 	<dependencies>
 		<dependency>
 			<groupId>${project.groupId}</groupId>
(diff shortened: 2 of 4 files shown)


californium-core/pom.xml
```diff
@@ -37,6 +37,35 @@
 	<build>
 		<plugins>
 			<plugin>
+				<groupId>org.apache.felix</groupId>
+				<artifactId>maven-bundle-plugin</artifactId>
+				<extensions>true</extensions>
+				<configuration>
+					<instructions>
+						<Export-Package>
+							org.eclipse.californium.core,
```
sophokles73: Do these exports correspond with the californium-osgi bundle?
Speaking of which, what is the relation between californium-osgi and californium-core after this change?

fipro78: These are the same exports as used in the californium-osgi bundle. It doesn't introduce a change in the relation between californium-osgi and californium-core.

Basically it is the pre-requisite to make californium-osgi a better OSGi citizen. Currently it includes californium-core and element-connector, which is not a good OSGi practice. If both bundles provide proper OSGi meta-data, they don't need to be wrapped anymore but can be simply dependent, which means you can simple have all three bundles in the runtime.

sophokles73: Ok, I guess a next step would then be to remove the `<Embed-Dependency>` tag from californium-osgi's pom file, right? Or should this be included here already?

fipro78: Yes that would be a next step. And no, I would not introduce this with this PR.

This PR is backwards compatible, only adds proper OSGi meta-data in all modules and does not break anything. The change in californium-osgi would be a breaking change for current consumers, because they suddenly need to add two additional bundles to their runtime.

I therefore suggest to make that change in another PR. Actually I would even introduce an even extended approach to change californium-osgi. But I'm still thinking about how this can be done and included in the build process.

sophokles73: ok, understood

sophokles73: When I build e.g. the californium-proxy module, then the resulting jar's manifest contains the following:

```
Manifest-Version: 1.0
Bnd-LastModified: 1490261180536
Build-Jdk: 1.8.0_121
Built-By: hak8fe
Bundle-Description: Cross-proxy module
Bundle-DocURL: http://www.eclipse.org/
Bundle-License: http://www.eclipse.org/org/documents/epl-v10.php, http:/
 /www.eclipse.org/org/documents/edl-v10.php
Bundle-ManifestVersion: 2
Bundle-Name: Californium (Cf) Proxy
Bundle-SymbolicName: californium-proxy
Bundle-Vendor: Eclipse Foundation
Bundle-Version: 2.0.0.SNAPSHOT
Created-By: Apache Maven Bundle Plugin
Export-Package: org.eclipse.californium.proxy;uses:="org.apache.http,org
 .eclipse.californium.core,org.eclipse.californium.core.coap,org.eclipse
 .californium.core.network,org.eclipse.californium.proxy.resources";vers
 ion="2.0.0",org.eclipse.californium.proxy.resources;uses:="com.google.c
 ommon.cache,org.eclipse.californium.core,org.eclipse.californium.core.c
 oap,org.eclipse.californium.core.network,org.eclipse.californium.core.s
 erver.resources";version="2.0.0"
Import-Package: com.google.common.cache;version="[15.0,16)",com.google.c
 ommon.collect;version="[15.0,16)",com.google.common.io;version="[15.0,1
 6)",com.google.common.primitives;version="[15.0,16)",org.apache.http,or
 g.apache.http.client,org.apache.http.client.protocol,org.apache.http.co
 nn,org.apache.http.entity,org.apache.http.impl,org.apache.http.impl.cli
 ent,org.apache.http.impl.nio,org.apache.http.impl.nio.reactor,org.apach
 e.http.message,org.apache.http.nio,org.apache.http.nio.protocol,org.apa
 che.http.nio.reactor,org.apache.http.params,org.apache.http.protocol,or
 g.apache.http.util,org.eclipse.californium.core;version="[2.0,3)",org.e
 clipse.californium.core.coap;version="[2.0,3)",org.eclipse.californium.
 core.network;version="[2.0,3)",org.eclipse.californium.core.network.con
 fig;version="[2.0,3)",org.eclipse.californium.core.server.resources;ver
 sion="[2.0,3)",org.eclipse.californium.proxy,org.eclipse.californium.pr
 oxy.resources
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.7))"
Tool: Bnd-3.0.0.201509101326
```
The imports for the Apache HttpComponents do not have any version information attached which will probably be a problem during resolution on a runtime, wouldn't it? In addition to that the HttpComponents artifacts do not seem to be OSGi bundles themselves, at least they are lacking the corresponding meta information in their manifests ...

fipro78: Yes and no. 

It is correct that the Apache HttpComponents used in the build are not provided as OSGi bundles directly. But there is also an OSGi version available (I really don't understand why there are two different versions). Well at least httpcore and httpclient are provided as OSGi bundles. http-nio seems not to be available as OSGi version. That means users that want to use californium-proxy will have an issue in an OSGi runtime because of the missing http-nio bundle. They will need to create an OSGi-ified version themselves. IMHO not a good practice by Apache HttpComponents that need to be discussed with the HttpComponents team.

Nevertheless the used dependency from Maven does not contain OSGi meta-data as you already noticed. We could either change the dependency to the OSGi versions of that bundles (I really have no clue why they have two versions) or keep it that way.

Keeping it that way simply means we have no lower version bound. So it is only checked if that package is provided by anybody. IMHO better than nothing. We just can't tell at runtime which bundle is wired if multiple versions are available in the runtime, and we can't specify the version we need at minimum.

sophokles73: > Nevertheless the used dependency from Maven does not contain OSGi meta-data as you already noticed. We could either change the dependency to the OSGi versions of that bundles (I really have no clue why they have two versions) or keep it that way.

Do you know the maven coordinates of these?

IMHO we definitely need to have the minimum version in the imports. So, I'd rather try to use OSGi-fied versions of the http client jars but as a fall back we could also add that information manually to the Felix plugin configuration, couldn't we?

fipro78: for httpclient it should be httpclient-osgi ... for http-nio there is no OSGi version available.

To be honest I don't know if we could specify the import versions manually. Could be, would need to check the felix bundle documentation. On the other hand the question is if we are able to know the version exported, as it doesn't have to be the bundle version.

sophokles73: I have taken a look at httpclient-osgi's content. It seems like they are using the same approach as we had taken for californium-osgi, i.e. they inline all content from httpclient and add some OSGi specific classes which introduce a dependency on the OSGi artifacts :-( So, it seems like we cannot depend on these "*-osgi" artifacts without introducing dependencies on OSGi itself ...

fipro78: OK, probably that is the reason for the OSGi-ified versions in Eclipse Orbit. But I think Californium can not rely on these artifacts. At least I am not sure if they are provided in a Maven Repository.

sophokles73: They are available from maven central but again, IMHO we should not declare those as dependencies because they would create a transitive dependency on OSGi artifacts. So, IMHO we (i.e. you :-)) need to refrain to setting the minimum version information explicitly in the maven bundle plugin config ...

fipro78: I think there is a misunderstanding. In Eclipse Orbit you can find the base distributions (httpclient and httpcomponents) with the OSGi meta-data added by someone who needed these base bundles for an Eclipse plug-in. 
Nevertheless I understand that the risk for incorrect transitive dependencies because of the Maven dependency resolution would be a risk for other projects. I also noticed that element-connector has a similar issue because netty-all does not contain OSGi meta-data although all single netty JARs contain the meta-data. 
I also changed the Bundle-SymbolicName to follow the OSGi best practices.

Let me know what you think.

sophokles73: We cannot depend on artifacts that are only available from the Eclipse repo. All our dependencies must be available from Maven Central.
Regarding netty, we could replace the dependency on netty-all with dependencies for the individual artifacts.

fipro78: I also thought about replacing netty-all with the required individual artifacts. But I was not sure about the changes caused for the current consumers, as it reduces the dependencies (IMHO something good) but also means consumers would need to change their pom.xml files in case they need the whole netty-all and until now relied on getting it transitively.

sophokles73: With the approach taken by you now, i.e. manually defining the version boundaries, we do not need to do this anyway, right? It is up to the user to provide an implementation of netty as he sees fit.

fipro78: yes with this approach actually the same meta-data is generated as with using the individual artifacts. it just means that if the californium project decides to update the dependencies, the boundaries need to updated accordingly, which would be done automatically with the individual bundles

sophokles73: understood

PR #250, opened 2017-03-23
Use LoopbackAddress instead of LocalHost.

boaks:
LoopbackAddress is used in 2.0.x succesfull. The LocalHost is mapped to
the ip-address of the host, which may be blocked by a firewall. 


boaks:
Only 3 symbolic names change, and already used in main/2.0.x.
So I merge it without review :-).

PR #251, opened 2017-03-23
Use "localhost" as destination instead of "any.

boaks:
Use "localhost/127.0.0.1" as destination instead of "any/0.0.0.0".
Use the ConnectorTestUtil for this replacment and unified message
creation.

Fixes the broken 2.0.x builds on hudson.

### DIFF
--- a/element-connector/src/test/java/org/eclipse/californium/elements/tcp/TcpConnectorTest.java
+++ b/element-connector/src/test/java/org/eclipse/californium/elements/tcp/TcpConnectorTest.java
@@ -11,22 +11,19 @@
  * http://www.eclipse.org/org/documents/edl-v10.html.
  * <p>
  * Contributors:
- * Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
+ *    Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - use ConnectorTestUtil
  ******************************************************************************/
 package org.eclipse.californium.elements.tcp;
 
 import static org.hamcrest.CoreMatchers.*;
 import static org.junit.Assert.*;
+import static org.eclipse.californium.elements.tcp.ConnectorTestUtil.*;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
 import java.net.InetSocketAddress;
-import java.net.ServerSocket;
-import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Random;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.californium.elements.Connector;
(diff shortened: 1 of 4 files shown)


sophokles73:
IMHO we should make sure that we always use the loopback device for running tests. Wouldn't that solve our problems?

boaks:
Yes, sure.

And for UDP #241 would solve this for almost all tests (except those who need the OS IP-stack) ;-).

element-connector/src/test/java/org/eclipse/californium/elements/tcp/TcpConnectorTest.java
```diff
@@ -96,25 +91,24 @@ public void serverClientPingPong() throws Exception {
...
 		clientCatcher.blockUntilSize(1);
 		assertArrayEquals(msg.getBytes(), clientCatcher.getMessage(0).getBytes());
 	}
 
 	@Test
 	public void singleServerManyClients() throws Exception {
-		int port = findEphemeralPort();
 		int clients = 100;
-		TcpServerConnector server = new TcpServerConnector(new InetSocketAddress(port), NUMBER_OF_THREADS,
+		TcpServerConnector server = new TcpServerConnector(new InetSocketAddress(0), NUMBER_OF_THREADS,
```
sophokles73: shouldn't this be `new InetSocketAddress(InetAddress.getLoopbackAddress(), 0)` now?

I think there are several more occurrences ...

boaks: The server could bind to "any", that's no problem.
But the destination for the message must then be either "localhost" or the "hostname" (which seems not to work for hudson). So I adjust the destination in the ConnectorTestUtil.

boaks:
The other tests (UDP) mainly use explizitly the "LoopbackAddress" for the server.
I can change this, but as said above, it's only required for the destination.

sophokles73:
> The other tests (UDP) mainly use explizitly the "LoopbackAddress" for the server.

I do not understand what this is supposed to mean.

boaks:
This PR fixes the issue, that sending a message to "any/0.0.0.0" seems not to work any longer on hudson. Therefore the destination (if "any/0.0.0.0") is replaced by "localhost/127.0.0.1". this PR doesn't change the address used by the server, because that didn't cause the test to fail.

In your remarks:

> shouldn't this be new InetSocketAddress(InetAddress.getLoopbackAddress(), 0) now?

I recognized that your idea of fixing it, is to change the server address to "localhost/127.0.0.1" instead of replacing only the destination.

Therefore I checked, how the other tests are doing it an found, that they "mainly" use the "localhost/127.0.0.1" as server address. So if, wanted, I can change the TCP tests accordingly. 
 
But in the spirit to change/fix only was broken, the replace of the destination seems for me to be the right approach.

sophokles73:
Understood. LGTM

boaks:
Just to say: if we want to use LoopbackAddress, there is no objection from my side. It may be easier to understand. If so, then I will just provide an other PR.

sophokles73:
+1 for using LoopbackAddress consistently across all tests

PR #252, opened 2017-03-24
Use LoopbackAddress ("localhost/127.0.0.1") for server.

boaks:
Make test simpler by omit "any/0.0.0.0" address for server.


sophokles73:
LGTM
don't forget to also apply to master, 1.0.x if feasible :-)

PR #254, opened 2017-03-27
Add simple file server to demo-apps.

boaks:
Mainly intended for demonstrating LWM2M firmware update.

### DIFF
--- a/demo-apps/cf-simplefile-server/src/main/java/org/eclipse/californium/examples/SimpleFileServer.java
+++ b/demo-apps/cf-simplefile-server/src/main/java/org/eclipse/californium/examples/SimpleFileServer.java
@@ -1,0 +1,253 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial creation
+ *                                      derived from HelloWorldServer example
+ ******************************************************************************/
+package org.eclipse.californium.examples;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.SocketException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.eclipse.californium.core.CoapResource;
+import org.eclipse.californium.core.CoapServer;
+import org.eclipse.californium.core.coap.CoAP;
+import org.eclipse.californium.core.coap.MediaTypeRegistry;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.coap.Response;
+import org.eclipse.californium.core.network.CoapEndpoint;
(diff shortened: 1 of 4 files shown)


demo-apps/cf-simplefile-server/src/main/java/org/eclipse/californium/examples/SimpleFileServer.java
```diff
@@ -0,0 +1,202 @@
...
+					response.getOptions().setContentFormat(accept);
+					exchange.respond(response);
+				} else {
+					LOG.log(Level.WARNING, "File {0} could not be read in!", file.getAbsolutePath());
+					exchange.respond(CoAP.ResponseCode.INTERNAL_SERVER_ERROR);
+				}
+			} catch (IOException ex) {
+				LOG.log(Level.WARNING, "File {0}: {1}", new Object[] { file.getAbsolutePath(), ex });
+				exchange.respond(CoAP.ResponseCode.INTERNAL_SERVER_ERROR);
+			} catch (OutOfMemoryError error) {
```
sophokles73: When do we expect this to happen?

demo-apps/cf-simplefile-server/src/main/java/org/eclipse/californium/examples/SimpleFileServer.java
```diff
@@ -0,0 +1,202 @@
...
+		}
+	}
+
+	/*
+	 * Constructor for a new simple file server. Here, the resources of the
+	 * server are initialized.
+	 */
+	public SimpleFileServer() throws SocketException {
+
+		add(new FileResource("data", new File("data")));
```
sophokles73: should probably use a system property or command line arg for the root path, shouldn't it?

boaks: Command line arguments (and defaults) are a possibility.
But though I'm using a similar approach now for months, nobody asked me to change the file nor coap  path. Therefore to keep it really simple, I compiled in the values.

demo-apps/cf-simplefile-server/src/main/java/org/eclipse/californium/examples/SimpleFileServer.java
boaks: IOException of OutOfMemoryError?
It's easy, I just want to have some logging and defined responses in the "exceptional" cases even if they never occur. But no problem we can remove one or both

demo-apps/cf-simplefile-server/src/main/java/org/eclipse/californium/examples/SimpleFileServer.java
sophokles73: I know, but the demo apps are intended to be used from the command line (they all end up a single jar file apps in the *run* folder) and therefore IMHO they should support at least the bare minimum config without requiring users to recompile, don't you think?

demo-apps/cf-simplefile-server/src/main/java/org/eclipse/californium/examples/SimpleFileServer.java
sophokles73: OutOfMemory. Have you made any good experience with catching it? My assumption would be that the system will already have become unstable when it occurs so there is not much point in catching it, I guess.

demo-apps/cf-simplefile-server/src/main/java/org/eclipse/californium/examples/SimpleFileServer.java
boaks: OK, I add command line argument.

demo-apps/cf-simplefile-server/src/main/java/org/eclipse/californium/examples/SimpleFileServer.java
boaks: OutOfMemory. Sure. In cases where "one big blob" is allocated, there is a small chance :-)

> byte[] content = new byte[(int) length]; 

But to keep it simple, I remove the OutOfMemory catch :-)

PR #259, opened 2017-03-30
Add conditional logging to only on failure.

boaks:
I'm not sure, if this is really the right approach. In some cases the timing would also be useful to analyse the failure.


boaks:
No, I don't think, saving some logging lines will be worth adding complexity :-).

PR #263, opened 2017-03-31
Introduce thread safe values passing.

boaks:
Relax retransmission timing.
The failure may be caused by a "real" retransmission instead of an one
with updated payload.


boaks:
https://hudson.eclipse.org/californium/job/1.0.x/26/

Also caused by the timing issue. So I give this fix a try :-)

Issue #256, opened 2017-03-29
Multithread safe implementation of Request/Response/Message


boaks:
The failed "2.0.x-nightly #48" may have other reasons, but I don't see, that the `Request`/`Response`/`Message` is implemented really "thread safe". Assuming, that on thread is waiting for the response and a other thread is reporting the response using Request.setResponse(Response), the access to `lock` is asymmetric!

Write is synchronized:

	public Response waitForResponse(long timeout) throws InterruptedException {
                ...
		synchronized (this) {
			if (lock == null) {
				lock = new Object();
			}
		}

Read, not:

	public void setResponse(Response response) {
		this.response = response;

		// only for synchronous/blocking requests
		if (lock != null) {
			synchronized (lock) {
				lock.notifyAll();
			}
		}

So, if `setResponse` doesn't see the current `lock`, it doesn't synchronize and therefore the `response` (and other state as `acknowledged`) may not be visible the other thread.

Or just assume, the `response` arrives before the wait creates the `lock`, I don't see how the states a synced between the threads.

boaks:
I would go for an simple synchronized approach for such state without layz lock.

boaks:
Done with #264

PR #264, opened 2017-03-31
Fix 2.0.x nightly 48

boaks:
#256 
Adjust thread safe access of Message/Response state.
(Apply code formatter) 

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
@@ -23,6 +23,7 @@
  *                                                    Message.getPayloadTracingString(). 
  *                                                    (for message tracing)
  *                                                    set scheme on setOptions(URI)
+ *    Achim Kraus (Bosch Software Innovations GmbH) - apply source formatter
  ******************************************************************************/
 package org.eclipse.californium.core.coap;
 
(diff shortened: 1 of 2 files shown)


californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
```diff
@@ -588,21 +581,17 @@ public Response waitForResponse() throws InterruptedException {
...
 			while (this.response == null && !isCanceled() && !isTimedOut() && !isRejected()) {
-				lock.wait(timeout);
+				wait(timeout);
 				long now = System.currentTimeMillis();
 				// timeout expired?
-				if (timeout > 0 && expired <= now) {
-					// break loop since response is still null
-					break;
+				if (timeout > 0) {
+					timeout = expired - now;
```
sophokles73: what's wrong with the previous condition? IMHO you shouldn't assign a value to the `timeout` parameter ...

boaks: > 	 * @param timeout the maximum time to wait in milliseconds.
> ...
> 			while (this.response == null && !isCanceled() && !isTimedOut() && !isRejected()) {
> 				lock.wait(timeout);
> 

So, if the while loop, this potentially waits in the sum for longer then timeouts (just the last wait may cause this, so checking after the wait(9 is not enough).
But I will change it not tu update the parameter.

sophokles73: agreed :-)

PR #268, opened 2017-04-05
Spelling fixes

scop:


boaks:
Thanks! 
Though the changes are mostly single lines, I would say we could merge them and "other left open PRs! could then be just rebased. Any objections? If none, I would go to merge this 6. april.

PR #270, opened 2017-04-06
Messageexchangestore cleanup

boaks:
Add final, volatile and remove not longer required synchronize.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
@@ -18,13 +18,14 @@
  *                                                    longer, so that MID (observe)
  *                                                    or token (blockwise) may
  *                                                    be reused for an other
- *                                                    exchange. 
+ *                                                    exchange.
  *    Achim Kraus (Bosch Software Innovations GmbH) - apply formatter
+ *    Achim Kraus (Bosch Software Innovations GmbH) - cleanup synchronization
+ *                                                    integrate clear() into stop()
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
 import java.net.InetSocketAddress;
-import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -61,13 +62,12 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 	private final ConcurrentMap<KeyToken, Exchange> exchangesByToken = new ConcurrentHashMap<>();
 
 	private final NetworkConfig config;
+	private final TokenProvider tokenProvider;
 	private boolean running = false;
-	private Deduplicator deduplicator;
+	private volatile Deduplicator deduplicator;
+	private volatile MessageIdProvider messageIdProvider;
 	private ScheduledFuture<?> statusLogger;
 	private ScheduledExecutorService scheduler;
-	private MessageIdProvider messageIdProvider;
-	private TokenProvider tokenProvider;
-	private SecureRandom secureRandom;
 
 	/**
 	 * Creates a new store for configuration values.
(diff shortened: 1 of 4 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -61,13 +61,12 @@
 	private final ConcurrentMap<KeyToken, Exchange> exchangesByToken = new ConcurrentHashMap<>();
 
 	private final NetworkConfig config;
+	private final TokenProvider tokenProvider;
 	private boolean running = false;
-	private Deduplicator deduplicator;
+	private volatile Deduplicator deduplicator;
```
sophokles73: is there a particular reason you want/need to declare these `volatile`?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -248,7 +235,7 @@ public boolean registerOutboundRequest(final Exchange exchange) {
 			throw new IllegalArgumentException("exchange does not contain a request");
 		} else {
 			int mid = registerWithMessageId(exchange, exchange.getCurrentRequest());
-			if (mid > Message.NONE) {
+			if (Message.NONE != mid) {
```
sophokles73: just out of interest: do you suspect a problem with the original comparison? FMPOV both alternatives should do the trick, wouldn't they?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -271,35 +258,24 @@ public boolean registerOutboundRequestWithTokenOnly(final Exchange exchange) {
...
-		int size;
-		synchronized (exchangesByToken) {
-			removed = exchangesByToken.remove(token, exchange);
-			size = exchangesByToken.size();
-		}
+		boolean removed = exchangesByToken.remove(token, exchange);
 		if (removed) {
-			LOGGER.log(Level.FINE, "removing exchange for token {0}, remaining exchanges by tokens: {1}",
-					new Object[] { token, size });
+			LOGGER.log(Level.FINE, "removing exchange for token {0}", new Object[] { token });
```
sophokles73: you generously removed the information about the remaining number of exchanges by token, is that information logged elsewhere now or is this the price we have to pay for getting rid of the synchronization?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -271,35 +258,24 @@ public boolean registerOutboundRequestWithTokenOnly(final Exchange exchange) {
...
+			removedExchange = exchangesByMID.remove(messageId);
+		} else if (exchangesByMID.remove(messageId, exchange)) {
+			removedExchange = exchange;
+		} else {
+			removedExchange = null;
 		}
 		if (null != removedExchange) {
-			LOGGER.log(Level.FINE, "removing exchange for MID {0}, remaining exchanges by MIDs: {1}",
-					new Object[] { messageId, size });
+			LOGGER.log(Level.FINE, "removing exchange for MID {0}", new Object[] { messageId });
```
sophokles73: see above

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -353,12 +325,8 @@ public boolean registerOutboundResponse(final Exchange exchange) {
 	 * Purges all registered exchanges from this store.
 	 */
 	public void clear() {
-		synchronized (messageIdProvider) {
-			synchronized (exchangesByToken) {
-				exchangesByMID.clear();
-				exchangesByToken.clear();
-			}
-		}
+		exchangesByMID.clear();
```
sophokles73: IMHO we need to keep synchronization here in order to prevent cases where thread A invokes `registerOutbound` and thread B invokes `clear` and any of the method invocations in `clear` are interwoven with the registration invocations in `registerOutbound`, don't you think?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -375,8 +343,6 @@ public synchronized void start() {
 						InMemoryMessageIdProvider.class.getName());
 				messageIdProvider = new InMemoryMessageIdProvider(config);
 			}
-			secureRandom = new SecureRandom();
-			secureRandom.nextInt(10); // trigger self-seeding of the PRNG
```
sophokles73: nice catch :+1:

californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/CropRotation.java
```diff
@@ -48,12 +50,12 @@
 	private boolean running = false;
 	private ScheduledExecutorService executor;
 
-	private ExchangeMap[] maps;
-	private int first;
-	private int second;
+	private final ExchangeMap maps[];
+	private volatile int first;
+	private volatile int second;
```
sophokles73: this seems to have been changed to allow for multi-threaded executors being used, right?
However, I think we should rather prevent multiple rotation tasks being scheduled to run in parallel instead.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: From my view, the intended usage of this class according threads seems to be not clear, or at least not documented.
The `deduplicator` write-access is synchronized and done before or with `start`.
The `deduplicator` read-access is not synchronized (`find `or `findPrevious`). I don't see a comment, that this methods are only allowed to be called from threads which are created after the `start` and even if, it would be hard to ensure that.
So in my opinion the easiest way to ensure, that other threads read the right value is `volatile`.
A other approach would be to introduce a "synchronized getter".

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: My idea of changing that, was to have a defined value range for `mid` and a defined usage pattern.
There is code, which uses `Message.NONE` just as values (therefore `!=` or `==`) and others as the negative value (therefore `>`). There was even a method (`MessageIdTracker.getNextMessageId()`), which returns `-1`. 
So I replaced the `-1` and go for `!=` or `==`.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: Yes :-). And it's only logged, if the health stuff is enabled.
 
My main concern was the nested `synchronized` in `clear()`

> 		synchronized (messageIdProvider) {
> 			synchronized (exchangesByToken) {

and going over the code, I only saw, that `synchronized` is used to have "atomic logging of the size". So I removed it.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: OK, T1 is adding to `exchangesByMID` and then to `exchangesByToken`.
And in between T2 invokes `clear()`.
My understand was, that `clear()` is only useful on shutdown and so I don't see that this will end in too much additional problems. Or is there any other use-case to invoke `clear()`?

californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/CropRotation.java
boaks: Write-access to `first` and `second` is synchronized, while `read` not.
As far as I know, this doesn't ensure, the `read` gets the right values.
Therefore I added `volatile`. Alternatively the read could be done also `synchronized`, but

> 		Exchange prev = maps[f].get(key);
> 		if (prev != null || f == s)
> 			return prev;
> 		prev = maps[s].get(key);
> 

the compare `f == s` indicates, that a none `synchronized` solution was intended.

boaks:
By the way:
Should we first try to have a general agreement on the "right way to do multi threading"?
The pattern, having only the write access synchronized, should this be used?
Or to rely on reading values only from Threads started later?
Or to document the thread usage?
I would appreciate that, at least for the mid future :-).

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
sophokles73: I see.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
sophokles73: ok, I can live with that.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
sophokles73: I agree in that invoking `clear()` makes most sense during shutdown. In fact, the method currently is invoked by `stop()` only, so we could simply reduce its visibility to `private`.

californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/CropRotation.java
sophokles73: Oh, I see. It is more about the `find` methods' unsynchronized access to the fields, understood.

sophokles73:
LGTM

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
sophokles73: IMHO we should either make this `private` or keep the synchronization.

boaks: I will go for private ...

boaks:
I integrated the `clear()` into `stop()` to indicate, that its not longer intended to be called outside of `stop()`. If it gets required again, then we must re-establish a proper synchronization.

sophokles73:
LGTM, don't forget to merge into master as well :-)

PR #272, opened 2017-04-07
Add long term testing.

boaks:
Add long term test runner for debugging a rarely failing tests.

### DIFF
--- a/element-connector/src/test/java/org/eclipse/californium/elements/runner/RepeatingTestRunner.java
+++ b/element-connector/src/test/java/org/eclipse/californium/elements/runner/RepeatingTestRunner.java
@@ -1,0 +1,194 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations - initial implementation
+ ******************************************************************************/
+package org.eclipse.californium.elements.runner;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.junit.runner.notification.Failure;
+import org.junit.runner.notification.RunListener;
+import org.junit.runner.notification.RunNotifier;
+import org.junit.runners.BlockJUnit4ClassRunner;
+import org.junit.runners.model.InitializationError;
+
+/**
+ * Runner for tests under debugging.
+ * 
+ * Runs a test repeated until it fails or the maximum number of repeats is
+ * reached.
+ * 
+ * <pre>
+ * "org.eclipse.californium.elements.runner.RepeatingTestRunner.repeats", maximum number of repeats for test.
+ * 0 := repeat test until failure. Default 100.
+ * "org.eclipse.californium.elements.runner.RepeatingTestRunner.alife", interval in milliseconds.
+ * 0 := disabled. Default: 1000
+ * </pre>
+ * 
+ * For execution with maven {@code -Dtest="XyzAbcTest" -DfailIfNoTests=false} may be used.
+ */
+public class RepeatingTestRunner extends BlockJUnit4ClassRunner {
+
+	public static final Logger LOGGER = Logger.getLogger(RepeatingTestRunner.class.getName());
+
+	/**
+	 * Final for logging mega bytes.
+	 */
+	private static final int MEGA_BYTE = 1024 * 1024;
+	/**
+	 * Default maximum number of repeats. Current value 100.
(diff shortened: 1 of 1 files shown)


element-connector/src/test/java/org/eclipse/californium/elements/runner/LongTermRunner.java
```diff
@@ -0,0 +1,191 @@
...
+	 */
+	private static final int DEFAULT_ALIFE_INTERVAL_IN_MILLISECONDS = 1000;
+	/**
+	 * Maximum number of repeats.
+	 * 
+	 * May be set via property
+	 * "org.eclipse.californium.runner.LongTermRunner.repeats". Default
+	 * {@link #DEFAULT_MAXIUM_REPEATS}.
+	 */
+	private final long maxiumRepeates;
```
sbernard31: typo : maximum

element-connector/src/test/java/org/eclipse/californium/elements/runner/LongTermRunner.java
```diff
@@ -0,0 +1,191 @@
...
+	/**
+	 * Final for logging mega bytes.
+	 */
+	private static final int MEGA_BYTE = 1024 * 1024;
+	/**
+	 * Default maximum number of repeats. Current value 100.
+	 * 
+	 * @see #maxiumRepeates
+	 */
+	private static final int DEFAULT_MAXIUM_REPEATS = 100;
```
sbernard31: typo maximum

element-connector/src/test/java/org/eclipse/californium/elements/runner/LongTermRunner.java
```diff
@@ -0,0 +1,191 @@
...
+ * <pre>
+ * "org.eclipse.californium.elements.runner.LongTermRunner.repeats", maximum number of repeats for test.
+ * 0 := repeat test until failure. Default 100.
+ * "org.eclipse.californium.elements.runner.LongTermRunner.alife", interval in milliseconds.
+ * 0 := disabled. Default: 1000
+ * </pre>
+ * 
+ * @see LongTerm
+ */
+public class LongTermRunner extends BlockJUnit4ClassRunner {
```
sbernard31: Why not calling this DebugRunner or something like that ?

element-connector/src/test/java/org/eclipse/californium/elements/category/LongTerm.java
```diff
@@ -0,0 +1,32 @@
...
+ * 
+ * Typical tests using this category are used with {@link LongTermRunner}.
+ * 
+ * <code>
+ * &#64;RunWith(LongTermRunner.class)
+ * &#64;Category(LongTerm.class)
+ * public class AbCdTest {
+ * </code>
+ */
+public interface LongTerm {
```
sbernard31: Why do we need a new Category ? If I understand well this idea is to be able to repeat an unstable test to try to reproduce error right ? so we can just use @RunWith(LongTermRunner.class).

boaks: If we introduce a Category, we can select to run this test only in the `pom.xm` (or almost only, because the other packages may still run their tests.)
See the branch "hudson_explore".

element-connector/src/test/java/org/eclipse/californium/elements/runner/LongTermRunner.java
boaks: thanks!

element-connector/src/test/java/org/eclipse/californium/elements/runner/LongTermRunner.java
boaks: Should we chose the name according what it **does** or what it is **intended** to used for?
In both cases, the "other" need to be documented :-).

element-connector/src/test/java/org/eclipse/californium/elements/category/LongTerm.java
sbernard31: If you just want to run only one test you can use :
`mvn test  -Dtest="SmallServerClientTest"  -DfailIfNoTests=false`
or
`mvn clean install  -Dtest="SmallServerClientTest"  -DfailIfNoTests=false`
without modifying pom.xml and without category.

HTH :)

element-connector/src/test/java/org/eclipse/californium/elements/runner/LongTermRunner.java
sbernard31: Ok, so maybe something like RepeatingTestRunner ?

boaks: Ok.

element-connector/src/test/java/org/eclipse/californium/elements/category/LongTerm.java
boaks: > -DfailIfNoTests=false

Neither google nor I know something about this :-)
Could you provide a explaining link?

sbernard31: https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#failIfNoTests

It's not clear to me but sometime build failed if there is not tests in a project. This happened often with you use -Dtest.

boaks: OK, I remove the introduced category.

boaks:
Requested changes done.

sbernard31:
LGTM.

PR #273, opened 2017-04-07
Move time to begin of line.

boaks:


boaks:
If there are no objections, I would go to merge this tomorrow.
In my opinion, it makes it easier to detect timing dependencies.

Issue #275, opened 2017-04-07
Processing out of order blockwise responses and retry the request 


boaks:
hudson:
https://hudson.eclipse.org/californium/job/2.0.x/146/

The build failed:
Simplified sequence:
- GET BLOCK 0
- GET BLOCK 0 (Retry)
- GET BLOCK 0 (Retry)
- ACK BLOCK 0
- GET BLOCK 1 (Dropped by test intention!)
- ACK BLOCK 0 (Retry)
- ACK BLOCK 0 (Retry)

My idea:
The `ACK BLOCK 0 (Retry)` stops the retransmission of the Exchange in the ReliabilityLayer.receiveResponse.
The BlockwiseLayer ignores the wrong block.

> 249 FINE [DirectDatagramSocketImpl]: (E199,T249) 127.0.0.1:8,234 ==> :8,236 [CON-GET    MID=  332, Token=ffd06f7a0af2b6a7, OptionSet={"Uri-Host":"localhost", "Uri-Path":"test"}, no payload] - (org.eclipse.californium.elements.util.DirectDatagramSocketImpl.java:817) receive() in thread Thread-1 at (03:54:33.021)
> 249 FINE [DirectDatagramSocketImpl]: (E201,T249) 127.0.0.1:8,234 ==> :8,236 [CON-GET    MID=  332, Token=ffd06f7a0af2b6a7, OptionSet={"Uri-Host":"localhost", "Uri-Path":"test"}, no payload] - (org.eclipse.californium.elements.util.DirectDatagramSocketImpl.java:817) receive() in thread Thread-1 at (03:54:33.628)
> 249 FINE [DirectDatagramSocketImpl]: (E202,T249) 127.0.0.1:8,234 ==> :8,236 [CON-GET    MID=  332, Token=ffd06f7a0af2b6a7, OptionSet={"Uri-Host":"localhost", "Uri-Path":"test"}, no payload] - (org.eclipse.californium.elements.util.DirectDatagramSocketImpl.java:817) receive() in thread Thread-1 at (03:54:33.628)
[...]

PR #276, opened 2017-04-07
Add blockwise test with miss ordered responses.

boaks:
Issue #275

Not intended to be merged before the issue is fixed!


boaks:
see #277

PR #281, opened 2017-04-10
Fix issue 275 3

boaks:
Return null for ACK with mismatching MID.

Fix for issue #275. Idea by sbernard31.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -28,6 +28,7 @@
  *                                                    remove exchange by MID on notify. 
  *                                                    Add Exchange for save remove.
  *    Achim Kraus (Bosch Software Innovations GmbH) - make exchangeStore final
+ *    Achim Kraus (Bosch Software Innovations GmbH) - return null for ACK with mismatching MID
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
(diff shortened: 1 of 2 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -234,6 +235,13 @@ public Exchange receiveResponse(final Response response, final CorrelationContex
 			return null;
 		} else if (correlationContextMatcher.isResponseRelatedToRequest(exchange.getCorrelationContext(), responseContext)) {
 
+			if (response.getType() == Type.ACK && exchange.getCurrentRequest().getMID() != response.getMID()) {
```
sophokles73: I have a very hard time following and reviewing your PRs, @boaks. Most of the time, it is very hard for me to understand the context of the proposes change(s). More detail information in the description fields and within the code would help a lot.
Please add a (detailed) comment describing under which circumstances you expect this to happen and why it makes sense to return `null`.

boaks:
Summary of issue #275 added as comment.

Sometimes errors are complex and there are more then one way to fix it. Though my first approach was also in my opinion too "complicated", I asked for other proposals and therefore prepared the other PRs.

boaks:
@sbernard31 
could you check, if this PR is according your intention for fixing issue #275?

@sophokles73 
though the failing scenario isn't too simple, I hope the comment is enough.

If you both agree, I would like to merge this soon.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
sophokles73: Is there a case where we do want to accept responses with the same token but an MID that differs from the request for CON/NON?

boaks: An ACK is in the first place matched by the MID. Only for "piggy-backed" ACKs, the token may also be used. The current implementation inverts this and therefore cause this erroneous behaviour.

The processing of other responses (CON/NON) is not changed by this PR.

sophokles73: My point is: can't we simply omit the check for the type being `ACK`, i.e.

````
} else if (correlationContextMatcher.isResponseRelatedToRequest(exchange.getCorrelationContext(), responseContext)) {
 
+			if (exchange.getCurrentRequest().getMID() != response.getMID()) {
````

Comment (boaks) [inline californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java]: **NO!**

For separate responses the MID is not used for matching request/response!

Comment (sophokles73) [inline californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java]: Sure, you're right.

File: californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java [hunk 2]
```diff
@@ -274,6 +276,66 @@ public void testGETLateNegotiationAndLostACK() throws Exception {
...
+	 * |                                                          |
+	 * | CON [MID=1236], GET, /status, 2:2/0/64           ------> |
+	 * |                                                          |
+	 * | <------   ACK [MID=1239], 2.05 Content, 2:2/0/64         |
+	 * </pre>
+	 * 
+	 * @throws Exception if the test fails.
+	 */
+	@Test
+	public void testGETWithMissOrderedResponses() throws Exception {
```
sophokles73: this should rather be `testGETWithDisorderedResponses`

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
```diff
@@ -274,6 +276,66 @@ public void testGETLateNegotiationAndLostACK() throws Exception {
...
+		System.out.println("Blockwise GET with responses miss ordered:");
```
sophokles73: `disordered`

boaks:
Now disordered is used for the test naming.

PR #282, opened 2017-04-11
Correct exchange states thread visibility

boaks:
Exchange is documented as not "thread safe", but used from different threads. This PR could not provide thread safety, it just intend to provided thread visibility for the fields.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
@@ -21,12 +21,15 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - add calculateRTT
  *                                                    use nanoTime instead of 
  *                                                    currentTimeMillis
+ *    Achim Kraus (Bosch Software Innovations GmbH) - ensure states visibility for
+ *                                                    different threads
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
 import java.util.Arrays;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.californium.core.Utils;
 import org.eclipse.californium.core.coap.BlockOption;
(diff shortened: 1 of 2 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
```diff
@@ -466,8 +489,9 @@ public boolean isComplete() {
 	public void setComplete() {
 		this.complete = true;
 		ExchangeObserver obs = this.observer;
-		if (obs != null)
+		if (null != obs) {
```
sophokles73: @boaks, you seem to prefer this pattern of putting the *constant* to compare with in the first place. Personally, I find this much harder to read. So, unless you have a very strong opinion on this and I'd rather have you not change these comparisons but instead stick to the approach used throughout the rest of the code. Otherwise we should discuss on the mailing list and have a vote on whether we want to adopt this pattern in general.

boaks: :-) my be I'm too influenced from MISRA.

Would the use of `final ExchangeObserver obs` be acceptable?

sophokles73: I don't have a problem with that at all. I do not understand, though, how this is related. Is it?

boaks: It makes the left side of an compare final.

sophokles73: obviously, the question is: is there a particular advantage of doing so?

boaks: Only for those, who switch between the programming languages.

sophokles73: Ok, so can we stick to the pattern already in use?

boaks:
So back to the old pattern :-).

californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
```diff
@@ -466,8 +489,9 @@ public boolean isComplete() {
 	public void setComplete() {
 		this.complete = true;
 		ExchangeObserver obs = this.observer;
-		if (obs != null)
+		if (obs != null) {
```
sophokles73: is there a particular reason why we do not simply invoke `completeCurrentRequest` at this point?

boaks: I don't know, insert such a call after `this.complete = true;` should work on my opinion. 
But though this PR's intention is to make the states visible to other threads , I didn't redesign the interface nor implementation.

PR #284, opened 2017-04-12
Add "cause" logging on "stop".

boaks:
Cleanup stop and log the "cause" when stopping.
Intended to analyze, if "stop" unlocks a deadlock.
(Debugging SmallServerClientTest)


boaks:

Issue #258, opened 2017-03-30
CON Notifies with updated payload but same MID


boaks:
Analysing `ObserveServerSideTest.testQuickChangeAndTimeout()` I got aware, that californium sends notifies with updated payload but the same MID.

<-----   CON [MID=462, T=0b], 2.05, observe(2), ETags(0003)
<-----   CON [MID=462, T=0b], 2.05, observe(3), ETags(0004)
<-----   CON [MID=462, T=0b], 2.05, observe(4), ETags(0005)
<-----   CON [MID=462, T=0b], 2.05, observe(5), ETags(0006)
<-----   CON [MID=462, T=0b], 2.05, observe(6), ETags(0007)

So my natural understanding would be, to change the MID, if the payload is changed.
Does someone know, why the MID is reused?

boaks:
e-mail to core@ietf.org:
-------------------------------------------------------------------
Hello CORE,

I have some questions according the intended message flow for an observe (RFC7641) which uses CON notifies and the resource is changing while retransmitting the notify and waiting for the ACK. I got aware that (at least one) implementation just modifies the payload in that case leaving the MID an retransmission time from the already pending notify.

Sequence:
CoAP client                                                                                     CoAP server
---------- GET R, CON, MID 1, TOKEN ABCD, Observe 0               ------------>
<-------- 2.05, ACK, MID 1, TOKEN ABCD, Observe 10, "state 1"   -------------
...
<-------- 2.05, CON, MID 65, TOKEN ABCD, Observe 12, "state 2" ------------          (R changed from "state 1" to "state 2")
<-------- 2.05, CON, MID 65, TOKEN ABCD, Observe 12, "state 2" ------------          (retransmission, because ACK missing)
...
<-------- 2.05, CON, MID 65, TOKEN ABCD, Observe 14, "state 3" ------------          (R changed from "state 2" to "state 3")  
----------                                   ACK, MID 65                                     ------------>

So:
Is it intended to (re-)use a MID with a modified message? (I personally would say, this is a NO GO!).
Should the retransmission timing/counter from the first notify be continued?
Should this continuing also prevent the second change to be send delayed with the next retransmission? 
Or should the second change simply wait until the first timeout all retransmission?

There may be even different answers/intended usage, depending on the usage of the CON notify:
If it's used the test the liveliness of the observe or If it's used to ensure (in my opinion a pseudo) safe state transfer  
 
[...]

PR #285, opened 2017-04-12
Add newMID expectation.

boaks:
Add MID test for retransmitted notifies with changed and unchanged payload.

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
@@ -17,6 +17,7 @@
  *    Daniel Pauli - parsers and initial implementation
  *    Kai Hudalla - logging
  *    Kai Hudalla (Bosch Software Innovations GmbH) - use static reference to Serializer
+ *    Achim Kraus (Bosch Software Innovations GmbH) - apply source formatter
  ******************************************************************************/
 package org.eclipse.californium.core.test.lockstep;
 
@@ -51,7 +52,6 @@ import org.eclipse.californium.elements.RawData;
 import org.eclipse.californium.elements.RawDataChannel;
 import org.eclipse.californium.elements.UDPConnector;
 
-
 public class LockstepEndpoint {
 
 	private static boolean DEFAULT_VERBOSE = false;
@@ -78,6 +78,7 @@ public class LockstepEndpoint {
 		this.incoming = new LinkedBlockingQueue<RawData>();
 		this.connector = new UDPConnector(new InetSocketAddress(InetAddress.getLoopbackAddress(), 0));
 		this.connector.setRawDataReceiver(new RawDataChannel() {
+
 			public void receiveData(RawData raw) {
 				incoming.offer(raw);
 			}
@@ -87,7 +88,7 @@ public class LockstepEndpoint {
 
 		try {
 			connector.start();
-//			Thread.sleep(100);
+			// Thread.sleep(100);
 		} catch (Exception e) {
 			throw new RuntimeException(e);
 		}
@@ -112,15 +113,15 @@ public class LockstepEndpoint {
 	public boolean isVerbose() {
 		return verbose;
 	}
-	
+
 	public int getPort() {
 		return connector.getAddress().getPort();
 	}
-	
+
 	public InetAddress getAddress() {
 		return connector.getAddress().getAddress();
 	}
-	
+
 	public Object get(String var) {
 		return storage.get(var);
 	}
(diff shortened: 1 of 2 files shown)


boaks:
This test extension is related to issue #258 
The reused MIDs didn't cause an failure, because the MIDs are not tested.
With this PR wrong reused MIDs would be detected.

boaks:

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
```diff
@@ -722,6 +725,49 @@ public ResponseExpectation loadMID(final String var) {
 			super.loadMID(var);
 			return this;
 		}
+		
+		public ResponseExpectation newMID(final String var) {
```
sophokles73: Maybe you can add a JavaDoc describing what the expectation actually is?

boaks: Should I really "break" the javadoc practice used by this class :-)?

(I will add a javadoc.)

boaks:
Javadoc added.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
```diff
@@ -403,31 +404,30 @@ public void testQuickChangeAndTimeout() throws Exception {
...
 		testObsResource.change("Second notification " + generateRandomPayload(10));
-		client.expectResponse().type(respType).code(CONTENT).token(tok).checkObs("B", "C").payload(respPayload).go();
+		client.expectResponse().type(respType).code(CONTENT).token(tok).newMID("MID").checkObs("B", "C").payload(respPayload).go();
 		// client does not ACK the CON notification
 
 		respType = NON;
 		testObsResource.change("NON notification 1 " + generateRandomPayload(10));
 		// server re-transmits unACKed CON notification but client does not reply
-		client.expectResponse().type(CON).code(CONTENT).token(tok).checkObs("B", "B").payload(respPayload).go();
+		client.expectResponse().type(CON).code(CONTENT).token(tok).newMID("MID").checkObs("B", "B").payload(respPayload).go();
```
sophokles73: re-transmission of the CON notification will probably use the same MID as the original CON notification, wouldn't it?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
```diff
@@ -403,31 +404,30 @@ public void testQuickChangeAndTimeout() throws Exception {
...
 
 		testObsResource.change("NON notification 2 " + generateRandomPayload(10));
 		// server re-transmits unACKed CON notification but client does not reply
-		client.expectResponse().type(CON).code(CONTENT).token(tok).checkObs("B", "B").payload(respPayload).go();
+		client.expectResponse().type(CON).code(CONTENT).token(tok).newMID("MID").storeMID("MID_R").checkObs("B", "B").payload(respPayload).go();
```
sophokles73: same here

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
```diff
@@ -403,31 +404,30 @@ public void testQuickChangeAndTimeout() throws Exception {
...
 
+		// server re-transmits unACKed CON notification with unmodified payload and (repeated) MID
+		client.expectResponse().type(CON).code(CONTENT).token(tok).loadMID("MID_R").loadObserve("B").payload(respPayload).go();
```
sophokles73: and again

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
```diff
@@ -403,31 +404,30 @@ public void testQuickChangeAndTimeout() throws Exception {
...
+		
 		testObsResource.change("NON notification 3 " + generateRandomPayload(10));
 		// server re-transmits unACKed CON notification but client does not reply
-		client.expectResponse().type(CON).code(CONTENT).token(tok).checkObs("B", "B").payload(respPayload).go();
-
-		testObsResource.change("NON notification 4 " + generateRandomPayload(10));
-		// server re-transmits unACKed CON notification but client does not reply
-		client.expectResponse().type(CON).code(CONTENT).token(tok).checkObs("B", "B").payload(respPayload).go();
+		client.expectResponse().type(CON).code(CONTENT).token(tok).newMID("MID").checkObs("B", "B").payload(respPayload).go();
```
sophokles73: and again

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
boaks: If the payload is changed, a new MID is required!

That's the above mentioned issue #258 

See 

https://github.com/eclipse/californium/issues/258#issuecomment-292884306

boaks:
Just to say, if the "newMID" would have been wrong, then the corrected implementation of the ObserveLayer would be also wrong.

In my opinion, it's generally: 
same message, same MID
different message, different MID

And different payload is different message.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
sophokles73: I see. Does this mean that Cf (as a server hosting resources that are observed by a client) already implements the behavior described in 4.5.2 of RFC7641?

My understanding of the way the test case has been implemented so far is that the server did *not* yet support the behavior mandated by RFC7641, or am I mistaken?

boaks: > I see. Does this mean that Cf (as a server hosting resources that are observed by a client) already implements the behavior described in 4.5.2 of RFC7641?

Since PR #279 the behaviour it's fixed :-).

And this PR is only the promised unit test. 

https://github.com/eclipse/californium/issues/258#issuecomment-292949575
> If the PR is agreed, I would **extend this test to check the mids.**

sophokles73: Ok, I have lost track amongst all the PRs and issues and cross references ...
I'll trust you on this one then ;-)

sophokles73:
LGTM

PR #288, opened 2017-04-18
Log source line

boaks:
Correct line number determination.

Additional small adjust on formatting output when logging elements are selected via policy.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CaliforniumFormatter.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CaliforniumFormatter.java
@@ -19,6 +19,7 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - move time to begin of line
  *                                                    use append for each
  *                                                    stacktrace line
+ *    Achim Kraus (Bosch Software Innovations GmbH) - search for line number of caller
  ******************************************************************************/
 package org.eclipse.californium.core;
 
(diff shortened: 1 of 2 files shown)


sophokles73:
If the algorithm for determining the line number is so fragile and unreliable, what sense does it make to run a unit test on it which I would expect to fail every now and then if your assumption is correct?

boaks:
> If the algorithm for determining the line number is so fragile and unreliable, 

The current implementation just takes the 7. or 8. stacktrace element.

```
		StackTraceElement[] stack = Thread.currentThread().getStackTrace();
		if (throwable != null && stack.length > 7) {
			lineNo = stack[7].getLineNumber();
		} else if (stack.length > 8) {
			lineNo = stack[8].getLineNumber();
		} else {
			lineNo = -1;
		}

```

In my opinion this is fragile, because it depends on the number of methods on the stack before the formatter is called. The test shows, that the current implementation therefore fails on some `log` methods. Additionally it depends on the `Handler.publish` implementation, which may also change the number of methods on the stack before the formatter is called. (Note: the implementation of `TestHandler.publish` has also influence on the failing tests.)

This PR replaces implementation by searching for the class/method in the stacktrace elements instead of rely, that it's either 7 or 8. In my opinion this is a reliable implementation (and the test indicates, that it's working).

sophokles73:
I see, makes sense.

PR #291, opened 2017-04-21
Use synchronized to access exchange retry fields.

boaks:
According the pattern in `ObservationLayer`, the access to the `exchange` should be synchronized.


boaks:

sophokles73:
LGTM

PR #292, opened 2017-04-21
Introduce a buffered logging handler for asynchronous writing.

boaks:


boaks:
> 35 times with direct logging, 24 times with buffered logging

So the difference is not that big as expected. Seems not really to pay off.

PR #294, opened 2017-04-24
Add correlation context and principal to TLS connectors

boaks:
Extract correlation context and principal.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/MapBasedEndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/MapBasedEndpointContext.java
@@ -51,8 +51,8 @@ public class MapBasedEndpointContext extends AddressEndpointContext {
 	 * 
 	 * @param peerAddress peer address of endpoint context
 	 * @param peerIdentity peer identity of endpoint context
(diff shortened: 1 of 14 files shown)


boaks:
A little more than the half of the changes are new unit tests.

So @sophokles73, though @sbernard31 is busy with analysing the memory leak (see issue #429), may be you can spent some time on this PR in the next weeks :-),

sbernard31:
I looked at this really quickly.
All of this is mainly about TCP connector which is (if I well remember) pretty experimental.
Just to say, I'm not against integrated this.

boaks:
Yes, your right.  And the spec. seems to be still a draft.

On the other side, it only affects the TCP/TLS stuff ( I see no dependency, which affects function to change for this). So that's for me the difference to BERT, which may affect the UDP blockwise stuff and therefore that's on a separate branch.

sbernard31:
Any reasons why this is not still not merge ?

boaks:
I didn't want to influence other PRs. 
Currently I work on moving the rpk trust stores.
 And when I'm finished with that, then I would go to rebase and retest this and have it merged.

Issue #290, opened 2017-04-21
Make timeout used by unit test configurable to setup them more relaxed on eclipse-hudson


boaks:
Let me first give a very brief summary about my experience with the unit tests:
4 weeks ago I introduced the `DirectDatagramSocketImpl` to prevent messaging loss from the native IP-stack.
==> But the tests still rarely failed.

2 weeks ago I introduced the `RepeatingTestRunner` to repeat a rarely failing test to debug it. I chose the `SmallServerClientTest` to do so. My result:
==>	It fails rarely, but I can’t see a deterministic pattern
==>	It seems to fails more frequently, if the logging level is higher 
==>	It seems to fails less frequently, if the timeout waiting for response is increased.

This week I tried a `BufferedLogHandler` to reduce the timing influence from writing the log messages to System.err. 
==> It seems to make the test a little more stable 

So my final conclusion: 
Analyzing the logs, the “pause” seems to be moving around, sometimes the messages are just delivered after the “destroy” (and so to late), sometime before but after the `waitForResponse()`. Sometimes you see, that the timeline in the log is corrupted. This could be potential the result of many issues, californium (longtime locks or busy threads, but I didn’t found that until now), the platform or the host.

So I would go for relaxing the timeout when executing the tests on Hudson.
And therefore, I think we should add a configurable function, which calculates the effective timeout. 
e.g.:

> ((timeout * scale) + bias)

There are sure other alternative strategies (use more deterministic CI host), so I would appreciate your feedback before spending too much time in this approach.

boaks:
Just to say:
relaxing timeouts will not solve all issues.

1.0.x-nightly, 26
09:15:31.909: [CON-GET    MID=36648, Token=95, OptionSet={"Observe":0,  
09:15:32.109: [CON-GET    MID=36648, Token=95, OptionSet={"Observe":0,
09:15:32.259: [ACK-2.05   MID=36648, Token=95,

java.lang.AssertionError: Must not have option number 6 but has

If the test execution itself is delayed, other threads may resend a message, which the test doesn't expect.

boaks:
To get more facts I prepared the "hudson_explore" branch to execute the `SmallServerClientTest` with 100 repeats every time and a random selection of:

Random Setup      | Alternative 1      | Alternative 2 
----------------- | ------------------ | -----------------
Response timeout: | either 1000ms | or 2000ms
Logging: | either `direct` | or `buffered`

[...]

PR #295, opened 2017-04-25
Lockstep deduplicator

boaks:
Reduce failures caused by unintended retransmission as 1.0.x-nightly, 26:

https://github.com/eclipse/californium/issues/290#issuecomment-296217897

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
@@ -21,6 +21,7 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - add newMID to ResponseExpectation
  *    Achim Kraus (Bosch Software Innovations GmbH) - correct mid check. issue #289
  *    Achim Kraus (Bosch Software Innovations GmbH) - use option names for logging
+ *    Achim Kraus (Bosch Software Innovations GmbH) - rename loadMID into sameMID
  ******************************************************************************/
 package org.eclipse.californium.core.test.lockstep;
 
@@ -230,7 +231,7 @@ public class LockstepEndpoint {
 			return this;
 		}
 
-		public MessageExpectation loadMID(final String var) {
+		public MessageExpectation sameMID(final String var) {
 			expectations.add(new Expectation<Message>() {
 
 				public void check(Message message) {
@@ -724,8 +725,8 @@ public class LockstepEndpoint {
 		}
 
 		@Override
-		public ResponseExpectation loadMID(final String var) {
-			super.loadMID(var);
+		public ResponseExpectation sameMID(final String var) {
+			super.sameMID(var);
 			return this;
 		}
 
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
@@ -22,6 +22,15 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - correct mid check. issue #289
  *    Achim Kraus (Bosch Software Innovations GmbH) - use option names for logging
  *    Achim Kraus (Bosch Software Innovations GmbH) - rename loadMID into sameMID
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add smart deduplication filter.
+ *                                                    filter messages based on the last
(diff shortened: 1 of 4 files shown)


boaks:

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
```diff
@@ -212,9 +259,61 @@ public void setDestination(InetSocketAddress destination) {
...
+	 * {@link MessageExpectation#sameBoth(String)}.
+	 * 
+	 * @param midExpectation MID expectation
+	 * @return next received message
+	 * @throws InterruptedException if waiting for message is interrupted.
+	 */
+	public Message receiveNextMessage(MIDExpectation midExpectation) throws InterruptedException {
+		while (true) {
+			RawData raw = incoming.poll(2, TimeUnit.SECONDS); // or take()?
+			assertNotNull("Did not receive a message (but nothing)", raw);
```
sophokles73: can we instead use something like `did not receive message within expected time frame (2 secs)`?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
```diff
@@ -212,9 +259,61 @@ public void setDestination(InetSocketAddress destination) {
...
+
+			Message msg = parser.parseMessage(raw);
+			if (null != midExpectation && null != lastIncomingMessage && lastIncomingMessage.getMID() == msg.getMID()
+					&& lastIncomingMessage.getType() == msg.getType() && !midExpectation.expectMID(msg)) {
+				// received message with same MID but not expected
+				// => drop message!
+				print("Deduplication drop " + msg);
```
sophokles73: maybe `discarding duplicate message: `

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
```diff
@@ -1422,4 +1624,10 @@ public void go() {
 		 */
 		public void go() throws Exception;
 	}
+
+	public static interface MIDExpectation {
```
sophokles73: can we use camel case: `MidExpectation`?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
```diff
@@ -1422,4 +1624,10 @@ public void go() {
...
+
+		public boolean expectMID(Message message);
```
sophokles73: same here: `expectMid`

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
boaks: Sure, the text was just take as it is.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
boaks: OK

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
boaks: OK 
(I chose MID, because `loadMID` used upper case letters also).

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
boaks: OK,
so keep `load/store/newMID` or change them also?

sophokles73: so, if `expectMID` is to be seen in the context of these other existing methods, I would say that we probably should keep the name `expectMID` in order to not confuse people too much.
What about the interface name itself? I assume that user code does not work directly with it, right? Thus, I'd propose to use correct camel case for the interface at least ...

boaks: OK.

boaks:
Message improved, interface name in CaMel.

sophokles73:
LGTM

PR #299, opened 2017-04-28
Restart executor on start() after stop().

boaks:
Reset "internal" executor to null in stop(), to restart it on start().
Add unit test for restart() using an internal executor. 


boaks:
If no objections, I will go to merge this next week.

sophokles73:
LGTM

Issue #56, opened 2016-06-08
The variable canceled of CoAPObserverRelation is not set to true after calling the reregister()


popoyatou:
Hello,

We are using californium in our software and we are facing the following situation:
The communication of the network is very bad and the devices could stop working at any moments. So after establishing a relationship of ovservation, we have to check every X minutes about the status of the CoAPObserverRelation and if the relation is cancelled, we need to reregister with the same token. (The automatic reregister after the max-age is passed may not work because the device is not very stable) 
However, it seems that after calling coapObserverRelation.reregister(), the value of coapObserverRelation.isCanceled() is still true (which should be set false) after an ACK is received. Is it possbile a bug that should be fixed?

Thank you very much.
Jie

sophokles73:
That's probably true.
Let me try to better understand your scenario.
1. client issues a GET request with the observe option set
2. server's reachability over the network is very bad, so the request may get re-transmitted by the client several times before it finally times out and thus `CoapObserveRelation.isCanceled()` (correctly) returns `true`
3. client tries to (re-)establish the observation by means of invoking `CoapObserveRelation.reregister()`
4. the new request succeeds and the server responds with an ACK but on the client side the `CoapObserveRelation.isCanceled()` method still returns `true`

If this is what you are experiencing then I think you are right in assuming that `isCanceled()` shouldn't return `true` anymore. However, I am not sure if @mkovatsc originally intended an observe request that has been canceled (either explicitly or implicitly) to be *re-animated* by invoking `CoapObserveRelation.reregister()`. The client might send a new request for that purpose as well and the fact that `reregister()` is declared public is the actual bug.

@mkovatsc maybe you can shed some light on this?

popoyatou:
Hi, Kai, yes, what you have understood is exactly our case. I have noticed that the method reregister() was private before and became public until the recent versions and thats why I have this assumption. And by checking the code, I didnt see any place to setCanceled(false).  For our case, this change is very good because we would like to reregister the observation with the same token. (otherwise, maybe we need to cancel the observation first and then start another observation?) Because each device can only have limited number of observers and reregistering with the same token can prevent wasting the observer positions.

sophokles73:
Yes, I see your point and to me it also makes sense to reuse the `CoapObserveRelation` in the way you intend to do. I just want to get @mkovatsc opinion about it because he has more knowledge about this area of Californium.

If we agree that the object should/culd be reused then fixing this should be easy ...

boaks:
[...]

PR #304, opened 2017-05-15
Add precondition check to reregister.

boaks:
Add also more unit tests for reregistration.

See issue #56 


sophokles73:
LGTM

Issue #187, opened 2016-12-29
Crash inside API CoapObserveRelation.reactiveCancel()


Rajsmit:
Hi, 

I found an crash inside API CoapObserveRelation.reactiveCancel(), when we tried to cancel a observer relation. Please check and provide your input.

12-29 15:31:26.781 6871-6871/? E/AndroidRuntime: FATAL EXCEPTION: main
                                                 Process: com.ikea.frekvens, PID: 6871
                                                 java.lang.RuntimeException: Unable to resume activity {com.ikea.frekvens/com.ikea.frekvens.home.activity.WelcomeActivity}: java.lang.IllegalArgumentException: MID must not be a 16 bit unsigned int
                                                     at android.app.ActivityThread.performResumeActivity(ActivityThread.java:2986)
                                                     at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3017)
                                                     at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2392)
                                                     at android.app.ActivityThread.access$800(ActivityThread.java:151)
                                                     at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1303)
                                                     at android.os.Handler.dispatchMessage(Handler.java:102)
                                                     at android.os.Looper.loop(Looper.java:135)
                                                     at android.app.ActivityThread.main(ActivityThread.java:5254)
                                                     at java.lang.reflect.Method.invoke(Native Method)
                                                     at java.lang.reflect.Method.invoke(Method.java:372)
                                                     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
                                                     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
                                                  Caused by: java.lang.IllegalArgumentException: MID must not be a 16 bit unsigned int
                                                     at org.eclipse.californium.core.network.Exchange$KeyMID.<init>(Exchange.java:552)
                                                     at org.eclipse.californium.core.network.Exchange$KeyMID.fromOutboundMessage(Exchange.java:623)
                                                     at org.eclipse.californium.core.network.UdpMatcher$ExchangeObserverImpl.completed(UdpMatcher.java:397)
                                                     at org.eclipse.californium.core.network.Exchange.setComplete(Exchange.java:441)
                                                     at org.eclipse.californium.core.network.stack.ExchangeCleanupLayer$CancelledMessageObserver.onCancel(ExchangeCleanupLayer.java:27)
[...]

PR #308, opened 2017-05-17
[#187] Prevent memory-leak caused by canceled but registered exchanges.

sophokles73:
If an outbound request gets canceled before it is sent it might still
get registered with the ExchangeStore. In order to prevent a memory leak
in the ExchangeStore we therefore make sure that the Exchange gets
"completed" again if it is canceled before it is sent out via the
Connector in the Outbox.


boaks:
LGTM.

PR #309, opened 2017-05-17
Fix retrans last server flight

boaks:
Tests and fix.

### DIFF
--- a/scandium-core/pom.xml
+++ b/scandium-core/pom.xml
@@ -28,6 +28,11 @@
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-all</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.hamcrest</groupId>
+			<artifactId>hamcrest-library</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 	<build>
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -39,6 +39,8 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - introduce synchronized getSocket()
  *                                                    as pair to synchronized releaseSocket().
  *    Achim Kraus (Bosch Software Innovations GmbH) - restart internal executor
+ *    Achim Kraus (Bosch Software Innovations GmbH) - processing retransmission of flight
+ *                                                    after last flight was sent.
  ******************************************************************************/
 package org.eclipse.californium.scandium;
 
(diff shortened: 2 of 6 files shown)


boaks:
RFC6347, 4.2.4, page 23, bottom

"the node that transmits the last flight (the server in an ordinary handshake or the client in a resumed
handshake) MUST respond to a retransmit of the peer’s last flight with a retransmit of the last flight."

was broken.

Processing Raw Public Key CERTIFICATE handshake messages with of epoch 0 without session causes Exceptions. Therefore introduce isNewClientHello to select CLIENT HELLO of epoch 0 only.

boaks:
I would like to merge this. Any objection?

sbernard31:
It sounds good to me but I have some questions maybe not directly linked to this PR :

1) It seems that [we resend the last flight without incremented](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java#L220) the sequence number of the record (I talked about the record one not the handshake message one)? And I'm not sure to know if this is normal. The spec seems to say that any retransmission should use a new sequence number (https://tools.ietf.org/html/rfc6347#section-4.1)

2) About the same [part of code](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java#L220), We don't check the message to process, any message trigger a retransmission of the last flight I found this strange.

3) I also notice that handshaker has no state, e.g. we can handle CERTIFICATE message while we are waiting for a CLIENT_HELLO message. I just want to know if I am the only on we found this strange ?

boaks:
To 1.

I'm always a bit confused about the sequence numbers :-).

The outer one is the record sequence number, it's incremented, as my wiresharks shows
(implemented in `DTLSFlight.setNewSequenceNumbers()`).

The inner one is the (handshake) message sequence number is NOT incremented, it's just repeated, as my wiresharks shows also.  RFC6347, 4.2.2

> Whenever each new message is generated, the message_seq value is incremented by one. Note that in the case of a rehandshake, this implies that the HelloRequest will have message_seq = 0 and the ServerHello will have message_seq = 1. When a message is retransmitted, the same message_seq value is used.

The retransmitted ClientHello in the sequence below that shows also seq=0.  

For the other questions: good issues, but I'm not sure, If we should add them to this PR.

sbernard31:
I agree about 2. and 3. we should not add it to this PR.

About 1.
It seems we agree that retransmission involve sequence number incrementation for record but not for handshake message.

So I think there is a bug about the last flight retransmission : 
```java
if (lastFlight != null) {
	// we already sent the last flight (including our FINISHED message),
	// but the client does not seem to have received it because we received
	// its finished message again, so we simply retransmit our last flight
	LOGGER.log(Level.FINER, "Received client's ({0}) FINISHED message again, retransmitting last flight...",
			getPeerAddress());
	recordLayer.sendFlight(lastFlight);
	return;
}	
```
`recordLayer.sendFlight(lastFlight)` seems to not increment sequence number.

boaks:
OK, now I understand ;-).

I'll try to add  `DTLSFlight.setNewSequenceNumbers()` .

boaks:
Now with incremented record sequence numbers.

boaks:
And with asserts on the record sequence numbers :-).

boaks:
I would still like to merge this. So any left objection?

sbernard31:
:+1:

boaks:
@sbernard31 

I would like to cherrypick this to the master.
It seems, that the reencrypting of your commit in 2.0.x
46e5c3355ce69efabe44014b10b682ffcff60b8a
is requried. Any reason, not do this in the master?

boaks:
Check branch "fix_retrans_last_flight" for required changes in master.

sbernard31:
>Any reason, not do this in the master?

No, I think we can cherry pick it in master.

Issue #305, opened 2017-05-15
Start CoAP retransmission timer when connector reports, that the data is sent.


boaks:
Change to start of the retransmission timer for CoAP on sent reported from the connector.
Helps to lower retransmission caused by busy connectors.

boaks:
master, failed build 230:

> 03:49:30.668: [0] === GET ==>
> 03:49:30.670: [1] <== ACK === (drop 1)
> 03:49:30.968: [2] === GET ==> (1. retransmission, drop 2)
> 03:49:31.268: [3] === GET ==> (2. retransmission, drop 3)
> 03:49:31.568: [4] === GET ==> (3. retransmission)
> (.......... delay ..............)
> 03:49:32.491: [5] <== ACK === (1. retransmission) overlapping with [6]
> 03:49:32.492: [6] === GET ==> (4. retransmission, drop 3)
> 
> Blockwise GET timed out after 1824ms 
> 03:49:30.668 + 1.824 => 03:49:32.512 Timeout
> 
> Possible times calculated:
> 03:49:31.568: [4] === GET ==> (3. retransmission)
> 03:49:31.868: Retransmission by timer task
> 03:49:32.168: Timeout
> 03:49:32.491: [5] <== ACK === (1. retransmission) too late
>  

If the timeout would start **AFTER** the sent, then this test may have succeeded.

boaks:
Fixed.

PR #312, opened 2017-05-24
Implement issue #305.

boaks:
Start retransmission timer after connector sends message.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -34,6 +34,8 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - use exchange.calculateRTT
  *    Achim Kraus (Bosch Software Innovations GmbH) - make exchangeStore in
  *                                                    BaseMatcher final
+ *    Achim Kraus (Bosch Software Innovations GmbH) - use new MessageCallback functions
+ *                                                    issue #305
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -631,15 +633,7 @@ public class CoapEndpoint implements Endpoint {
 				exchange.setComplete();
 
 			} else {
-				// create callback for setting correlation context
-				MessageCallback callback = new MessageCallback() {
-
-					@Override
-					public void onContextEstablished(final CorrelationContext context) {
-						exchange.setCorrelationContext(context);
-					}
-				};
-				RawData message = serializer.serializeRequest(request, callback);
+				RawData message = serializer.serializeRequest(request, new RequestCallback(exchange, request));
 				connector.send(message);
 			}
 		}
@@ -665,7 +659,7 @@ public class CoapEndpoint implements Endpoint {
 				if (null != exchange) {
 					correlationContext = exchange.getCorrelationContext();
 				}
-				connector.send(serializer.serializeResponse(response, correlationContext));
+				connector.send(serializer.serializeResponse(response, correlationContext, new MessageCallbackForwarder(response)));
 			}
 		}
 
@@ -690,7 +684,7 @@ public class CoapEndpoint implements Endpoint {
 				if (null != exchange) {
 					correlationContext = exchange.getCorrelationContext();
 				}
-				connector.send(serializer.serializeEmptyMessage(message, correlationContext));
+				connector.send(serializer.serializeEmptyMessage(message, correlationContext, new MessageCallbackForwarder(message)));
 			}
 		}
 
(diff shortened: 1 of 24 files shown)


boaks:
I would like to merge this. Any objection?

sbernard31:
I will review it today.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -237,6 +240,12 @@ public void onTimeout() {
 				public void onCancel() {
 					clearBlock1Status(key);
 				}
+
+				@Override
+				public void onSendError(Throwable error) {
+					copyToken();
+					clearBlock1Status(key);
+				}
```
sbernard31: Should we copyToken() on "onSent" too ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1342,10 +1344,8 @@ private void sendMessage(final RawData message, final DTLSSession session) {
 					session.getSequenceNumber(),
 					new ApplicationMessage(message.getBytes(), message.getInetSocketAddress()),
 					session);
-			if (message.getMessageCallback() != null) {
-				message.getMessageCallback().onContextEstablished(ctx);
-			}
```
sbernard31: Even if the API look better, I have some concerns about correlation context is now sent to upper layer only after record was sent. I think this could involve some race condition issues.  WDYT ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: I toke the implementation from `onReject()`/`onTimeout()`.
Without the `onSendError()` it would be a `onTimeout()`, so it should be OK.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -631,21 +633,13 @@ public void sendRequest(final Exchange exchange, final Request request) {
...
-				};
-				RawData message = serializer.serializeRequest(request, callback);
+				RawData message = serializer.serializeRequest(request, new RequestCallback(exchange, request));
 				connector.send(message);
 			}
 		}
 
 		@Override
-		public void sendResponse(Exchange exchange, Response response) {
+		public void sendResponse(final Exchange exchange, Response response) {
```
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: I will change the order. But even with that change, everything must then be done synchronous to the callback. I hope this doesn't slow down too much.
(My idea was to catch IOExceptions or other issue and report them as onSendError instead of onSent.)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: I'm not sure you understand the point. (or maybe it's me :p)
I would like to know if we should add a `copyToken` for the `onSent` event ? (not onSendError)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: Changing the order is not ok. I think "sent" event should be called after we really send the message. But I also think that context established still make sense before.

I mean the handshake is done => we raise `contextEstablised`
then, we send app data => we raise `message sent`

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: :-) OK, it was me, who didn't read.
Yes, this makes sense.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: After my answer to your first comment about the order I was already starting to think in the same direction as your second comment :-).
`MassageCallback` requires 3 methods, stay with `onContextEstablished()` and the 2 additional ones `onSent()` and `onSendError()`. I would stay with calling `onContextEstablished()` before sending and `onSent()` afterwards or `onSendError()`, if sending fails.

boaks:
Now with kept "onContextEstablished" and added "onSent()" and "onError()".

boaks:
Now with kept "onContextEstablished" and added "onSent()" and "onError()".

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: Done. "onContextEstablished" exists again.

sbernard31:
:+1:

Issue #311, opened 2017-05-24
ExchangeObserverImpl.contextEstablished(final Exchange exchange) 


boaks:
```
if (exchange.getRequest() != null) {
   observationStore.setContext(exchange.getRequest().getToken(), exchange.getCorrelationContext());
}
keyToken token = KeyToken.fromOutboundMessage(exchange.getCurrentRequest());
exchangeStore.setContext(token, exchange.getCorrelationContext());
```

Is there a reason, why the correlation context in the "observationStore" is (repeated) adjusted for all "current requests" (e.g. each block in a blockwise)?
I would like to optimize this reducing that to observes and the first block.

boaks:
After a discussion with @sbernard31 about the usage of the `onContextEstablished` (it must be ensured, that all operations according the correlation context are finished before returning from the callbac, otherwise there may be race condition with the response and those operations), I have concerns about the API of `ObservationStore `and `MessageExchangeStore ` `setContext`. Currently every request sent with a correlation context will call this `setContext`. As far as I has seen, the implementation then execute a lookup, to see, if somethng should be stored.

**This causes a lookup for every request.**

Currently this lookup would only be required for observes (therefore #314 does some optimization for the observe store, where this seems to be obvious). For the  `MessageExchangeStore ` this optimization would be only possible, if the documation of setContext is change to limit this for observes. If this limitation is not wanted, I would vote for replace the token or keytoken with the request, so that the implementation could use the request for faster filtering.

sophokles73:
Just as a starting point, I am posting @sbernard31's comment from #314:

> For me it's more important to know, which cases would require a setContext XyzStorage call. For now, it seems to be observe. Are there other use cases?

What do you call `XyzStorage` ? we need context for all Request/Response correlation (observe or not), so for all exchanges.

> Why do we need setContext in both storages? 

Both storages (ObservationStore and MessageExchangeStore) store exchanges. When a correlation context is established we must associate it to an exchange. So it should be stored (persisted for persistent store) that's why there is this `setContext` method. For now, I think there is only persistent implementation of `ObservationStore`, as far as I know `MessageExchangeStore` exist only in memory.  

> Why is one called for the request, the other one for currentRequest? 

About `ObservationStore`, we only store observe exchange (we don't store magical blockwise exchange) so we are only interested by the `request` (as `currentrequest` is about blockwise).

[...]

PR #314, opened 2017-05-29
Optimize correlation context processing.

boaks:
Set correlation context in observe store only for observes. Issue #311

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
@@ -23,6 +23,9 @@
  *                                                    currentTimeMillis
  *    Achim Kraus (Bosch Software Innovations GmbH) - ensure states visibility for
  *                                                    different threads
+ *    Achim Kraus (Bosch Software Innovations GmbH) - forward CorrelationContext only
+ *                                                    for the first time set.
+ *                                                    issue #311
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -156,7 +159,7 @@ public class Exchange {
 	private volatile int failedTransmissionCount = 0;
 
 	// handle to cancel retransmission
-	private AtomicReference<ScheduledFuture<?>> retransmissionHandle = new AtomicReference<ScheduledFuture<?>>();
+	private final AtomicReference<ScheduledFuture<?>> retransmissionHandle = new AtomicReference<ScheduledFuture<?>>();
 
 	// If the request was sent with a block1 option the response has to send its
 	// first block piggy-backed with the Block1 option of the last request block
@@ -172,7 +175,7 @@ public class Exchange {
 	// protocol stage executor
 	private volatile boolean customExecutor = false;
 
-	private volatile CorrelationContext correlationContext;
+	private final AtomicReference<CorrelationContext> correlationContext = new AtomicReference<CorrelationContext>();
 
 	/**
 	 * Creates a new exchange with the specified request and origin.
@@ -198,7 +201,7 @@ public class Exchange {
 		// might only be the first block of the whole request
 		this.currentRequest = request;
 		this.origin = origin;
-		this.correlationContext = ctx;
+		this.correlationContext.set(ctx);
 		this.nanoTimestamp = System.nanoTime();
 	}
 
@@ -583,10 +586,13 @@ public class Exchange {
 	 * @param ctx the correlation information
 	 */
 	public void setCorrelationContext(final CorrelationContext ctx) {
-		correlationContext = ctx;
-		ExchangeObserver obs = this.observer;
-		if (obs != null) {
-			obs.contextEstablished(this);
+		if (correlationContext.compareAndSet(null, ctx)) {
+			ExchangeObserver obs = this.observer;
+			if (obs != null) {
+				obs.contextEstablished(this);
+			}
+		} else {
+			correlationContext.set(ctx);
 		}
 	}
 
@@ -598,7 +604,7 @@ public class Exchange {
 	 *         information is available.
 	 */
 	public CorrelationContext getCorrelationContext() {
-		return correlationContext;
+		return correlationContext.get();
 	}
 
 	/**
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -31,6 +31,8 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - return null for ACK with mismatching MID
  *    Achim Kraus (Bosch Software Innovations GmbH) - release all tokens except of
  *                                                    starting observe requests
+ *    Achim Kraus (Bosch Software Innovations GmbH) - optimize correlation context
+ *                                                    processing. issue #311
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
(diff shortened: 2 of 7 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -396,11 +398,12 @@ public void completed(final Exchange exchange) {
 
 		@Override
 		public void contextEstablished(final Exchange exchange) {
-
-			if (exchange.getRequest() != null) {
-				observationStore.setContext(exchange.getRequest().getToken(), exchange.getCorrelationContext());
+			Request currentRequest = exchange.getCurrentRequest(); 
+			Request request = exchange.getRequest(); 
+			if (request == currentRequest && request != null && request.isObserve()) {
+				observationStore.setContext(request.getToken(), exchange.getCorrelationContext());
```
sbernard31: We check that `request == currentRequest`.
This happens only for blockwise transfer.
For block2 it's ok as this avoid to set context for each block2 request.
For block1 this means that correlation context will only be set  when the request will be completely sent. I think this is a problem. I think we want to set the correlation as soon as possible.

Should we just avoid to call `contextEstablished` if correlation context is already set or just when it changed  ?
I mean in Exchange :
```java
public void setCorrelationContext(final CorrelationContext ctx) {
    if (correlationContext != ctx) { // or correlationContext != null depending the behavior we want
       	correlationContext = ctx;
        if (obs != null) {
            obs.contextEstablished(this);
        }
    }
}
```

sophokles73: This makes sense to me. in fact, I wonder if it is (or should be) possible for the correlation context to change once it has been established? Currently, I do not think so and therefore think that we should go with @sbernard31 's approach ...

boaks: `correlationContext != ctx`

I guess, this will always fire :-). (currently it's implemented as always a new object).

So, this triggers the question, about the natures of correlation contexts. 
Should we replace the "==" by "some equals"?
Should the connector reuse the CorrelationContext object?

By the way, my last comment on issue #311 will still be valid even with this improvement.

https://github.com/eclipse/californium/issues/311#issuecomment-305957631

sophokles73: If we agree that a correlation context *cannot* change, then we should be able to simply check for *not null*, shouldn't we?

boaks: Do you propose, to assume, that, if the correlation context is set once for an exchange, it cannot change? I don't agree, simply because I don't know all the cases where a exchange is used.

There is currently a ongoing discussion about the request response matching in the ietf core mailing list. Before there is a conclusion there, I would not like to fix the behaviour in that way.
(https://www.ietf.org/mail-archive/web/core/current/msg08814.html).

So let me point again to first discuss issue #311.
For me it's more important to know, which cases would require a setContext XyzStorage call. For now, it seems to be observe. Are there other use cases? Why do we need setContext in both storages? Why is one called for the request, the other one for currentRequest? For me, this seems to be wired enough, and I would first go for understanding the usage of "setContext" in the storages.

sbernard31: > For me it's more important to know, which cases would require a setContext XyzStorage call. For now, it seems to be observe. Are there other use cases?

What do you call `XyzStorage` ? we need context for all Request/Response correlation (observe or not), so for all exchanges.

> Why do we need setContext in both storages? 

Both storages (ObservationStore and MessageExchangeStore) store exchanges. When a correlation context is established we must associate it to an exchange. So it should be stored (persisted for persistent store) that's why there is this `setContext` method. For now, I think there is only persistent implementation of `ObservationStore`, as far as I know `MessageExchangeStore` exist only in memory.  

> Why is one called for the request, the other one for currentRequest? 

About `ObservationStore`, we only store observe exchange (we don't store magical blockwise exchange) so we are only interested by the `request` (as `currentrequest` is about blockwise).

About `MessageExchangeStore`, we store every thing except "long term exchange relation" so in that case we care about blockwise transfert so I suppose this is the idea about why we use `currentRequest` instead of `request`.

FMPOV, setting correlationContext on observation store only for observe request make sense to me. (good optimization)
Without thinking too much , raising the event `onCorrelationContext` only when it really change can not be a bad modification. (using equals not != of course :p)

But if we think more about that, Exchange is used for :
 - Request/Response.
 - ObserverRequest/Response/notification
 - Resquest/Response + magical blockwise
For me in all case, the correlation context should not changed.
With our current implementation, the only case where context is changed (I mean not initially set) is when we do blockwise transfer. The question is "Should we allow to change context for a whole blockwise transfer?" 
Personally, I'll be tented to say "no".
It's not clear to me but the [RFC 7959](https://tools.ietf.org/html/rfc7959#section-7) seems to say "no" too :
_" Where access to a resource is only granted to clients making use of
   specific security associations, all blocks of that resource MUST be
   subject to the same security checks; "_

boaks: I used `XyzStorage` for both `ObservationStore `and `MessageExchangeStore`.

Currently, I've only saw implementation for observes (californium and leshan). 

https://github.com/eclipse/californium/issues/311#issuecomment-305957631

I think together "for all request" and "processing must be completed before returning" this is very dangerous for the performance. Therefore I would go for changing the setContext API to enable the implementation to select the way of processing the context (currently only filter for observes).

sophokles73: I agree with @sbernard31 and I am sorry, but I am not able to understand the point of your latest comment, @boaks. Can you please elaborate/rephrase what your intention is?

boaks: @sophokles73  read #311, and I'm pretty sure, you can be more precise in what you not understand or where you have a different view.

I created this issue #311 to discuss this topic there even before this PR. This PR was only the part of obvious optimizations. From the available source alone It's hard to understand, what exactly should be achieved with or must be obeyed for the context correlation processing.

To use (null == correlationContext) for further optimizations would only optimize blockwise. I think we need more optimizations here.

So: start reading #311 and discuss there!

sbernard31: @boaks ok so let's restart to discuss on #311....

boaks:
The PR is updated according the discussion in #311.

If no objections, I will merge it next week.

californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
```diff
@@ -583,10 +586,13 @@ public void setCustomExecutor() {
...
-		ExchangeObserver obs = this.observer;
-		if (obs != null) {
-			obs.contextEstablished(this);
+		if (correlationContext.compareAndSet(null, ctx)) {
+			ExchangeObserver obs = this.observer;
+			if (obs != null) {
+				obs.contextEstablished(this);
+			}
+		} else {
+			correlationContext.set(ctx);
```
sbernard31: Should we really set the new context if there is already one ?

boaks: That's the closest implementation to the current one.

sbernard31: I agree. But this is not really my question ?
I suppose by this response you mean that this is out of scope. That's ok. 
But I re-ask the question here as nobody really answer in https://github.com/eclipse/californium/issues/311#issuecomment-306852749

boaks: https://github.com/eclipse/californium/issues/311#issuecomment-306852749

> My response is "no, it must not change".

https://github.com/eclipse/californium/issues/311#issuecomment-306860252

> I do not see how it could, actually. Am I missing something?


https://github.com/eclipse/californium/issues/311#issuecomment-306860933

> It could be changed each time we send a blockwise request. (If I read the code correctly)

So I don't see, that this discussion is finished.

https://github.com/eclipse/californium/issues/311#issuecomment-306839999

> But I would feel more comfortable, if the discussion in the mailing list came to a conclusion.

That's still my position. Change enough to optimize the "critical callback", but less as possible while it is discussed.

sbernard31: I think you put too many hope on this discussion ;).
But I'm ok to let this point beside for now.

boaks: So LGTM?

sbernard31:
:+1:

PR #317, opened 2017-05-31
Stop retransmissions when we receive expected message.

sbernard31:
We should stop handshake message retransmission as soon as we receive the first message of the expected flight.
(see #228 for discussion about this)

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -1639,6 +1639,10 @@ public class DTLSConnector implements Connector {
 			public void sendFlight(DTLSFlight flight) {
 				sendHandshakeFlight(flight, connection);
 			}
+
+			public void cancelRetransmissions() {
+				connection.cancelPendingFlight();
+			}
 		};
 	}
 
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -1641,7 +1641,11 @@ public class DTLSConnector implements Connector {
 			}
 
 			public void cancelRetransmissions() {
-				connection.cancelPendingFlight();
+				// TODO remove this check when this experimental feature will be
+				// not experimental anymore ^^
+				if (config.isEarlyStopRetransmission()) {
+					connection.cancelPendingFlight();
+				}
 			}
 		};
 	}
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -67,6 +67,11 @@ public final class DtlsConnectorConfig {
 	private X509Certificate[] trustStore = new X509Certificate[0];
 
 	/**
+	 * Experimental feature : Stop retransmission at message receipt
+	 */
+	private boolean earlyStopRetransmission = true;
+
+	/**
 	 * The maximum fragment length this connector can process at once.
 	 */
 	private Integer maxFragmentLengthCode = null;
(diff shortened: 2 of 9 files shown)


sbernard31:
I tested it with #305 by simulating some pskstore latency. It behaves much better.

boaks:
RFC6347, 4.2.4

> 3. The implementation receives the next flight of messages: if this
> is the final flight of messages, the implementation transitions to
> FINISHED. If the implementation needs to send a new flight, it
> transitions to the PREPARING state. Partial reads (whether
> partial messages or only some of the messages in the flight) do
> not cause state transitions or timer resets.

I don't know, why "Partial reads (whether partial messages or only some of the messages in the flight) do not cause state transitions or timer resets." is specified, but before we merge this PR, it would be good to have some infos about that.

sbernard31:
I will ask this on the tls@ietf.org mailing list but I'm not sure this can be applied to the letter to our implementation as we don't handle this exactly as the state machine described in this section.

sbernard31:
Here a pertinent [answer](https://mailarchive.ietf.org/arch/msg/tls/oMQTOOHdujD03XDix9r4nTp59PM).

The problem is about the last flight.
A possible solution is to not stop retransmission on CHANGE CIPHER SPEC as last flight is always CHANGE CIPHER SPEC/ FINISHED ?

boaks:
OK. 
LGTM.

Just as remark: if it's configurable, it would be easily possible to switch back in case of troubles.

sbernard31:
I made the modification to not stop retransmissions on CIPHER SPEC CHANGE message.

About make it configurable, I feel this a bit strange.
If we really want to allow a workaround in case this could be a problem.
I prefer to allow  to override `getRecordLayerForPeer()` rather adding a new DTLS config parameters.

boaks:
The idea of making it configurable is mainly triggered to allow to easily check both "strategies", early and late stops of retransmissions to see, which approach works better in a setup. I don't expect, that one works and the other not. I just expect, that the one has sometime advantages and sometimes the other. Overwritting would make it possible to change behaviour, but it's not that easy as adjust the configuration.
If `DTLSConnector` would have a flag, the new `cancelRetransmissions `could just use that to either call or not call `connection.cancelPendingFlight()`

sbernard31:
> I just expect, that the one has sometime advantages and sometimes the other

I don't agree. I see the current PR as an optimization (bandwidth) and also a way to not disturb a not so stable DTLS implementation.
As we seem not so confident about this PR. I can understand you want a way to deactivate it, waiting we make sure this is the right way to do, but I hope we will not keep this forever.

So I will add a "temporary" config option to deactivate this features. (Hoping we will remove it before the 2.0.0 release)

boaks:
(Hoping we will remove it before the 2.0.0 release)

agreed :-).

sbernard31:
Config added.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1641,7 +1641,10 @@ public void sendFlight(DTLSFlight flight) {
 			}
 
 			public void cancelRetransmissions() {
-				connection.cancelPendingFlight();
+				// TODO remove this check when this experimental feature will be
+				// not experimental anymore ^^
+				if (config.isEarlyStopRetransmission())
```
boaks: {
}
?

sbernard31: Personally,  I prefer without brackets for "single line if", but if you prefer with it I can add it.

boaks: I just prefer a "unique format" :-), 
therefore I think, with {} is, what is configured in the formatter.

sbernard31: Good try about formatter, but this is not the case ;)

boaks:
LGTM

boaks:
@sbernard31 
Should I merge and pick it to master?

sbernard31:
Integrated in 2.0.x branch (commit 660834f97d01813b9f5e4815b0f2e2a5a1e3c26d, f7a83d2bc3b3a8694a3e5e7cc1073f4965a1431e, d4ae205342d0bf7a16679a82ca811ce3e297e79a, 41180981f90d7ac441d5263a9c1efee5f72842f8)

boaks:
@sbernard31 
Is this improvement intended for master also?

sbernard31:
I don't know...
As it is "experimental" we can wait to confirm if it really works before to ask ourself if we back-port it in master ?

PR #319, opened 2017-06-09
Bert branch new

raipratheek:
This is my humble effort to include BERT option to Californium. I badly needed this feature in Californium for my Master Thesis work. 
The changes are based on the draft https://tools.ietf.org/html/draft-ietf-core-coap-tcp-tls-07
I have tested this change in my environment and it works fine.
Please provide your valuable feedback :)

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -290,7 +290,7 @@ public class BlockwiseLayer extends AbstractLayer {
 
 					if (status.getCurrentSzx() == 7) {
 						// For BERT option.
-						status.setCurrentNum(status.getCurrentNum() + (preferredBlockSize / 1024));
+						status.setCurrentNum(status.getCurrentNum() + (request.getPayloadSize() / 1024));
 					} else {
 						status.setCurrentNum(status.getCurrentNum() + 1);
 					}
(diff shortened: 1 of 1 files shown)


boaks:
First thanks for your contribution!

Currently we use the branch 2.0.x for development, it will become master with a 2.0 release.
So I think, your changes should better got to 2.0.x.

Our design pattern for the "Layers" in `org.eclipse.californium.core.network.stack` prefers to isolate TCP/UDP specific implementations in specialized classes used in the `CoapTcpStack` and `CoapUdpStack` constructors (e.g. `TcpObserveLayer` and `ObserveLayer`). For `BlockwiseLayer` I would guess, that subclassing "TcpBlockwiseLayer extends BlockwiseLayer" would limit the required changes in `BlockwiseLayer` and moves the required new implementation in a separated class to the `CoapTcpStack` only.

Depending on the resulting amount of changes required in  `BlockwiseLayer`, I'm not sure if the other committers could be convinced, that this TCP extension doesn't bring in riscs to the UDP part.

So we will see.

sophokles73:
I agree with @boaks in that these changes should (if at all) be targeted at the 2.0.x branch.
That said, I do not think that we are currently in a position to support BERT when we do not even have a working implementation of TCP transport. FMPOV we first should worry about getting a TCP implementation based on the latest draft up and running. Once we have that we can add support for BERT. Would you be interested in helping with that?

raipratheek:
Thank you for your views. I would definitely like to contribute to the project :)

boaks:
FMPOV BERT is one part of the missing stuff. I would not make it mandatory, that the other missing parts (e.g. signalling) must be also available.

But currently I don't want to risc the UDP stuff for TCP stuff. So let's see, how much in the common `BlockwiseLayer` must be changed. If the risc seems to get to high, I would start thinking of an "preview branch".

raipratheek:
new PL created for BERT. Hence closing this.

PR #320, opened 2017-06-13
Add get and has to support configuration completion.

boaks:
See leshan PR # 350.


boaks:
@sbernard31 
Does this now work for leshan?

sophokles73:
why do you need getters on the `Builder`?

boaks:
https://github.com/eclipse/leshan/pull/350

I don't know, if its really needed or more an experiment.

Issue #173, opened 2016-12-07
Do not identify observation by token only.


sbernard31:
Currently, Observation is identify by token in observation store.

We should identify it by token + a peer identifier.

I recommend that peer identifier was : 
    - peer-address for unsecure connection
    - DTLS identity for secure connection or session if there is not authentication

As there is no real consensus on this, I suppose we should find a way to allow users to bring its own behavior.

sophokles73:
Ok, here's a proposal:

we introduce a new interface `EndpointIdentifier` which we use in place of `InetSocketAddress` everywhere throughout the system. We can then have multiple implementations of it, e.g. one based on `InetSocketAddress` when using UDP only and another one based on the DTLS client identity or session ID. We might even evolve the `CorrelationContext` into this concept.

WDYT?

sbernard31:
It sounds good.

sophokles73:
Just to be clear, this basically means replacing the InetAddress/port combination everywhere throughout Californium/Scandium with a source and destination `EndpointIdentifier`, e.g. the `Message` class would no longer have `getSourceAddress()` nor `getSourcePort()` methods but only a `getSourceEndpointIdentifier()` method which would not necessarily contain an IP address and port at all, e.g. if the transport would be SMS or NB-IoT where addressing is not IP based AFAIK ...

Still sounds like a good idea? My feeling is that we would need to change around 30 percent of the classes in the project ... don't get me wrong, I believe this is the right thing to do but it will take some time to actually do it ...

sbernard31:
I thought you just plan to change the way we identify observation not all messages.
I understand the need to add a sourcepointIdentifier, but I'm not sure this is a good idea to replace data like sourceAddress/sourcePort. This seems to be valuable data.

sophokles73:
That data would still be available, but only as part of the `EndpointIdentifier`, i.e. you could do something like `UdpEndpointIdentifier.getAddress()` to retrieve the peer's IP address/port.

boaks:
Though currently the observe is the trigger, I could life with reduce the scope to that for the first step.
But in general I think, that it should be used later consequently.

sbernard31:
Let's think about the API.

Currently the API looks like this :
```java
// create request
Request request = new Request(Code.GET);
request.setObserve();
request.setURI("my/path");
request.setDestination(destinationAddress)

// listen new notification
endpoint.addNotificationListener(new NotificationListener() {
    @Override
    public void onNotification(Request request, Response response) {
[...]

PR #322, opened 2017-06-15
Add endpoint identifier.

sbernard31:
Here is a try to define an identifier for foreign endpoint.
The PR was largely inspired by this long [discussion](https://github.com/eclipse/californium/issues/173#issuecomment-290360652).

This is a work in progress and should not be merged now.
This first step introduces the concept of endpoint identifier and uses it to identify observation.

Next step : Adding `Principal` to `CorrelationContext` to be able to define `PrincipalEndpointIdentifier` to be able to test that in Leshan.

While this first step, I had the feeling that maybe `CorrelationContext` could be not necessary at CoAP side anymore (only at Connector side). This is a track I would explore. This could simplify the code and also solve some questions like #311.

Another point will be to change TokenProvider (replace `reserve/release` by `random generation`)  to solve issue raised by Achim in #173.

Even, if this is a work in progress, I strongly advice to try to review this PR each time I add new commits. This is the only way to make this review digestible.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
@@ -31,6 +31,7 @@ import java.util.logging.Logger;
 import org.eclipse.californium.core.coap.MessageObserverAdapter;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
+import org.eclipse.californium.core.identifier.EndpointIdentifier;
 import org.eclipse.californium.core.network.Exchange.KeyToken;
 import org.eclipse.californium.core.network.Exchange.Origin;
 import org.eclipse.californium.core.network.config.NetworkConfig;
@@ -118,7 +119,7 @@ public abstract class BaseMatcher implements Matcher {
 	 * 
 	 * @param request observe request.
 	 */
-	protected final void registerObserve(final Request request) {
+	protected final void registerObserve(final Exchange exchange, final Request request) {
 
 		// We ignore blockwise request, except when this is an early negotiation
 		// (num and M is set to 0)
@@ -127,7 +128,6 @@ public abstract class BaseMatcher implements Matcher {
 			// add request to the store
 			final KeyToken idByToken = KeyToken.fromOutboundMessage(request);
 			LOG.log(Level.FINER, "registering observe request {0}", request);
-			observationStore.add(new Observation(request, null));
 			// remove it if the request is cancelled, rejected, timedout, or send error
 			request.addMessageObserver(new MessageObserverAdapter() {
 				@Override
@@ -137,8 +137,13 @@ public abstract class BaseMatcher implements Matcher {
 				
 				@Override
 				protected void failed() {
-					observationStore.remove(request.getToken());
+					observationStore.remove(request.getDestinationEndpoint(), request.getToken());
 					exchangeStore.releaseToken(idByToken);
+				}
+
+				public void onDestinationEndpointDefined(EndpointIdentifier destination) {
+					LOG.log(Level.FINER, "registering observe request {0}", request);
+					observationStore.add(destination, new Observation(request, exchange.getCorrelationContext()));
 				}
 			});
 		}
@@ -158,7 +163,8 @@ public abstract class BaseMatcher implements Matcher {
 		final Exchange.KeyToken idByToken = Exchange.KeyToken.fromInboundMessage(response);
 		Exchange exchange = null;
 
-		final Observation obs = observationStore.get(response.getToken());
+		final EndpointIdentifier sourceEndpoint = response.getSourceEndpoint();
+		final Observation obs = observationStore.get(sourceEndpoint, response.getToken());
 		if (obs != null) {
 			// there is an observation for the token from the response
 			// re-create a corresponding Exchange object for it so
@@ -185,7 +191,7 @@ public abstract class BaseMatcher implements Matcher {
 						LOG.log(Level.FINE,
 								"Response to observe request with token {0} does not contain observe option, removing request from observation store",
 								idByToken);
-						observationStore.remove(request.getToken());
+						observationStore.remove(sourceEndpoint, request.getToken());
 						exchangeStore.releaseToken(idByToken);
 					} else {
 						notificationListener.onNotification(request, resp);
@@ -199,7 +205,7 @@ public abstract class BaseMatcher implements Matcher {
 
 				@Override
 				protected void failed() {
-					observationStore.remove(request.getToken());
+					observationStore.remove(sourceEndpoint, request.getToken());
 					exchangeStore.releaseToken(idByToken);
 				}
 			});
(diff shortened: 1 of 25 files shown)


boaks:
To be frank:
Though the discussion about RFC7252, section 9.1.2 has started again, I would not change the API too much until that discussion seems to get a direction "with future".

So in my view, providing the "intended credentials or parts of the credential as identity" ahead when sending a Message, should overcome the callbacks with that information at all. The calback are only required, if you bind the matching to the "created session". 
Therefore I would **not add** the callbacks for the "destination endpoint". To use these credentials as identity, these credential could not be reused (e.g. groups of devices with same credentials), but until someone came up with that, I would not care too much about that.

For notifies received from changed addresses, the credential are also available on receive and could be used to find the proper exchange (and adjust the ip-address for future communication to that device). Again, this requires unique credentials for all devices.

boaks:
Not to change the API to much, my intention was to put that "identity" into the correlation context and add a optional correlation context also for sending a request.

sbernard31:
>  I would not change the API too much until that discussion seems to get a direction "with future".

I'm not so confident. I don't think a final solution will be proposed soon :/

>The calback are only required, if you bind the matching to the "created session".

Callback is necessary for use case where peer don't use authentication, or where the identifier chosen could not be know by the user before (e.g. session ID)

>Therefore I would not add the callbacks for the "destination endpoint"

At short term, I need it for backward compatibility. For reason exposed above, I'm not sure we will be able to remove it. But If we can I agree to remove it. (Less code we have better, it is)

> To use these credentials as identity, these credential could not be reused (e.g. groups of devices with same credentials), 

That's why having a EndpoinIdentifierFactory is a good way to let users choose what they want to use as "identifier"

> Not to change the API to much, my intention was to put that "identity" into the correlation context and add a optional correlation context also for sending a request.

I rather prefer to remove the CorrelationContext to CoapStack and just let CorrelationContext at element-connector level.

Unless you have strong opinion against my first step, I will continue in that way more in a POC spirit. We have this "identifier" issue for a long time ago and I would like to move forward. I'm sure I will face issue I don't think about for now.

boaks:
> I'm not so confident. I don't think a final solution will be proposed soon :/

I agree, that it will not be proposed soon. 

> Callback is necessary for use case where peer don't use authentication, or where the identifier chosen could not be know by the user before (e.g. session ID)

That's exactly the **NONE** use case for me. But if you want to solve that, try it. 
I just don't understand, why additional values in the correlation context will not do it.

sbernard31:
> I just don't understand, why additional values in the correlation context will not do it.

You mean instead of using endpoint identifier ?

boaks:
Yes. 
Add the "stringified credential identity" into it and add a optional correlation context to the send(request).

sbernard31:
And we could use a partial/incomplete correlation context on send() ?
Is this context will be used as identifier for observe ? (correlationContext+token)

boaks:
> And we could use a partial/incomplete correlation context on send() ?

Therefore the exchangeable correlation context matcher was introduced. My idea was to make a matcher, which uses the identity.

> Is this context will be used as identifier for observe ? (correlationContext+token)

My  idea was to use the "identity" to adjust the "address".

sbernard31:
> My idea was to use the "identity" to adjust the "address".

That's mean you identify observation by (address+token) ?

boaks:
The `UdpMatcher.receiveResponse()` currently uses address + token to lookup for the exchange.

So my idea (not detailed right now) was, that a correlation context may also be used for adjust addresses before this lookup.

sbernard31:
So you should have a kind of index {identity => keys (address+token) }?
Then update keys with the new address. Hopping that news address was not already used.
Does it simpler to just use identity+token instead ?

boaks:
My basic idea is to add a "identifier" (say the psk identity) to the correlation context and pass it in with the request. The connector the decided, if the "identifier" matches using the correlation context matcher. On receiving a message, the connector adds the identity to the correlation context.
The Exchange-Matcher tries to match the incoming message as it does now (so normal request/response mapping will not work with address changes). If no exchange is found, but the message is identified as notify, the ObservationStore is considered. To possibly used the identity there (and to solve issue #173), the ObservationStore API must be extended. Though its no detailed design, sure there may occur some follow-up issues. But for now this seems to me to be the smallest change with the required effect. It depends then on the  ObservationStore implementation, how to match, if no identity is used. If someone would like to us the token only approach, it would be his decision.

californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
```diff
@@ -158,7 +163,8 @@ protected final Exchange matchNotifyResponse(final Response response, final Corr
 		final Exchange.KeyToken idByToken = Exchange.KeyToken.fromInboundMessage(response);
 		Exchange exchange = null;
 
-		final Observation obs = observationStore.get(response.getToken());
+		final EndpointIdentifier sourceEndpoint = response.getSourceEndpoint();
```
boaks: But, where is that "reponse source endpoint" set during receiving a response?
That's the core of my question, on what "protocol artifact" your are going to build the identity.

sbernard31: This is set in [coapEndpoint](https://github.com/eclipse/californium/blob/endpoint_identifier/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java#L748)

And the [`IdentifierFactory`](https://github.com/eclipse/californium/blob/endpoint_identifier/californium-core/src/main/java/org/eclipse/californium/core/identifier/IdentifierFactory.java) is responsible to define what should be used as Identifier.

(For Leshan I want to use Principal)

sbernard31: The `IdentifierFactory` is responsible to translate Identifier in context and context in Identifier.
(translation is centralized in this class)

I have in mind that context could be a connector concept and Identifier a CoAP stack one.

boaks: OK.

So the main difference seems to be, that I would just pass the `CorrelationContext `into the `ObserveStore`, and you would prefer to have a "IdentifierFactory" to convert that `CorrelationContext` ahead into a `EndpointIdentifier`.

californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
```diff
@@ -137,9 +137,14 @@ public void onCancel() {
...
 				@Override
 				protected void failed() {
-					observationStore.remove(request.getToken());
+					observationStore.remove(request.getDestinationEndpoint(), request.getToken());
 					exchangeStore.releaseToken(idByToken);
 				}
+
+				public void onDestinationEndpointDefined(EndpointIdentifier destination) {
+					LOG.log(Level.FINER, "registering observe request {0}", request);
+					observationStore.add(destination, new Observation(request, exchange.getCorrelationContext()));
```
boaks: If we delay the `add` until `onDestinationEndpointDefined`, I'm not sure if we require the `remove `in `failed()`

californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
sbernard31: Yep.

californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
sbernard31: I'm not sure the destination endpoint is defined just after an handshake success, but the APPLICATION _DATA sending can fail.

sbernard31:
This will never be merged.
I hope the idea to have a better identifier for observations will be not dropped too :)

PR #324, opened 2017-06-20
Fix rare race condition in block1-blockwise transfe.

boaks:
If the generated token is copied after sending the message, the response
may already being processed without that token in Block1BlockwiseStatus.
Therefore introduce setBuilt()/onBuilt() callback to MessageObservers.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Message.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Message.java
@@ -25,6 +25,12 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - use unmodifiable facade
  *                                                    instead of create it on
  *                                                    every getMessageObservers()
+ *    Achim Kraus (Bosch Software Innovations GmbH) - make more messaging states thread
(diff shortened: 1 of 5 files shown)


boaks:
If someone has a better naming then `setBuilt()`/`onBuilt()`, don't hesitate to make proposals.

boaks:

sophokles73:
> If someone has a better naming then setBuilt()/onBuilt(), don't hesitate to make proposals.

What about `beforeSend()`?

boaks:
> // when the request for transferring the first block
> // has been sent out, we copy the token to the
> // original request so that at the end of the
> // blockwise transfer the Matcher can correctly
> // close the overall exchange
> 

So my intention was to call the `setBuilt()` also when the request is canceled to be prepared for the cleanup. So `beforeSend()` may be over-/miss-interpreted.

sophokles73:
then maybe `readyToSend()`?

boaks:
OK, that sounds good for me.

sbernard31:
👍

PR #325, opened 2017-06-20
Fix rare race condition for MID on responses.

boaks:
Proactive observe cancellation may cause errors, if they cancel not
completely created notifies (before the MID is assigned).

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -40,6 +40,8 @@
  *                                                    rare race condition in block1wise
  *                                                    when the generated token was copied
  *                                                    too late (after sending). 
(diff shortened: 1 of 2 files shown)


boaks:

sbernard31:
I do not get this one.

boaks:
We had this already for requests:
The MID is assigned during the "send process in the ExchangeMatcher". If parallel something calls "cancel", you may see a `IllegalArgumentException("MID must be a 16 bit unsigned int: " + mid);`. I saw this only once on my test machines with the `UdpMatcher.removeNotificationsOf()` in the stacktrace during 
`MemoryLeakingHashMapTest.testObserveProactive()`.

sbernard31:
I suppose it's ok.

We faced a lot of race condition, I agree that the best way at short/mid term is to fix it in that way.
But this make me feel that we could have a global design issue. At long term we should maybe consider to introduce something like `StripedExecutor` like for DTLS. But I'm not sure we have a good candidate for the "stripe", maybe the endpoint Identifier ? I don't know  I'm just thinking ...

PR #326, opened 2017-06-20
Add isEmptyWithDump() for more details in logs while testing.

boaks:

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
@@ -24,6 +24,10 @@
  *                                                    integrate clear() into stop()
  *    Achim Kraus (Bosch Software Innovations GmbH) - remove setContext().
  *                                                    issue #311
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add dump() and call it from
+ *                                                    isEmpty depending on the log-level.
+ *                                                    Intended to be used for logging
+ *                                                    while testing.
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -32,6 +36,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executors;
@@ -124,6 +129,7 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 		StringBuilder b = new StringBuilder("MessageExchangeStore contents: ");
 		b.append(exchangesByMID.size()).append(" exchanges by MID, ");
 		b.append(exchangesByToken.size()).append(" exchanges by token, ");
+		b.append(deduplicator.size()).append(" MIDs, ");
 		return b.toString();
 	}
 
@@ -163,8 +169,16 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 
 	@Override
 	public boolean isEmpty() {
-		LOGGER.finer(dumpCurrentLoadLevels());
-		return exchangesByMID.isEmpty() && exchangesByToken.isEmpty() && deduplicator.isEmpty();
+		boolean empty = exchangesByMID.isEmpty() && exchangesByToken.isEmpty() && deduplicator.isEmpty();
+		// enable dump exchanges with WARNING, if log level is FINEST
+		boolean verbose = !empty && LOGGER.isLoggable(Level.FINEST);
+		dump(verbose ? Level.WARNING : Level.FINER, 3);
+		return empty;
+	}
+
+	@Override
+	public String toString() {
+		return dumpCurrentLoadLevels();
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/MessageExchangeStore.java
```diff
@@ -181,6 +184,15 @@
 	boolean isEmpty();
 
 	/**
+	 * Checks if there are any exchanges currently being registered in this
+	 * store and dump these exchanges. Intended to be used for more details in
+	 * log when testing.
+	 * 
+	 * @return {@code true} if there no exchanges registered.
+	 */
+	boolean isEmptyWithDump();
```
sbernard31: Should we really add this to the interface ? Maybe add this to the Implementation is enough ?

boaks: Yes, we should really add it there.

sbernard31: Why ? It seems to me that this is for unit testing purpose only.  And AFAIK, we have only 1 implementation to test and I bet we will not add a new one at short/mid-term.

boaks: For me this seems to be the classic "waste of time" discussion!

First, the main cause of this discussion seems to be, that the `MessageExchangeStore` / `InMemoryMessageExchangeStore` is stale but just not removed. In consequence, the discussion goes for "how should the new stuff be introduced?" Assuming, that this pattern (interface/implementation) is still valid or that it is already deprecated?

The arguments for placing the `isEmptyWithDump` are the same as for the already available `isEmpty`. So from my point of view, the task "adding more logging" MUST place the `isEmptyWithDump` in the same way then `isEmpty`. 
Redesign `MessageExchangeStore` / `InMemoryMessageExchangeStore` is a different task and may change that placing again.

sbernard31: I suppose `MessageExchangeStore` interface was introduced to let users implement their own `MessageExchangeStore`. This adds flexibility, so it  sounds good idea.

An interface is a contract with our users, if we add this method we force users to implement it although we don't use it. This seems a bad idea.  Especially because isEmptyWithDump has a really limited scope of usefulness.

FMPOV, even having `isEmptyWithDump` in `InMemoryExchangeStore` sounds strange. I will just go for adding getter for  `ExchangeMaps (only in implementation)` and let `Tests` classes the responsibility to log it.

> The arguments for placing the isEmptyWithDump are the same as for the already available isEmpty

I don't know the arguments about why we add `isEmpty` in the interface. It's seems that we used it only in tests. I maybe validated a PR with this addition. Maybe I didn't notice it because having an `isEmpty` method on a `Store` interface is usual.

I share my thought with you. If you still disagree, too bad :). I suppose my arguments was not strong enough.

Anyway, I don't think that thinking about "what is part of the API" or "what is the responsability of a class" is a "waste of time". No offense but what kind of feedbacks would you expect when you open a PR ? Personally I check for bugs, code readability and code organization :)

sophokles73: I tend to agree with Simon that `isEmptyWithDump` should not be part of the interface. I also do not see why we need a separate method for this at all. Why don't we simply make `InMemoryMessageExchangeStore.isEmpty` log more detailed information if a finer log level is active, e.g. FINEST?

boaks: :-) Because in the meantime I poll "isEmpty()" and fail with a more relaxed timeout :-).

Though the polling would cause a lot of logging, I introduced the isemptyWithDump for the final check, to see, if it's not empty, what is left.

Please remember:
The eclipse CI build infrastructure doesn't grant a minimum CPU quote and starvation seems to affect some, but not all threads. So the test thread may execute while the sweeper may starve much more. Therefore the relaxed times are required.
   
Currently I just merge this into the "hudson_debug" branch, if I want to analyse a specific failing test. Any other hudson build failures with not empty exchange stores, I simply ignore in the meantime, because with no additional information it's a waste of time.

boaks:
I changed the "isEmptyWithDump" into a variant selecting the dump dynamically via the logging level.
Just to mention: that approach, with dynamical log level will not work with slf4j (at least I don't know the trick there :-) ).

sbernard31:
This is more acceptable like that for me but I still think it's better to keep the "core" code as simple as possible and let users be able to explore ExchangeStore content.

This is better for us for maintainability and better for users for flexibility.

Anyway, does it make sense to integrate it like that if we know we will go with slf4j and this doesn't work with it ?

sophokles73:
> Just to mention: that approach, with dynamical log level will not work with slf4j (at least I don't know the trick there :-) )

If this is about whether it is possible to determine the log level during runtime you can relax, that's possible with slf4j as well. `org.slf4j.Logger` has several `isXXXEnabled` methods where `XXX` can be any of the log levels ...

californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
```diff
@@ -0,0 +1,98 @@
...
+		Level level = STORE_LOGGER.getLevel();
+		try {
+			waitUntilDeduplicatorShouldBeEmpty(exchangeLifetime, sweepInterval, new CheckCondition() {
+
+				@Override
+				public boolean isFulFilled() throws IllegalStateException {
+					return clientExchangeStore.isEmpty() && serverExchangeStore.isEmpty();
+				}
+			});
+			STORE_LOGGER.setLevel(Level.FINEST);
```
boaks: That's the challenge for slf4j; setLevel.

boaks:
https://ci.eclipse.org/californium/job/master/288/display/redirect?page=changes

> Server side message exchange store still contains exchanges


So, without more information about the "still contained exchanges" such rare errors could not be fixed, at least not by me  ;-(.

sbernard31:
Replaced by #504.

PR #327, opened 2017-06-20
Redesign test to introduce lower notify rate for blockwise notifies.

boaks:
Ensure also longer timeout for blockwise notifies to ensure, that a
retransmitted message doesn't cause the test to fail.


boaks:

PR #328, opened 2017-06-21
Introduce configurable message ID trackers.

boaks:
Rename MessageIdTracker to HashedMessageIdTracker. Introduce
MessageIdTracker interface to support different implementations.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/GroupedMessageIdTracker.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/GroupedMessageIdTracker.java
@@ -13,12 +13,14 @@
  * Contributors:
  *    Bosch Software Innovations - initial creation
  *                                 (derived from MessageIdTracker)
+ *    Achim Kraus (Bosch Software Innovations GmbH) - introduce MessageIdTracker
+ *                                                    interface
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
-import java.security.SecureRandom;
 import java.util.concurrent.TimeUnit;
 
+import org.eclipse.californium.core.coap.Message;
 import org.eclipse.californium.core.network.config.NetworkConfig;
 
 /**
@@ -37,13 +39,10 @@ import org.eclipse.californium.core.network.config.NetworkConfig;
  * little longer to use the first MIDs of a group because they freed with the
  * lease of the last MID of the group.
  */
-public class GroupedMessageIdTracker {
+public class GroupedMessageIdTracker implements MessageIdTracker {
 
-	private static final int TOTAL_NO_OF_MIDS = 1 << 16;
 	/**
 	 * Number of groups.
-	 * 
-	 * Configurable using {@link NetworkConfig.Keys#MID_PROVIDER_GROUPS}.
 	 */
 	private final int numberOfGroups;
 	/**
@@ -55,7 +54,7 @@ public class GroupedMessageIdTracker {
 	 * 
 	 * @see System#nanoTime()
 	 */
-	private final long exchangeLifetime;
+	private final long exchangeLifetimeNanos;
 	/**
 	 * Array with end of lease for MID groups. MID divided by
 	 * {@link #sizeOfGroups} is used as index. Values in nanoseconds.
@@ -69,34 +68,26 @@ public class GroupedMessageIdTracker {
 	private int currentMID;
 
 	/**
-	 * Creates a new tracker based on configuration values.
-	 * <p>
+	 * Creates a new MID group based tracker.
+	 * 
 	 * The following configuration values are used:
 	 * <ul>
-	 * <li>{@link org.eclipse.californium.core.network.config.NetworkConfig.Keys#MID_PROVIDER_GROUPS}
+	 * <li>{@link org.eclipse.californium.core.network.config.NetworkConfig.Keys#MID_TRACKER_GROUPS}
 	 * - determine the group size for the message IDs. Each group is marked as
 	 * <em>in use</em>, if a MID within the group is used.</li>
 	 * <li>{@link org.eclipse.californium.core.network.config.NetworkConfig.Keys#EXCHANGE_LIFETIME}
 	 * - each group of a message ID returned by <em>getNextMessageId</em> is
 	 * marked as <em>in use</em> for this amount of time (ms).</li>
-	 * <li>{@link org.eclipse.californium.core.network.config.NetworkConfig.Keys#USE_RANDOM_MID_START}
-	 * - if this value is {@code true} then the message IDs returned by
-	 * <em>getNextMessageId</em> will start at a random index. Otherwise the
-	 * first message ID returned will be {@code 0}.</li>
 	 * </ul>
 	 * 
-	 * @param config the configuration to use.
+	 * @param initialMid initial MID
+	 * @param config configuration
 	 */
-	public GroupedMessageIdTracker(final NetworkConfig config) {
-		numberOfGroups = config.getInt(NetworkConfig.Keys.MID_PROVIDER_GROUPS);
+	public GroupedMessageIdTracker(int initialMid, NetworkConfig config) {
+		exchangeLifetimeNanos = TimeUnit.MILLISECONDS.toNanos(config.getLong(NetworkConfig.Keys.EXCHANGE_LIFETIME));
+		numberOfGroups = config.getInt(NetworkConfig.Keys.MID_TRACKER_GROUPS);
+		currentMID = initialMid;
 		sizeOfGroups = (TOTAL_NO_OF_MIDS + numberOfGroups - 1) / numberOfGroups;
-		exchangeLifetime = TimeUnit.MILLISECONDS.toNanos(config.getLong(NetworkConfig.Keys.EXCHANGE_LIFETIME));
-		boolean useRandomFirstMID = config.getBoolean(NetworkConfig.Keys.USE_RANDOM_MID_START);
-		if (useRandomFirstMID) {
-			currentMID = new SecureRandom().nextInt(TOTAL_NO_OF_MIDS);
-		} else {
-			currentMID = 0;
-		}
 		midLease = new long[numberOfGroups];
 	}
 
@@ -109,16 +100,17 @@ public class GroupedMessageIdTracker {
 	public int getNextMessageId() {
 		final long now = System.nanoTime();
 		synchronized (this) {
-			int mid = currentMID % TOTAL_NO_OF_MIDS;
+			// mask mid to the 16 low bits
+			int mid = currentMID & 0x0000FFFF;
 			int index = mid / sizeOfGroups;
 			int nextIndex = (index + 1) % numberOfGroups;
 			if (midLease[nextIndex] < now) {
-				midLease[index] = now + exchangeLifetime;
+				midLease[index] = now + exchangeLifetimeNanos;
 				++currentMID;
 				return mid;
 			}
 		}
-		return -1;
+		return Message.NONE;
 	}
 
 	/**
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
@@ -15,11 +15,17 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - remove nested synchronize.
  *                                                    reduce blocking on 
  *                                                    tracker.getNextMessageId()
+ *    Achim Kraus (Bosch Software Innovations GmbH) - introduce MessageIdTracker
+ *                                                    interface and rename old
+ *                                                    MessageIdTracker to
+ *                                                    MapBasedMessageIdTracker.
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
 import java.net.InetSocketAddress;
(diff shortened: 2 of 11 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/GroupedMessageIdTracker.java
```diff
@@ -113,12 +94,12 @@ public int getNextMessageId() {
...
 			int nextIndex = (index + 1) % numberOfGroups;
 			if (midLease[nextIndex] < now) {
-				midLease[index] = now + exchangeLifetime;
+				midLease[index] = now + exchangeLifetimeNanos;
 				++currentMID;
 				return mid;
 			}
 		}
-		return -1;
+		return Message.NONE;
```
sophokles73: :+1:

californium-core/src/main/java/org/eclipse/californium/core/network/HashedMessageIdTracker.java
```diff
@@ -0,0 +1,92 @@
...
+ * 
+ * <pre>
+ * The same Message ID MUST NOT be reused (in communicating with the
+   same endpoint) within the EXCHANGE_LIFETIME (Section 4.8.2).
+ * </pre>
+ */
+public class HashedMessageIdTracker implements MessageIdTracker {
+
+	private final Map<Integer, Long> messageIds;
+	private final long exchangeLifetime; // milliseconds
```
sophokles73: any reason why these are millis whereas the `GroupedMessageIdTracker` uses nanos?

californium-core/src/main/java/org/eclipse/californium/core/network/HashedMessageIdTracker.java
```diff
@@ -0,0 +1,92 @@
...
+ * <p>
+ * According to the
+ * <a href="https://tools.ietf.org/html/rfc7252#section-4.4">CoAP spec</a>
+ * 
+ * <pre>
+ * The same Message ID MUST NOT be reused (in communicating with the
+   same endpoint) within the EXCHANGE_LIFETIME (Section 4.8.2).
+ * </pre>
+ */
+public class HashedMessageIdTracker implements MessageIdTracker {
```
sophokles73: I think that the term `Hashed` is a little misleading here. To me it suggests that some magic happens making sure that no sensitive data is kept explicitly but only hashed values (which obviously is not the case).

So, maybe call it `MapBasedMessageIdTracker` or `SimpleMessageIdTracker`?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
```diff
@@ -31,25 +37,64 @@
...
+	private final long exchangeLifetimeMillis;
 
 	/**
 	 * Creates an new provider for configuration values.
 	 * 
-	 * @param config the configuration to use. In particular, the <em>EXCHANGE_LIFETIME</em>
-	 *         configuration parameter is used as the period of time a message ID is marked as
-	 *         <em>in use</em> when it is allocated for a message exchange.
+	 * The following configuration values are used:
+	 * <ul>
```
sophokles73: what about `NetworkConfig.Keys.MID_TRACKER`?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
```diff
@@ -31,25 +37,64 @@
...
-		this.config = config;
-		trackers = new LeastRecentlyUsedCache<>(
-				config.getInt(NetworkConfig.Keys.MAX_ACTIVE_PEERS, 150000),
-				config.getLong(NetworkConfig.Keys.MAX_PEER_INACTIVITY_PERIOD, 10 * 60)); // 10 minutes
+		String textualMode = null;
+		TrackerMode mode = TrackerMode.GROUPED;
+		try {
+			textualMode = config.getString(NetworkConfig.Keys.MID_TRACKER);
+			mode = TrackerMode.valueOf(textualMode);
+		} catch (Throwable t) {
```
sophokles73: why do we need to catch an exception here? Can't we simply check for `null`s?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
```diff
@@ -31,25 +37,64 @@
...
+			if (null == textualMode) {
+				throw new IllegalArgumentException("Tracker mode unknown!");
+			} else {
+				throw new IllegalArgumentException("Tracker mode '" + textualMode + "' not supported!");
+			}
+		}
+		this.mode = mode;
+		this.exchangeLifetimeMillis = config.getLong(NetworkConfig.Keys.EXCHANGE_LIFETIME);
```
sophokles73: I do not really like the fact that we need to determine the union of all properties used by all tracker implementations here. Shouldn't we simply pass the `NetworkConfig` to the particular tracker instance (as it was before your change)?

californium-core/src/main/java/org/eclipse/californium/core/network/NullMessageIdTracker.java
```diff
@@ -0,0 +1,58 @@
...
+ * <a href="https://tools.ietf.org/html/rfc7252#section-4.4">CoAP spec</a>
+ * 
+ * <pre>
+ * The same Message ID MUST NOT be reused (in communicating with the
+   same endpoint) within the EXCHANGE_LIFETIME (Section 4.8.2).
+ * </pre>
+ * 
+ * This implementation just increments the MIDs and ignores the RFC7252, 4.8.2.
+ */
+public class NullMessageIdTracker implements MessageIdTracker {
```
sophokles73: Why is this a `Null` `MessageIdTracker`? What about `SingleCounterMessageIdTracker` or something that gives a hint how it works?

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
```diff
@@ -57,12 +60,27 @@
...
+	 */
+	public static final int DEFAULT_MID_TRACKER_GROUPS = 16;
+
+	/**
+	 * The default exchange lifetime in milliseconds.
+	 * <p>
+	 * The default value is 247s.
 	 */
-	public static final int DEFAULT_MID_PROVIDER_GROUPS = 16;
+	public static final long DEFAULT_EXCHANGE_LIFETIME = 247 * 1000;
```
sophokles73: I have often thought about making the `EXCHANGE_LIFETIME` a derived value from the other config params since the way we currently do it leaves a lot of room for mis-configuration ...

californium-core/src/main/java/org/eclipse/californium/core/network/HashedMessageIdTracker.java
boaks: I will go for `MapBasedMessageIdTracker`. (next week :-) )

californium-core/src/main/java/org/eclipse/californium/core/network/HashedMessageIdTracker.java
boaks: The millis are used by the previous implementation. I didn't spent too much time in normalize that, because I guess, this implementation will be removed in the future.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
boaks: OK.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
boaks: `valueOf` may throw a IllegalArgumentException, or?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
boaks: My idea was mainly triggered by the "mid" intialization, where the code was cloned to the implementations. So currently we use 4 values, two must be evaluated at this level. For the other two, I felt, that the config is somehow more. But it can be changed back for the exchangelifetime and midgroups.

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
boaks: The value (including a assumed maxium processing delay) is almost a guess. But, yes we can calculate the default from other provided values.

californium-core/src/main/java/org/eclipse/californium/core/network/NullMessageIdTracker.java
boaks: `Null` implies my own opinion about the implementation. It does **null** tracking :-) .
I want to keep this name and the config value NULL to express that.

boaks:
@sophokles73 
FMPOV the comments should be now included in the PR.
(I kept the naming for the Null...)

californium-core/src/main/java/org/eclipse/californium/core/network/NullMessageIdTracker.java
sophokles73: I see

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageIdProvider.java
sophokles73: Yes, but ni most enums we defined in Cf we have defined a static method for getting a value for e.g. its code which then returns `null` if the code is not defined ... but I guess this is a matter of preference. Personally, I think this should at least be narrowed down to catching `IllegalArgumentException` explicitly for clearness...

boaks:
Now catches IllegalArgument and NullPointer instead of Throwable.

californium-core/src/main/java/org/eclipse/californium/core/network/NullMessageIdTracker.java
```diff
@@ -0,0 +1,58 @@
...
+		currentMID.set(initialMid);
+	}
+
+	/**
+	 * Gets the next message ID.
+	 * 
+	 * @return a message ID.
+	 */
+	public int getNextMessageId() {
+		return currentMID.getAndIncrement() % TOTAL_NO_OF_MIDS;
```
sbernard31: This doesn't  works when `currentMID` will exceed Integer.MAX_VALUE, it will become negative (MIN_VALUE).
The modulo of this negative number will be negative too.
Here is what happen
```
 currentMID | NextMessageID
----------------------------
 2147483644 | 65532
 2147483645 | 65533
 2147483646 | 65534
 2147483647 | 65535
-2147483648 | 0
-2147483647 | -65535
-2147483646 | -65534
-2147483645 | -65533
```

You should do something like : 
```java
     if (currentMID.get() + 1 >= TOTAL_NB_OF_MIDS) {
         currentMID.set(0);
     } else {
         currentMID.incrementAndGet();
    }
}
```
And do a modulo in constructor.

sophokles73: Hmm, aren't the numbers correct when interpreting them as unsigned 16bit integers, i.e. 

the binary representation of -65535 as a signed 32bit integer is

11111111 11111111 00000000 00000001, isn't it? So, if you just use the 16 least significant bits and interpret them as an unsigned int, then you get
00000000 00000001 = 1
right?

sbernard31: At encoding, with our current code, when you will write the first 16bits, this will works. I mean message will be correctly encoded. But you could imagine an encoder which raise an error if the MID > 65536, it wouldn't surprise me.

And more important, all the time before the encoding you will use a negative value. I mean Message.getMID() will return a negative value, Log will display a negative value. That doesn't sound good.

boaks:
@sbernard31 
Thanks! It would have taken a long time to detect that, because we would have to sent a lot of messages without being interrupted ;-).

californium-core/src/main/java/org/eclipse/californium/core/network/NullMessageIdTracker.java
sophokles73: very true :-)

boaks:
I cleanup the MID integer overflow for the other trackers as well.

sbernard31:
It's completely a detail, but I consider the hexa or binary notation for bitmask more readable:
```java
// mask result to the 16 low bits
return currentMID.getAndIncrement() & 0x0000FFFF;
```
instead of
```java
// mask result to the 16 low bits
return currentMID.getAndIncrement() & (TOTAL_NO_OF_MIDS - 1);
```

boaks:
Would it be OK to introduce a symbolic mask, say MID_MASK = 0x0000FFFF ?

sbernard31:
As the comment above is _"mask result to the 16bits low bits"_, I think it's just fine to use   `0x0000FFFF` instead of a constant/attrbute `MID_MASK`.

boaks:
Done with 0x0000FFFF

sbernard31:
LGTM

PR #331, opened 2017-06-21
Add parameter to enable coap+tcp server also.

boaks:
Use "coap+tcp:" as parameter when starting the HelloWorldServer to
enable CoAP over TCP connectors.
Use "coap+tcp://<host>/helloWorld" for the client to test CoAP over TCP.

### DIFF
--- a/demo-apps/cf-helloworld-server/src/main/java/org/eclipse/californium/examples/HelloWorldServer.java
+++ b/demo-apps/cf-helloworld-server/src/main/java/org/eclipse/californium/examples/HelloWorldServer.java
@@ -13,6 +13,7 @@
  * Contributors:
  *    Matthias Kovatsch - creator and main architect
  *    Kai Hudalla (Bosch Software Innovations GmbH) - add endpoints for all IP addresses
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add TCP parameter
  ******************************************************************************/
 package org.eclipse.californium.examples;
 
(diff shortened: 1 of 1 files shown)


boaks:
Before the merge, I will log a hint, that tcp is still experimental.

demo-apps/cf-helloworld-server/src/main/java/org/eclipse/californium/examples/HelloWorldServer.java
```diff
@@ -27,22 +28,30 @@
...
     public static void main(String[] args) {
-        
-        try {
 
+        try {
             // create server
+            boolean udp = true;
+            boolean tcp = false;
+            if (0 < args.length) {
+                tcp = args[0].equalsIgnoreCase("coap+tcp:");
```
sophokles73: do you really want to include the `:` in the arg?
maybe the arg should better be named something like `enable-tcp`?

boaks: My intention was, that if we even add "interfaces" as "coap://192.168.10.110:5683" or "coaps+tcp://192.168.5.50:5684" to the argument list, the "coap+tcp:" would be the closest choice for now.

sophokles73: you mean, in order to e.g. specify the interface and port to bind to, right?

boaks: yes, for a future extension.

boaks: Please note:
this PR was mainly prepared, because a user had problems to setup the TCP communication. So I thought, it would be a good idea to provide a "working setup".

PR #332, opened 2017-06-22
Corrected LinkFormat filtering with multiple query components, added tests.

jeeger:
The old matching code matched queries with multiple components incorrectly, returning `true` on the first matching resource. Thus, querying with `foo&bar=one` would return all resources that have the attribute `foo`, without matching on the `bar=one` query.

The new matches() code fixes this.


boaks:
First, thank you for your contribution!

But to use it, eclipse requires your ECA. Did you sign it already?

If not, see https://www.eclipse.org/legal/ECA.php

PR #333, opened 2017-06-22
Fixed LinkFormat.matches() for multiple queries, added tests.

jeeger:
The current implementation of matches() returns superfluous matches if the query consists of multiple elements. For example, querying `foo&bar=one` returns all resources that have the `foo` attribute, instead of only those that have both `foo` and `bar=one`.

I have also added tests for these queries.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
@@ -161,7 +161,10 @@ public class LinkFormat {
 	 *
 	 * Queries are interpreted according to <a href="https://tools.ietf.org/html/rfc6690#section-4.1">RFC 6690</a>,
 	 * section 4.1, with the important difference that more than one query can be passed to the function. The
-	 * resource only matches the list of queries if the resource matches every query in the list.
+	 * resource only matches the list of queries if the resource matches every query in the list. This functionality
+	 * is required to implement resource directory filtering according to the
+	 * <a href="https://tools.ietf.org/html/draft-ietf-core-resource-directory-11#section-7">Resource directory</a>
+	 * draft, which requires support for matching multiple attributes.
 	 *
 	 * @param resource The resource to match.
 	 * @param queries The list of queries to match the resource with.
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
@@ -19,20 +19,13 @@
  ******************************************************************************/
 package org.eclipse.californium.core.coap;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Scanner;
-import java.util.Set;
-import java.util.concurrent.ConcurrentSkipListSet;
-import java.util.regex.Pattern;
-
 import org.eclipse.californium.core.WebLink;
 import org.eclipse.californium.core.server.resources.Resource;
 import org.eclipse.californium.core.server.resources.ResourceAttributes;
+
+import java.util.*;
+import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.regex.Pattern;
 
 
 public class LinkFormat {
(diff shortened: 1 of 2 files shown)


californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
```diff
@@ -162,66 +155,64 @@ public static StringBuilder serializeAttribute(String key, List<String> values)
...
-						// handle case like rt=[Type1 Type2]
-						if (actual.indexOf(" ") > -1) { // if contains white space
-							String[] parts = actual.split(" ");
-							for (String part : parts) { // check each part for match
-								if (part.equals(expected)) {
-									return true;
-								}
+							// Wildcard query
+							if (actual.equals(shortened)) {
+								matched = true;
```
boaks: Would a `break` here optimize the loop?

californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
```diff
@@ -162,66 +155,64 @@ public static StringBuilder serializeAttribute(String key, List<String> values)
...
 							}
+						} else if (actual.equals(expected)) {
+							// Regular query
+							matched = true;
```
boaks: Same here; would a `break` here optimize the loop?

jeeger:
Added break statements.

boaks:
I'm struggling a little to see the RFC your PR is based on.

RFC6690, section 4.1, Query Filtering

>    A server implementing this specification MAY recognize the query part
>    of a resource discovery URI as a filter on the resources to be
>    returned.  The path and query components together should conform to
>    the following level-4 URI Template [RFC6570]:
> 
>        /.well-known/core{?search*}
> 
>    where the variable "search" is a 1-element list that has a single
>    name/value pair, where
> 
>    o  name is either "href", a link-param name defined in this
>       specification, or any other link-extension name, and
> 
>    o  value is either a Complete Value String that does not end in an
>       "*" (%2A), or a Prefix Value String followed by an "*" (%2A). 

"search" is a 1-element list that has a **single** name/value pair

So I'm not sure, ion which RFC your multi attribute example is based. Can you help me out?

jeeger:
Hello!

The `*` indicates a multi-element list of type "search" (as indicated by https://tools.ietf.org/html/rfc6570#section-2.4.2). So, the `search` parameter can be repeated an arbitrary number of times. 

The example for the "composite" type in my linked RFC seems to confirm this.

boaks:
I have to confess, that I'm not too used to that.
RFC6570, section 2.4.2, page 17,

> find{?year*}       find?year=1965&year=2000&year=2012

So if a logical `and` is applied to multiple attributes, how should the above example from the RFC6570 resolve?
If multiple filter attributes are defined, I think there should be a definition, if they applied `and` or `or`.

jeeger:
Hm, on a closer reading, this seems to indicate that the value is "interpolated" into the query, not repeated. So `?list*`, with `list` as a list containing a single key-value pair, this would expand to `?key=value`. 

So, for `/well-known/.core`, passing multiple values is not allowed. 

However, the resource directory [RFC ](https://tools.ietf.org/html/rfc6570#section-2.4.2) specifies that multiple values need to be supported.

The proper way would probably be keeping this more general function, and accepting only one query in `/well-known/.core` filtering, and not doing this for the resource directory.

boaks:
I wrote a e-mail to the ietf core-mailing list.
 
https://www.ietf.org/mail-archive/web/core/current/msg08848.html

Maybe we receive an answer.

boaks:
RFC6570, 1.2, page 7

> An explode ("*") modifier indicates that the variable is to be treated as a composite value,
> consisting of either a list of names or an associative array of
> (name, value) pairs, that is expanded as if each member were a
> separate variable (Section 2.4.2).

So for me `*` seems not to indicate "multiple occurrence", it seems to indicates "explode the list" .
And with RFC6690, 4.1, specifies the list as 1-element list, for me it seems, that multiple search expressions are out of scope of RFC6690.

jeeger:
Yes, that was also my final interpretation. However, this pull request should still be accepted to make the resource directory search work correctly.

Citing from the [resource directory draft](https://tools.ietf.org/html/draft-ietf-core-resource-directory-11#section-7):

```
   Multiple query parameters MAY be included in a lookup, all included
   parameters MUST match for a resource to be returned.  The
   character'*' MAY be included at the end of a parameter value as a
   wildcard operator.

   RD Lookup requests MAY use any set of query parameters to match the
   registered attributes and relations.
```

There should be a consistency check for `/.well-known/core` filtering that returns an error (4.02?) if more than one search query is passed, and the corrected matching function should be used for resource directory searches.

I will add additional documentation to the `matches` function, to indicate that the function does a boolean `AND` operation on all passed queries.

Would that be acceptable?

boaks:
> There should be a consistency check for /.well-known/core filtering that returns an error (4.02?) if more than one search query is passed, and the corrected matching function should be used for resource directory searches.

I'm not sure, if I understand, what you propose:
If more then 1 search is provided in the query option, then an error is returned?
(Will this be implemented in the `DiscoveryResource`?) 

But what does the "corrected matching function" then do? Currently its only called by that discovery, and therefore I don't understand, why you want to support multiple search query options. Do you plan to build a "resource directory" based on californium and do you require that function for that?    

> I will add additional documentation to the matches function, to indicate that the function does a boolean AND operation on all passed queries.

> Would that be acceptable?

From my side, yes. 
(If we don't receive an "opposite" answer from the core-list.)  

Maybe one of the other committer provides also his opinion.

jeeger:
> If more then 1 search is provided in the query option, then an error is returned? (Will this be implemented in the DiscoveryResource?)

Yes, I guess so. [RFC 6690](https://tools.ietf.org/html/rfc6690#section-4.1) specifies the requirement level for queries as MAY, so I think this would be a valid implementation, and better than returning data that just matches the first query element while ignoring the rest.

> Do you plan to build a "resource directory" based on californium and do you require that function for that?

The `californium.tools` [repository](https://github.com/eclipse/californium.tools) contains an implementation of the resource directory draft in `cf-rd`. This implementation  uses the `matches()` function as well. The resource directory needs to support more complex queries than regular CoAP clients, so the "extended"`matches()` function is required. This is where I originally noticed the strange behavior of the `matches()` function.

PS: Thanks for your feedback! I had misunderstood the spec here.

boaks:
> I will add

Do you still work on this? Adding the documentation and the check in `DiscoveryResource`.

If not, I would announce to merge this PR, if nobody has an objection, and add the documentation and check afterwards.

jeeger:
Hey!

I'll add the documentation now, and do the "check for single query element" in another pull request.

boaks:
OK!

boaks:
Maybe you add your link

https://tools.ietf.org/html/draft-ietf-core-resource-directory-11#section-7

also to the documentation?

jeeger:
Yep, I'll do it today.

boaks:
LGTM. If no objection, I will go to merge this.
(Rebase, squash and merge)

boaks:
Rebased and squashed to commit 

20f2a6583279ddf0ccad6b826edc885df9ff17e4

on master.

@jeeger 
Usually a fix or extension is prepared on a special branch.
This is done to be able to rebase on new masters. Otherwise you may get decoupled from the current master. If you want to get in sync with master again, rename your local master branch, fetch, and checkout master again, or fetch/pull foreced.

PR #334, opened 2017-06-22
Report error callback, if DTLS handshake fails.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -50,6 +50,12 @@
  *                                                    cancel flight only, if they
  *                                                    should not be retransmitted
  *                                                    anymore.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - call handshakeFailed on 
+ *                                                    terminateOngoingHandshake,
+ *                                                    processAlertRecord, 
+ *                                                    handleTimeout,
+ *                                                    and add error callback in
+ *                                                    newDeferredMessageSender.
  ******************************************************************************/
 package org.eclipse.californium.scandium;
 
@@ -598,7 +604,8 @@ public class DTLSConnector implements Connector {
 					"Aborting handshake with peer [{0}]: {1}",
 					new Object[]{peerAddress, cause.getMessage()});
 			}
-			DTLSSession session = connection.getOngoingHandshake().getSession();
+			Handshaker handshaker = connection.getOngoingHandshake();
+			DTLSSession session = handshaker.getSession();
 			AlertMessage alert = new AlertMessage(AlertLevel.FATAL, description, peerAddress);
 			if (!connection.hasEstablishedSession()) {
 				terminateConnection(connection, alert, session);
@@ -607,6 +614,7 @@ public class DTLSConnector implements Connector {
 				send(alert, session);
 				connection.terminateOngoingHandshake();
 			}
+			handshaker.handshakeFailed(cause);
 		}
 	}
 
@@ -765,6 +773,8 @@ public class DTLSConnector implements Connector {
 		record.setSession(session);
 		try {
 			AlertMessage alert = (AlertMessage) record.getFragment();
+			Handshaker handshaker = connection.getOngoingHandshake();
+			HandshakeException error = null;
 			LOGGER.log(
 					Level.FINEST,
 					"Processing {0} ALERT from [{1}]: {2}",
@@ -774,6 +784,7 @@ public class DTLSConnector implements Connector {
 				// (http://tools.ietf.org/html/rfc5246#section-7.2.1)
 				// we need to respond with a CLOSE_NOTIFY alert and
 				// then close and remove the connection immediately
+				error = new HandshakeException("Received 'close notify'", alert);
 				terminateConnection(
 						connection,
 						new AlertMessage(AlertLevel.WARNING, AlertDescription.CLOSE_NOTIFY, alert.getPeer()),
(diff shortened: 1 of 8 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SessionListener.java
```diff
@@ -62,4 +64,13 @@ void sessionEstablished(Handshaker handshaker, DTLSSession establishedSession)
...
 	 */
 	void handshakeCompleted(InetSocketAddress peer);
+
+	/**
+	 * Indicates that a handshake with a given peer has failed.
+	 * 
+	 * @param peer the IP address and port of the peer the handshake has failed
+	 * @param exception the occurred handshake exception, which caused the handshake to fail
+	 */
+	void handshakeFailed(InetSocketAddress peer, HandshakeException exception);
```
sophokles73: The handshake may as well fail due to any other exception, don't you think? Is there anything specific we need to to with the `HandshakeException`, i.e. is it important that it is a `HandshakeException` and not a plain `Throwable`?

boaks: I did this, because as far as I understood it, the handshake termination is triggered by such an Exception. But I can change the parameter type to Throwable.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
```diff
@@ -263,4 +265,9 @@ public void sessionEstablished(final Handshaker handshaker, final DTLSSession es
 	public void handshakeCompleted(final InetSocketAddress peer) {
 		// nothing to do
 	}
+
+	@Override
+	public void handshakeFailed(final InetSocketAddress peer, HandshakeException exception) {
+		// nothing to do
```
sbernard31: Should we remove connection from store ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -785,6 +795,9 @@ private void processAlertRecord(final Record record, final Connection connection
 					errorHandler.onError(alert.getPeer(), alert.getLevel(), alert.getDescription());
```
sbernard31: Not completely linked to this issue, but now we have MessageCallback.onError()`  + `SessionListener.handshakeFailed()`. Do we still need errorHandler ?

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
```diff
@@ -1507,9 +1507,12 @@ public void onError(InetSocketAddress peerAddress, AlertLevel level, AlertDescri
 			}
 		});
 		client.start();
-		client.send(new RawData("Hello".getBytes(), serverEndpoint));
+		SimpleMessageCallback callback = new SimpleMessageCallback();
+		RawData data = RawData.outbound("Hello".getBytes(), serverEndpoint, null, callback, false);
+		client.send(data);
 
 		assertTrue(latch.await(MAX_TIME_TO_WAIT_SECS, TimeUnit.SECONDS));
+		assertThat("Error callback missing", callback.getError(MAX_TIME_TO_WAIT_SECS), is(notNullValue()));
```
boaks: Thanks!

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: I didn't get it. The one is for catching a failed handshake, the other is to report a send error, which may have other causes (even with successful handshake the send may fail).

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
boaks: As far as I understood, this is already done before. I didn't want to change the current call logic.

sbernard31: Ok, it seems to be done in `terminateConnection`.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: I talked about : [`ErrorHandler`](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/ErrorHandler.java)
It was introduce long time ago to get error feedback from DTLS Layer, but it seems to me that what you introduce (`MessageCallback.onError`+`SessionListener.handshakeFailed`) do the job better. So we could remove it now ?

boaks: The ErrorHandler provides details about the received alert message, which will be only available for "HandshakeException" (containing a reference to the alert message). So with the above change to Throwable, a "instance of" check would be required.

sbernard31: This sounds acceptable. I tried a longtime ago to use ErrorHandler and this is not so easy as you don't have so much context. Handshake errors available on `MessageCallback.onError` is really better !

boaks: There are tests based on AlertMessages. To change them, givenAnEstablishedSession must be extended. Currently I detected some (very small) race conditions in the flight cancel. So I think elimination of the ErrorHandler should be shifted ;-).

sbernard31: When I made this kind of feedback, it doesn't mean that it should be done "immediately in this PR by this PR authors". It's mainly to know on what we agree.
I will open an issue and we can do that later.

boaks:

sbernard31:
LGTM

PR #336, opened 2017-06-23
Add get on incomplete DTLS configuration for DTLSconfiguration.Builder

sbernard31:
The idea was to allow to external lib to complete a user configuration.

I'm sure this is not so clear, so you can have a look at Leshan issues (https://github.com/eclipse/leshan/pull/350, https://github.com/eclipse/leshan/pull/356) to better understand the needs.

### DIFF
--- a/demo-apps/cf-secure/src/main/java/org/eclipse/californium/examples/SecureClient.java
+++ b/demo-apps/cf-secure/src/main/java/org/eclipse/californium/examples/SecureClient.java
@@ -69,7 +69,7 @@ public class SecureClient {
 			Certificate[] trustedCertificates = new Certificate[1];
 			trustedCertificates[0] = trustStore.getCertificate("root");
 
-			DtlsConnectorConfig.Builder builder = new DtlsConnectorConfig.Builder(new InetSocketAddress(0));
+			DtlsConnectorConfig.Builder builder = new DtlsConnectorConfig.Builder();
 			builder.setPskStore(new StaticPskStore("Client_identity", "secretPSK".getBytes()));
 			builder.setIdentity((PrivateKey)keyStore.getKey("client", KEY_STORE_PASSWORD.toCharArray()),
 					keyStore.getCertificateChain("client"), true);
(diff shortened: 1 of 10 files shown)


boaks:
To be even more flexible, e.g. removing parameters from the users choice at all, what do you think to move this work into a own configuration into leshan and the just in the last step, when creating the DTLSConnector, then use that new configuration to fill in the data into the configuration provided by californium?

sbernard31:
I thought about that but this requires that each time the DTLSConnectorConfig changes we update (e.g. adding/removing earlyStopRetransmission)  the "config" at Leshan side.

For now, this PR  seems to me an acceptable tradeoff. (not so impacting at californium side and flexible enough to allow the kind of thing we would like to do in Leshan)

boaks:
Is there already some code in leshan, which shows, how this flexibility is then used?

sbernard31:
[here](https://github.com/eclipse/leshan/pull/356/commits/8cdfa727ce47b18a616ec6022d73419117a5cce7#diff-ab2b2333ae602c4ca955af3e2805c488R289) the flexibility for californium users.

[here]() the benefits for Leshan users.

boaks:
Hm, one of the links references just this PR?

But anyway, is see the benefits for leshan and so LGTM.

sbernard31:
oops I wanted to point this : https://github.com/eclipse/leshan/pull/356#issuecomment-311954844

boaks:
Would it be possible to add the constructor 

`public Builder(InetSocketAddress address)`

again (additionally)?

sbernard31:
I replace it by a default value and a setter.
Why should we have a constructor with inetSocketAddress ?

boaks:
Just for "backwards compatibility" marked as deprecated.

sbernard31:
This API break is not so annoying to integrate and we have others API break with this version on MessageObserver and coapRequest.getURI but if you think this is valuable we can do that.

boaks:
After try to add it, I found, that the CleanerMessageObserver in CaliforniumLwM2mRequestSender will  be anyway red, because MessageObserver is changed ... so your right, there is no benefit.

PR #340, opened 2017-06-28
Adjust thread safe cancel of flight.

boaks:
Change implementation of cancel to prevent flights from being scheduled
for retransmission due to race conditions. Make TimerTask trigger
striped execution instead of direct executing handleTimeout to prevent
also that race condition.


sbernard31:
LGTM

PR #342, opened 2017-06-28
BERT option for branch 2.0.x

raipratheek:
Signed-off-by: raipratheek <raipratheek@gmail.com>
As discussed in Pull request #319, changes have been adapted to the 2.0.x branch.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
@@ -21,6 +21,7 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - add InputStream support for environments
  *                                                    without file access.
  *    Achim Kraus (Bosch Software Innovations GmbH) - add new keys for MID tracker
+ *    Pratheek Rai - Added TCP_NUMBER_OF_BULK_BLOCKS for BERT option.
  ******************************************************************************/
 package org.eclipse.californium.core.network.config;
 
@@ -194,6 +195,13 @@ public final class NetworkConfig {
 		public static final String TCP_CONNECTION_IDLE_TIMEOUT = "TCP_CONNECTION_IDLE_TIMEOUT";
 		public static final String TCP_CONNECT_TIMEOUT = "TCP_CONNECT_TIMEOUT";
 		public static final String TCP_WORKER_THREADS = "TCP_WORKER_THREADS";
+		
+		/**
+		 * If the value is greater than 1, this sets up the active use of BERT.
+		 * i.e. Messages will be sent with BERT option. The passive
+		 * receiving of BERT message is always enabled while using TCP connector.
+		 */
+		public static final String TCP_NUMBER_OF_BULK_BLOCKS = "TCP_NUMBER_OF_BULK_BLOCKS";
 	}
 
 	/**
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -26,7 +26,8 @@
  *                                                    to fix rare race condition in
  *                                                    block1wise, when the generated
  *                                                    token was copied too late 
- *                                                    (after sending). 
+ *                                                    (after sending).
+ *    Pratheek Rai - changes for BERT 
  ******************************************************************************/
 package org.eclipse.californium.core.network.stack;
 
@@ -196,7 +197,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		lower().sendRequest(exchange, requestToSend);
 	}
 
-	private Request startBlockwiseUpload(final Exchange exchange, final Request request) {
+	protected Request startBlockwiseUpload(final Exchange exchange, final Request request) {
 
 		final KeyUri key = getKey(exchange, request);
 
@@ -283,7 +284,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		}
 	}
 
-	private void handleInboundBlockwiseUpload(final Exchange exchange, final Request request) {
+	protected void handleInboundBlockwiseUpload(final Exchange exchange, final Request request) {
 
 		if (requestExceedsMaxBodySize(request)) {
 
@@ -367,7 +368,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		}
 	}
 
-	private void sendBlock1ErrorResponse(final KeyUri key, final Exchange exchange, final Request request,
+	protected void sendBlock1ErrorResponse(final KeyUri key, final Exchange exchange, final Request request,
 			final ResponseCode errorCode, final String message) {
 
 		BlockOption block1 = request.getOptions().getBlock1();
@@ -379,7 +380,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		lower().sendResponse(exchange, error);
 	}
 
-	private void handleInboundRequestForNextBlock(final Exchange exchange, final Request request,
+	protected void handleInboundRequestForNextBlock(final Exchange exchange, final Request request,
 			final KeyUri key, final Block2BlockwiseStatus status) {
 
 		synchronized (status) {
@@ -650,7 +651,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		}
 	}
 
-	private void sendNextBlock(final Exchange exchange, final Response response, final KeyUri key, final Block1BlockwiseStatus status) {
+	protected void sendNextBlock(final Exchange exchange, final Response response, final KeyUri key, final Block1BlockwiseStatus status) {
 
 		BlockOption block1 = response.getOptions().getBlock1();
 		int currentSize = status.getCurrentSize();
@@ -773,61 +774,8 @@ public class BlockwiseLayer extends AbstractLayer {
 
 					} else if (block2.isM()) {
 
-						int currentSize = status.getCurrentSize();
-						// do late block size negotiation
-						int newSize, newSzx;
-						if (block2.getSzx() > preferredBlockSzx) {
-							newSize = preferredBlockSize;
-							newSzx = preferredBlockSzx;
-						} else {
-							newSize = currentSize;
-							newSzx = status.getCurrentSzx();
-						}
-						int nextNum = status.getCurrentNum() + currentSize / newSize;
-
-						Request request = exchange.getRequest();
-
-						Request block = new Request(request.getCode());
-						// do not enforce CON, since NON could make sense over SMS or similar transports
-						block.setType(request.getType());
-						block.setDestination(request.getDestination());
-						block.setDestinationPort(request.getDestinationPort());
-
-						/*
-						 * WARNING:
-						 * 
-						 * For Observe, the Matcher then will store the same
-						 * exchange under a different KeyToken in exchangesByToken,
-						 * which is cleaned up in the else case below.
-						 */
-						if (!response.getOptions().hasObserve()) {
-							block.setToken(response.getToken());
-						}
-
-						// copy options
-						block.setOptions(new OptionSet(request.getOptions()));
-						block.getOptions().setBlock2(newSzx, false, nextNum);
-						if (response.getOptions().getETagCount() > 0) {
-							// use ETag provided by peer
-							block.getOptions().addETag(response.getOptions().getETags().get(0));
-						}
-
-						// make sure NOT to use Observe for block retrieval
-						block.getOptions().removeObserve();
-
-						// copy message observers from original request so that they will be notified
-						// if something goes wrong with this blockwise request, e.g. if it times out
-						block.addMessageObservers(request.getMessageObservers());
-						// add an observer that cleans up the block2 transfer tracker if the
-						// block request fails
-						addBlock2CleanUpObserver(block, key);
-
-						status.setCurrentNum(nextNum);
-
-						LOGGER.log(Level.FINER, "requesting next Block2 [num={0}]: {1}", new Object[]{ nextNum, block });
-						exchange.setCurrentRequest(block);
-						lower().sendRequest(exchange, block);
-
+						// request next block
+						requestNextBlock(exchange, response, key, status);
 					} else {
 
 						// we have received the last block of the block2 transfer
@@ -877,10 +825,72 @@ public class BlockwiseLayer extends AbstractLayer {
 			}
 		}
 	}
+	
+	/**
+	 * Sends request for the next response block.
+	 */
+	protected void requestNextBlock(final Exchange exchange, final Response response, final KeyUri key, final Block2BlockwiseStatus status) {
+		int currentSize = status.getCurrentSize();
+		// do late block size negotiation
+		int newSize, newSzx;
+		BlockOption block2 = response.getOptions().getBlock2();
+		if (block2.getSzx() > preferredBlockSzx) {
+			newSize = preferredBlockSize;
+			newSzx = preferredBlockSzx;
+		} else {
+			newSize = currentSize;
+			newSzx = status.getCurrentSzx();
+		}
+		int nextNum = status.getCurrentNum() + currentSize / newSize;
+
+		Request request = exchange.getRequest();
+
+		Request block = new Request(request.getCode());
+		// do not enforce CON, since NON could make sense over SMS or similar transports
+		block.setType(request.getType());
+		block.setDestination(request.getDestination());
+		block.setDestinationPort(request.getDestinationPort());
+
+		/*
+		 * WARNING:
+		 * 
+		 * For Observe, the Matcher then will store the same
+		 * exchange under a different KeyToken in exchangesByToken,
+		 * which is cleaned up in the else case below.
+		 */
+		if (!response.getOptions().hasObserve()) {
+			block.setToken(response.getToken());
+		}
+
+		// copy options
+		block.setOptions(new OptionSet(request.getOptions()));
+		block.getOptions().setBlock2(newSzx, false, nextNum);
+		if (response.getOptions().getETagCount() > 0) {
+			// use ETag provided by peer
+			block.getOptions().addETag(response.getOptions().getETags().get(0));
+		}
+
+		// make sure NOT to use Observe for block retrieval
+		block.getOptions().removeObserve();
+
+		// copy message observers from original request so that they will be notified
+		// if something goes wrong with this blockwise request, e.g. if it times out
+		block.addMessageObservers(request.getMessageObservers());
+		// add an observer that cleans up the block2 transfer tracker if the
+		// block request fails
+		addBlock2CleanUpObserver(block, key);
+
+		status.setCurrentNum(nextNum);
+
+		LOGGER.log(Level.FINER, "requesting next Block2 [num={0}]: {1}", new Object[]{ nextNum, block });
+		exchange.setCurrentRequest(block);
+		lower().sendRequest(exchange, block);
+
+	}
 
 	/////////// HELPER METHODS //////////
 
-	private static KeyUri getKey(final Exchange exchange, final Request request) {
+	protected static KeyUri getKey(final Exchange exchange, final Request request) {
 
 		if (exchange.isOfLocalOrigin()) {
 			return KeyUri.fromOutboundRequest(request);
@@ -889,7 +899,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		}
 	}
 
-	private static KeyUri getKey(final Exchange exchange, final Response response) {
+	protected static KeyUri getKey(final Exchange exchange, final Response response) {
 
 		if (exchange.isOfLocalOrigin()) {
 			return KeyUri.fromInboundResponse(exchange.getRequest().getURI(), response);
@@ -912,7 +922,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		}
 	}
 
-	private Block1BlockwiseStatus getInboundBlock1Status(final KeyUri key, final Exchange exchange, final Request request) {
+	protected Block1BlockwiseStatus getInboundBlock1Status(final KeyUri key, final Exchange exchange, final Request request) {
 
 		synchronized (block1Transfers) {
 			Block1BlockwiseStatus status = block1Transfers.get(key);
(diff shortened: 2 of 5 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
```diff
@@ -192,6 +192,8 @@
 		public static final String TCP_CONNECTION_IDLE_TIMEOUT = "TCP_CONNECTION_IDLE_TIMEOUT";
 		public static final String TCP_CONNECT_TIMEOUT = "TCP_CONNECT_TIMEOUT";
 		public static final String TCP_WORKER_THREADS = "TCP_WORKER_THREADS";
+		public static final String BERT_OPTION = "BERT_OPTION";
```
boaks: "option" is a used term in CoAP. Therefore I would prefer an other term/name. What's about "USE_BERT"?

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
```diff
@@ -192,6 +192,8 @@
...
+		public static final String BERT_OPTION_STEP_SIZE = "BERT_OPTION_STEP_SIZE";
```
boaks: What's about "NUMBER_OF_BLOCKS_IN_BULK_USING_BERT"?

boaks: Or "TCP_NUMBER_OF_BULK_BLOCKS" for both? If value >1 enable BERT, for <= 1 disabled?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -115,8 +115,8 @@
 	private final LeastRecentlyUsedCache<KeyUri, Block1BlockwiseStatus> block1Transfers;
 	private final LeastRecentlyUsedCache<KeyUri, Block2BlockwiseStatus> block2Transfers;
 	private int maxMessageSize;
-	private int preferredBlockSize;
-	private int preferredBlockSzx;
+	protected int preferredBlockSize;
+	protected int preferredBlockSzx;
```
boaks: I wouldn't change this to `protected` just for adjust them to BERT values. Instead, I would add the adjustment to the CoapTcpStack and correct the `config` there. I would also only create a `TcpBlockwiseLayer` instead of a `BlockwiseLayer`, if enabled.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
```diff
@@ -0,0 +1,394 @@
...
+import org.eclipse.californium.core.network.config.NetworkConfig.Keys;
+
+/**
+ * Blockwise layer for TCP. Extended mainly for BERT option. Capable of handling
+ * incoming blocks with SZX = 7.
+ * 
+ * @author Pratheek Rai.
+ *
+ */
+public class TCPBlockwiseLayer extends BlockwiseLayer {
```
boaks: Usually we use CaMeL-case. So I would rename the class to `TcpBlockwiseLayser`.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
```diff
@@ -0,0 +1,394 @@
...
+	/**
+	 * The number of 1024 Bytes packets to be sent as a block. Size of 1 BERT
+	 * block = bertStepSize * 1024 Bytes
+	 */
+	private int bertStepSize;
+
+	/**
+	 * Flag for BERT option.
+	 */
+	private boolean isBertEnabled;
```
boaks: If CoapTcpStack would already evaluate, if BERT should be sued or not, I think this flag could be removed.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
```diff
@@ -0,0 +1,394 @@
...
+			preferredBlockSize = BERT_INT_BLOCK_SIZE;
+		}
+	}
+
+	/**
+	 * Handles inbound blockwise upload from peer. Handling is done here only
+	 * for SZX = 7 all other handling is done in the BlockwikseLayer.
+	 */
+	protected void handleInboundBlockwiseUpload(final Exchange exchange, final Request request) {
+
```
boaks: This seems to be a lot of code copy from `BlockwiseLayer`. 
But as I meant previously, if we don't want to change the `BlockwiseLayer` itself too much, I agree that such a copy must then be used. May be we can get feedback from other committers, how much changes would be acceptable to omit code copies.

boaks: but it's not only a code copy, unfortunately it copies the "old" (master) implementation. In 2.0.x error checking and handling was added and I miss that.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
```diff
@@ -0,0 +1,394 @@
...
+				Response block = null;
+
+				for (int i = 0; ((i < bertStepSize) && (hasNextBlock)); i++) {
+					block = status.getNextResponseBlock(block2);
+					hasNextBlock = block.getOptions().getBlock2().isM();
+					currentNum = currentNum + 1;
+					block2 = new BlockOption(BERT_SZX, hasNextBlock, currentNum);
+					responseBlockBuilder.put(block.getPayload());
+				}
+
```
boaks: Moving this payload preparation into the `Block2BlockwiseStatus` would be a easier solution then break down the payload and accumulate it again. But again, as I meant previously, if we don't want to change the `BlockwiseLayer` itself too much, I agree that the response must be accumulated again.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
```diff
@@ -0,0 +1,394 @@
...
+	/**
+	 * Send request to peers.Handling is done here only when BERT option is
+	 * enabled otherwise handling is done in the BlockwikseLayer.
+	 */
+	public void sendRequest(final Exchange exchange, final Request request) {
+
+		Request firstRequest = request;
+		Request requestToSend;
+		if (isBertEnabled) {
```
boaks: May be removed, if the `TcpBlockwiseLayer `is conditionally create in `CoapTcpStack`.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
```diff
@@ -0,0 +1,394 @@
...
+
+			if (isTransparentBlockwiseHandlingEnabled()) {
+
+				BlockOption block2 = request.getOptions().getBlock2();
+				if (block2 != null && block2.getNum() > 0) {
+					isBertEnabled = false;
```
boaks: This would cause disable BERT permanently when ever a broken or paused blockwise is restarted in the middle.

Sequence:
Client read resource a blockwise. Response is split in 20 blocks, each grouped in a message with 4 blocks.
After 2 messages (8 blocks), the client lose the connection and recreate it.
Then the client start a new blockwise read from block 8 on.

=> as result of the `isBertEnabled = false` this would not only disable BERT for this transfer, it would permanently disable it until the next restart.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
```diff
@@ -0,0 +1,394 @@
...
+					super.sendRequest(exchange, request);
```
boaks: Calling `super.sendRequest()` will acutally send a message, so `lower().sendRequest()` will send a second message ...

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
```diff
@@ -0,0 +1,394 @@
...
+		}
+
+	}
+
+	/**
+	 * Send response to peers. Handling is done here only when BERT option is
+	 * enabled otherwise handling is done in the BlockwikseLayer.
+	 */
+	public void sendResponse(final Exchange exchange, final Response response) {
+
```
boaks: `BlockwiseLayer.sendResponse` in 2.0.x contains some more checks and error handling, this is missing here.

sophokles73:
@raipratheek which version of the coap-over-tcp draft is this supposed to be based on?

sophokles73:
I do not quite understand why we need subclasses for handling BERT blocks. The spec gives the impression, that the overall blockwise transfer works exactly as usual. My first idea would be that we only need to adapt the Blockwise Status classes accordingly, i.e. properly encapsulate the calculation of next block num, assembling messages etc in there while handling SZX=7 in a special way.

boaks:
The basic idea to create a subclass was to isolate the changes from the UDP BlockwiseLayer.

As far as I understood the last months, concentration on UDP limits the efforts for TCP. And therefore I recommended to move the changes from the original PR #319 into a separated class.

boaks:
@sophokles73 

Do you see any issues/violations according draft-ietf-core-coap-tcp-tls-09, section 6?
As far as I followed up draft-ietf-core-coap-tcp-tls-09, there is no change in BERT since it was included in that RFC draft. So the question about the version related to BERT seems for me to be not that important. 

Or do you know a change in the BERT definition in that draft?

sophokles73:
I just wanted to know what draft version we are referring to. IIRC our current TCP support is based on draft 4 so I wanted to know whether we are supporting BERT from that version or another one. But if BERT has remained unchanged since draft 4 all is fine, I guess.

raipratheek:
I have basically referred to Draft 09. https://tools.ietf.org/html/draft-ietf-core-coap-tcp-tls-09.
But I also referred to the previous versions and found no changes made.

boaks:
ver 4
<    used over a reliable byte stream transport.  While this suggests that
<    the Block-wise transfer protocol [I-D.ietf-core-block] is also no
<    longer needed, it remains applicable for a number of cases:
ver 9
>    used over a reliable transport.  While this suggests that the Block-
>    wise transfer protocol [RFC7959] is also no longer needed, it remains
>    applicable for a number of cases:

ver 4
<    the equivalent Option with SZX == 6, except that the payload is
ver 9
>    the equivalent Option with SZX == 6, except that the payload is also

ver 4
<    (NUM), more bit (M), and block size exponent (2**(SZX+4)) separated
<    by slashes.  E.g., a Block2 Option value of 33 would be shown as
ver 9
>    (NUM), more bit (M), and block size (2**(SZX+4)) separated by
>    slashes.  E.g., a Block2 Option value of 33 would be shown as

So the differences are very small changes in the wording.

The draft-ietf-core-coap-tcp-tls-09 contains also a 5.3.1 (max message size) and 5.3.2 (blockwise), which indirect enables BERT, but though we don't have CSM right now at all, this is not included.

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
raipratheek: I agree. We could check the TCP_NUMBER_OF_BULK_BLOCKS value and then decide if BERT is enabled.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
raipratheek: I agree.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
raipratheek: I prefer having this flag because I want to use TcpBlockwiseLayer in CoapTcpStack. I think it would help to receive incoming BERT messages even  when the BERT sending option is disabled. This could be useful until we have the Messaging implemented.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
raipratheek: I will update this.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
raipratheek: The idea is to have minimal changes in BlockwiseLayer code as UDP has greater priority than TCP at the moment.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
raipratheek: I will check this.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
raipratheek: I will check this.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
boaks: Good point!
Would it be possible to add some javadoc to the configuration, that it will only setup the active use of BERT, not the passive receiving of data?

raipratheek: Yes. I will add Javadoc for the same.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
raipratheek: What should be the behaviour in case  BERT is enabled ?  Should BERT be disabled only for the Current Message ? or should BERT not be disabled at all ?

boaks: Resume a blockwise transfer from the provided block on.
If SZX indicates BERT, use BERT for resume the blockwise transfer.

Resume just means, "transfer the payload from that block on".

raipratheek: got it.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
raipratheek: Can these error checks simple be copied from BlockwiseLayer to TcpBlockwiseLayer ? or does it require some additional handling ?

boaks: Just from a short static code check:
 - there are checks, if the blocksize is negotiated.
   Currently I'm not deep enough in the specs but I would go for adjust the blocksize , 
   which would require to switch from BERT to none BERT for that transfer. Maybe a
   spec analysis (coap over tcp and/or RFC7959)  and questions to the core list or 
   coap over tcp github issue would be required, if others request that. For my self,
   as long as your the main user and as long as it is experimental (what TCP is at all in the
   moment), I would not require to do that spec  work, just adjust the blocksize and add a
  comment.  

- there are checks about the content format. I think they could be copied.

- there are checks about the block number. I'm not sure, but I think they could be copied.
   (junit test may show, if this works proper)

- there are checks for transfer resumption, they should trigger something similar for BERT.  
  I don't know, if this could be copied.

All in all: to answer, if something could be copied, implies, a deeply understanding both implementations. For that, I would require also more time to get deeper in.

sophokles73:
> The basic idea to create a subclass was to isolate the changes from the UDP BlockwiseLayer.


I guess my whole point is that I do not really understand why there need to be substantial changes being made to the `BlockwiseLayer` at all. My understanding is that it should be possible to support BERT by means of some changes to `BlockNStatus` classes and some adjustments to `BlockwiseLayer` to delegate more of the assembly/split-up to the `BlockNStatus` classes. Or am I mistaken?

boaks:
> Or am I mistaken?

Technically your right. But FMPOV this is the result of the review and not something known before.
So it raises two question:
- would @raipratheek really do a "third PR", I don't know, and I wouldn't insist on that
- will there be committers, who will spent time in reviewing the changes in the BlockwiseLayer

As I also mentioned very early in this BERT discussion, I'm not sure, if TCP stuff should go to the "main 2.0.x" branch, or if we should keep them on a "experimental tcp" branch.

sophokles73:
So, I am hearing that you want to proceed with the PR as it is now and *maybe* later re-factor it the way I suggest, right?

boaks:
This depends mainly on @raipratheek, if he would be willing to prepare a 3. PR.
I would understand, that 2 PRs are enough invested time, so I'm open, how to continue.

raipratheek:
I think the first PR was more of modification to BlockNStatus and some modification to BlockwiseLayer. But the modification was done on the code from Master.
Since I got to know that UDP has the priority over TCP, I agree with the idea of separating TCP and UDP for now. The 2nd PR changes have no impacts on the UDP part of branch 2.0.x. ( atleast from MPOV).
I think for the suggested 3rd PR , I would require more time as I have to understand few of the scenarios of Block-wise transfer in CoAP and not to mention it will impact UDP code. 
Having said that I am open to working on the 3rd PR if the commiters think that is the only way forward :)

boaks:
> I think the first PR was more of modification to BlockNStatus and some modification to BlockwiseLayer.
 
Can you point us to a commit in PR #319, which changes the `BlockwiseStatus`?
(Sorry, I only see change in `BlockwiseLayer` in the commits there.) 

> Since I got to know that UDP has the priority over TCP, I agree with the idea of separating TCP and UDP for now. The 2nd PR changes have no impacts on the UDP part of branch 2.0.x.

To split stuff, which is only used for TCP into separated classes, is a design principal, which is valid even without priorities. 
**NOT** to change the `BlockwiseLayer` itself too much, that's the point where the priority comes in.

> Having said that I am open to working on the 3rd PR if the committers think that is the only way forward :)

FMPOV I would take this PR and put it in an separated TCP branch (together with my PR #294).

raipratheek:
> Can you point us to a commit in PR #319, which changes the BlockwiseStatus?
(Sorry, I only see change in BlockwiseLayer in the commits there.)
I mean to say the idea was similar to the 3rd PR suggestion. with reference to branch 2.0.x the 1st PR was changes to BlockNStatus and BlockwiseLayer. ( After refactoring of BlockwiseLayer code in branch 2.0.x)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TCPBlockwiseLayer.java
```diff
@@ -70,59 +68,84 @@ public TCPBlockwiseLayer(NetworkConfig config) {
...
+							"unexpected Content-Format");
 
-				exchange.setCurrentResponse(piggybacked);
-				lower().sendResponse(exchange, piggybacked);
+				} else if (!status.addBlock(request.getPayload())) {
+					sendBlock1ErrorResponse(key, exchange, request, ResponseCode.REQUEST_ENTITY_TOO_LARGE,
+							"body exceeded expected size " + status.getBufferSize());
+				}
 
-			} else {
```
boaks: Though this commit is based on an other commit, its hard to see the resulting change.
Therefore, depending on the amount and complexity of the changes, it's good to rebase an "squash" commits (squash means to create a commit, which collects the resulting changes of several commits. Depending on the git tools your using, there are several ways to squash).

Removing `} else {` may be wrong, the stuff below may then be executed, even if above an error was detected. 
But I'm unsure about that, because this is a change of a change and it's not obvious what is the  resulting change.

boaks:
Hi @raipratheek 

great! Commit e4abeb4d844e2e51415511bd7367880c32af58b9 looks good, but commit 3389328 seems to "merge" the old history in. 
If you reset your "bert2.0.x" branch "--hard" to the commit "e4abeb4d844e2e51415511bd7367880c32af58b9" and do a "forced push" on the branch "bert2.0.x", everything should be OK.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpBlockwiseLayer.java
```diff
@@ -0,0 +1,523 @@
...
+				} else if (!status.hasContentFormat(request.getOptions().getContentFormat())) {
+
+					sendBlock1ErrorResponse(key, exchange, request, ResponseCode.REQUEST_ENTITY_INCOMPLETE,
+							"unexpected Content-Format");
+
+				} else if (!status.addBlock(request.getPayload())) {
+					sendBlock1ErrorResponse(key, exchange, request, ResponseCode.REQUEST_ENTITY_TOO_LARGE,
+							"body exceeded expected size " + status.getBufferSize());
+				}
+
```
boaks: Ok, now I see and think `} else {` is missing.
Without, the "errors" above will fall into the "success" below.

raipratheek:
Done. Squashed all the changes for BERT :)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/TcpBlockwiseLayer.java
```diff
@@ -0,0 +1,523 @@
...
+					LOGGER.fine("outbound request contains block2 option, creating random-access blockwise status");
+					addRandomAccessBlock2Status(exchange, request);
+					handleRandomBlockAccess(exchange, request, block2.getNum());
+				} else if (requiresBlockwise(request)) {
+					// This must be a large POST or PUT request with BERT
+					// option.
+					ByteBuffer requestBuilder = ByteBuffer.allocate(bertStepSize * BERT_INT_BLOCK_SIZE);
+					boolean hasNextBlock = true;
+					KeyUri key = getKey(exchange, request);
+					firstRequest = startBlockwiseUpload(exchange, request);
```
boaks: `firstRequest` is only used within this scope, so I would consider to move the line 206 

> Request firstRequest = request;

into this block.

raipratheek:
License header added and review issues have been fixed.

boaks:
Hi @raipratheek 

I merged your PR to the new branch "tcp_experimental_features" 
(commit 344773cb58a0159975e7bf458bc0ee5922d648fe)
Thanks four your contribution and patience.

Issue #345, opened 2017-06-29
NPE in UDPConnector.stop()


sbernard31:
I faced an NPE :
```
java.lang.NullPointerException
	at org.eclipse.californium.elements.UDPConnector.stop(UDPConnector.java:206)
	at org.eclipse.californium.core.network.CoapEndpoint.stop(CoapEndpoint.java:455)
	at org.eclipse.californium.core.network.CoapEndpoint.start(CoapEndpoint.java:426)
```
in :
```java
String address = socket.getLocalSocketAddress().toString();
if (socket != null) {
    socket.close();
    socket = null;
}
LOGGER.log(Level.CONFIG, "UDPConnector on [{0}] has stopped.", address);
```

boaks:
@sbernard31 
So is this issue closed with PR #347?
(I picked the fix to 1.0.0 and master as well)

sbernard31:
Yep !

PR #347, opened 2017-06-30
Fix error stopping an connector, when socket failed to open.

boaks:
Issue #345


sbernard31:
LGTM

PR #349, opened 2017-07-05
Split dtls test.

boaks:
Move correlation tests into seperate test class.
Use ConnectorHelper for the correlation tests.


sbernard31:
LGTM

Issue #344, opened 2017-06-29
CorrelationContext


boaks:
I did a short test using the principal names as endpoint identity for DTLS and notifies.

This extents the exporting of the `CorrelationContext` over the scope of the implementation of `ObservationStore`. So I would like to have the usage and expectation more defined.

FMPOV a CorrelationContext should be considered as a "unmodifiable map of string to string". The current used types (e.g. `DtlsCorrelationContext`) should not be mandatory, I would even like to change them to Builders. That would remove some comfort methods (e.g.  `DtlsCorrelationContext.getEpoch()` ) but on the other side it would simplify to provide a CorrelationContext as input for requests and the save/restore mechanism in the `ObservationStore`.

So any thoughts about that?

sbernard31:
> I did a short test using the principal names as endpoint identity for DTLS and notifies.

What do you mean by _for DTLS_ ?

boaks:
"principal names" could only identify a endpoint using DTLS, as far as I know.
And for the normal request/response mapping, in my opinion the currently used address:port mapping should do it. So the "main pain" with the ip-address changes right now seems to be the notifies. Therefore I wrote "endpoint identity for DTLS and notifies".

sbernard31:
About request/response mapping, I never experiment it but I suspect that with NAT timeout/retransmission/deduplication you could sometime face some unexpected behavior.

But I agree the priority is observation, so let's focus on it.

It seems that your proposition target the same issue than #322.
We start a discussion about that there. I just missed the last response (sry about that). Should we continue the discussion to find the best solution with what we have in mind  here or at #322 ?

boaks:
> Should we continue the discussion to find the best solution with what we have in mind here or at #322 ?

Where ever you prefer.

But, let me just say, I think, that independent from the usage in #322, I think, if we use the correlation context  "external", as a external `ObservationStore` implementation would also imply, we should define some more about the rules to use it.

sbernard31:
> if we use the correlation context "external", as a external ObservationStore implementation would also imply, we should define some more about the rules to use it.

you mean for example limit it to an unmodifiable map of string to string?
For the unmodifiable map, I agree. 
For the string to string, I could agree. Data in correlation constraint will probably be serialized/deserialized at some point or another, So I ask myself if we should do that when we create the context (so using string=>string) or when we really need to serialize/deserialize it (and using string => object)?

---------
[...]

PR #350, opened 2017-07-05
Add endpoint id to correlation context for matching based on that.

boaks:
This introduces a endpoint id to the correlation context and is intended
as "work in progress". The test in the cf-secure demo shows, how
notifies could be processed with this approach. Currently the
ObservationStore is not extended and therefore this is also based on
unique tokens. 

NOT INTENDED TO BE MERGED!

### DIFF
--- a/demo-apps/cf-secure/src/test/java/org/eclipse/californium/secure/test/FixPskStore.java
+++ b/demo-apps/cf-secure/src/test/java/org/eclipse/californium/secure/test/FixPskStore.java
@@ -1,0 +1,51 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Achim Kraus (Bosch Software Innovations GmbH) - initial implementation.
+ *                                                    Modified variant of ManInTheMiddle
+ ******************************************************************************/
+
+package org.eclipse.californium.secure.test;
+
+import java.net.InetSocketAddress;
+import java.util.Arrays;
+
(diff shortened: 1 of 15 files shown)


demo-apps/cf-secure/src/test/java/org/eclipse/californium/secure/test/FixPskStore.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.eclipse.californium.secure.test;
+
+import java.net.InetSocketAddress;
+import java.util.Arrays;
+
+import org.eclipse.californium.scandium.dtls.pskstore.PskStore;
+import org.eclipse.californium.scandium.util.ServerNames;
+
+public class FixPskStore implements PskStore {
```
sbernard31: It seems the same as : org.eclipse.californium.scandium.dtls.pskstore.StaticPskStore

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -723,6 +725,7 @@ private void processApplicationDataRecord(final Record record) {
 	private void handleApplicationMessage(ApplicationMessage message, DTLSSession session) {
 		if (messageHandler != null) {
 			DtlsCorrelationContext context = new DtlsCorrelationContext(
+					session.getPeerIdentity().getName(),
```
sbernard31: There is no way to choose what to use as endpoint Identity.
What happens for unauthenticated peer ?

demo-apps/cf-secure/src/test/java/org/eclipse/californium/secure/test/FixPskStore.java
sbernard31: it seems to be the same than org.eclipse.californium.scandium.dtls.pskstore.StaticPskStore

boaks: OK, I tried the InMemory ...

boaks:
The PR was not intended to be merged and therefore only contained the principals of the idea.

In the meantime, I think just moving the `Principal` from `Request` to `Message` and use the `Message` for the `ObservationStore` will do it with even less changes

https://github.com/eclipse/californium/issues/344#issuecomment-314075391 

and would enable an `ObservationStore` implementer to add either your "identity/factory" approach or use the "principal's name" directly to do so.

boaks:
Not longer intended.

PR #355, opened 2017-07-10
Unable to restart blockwise after an incomplete one.

sbernard31:
The use case which helps me to discover that bug, is the same than #211.
After the scenario described in #211, if you try to do a GET with block2  (see test case of this PR) you will face to something like that : 

```
WARNING [BlockwiseLayer]: ignoring block2 response with wrong block number 0 (expected 1):
```

My understanding : we still have blockwise data for the given path because the first block transfer never finished and we never clean it.

The fix, I proposed in #211 which has been refused fix this scenario and I continue to think it should be integrated. (I re-add it in this PR)
But maybe we face a more larger problem with incomplete block transfer.
More tests need to be written to explore that.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -572,6 +572,18 @@ public class BlockwiseLayer extends AbstractLayer {
 					upper().receiveResponse(exchange, resp);
 				}
 			} else if (!response.hasBlockOption()) {
+				
+				// most recent none blockwise notification should stop current blockwise transfer
+				KeyUri key = getKey(exchange, response);
+				Block2BlockwiseStatus status = getBlock2Status(key);
+				if (response.getOptions().hasObserve() && status != null && status.isNotification()
+						&& status.getObserve() < response.getOptions().getObserve()) {
+					// complete current exchange
+					exchange.setComplete();
+					exchange.setCurrentRequest(exchange.getRequest());
+					// clear blockwise status
+					clearBlock2Status(key);
+				}
 
 				// This is a normal response, no special treatment necessary
 				exchange.setResponse(response);
@@ -893,7 +905,6 @@ public class BlockwiseLayer extends AbstractLayer {
 					}
 
 				} else {
-
 					// ERROR, wrong block number (server error)
 					// Canceling the request would interfere with Observe, so just ignore it
 					LOGGER.log(Level.WARNING,
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
@@ -320,7 +320,30 @@ public class ObserveClientSideTest {
 
 	/**
 	 * Verifies behavior of observing a resource which start a blockwise
-	 * transfer and receiving a notification without blockwise at the same time.
+	 * transfer and receiving a none-blockwise notification before previous
+	 * blockwise transfer finished.
+	 * 
+	 * <pre>
+	 *    (actual used MIDs and Tokens my vary!)
+	 *    ####### establish observe an start blockwise tranfer #############
+	 *    CON [MID=37434, T=cbeb45abdb65aba2], GET, /test, observe(0)    ----->
+	 *    <-----   ACK [MID=37434, T=cbeb45abdb65aba2], 2.05, 2:0/1/16, observe(1)
+	 *    CON [MID=37435, T=d2b1c9e97df24976], GET, /test, 2:1/0/16    ----->
+	 *    ####### receive none blockwise notification #############
+	 *    <-----   CON [MID=222, T=cbeb45abdb65aba2], 2.05, observe(2)
+	 *    ACK [MID=222]   ----->
+	 *    ####### ensure previous blockwise tranfer is stopped #############
+	 *    <-----   CON [MID=37435, T=d2b1c9e97df24976], 2.05, 2:1/1/16
+	 *    RST [MID=37435]   ----->
+	 *    ####### ensure next none-blockwise notification is received #############
+	 *    <-----   CON [MID=223, T=cbeb45abdb65aba2], 2.05, observe(3)
+	 *    ACK [MID=223]   ----->
+	 *    ####### ensure next blockwise notification is received #############
+	 *    <-----   CON [MID=224, T=cbeb45abdb65aba2], 2.05, 2:0/1/16, observe(4)
+	 *    ACK [MID=224]   ----->
+	 *    CON [MID=21291, T=61c34585da9003e7], GET, /test, 2:1/0/16    ----->
+	 *    <-----   ACK [MID=21291, T=61c34585da9003e7], 2.05, 2:1/0/16
+	 * </pre>
 	 * 
 	 * @throws Exception
 	 */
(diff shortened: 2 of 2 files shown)


boaks:
Thought I was unclear about the use-case, I tried to capture the traffic.

> "No.","Time","Source","Destination","Port","Length","Protocol","Info"
> "6","07:38:07.582141","127.0.0.1","127.0.0.1","49648","96","CoAP","CON, MID:40599, GET, TKN:16 8b d0 1a 8b 5a 27 2b, /test"
> "7","07:38:07.585194","127.0.0.1","127.0.0.1","49647","68","CoAP","ACK, MID:40599, Empty Message"
> "8","07:38:07.588260","127.0.0.1","127.0.0.1","49647","128","CoAP","CON, MID:111, 2.05 Content, TKN:16 8b d0 1a 8b 5a 27 2b, Block #0 (text/plain)"
> "9","07:38:07.588313","127.0.0.1","127.0.0.1","49647","106","CoAP","CON, MID:222, 2.05 Content, TKN:16 8b d0 1a 8b 5a 27 2b (text/plain)"
> "10","07:38:07.590949","127.0.0.1","127.0.0.1","49648","68","CoAP","ACK, MID:111, Empty Message"
> "11","07:38:07.594821","127.0.0.1","127.0.0.1","49648","98","CoAP","CON, MID:40600, GET, TKN:82 d9 f1 64 c9 0d 6d 71, End of Block #1, /test"
> "12","07:38:07.597091","127.0.0.1","127.0.0.1","49648","68","CoAP","ACK, MID:222, Empty Message"
> "13","07:38:07.597337","127.0.0.1","127.0.0.1","49647","124","CoAP","ACK, MID:40600, 2.05 Content, TKN:82 d9 f1 64 c9 0d 6d 71, Block #1 (text/plain)"
> "14","07:38:07.600374","127.0.0.1","127.0.0.1","49647","106","CoAP","CON, MID:223, 2.05 Content, TKN:16 8b d0 1a 8b 5a 27 2b (text/plain)"
> "15","07:38:07.603378","127.0.0.1","127.0.0.1","49648","68","CoAP","ACK, MID:223, Empty Message"
> "16","07:38:07.603716","127.0.0.1","127.0.0.1","49647","106","CoAP","CON, MID:224, 2.05 Content, TKN:16 8b d0 1a 8b 5a 27 2b (text/plain)"
> "17","07:38:07.605939","127.0.0.1","127.0.0.1","49648","68","CoAP","ACK, MID:224, Empty Message"
> "18","07:38:07.607443","127.0.0.1","127.0.0.1","49648","94","CoAP","CON, MID:40601, GET, TKN:9c 0d 89 c3 7c 1f 97 5a, /test"
> "19","07:38:07.607638","127.0.0.1","127.0.0.1","49647","68","CoAP","ACK, MID:40601, Empty Message"
> "20","07:38:07.607684","127.0.0.1","127.0.0.1","49647","124","CoAP","CON, MID:112, 2.05 Content, TKN:9c 0d 89 c3 7c 1f 97 5a, Block #0 (text/plain)"
> "21","07:38:07.609817","127.0.0.1","127.0.0.1","49648","68","CoAP","ACK, MID:112, Empty Message"
> 

I'm wondering, it seems, that there are two responses (one blockwise MID111, one none-blockwise MID222).

The basic violation at the end seems, that the response to the new request seems to be checked against the expected response for the first blockwise transfer. FMPOV this need more investigation ...

sbernard31:
> I'm wondering, it seems, that there are two responses (one blockwise MID111, one none-blockwise MID222).

Yes, all is explained in #211.

> The basic violation at the end seems, that the response to the new request seems to be checked against the expected response for the first blockwise transfer.

That's what I tried to explain by : _"we still have blockwise data for the given path because the first block transfer never finished and we never clean it."_
Even if I'm not sure what you mean by basic violation. It's a bug.

boaks:
I didn't saw the irritating two consecutive CON message from the same side in #211 as above messages 8 and 9. But this seems to be just a artifact of the test implementation with no further consequences.
The "violation", which causes the GET to break, is then printed as warning:

> WARNING: ignoring block2 response with wrong block number 0 (expected 1): 

So this indicates, as you stated, that there is still a blockwise transfer ongoing.
Your fix will solve the issue, but only if a none-blockwise response is involved.
I added a additional test, which
1. establish a observer with blockwise response
2. send a partial blockwise notiy
3. send a GET request

Though no "none-blockwise" response wasn't involve, the GET still fails.
So FMPOV, a new "GET" should finish the old blockwise transfer.

Is it possible to push my added test to your branch "block2_obs"?

sbernard31:
> Is it possible to push my added test to your branch "block2_obs"?

Sure.

boaks:
@sbernard31 
Maybe you add also your analysis of the test case.

boaks:
The fix https://github.com/eclipse/californium/pull/355/commits/d60a2c8ad51e9dca96779af947a9123435420ea0 is still in. But the tests will not longer fail, even if it is removed.

boaks:
My system reports with the fix https://github.com/eclipse/californium/commit/d60a2c8ad51e9dca96779af947a9123435420ea0 in

> MemoryLeakingHashMapTest.assertAllExchangesAreCompleted:152 Client side message exchange store still contains exchanges

If I remove the fix, the test didn't fail.
Can you check that on your system?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -558,7 +558,7 @@ public void receiveResponse(final Exchange exchange, final Response response) {
 			} else if (!response.hasBlockOption()) {
 				
 				KeyUri key = getKey(exchange, response);
-				Block2BlockwiseStatus status = getInboundBlock2Status(key, exchange, response);
+				Block2BlockwiseStatus status = block2Transfers.get(key);
```
boaks: Or consider `getBlock2Status`

sbernard31: you're right.

boaks:
Though I started to think, that there are (at least :-) two issues, I split my changes into an other branch.
1. issue: pending bockwise notify, deprecated by a new notify, so blockwise transfer not longer required
2. issue: pending bockwise notify, but taken over by a new request

So I would like to move my commits into an other PR and prepare this branch with your commits addressing issue 1.
This would require a forced push on the branch "block2_obs", would this be OK?
Or do you prefer to remove my commits by yourself?

sbernard31:
I agree about the "at least" 2 issues. 

1. should be fixed with commit d60a2c8.
2. is partially fixed by 396c0b3. (see comment below)

I'm OK to separate this PR in 2 distinct PRs. I deal with this one (clean the history and maybe write more specific tests about 1.).

About 2. I think about different blockwised interlacing use cases.

a) A pending blockwise notification interrupted by a new blockwise GET request.
=> with your fix, we should give the priority to the GET request. the notification blockwise transfer will be stopped. (I have concern about the notification exchange cleaning) 
=> it's ok for me.

b) A pending blockwise notification interrupted by a pending blockwise notification
=> priority to the most recent one, I think this is the way this is implemented now. (same concern about the exchange cleaning)
=> ok for me too.

c) A blockwise GET request. interrupted by a new blockwise GET request.
=> we should give the priority to the last GET, your fix do that. But should we report that the previous one is aborted maybe using `setSendError` ?

d) A blockwise GET request interrupted by a new notification.
=> This is the only case where I think the older (the GET) should have the priority. I mean If we receive a blockwise notification during a GET we ignore it.

WDYT ? Should we write tests for all this use cases ?

(I have others concerns about blockwise I will open issue to discuss about that.)

boaks:
Just to say, RFC7959, section 2.4, page 13, 
"The Block2 Option provides no way for a single endpoint to perform multiple concurrently proceeding block-wise response payload transfer (e.g., GET) operations to the same resource."

FMPOV, the "transparent blockwise mode" with "internal blockwise get for the left notify payload" will always cause some trouble. Currently I added the test case "blockwise-observer-get-notify", which also fails.

> WDYT ? Should we write tests for all this use cases ?

I'm not sure, if this really pays of. To be frank, GET interleaved with notifies will always tend to corrupt your state, because of race conditions you will not know, which is the newer state. Therefore, I would not recommend that and so it seems to be much work for something I would not recommend.

sbernard31:
I do not understand, what did you recommend exactly ?

boaks:
Don't use RFC7641 together with RFC7959.

sbernard31:
> Don't use RFC7641 together with RFC7959.

I agree, personally I will go for "Don't use RFC7641 at all".

As often, there is not only good Idea in production and unfortunately we have the bad idea to use RFC7641... and RFC7959 together ...
So I cannot just ignoring the problem.

sbernard31:
Ok I cleaned this PR.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
```diff
@@ -333,44 +356,52 @@ public void testBlockwiseObserveAndNotificationWithoutBlockwise() throws Excepti
...
 		// option value)
 		Response response = request.waitForResponse();
 		assertResponseContainsExpectedPayload(response, notifpayload);
 
+		// Send next block
+		server.sendResponse(CON, CONTENT).loadMID("SECOND_BLOCK_MID").loadToken("SECOND_BLOCK_TOK").block2(1, true, 16)
+				.payload(respPayload.substring(16, 32)).go();
+		// ensure client don't ask for block anymore
+		server.expectEmpty(RST, (Integer) server.get("SECOND_BLOCK_MID")).go();
+		// TODO ensure that blockdata buffer is cleared in blockwiselayer...
```
sbernard31: I didn't a find a good way to check blockdata in blockwiselayer is cleaned.
I previously workaround that by sending a GET request and see if it's work but now this test is in #357.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -546,6 +546,18 @@ public void receiveResponse(final Exchange exchange, final Response response) {
...
+				// most recent none blockwise notification should stop current blockwise transfer
+				KeyUri key = getKey(exchange, response);
+				Block2BlockwiseStatus status = getBlock2Status(key);
+				if (response.getOptions().hasObserve() && status != null && status.isNotification()
+						&& status.getObserve() < response.getOptions().getObserve()) {
+					// complete current exchange
+					exchange.setComplete();
+					exchange.setCurrentRequest(exchange.getRequest());
+					// clear blockwise status
+					clearBlock2Status(key);
```
boaks: After two days, I'm pretty sure, that I can't foresee, if the changes are proper working :-).

Just merge and check.

LGTM

boaks: After two day "working on the blockwise stuff" ...

sbernard31: Which part give you some concerns  ?
ClearBlock ?
ClearExchange ?
or both? 

Personally, I think the clearblock2 make senses and is useful.
The exchange one is not so clear mainly because of the design and not so useful as I know one of your previous fix resolve the cleaning exchange indirectly, this is just the "explicit" way.
I think we need a clearer way to clean exchange when we want to interrupt a blockwise transfer.

boaks: It was just my experience: I fix something, something else breaks.

PR #356, opened 2017-07-12
Fix block2 notiy

boaks:
Clear blockwise notify when new request is send.


boaks:
spelling ...

PR #357, opened 2017-07-12
Fix block2 notify

boaks:
Clear blockwise notify when new request is send.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
@@ -22,6 +22,8 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - use new introduced failed() 
  *                                                    instead of onReject() and
  *                                                    onTimeout().
+ *    Achim Kraus (Bosch Software Innovations GmbH) - don't cleanup on cancel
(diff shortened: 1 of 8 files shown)


sbernard31:
Even, if you are not so much exited by the idea, I would like to write tests for use cases described here : https://github.com/eclipse/californium/pull/355#issuecomment-314772196

I think we have tests case for :
a) A pending blockwise notification interrupted by a new blockwise GET request. (in this PR)
b) A pending blockwise notification interrupted by a new blockwise notification (in #355)

I would like to add : 
c) A blockwise GET request. interrupted by a new blockwise GET request.
d) A blockwise GET request interrupted by a new notification.

And maybe add a second test for b)

Can I merge this PR ?

boaks:
>  Even, if you are not so much exited by the idea

I just can't spent too much time in it ...

Currently I'm on d), but the trick in BaseMatcher to ignore the cancel doesn't work there. I think "cancelTransfer" would be required to distinguish the cases ;-(.

boaks:
I think we should merge it, because the "BlockStatus.cancelTransfer" may also be required in for the single block responses. 
But there will be an additional PR.

boaks:
Uups, or did you ment to merge #355?

sbernard31:
I talked about this PR. For #355, I wait you review it.

PR #359, opened 2017-07-13
Fix block2 notify - 2.

boaks:
Introduce "stop transfer" for blockwise transfers


boaks:
Not longer preferred.

PR #363, opened 2017-07-25
Fixes blockwise

boaks:
Correct several issues in blockwise transfer.


boaks:
Rebase reordered commits. So create PR again :-).

PR #367, opened 2017-07-27
Bert TCP Support

khannag:
This PR is to enable BERT TCP support in the master branch. We need these changes to speed up the performance of CoAP GETs (Observing substantial performance benefits when performing multi-block transfers with these changes). I am requesting here to enable BERT support on the current master branch since  the 2.0 branch is not released as yet and while I understand that current master may not be supporting other TCP features, this feature by itself lends a lot of performance benefit.

The code changes are leveraging the two PRs earlier submitted by @raipratheek. I am referring to PR#342 (https://github.com/eclipse/californium/pull/342) and PR# (https://github.com/eclipse/californium/pull/319). Basically, I am still using the pattern of PR#342 to include a separate class TcpBlockWiseLayer, but the actual changes in TcpBlockWiseLayer are much different (leveraged PR# 319 here), since I am working relative to the master branch.

Changes have been tested in our setup and are showing reducing in transfer time by orders of magnitude as mentioned earlier. Note that I used a value of 64 for TCP_NUMBER_OF_BULK_BLOCKS in my tests.

Signed-off-by: Gaurav Khanna <gkkhanna@amazon.com>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
@@ -194,6 +194,13 @@ public final class NetworkConfig {
 		public static final String TCP_CONNECTION_IDLE_TIMEOUT = "TCP_CONNECTION_IDLE_TIMEOUT";
 		public static final String TCP_CONNECT_TIMEOUT = "TCP_CONNECT_TIMEOUT";
 		public static final String TCP_WORKER_THREADS = "TCP_WORKER_THREADS";
+		
+		/**
+		 * If the value is greater than 1, this sets up the active use of BERT.
+		 * i.e. Messages will be sent with BERT option. The passive
+		 * receiving of BERT message is always enabled while using TCP connector.
+		 */
+		public static final String TCP_NUMBER_OF_BULK_BLOCKS = "TCP_NUMBER_OF_BULK_BLOCKS";			
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


khannag:
@boaks, @sophokles73 - It would be great if you could take a look at this PR when you get a chance.

sophokles73:
@khannag 

thanks for your interest in Cf which is also reflected by your submitting this PR. I do appreciate any kind of contribution to the project. However, the blockwise layer has been causing us a lot of pain and trouble to both develop and maintain and support. It is one of the hardest parts of Cf to actually understand and debug. That said, I am not very much inclined to accept any code that will further manifest its current (terrible) state. We already had a long discussion around #342 and I only agreed to accept the code because we *parked* it on an experimental branch only instead of merging it to 2.0.x. I have the very same concern with your PR as I had (and have) with #342, which basically is that I do not want to have all the (bad) code duplicated in a subclass when FMPOV the right way to approach the issue would be to better encapsulate the SZX handling required for downloading block1 transfers and handing out block2 response fragments in the `BlockwiseStatus` class(es). That way we should be able to keep a single `BlockwiseLayer` class which is capable of doing standard and BERT transfers. FMPOV this much better reflects the spirit of the BERT definition which seems to define it as an extension of the existing blockwise transfer by means of new parameter values instead of defining an alternative mechanism.

Do you think we can find some common ground to modify your PR into that direction?

@boaks  @sbernard31 I would also be interested in your opinion about this as well

boaks:
As I wrote in the other PR: currently we are not investing too much in TCP, just because there are not too many people involved at all. We hardly getting UDP working and I don't see the possibility in the near future to change that. For mid future the PR base 2.0.x seems to be in my opinion then something to start than.

khannag:
Thanks for the comments. I agree on the idea of a single BlockwiseLayer Class. I will review the 2.0.x branch and take a stab at the suggested changes - moving the SZX handling to BlockwiseStatus. I will share the changes as soon as I have it done.

boaks:
@khannag 

Currently it seems, that the next release will be the 2.0. So I'm not sure, if the work in 1.1 makes therefore sense. Are you depending on function from 1.1 which are changed too much for 2.0? Or would 2.0 also be fine for you?

boaks:
@khannag 

if you still have interest, just comment to reopen it.

PR #369, opened 2017-07-27
Fixes blockwise

boaks:
Correct several issues in blockwise transfer.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -33,6 +33,9 @@
  *                                                    sendRequest() for more details.
  *                                                    cancel also the pending requests
  *                                                    of the stale transfer.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - introduce stop transfer on
+ *                                                    received responses generally.
+ *                                                    cleanup stale functions.
  ******************************************************************************/
 package org.eclipse.californium.core.network.stack;
 
@@ -208,7 +211,7 @@ public class BlockwiseLayer extends AbstractLayer {
 					// of the notify to be abandoned so that the client receives
 					// the requested response but lose the notify. 
 					clearBlock2Status(key);
-					status.cancelTransfer();
+					status.completeTransfer(null);
 				}
 				
 				if (requiresBlockwise(request)) {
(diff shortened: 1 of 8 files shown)


boaks:
Seems that the order of the commits is mixed up.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -556,34 +558,69 @@ public void receiveResponse(final Exchange exchange, final Response response) {
...
+								Level.FINER,
+								"discarding outdated block2 transfer response for old notification {0}, current is {1}",
+								new Object[]{ status.getObserve(), response });
+						clearBlock2Status(key);
+						status.completeTransfer(exchange);
 					} else {
-						resp.setType(Type.NON);
+						LOGGER.log(
+								Level.FINER,
+								"discarding old notification [{0}] received during ongoing blockwise transfer: {1}",
```
sbernard31: The log talk about "notification" but It seems to me that we can have block of "normal" response.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
```diff
@@ -172,9 +198,9 @@ public void testBlockwiseObserve() throws Exception {
 		client.addNotificationListener(notificationListener);
 		client.sendRequest(request);
 
-		server.expectRequest(CON, GET, path).storeBoth("A").storeToken("T").go();
+		server.expectRequest(CON, GET, path).observe(0).storeBoth("A").storeToken("T").go();
 		server.sendResponse(ACK, CONTENT).loadBoth("A").observe(0).block2(0, true, 16).size2(respPayload.length()).payload(respPayload.substring(0, 16)).go();
-		server.expectRequest(CON, GET, path).storeBoth("B").block2(1, false, 16).go();
+		server.expectRequest(CON, GET, path).noOption(OptionNumberRegistry.OBSERVE).storeBoth("B").block2(1, false, 16).go();
```
boaks: I will remove the added checks.

boaks:
> There is a lot of change in c979964 and I'm not sure to understand the ins and outs of this.
> I try to rebase that branch on #373 and the tests I added fail because in this version request is not cancelled anymore. (cancelTransfer replaced by completeTransfer)

The issue with the "cancel" is demonstrated in PR #379. The first "trick", to ignore the "cancel" for "recreated" observe exchanges doesn't work, if the first notification (response to observe) is interrupted by a fresh notification. Therefore I started to cleanup that mechanism.

Maybe adding a "cancel" for "none observe get" will then cancel such a interrupted get.

boaks:
> I try to rebase that branch on #373 and the tests I added fail because in this version request is not cancelled anymore. 

Should work now.

sbernard31:
LGTM : I integrated it in master.

PR #374, opened 2017-08-04
Allow to get the target and the certificate path from X509CertPath 

AlexITC:
Signed-off-by: Alexis Hernandez <alexis22229@gmail.com>

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/X509CertPath.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/X509CertPath.java
@@ -45,7 +45,7 @@ public class X509CertPath implements Principal {
 	 *                                  X.509 certificates only.
 	 */
 	public X509CertPath(final CertPath certPath) {
-		if (certPath.getType() != TYPE_X509) {
+		if (!TYPE_X509.equals(certPath.getType())) {
 			throw new IllegalArgumentException("Cert path must contain X.509 certificates only");
 		} else if (certPath.getCertificates().isEmpty()) {
 			throw new IllegalArgumentException("Cert path must not be empty");
@@ -96,5 +96,23 @@ public class X509CertPath implements Principal {
 	@Override
 	public String getName() {
 		return target.getSubjectX500Principal().getName();
+	}
+
+	/**
+	 * Gets this certificate path.
+	 *
+	 * @return The path.
+	 */
+	public CertPath getPath() {
+		return path;
+	}
+
+	/**
+	 * Gets the asserted identity of this certificate path.
+	 *
+	 * @return The target certificate.
+	 */
+	public X509Certificate getTarget() {
+		return target;
 	}
 }


sophokles73:
IMHO it would be sufficient to add a *getCertPath* method to `org.eclipse.californium.scandium.auth.X509CertPath`. No need to introduce an additional field all along the stack ...

AlexITC:
@sophokles73 you are right, its quite simpler in this way, thanks!

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/X509CertPath.java
```diff
@@ -97,4 +97,18 @@ public static X509CertPath fromBytes(final byte[] encodedPath) {
 	public String getName() {
 		return target.getSubjectX500Principal().getName();
 	}
+
+	/**
+	 * Gets this certificate path.
+	 *
+	 * @return The path.
+	 */
+	public CertPath getPath() { return path; }
```
sophokles73: please put the method body on its own line (like in the the rest of the class)

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/X509CertPath.java
```diff
@@ -97,4 +97,18 @@ public static X509CertPath fromBytes(final byte[] encodedPath) {
...
+
+	/**
+	 * Gets the asserted identity of this certificate path.
+	 *
+	 * @return The subject.
```
sophokles73: this is not the subject but the whole certificate

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
```diff
@@ -664,7 +664,7 @@ public Principal getPeerIdentity() {
 	/**
 	 * Sets the authenticated peer's identity.
 	 * 
-	 * @param the identity
+	 * @param peerIdentity the identity
```
sophokles73: this doesn't seem to be related to the other changes.
next time, please create a separate PR for it

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/X509CertPath.java
AlexITC: the method body is already in the same line as the method signature, isn't it?

sophokles73: what I mean is:
````
public CertPath getPath() {
  return path;
}
````

AlexITC: sure, I placed it this way following the style of `getName`, I'll change it soon, thanks.

AlexITC:
@sophokles73 I just noticed the `getName` method was not compacted, Intellij was rendering it in as compacted, sorry for that, code was updated. Thanks!

PR #375, opened 2017-08-07
Add assumption for no unintended message retransmission.

boaks:
If a test fails on unintended retransmission (caused by execution time),
this changes the test "fails" into "ignore result".

Failure

https://hudson.eclipse.org/californium/job/1.0.x/95/

would therefore be ignored instead of being reported as error.

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/LockstepEndpoint.java
@@ -30,6 +30,14 @@
  *                                                    to accept multiple types (Type... types).
  *                                                    Changed reponseType in type(Type... types)
  *                                                    and storeType().
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add assumption for no unintended message
+ *                                                    retransmission. If a test fails on
+ *                                                    unintended retransmission (caused by
+ *                                                    execution time), this changes the test
+ *                                                    "fails" into "ignore result". Only
+ *                                                    support, if system property
+ *                                                    "org.eclipse.californium.junit.starving"
+ *                                                    is also set to true.
  ******************************************************************************/
 package org.eclipse.californium.core.test.lockstep;
 
@@ -46,7 +54,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
 import org.junit.Assert;
-
+import org.junit.Assume;
 import org.eclipse.californium.core.Utils;
 import org.eclipse.californium.core.coap.BlockOption;
 import org.eclipse.californium.core.coap.CoAP.Code;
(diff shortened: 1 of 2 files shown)


sbernard31:
I do not understand the point. Why should we ignore test results when we detect retransmission ?

boaks:
The idea is to **selective** ignore tests, if there is a **unintended** retransmission.

Currently this only occurs for CON notifies with changing payload.
If you look at the log of the failed test, it shows that the "observe: 1" is sent 3 times. 

https://hudson.eclipse.org/californium/job/1.0.x/95/testReport/junit/org.eclipse.californium.core.test.lockstep/ObserveServerSideTest/testEstablishmentAndTimeoutWithUpdateInMiddle/

but the intention was only 2 times. If that happens, the "retransmission counter" expires before the intended last retransmission is sent and therefore the "expectation" could not be longer fulfilled.

Yes, it brings in some uncertainty, if this is caused by bugs rather then by HIPP starvation. But from my experience over the last months, the eclipse HIPP starvation causes a lot more issue, then mass repeated execution in a more deterministic environment shows.

sbernard31:
I took so much time to understand this test...
I didn't know that notification retransmission works like that. (I mean using the same counter for different notification...)

About your modification, your idea is to ignore tests which could failed because of HIPP starvation. In that case do we want to ignore it or relaunch it ?

boaks:
> I didn't know that notification retransmission works like that. (I mean using the same retransmission counter for different notification...)

RFC 7641, 4.5.2. Advanced Transmission, page 19:

> 4. The new notification is transmitted with a new Message ID and the
> following transmission parameters: if the previous
> (re)transmission attempt timed out, retain its transmission
> parameters, increment the retransmission counter, and double the
> timeout; otherwise, initialize the transmission parameters as
> usual (see Section 4.2 of RFC 7252 [RFC7252]).

So I consider, "congestion control" was higher prioritized then "not losing notifies", because "not losing notifies" is not ensured by RFC 7641 at all.

> About your modification, your idea is to ignore tests which could failed because of HIPP starvation. In that case do we want to ignore it or relaunch it ?

Just ignore. But I think, adding a system property additionally to the optional setter would make the ignore even more "limited" to starving HIPP. I will add that  ...

sbernard31:
Ok sounds good.

Just a thought :
We have issue with [HIPP starvation](https://bugs.eclipse.org/bugs/show_bug.cgi?id=515076). It seems we can not find any good solution about that.

But what is the "Hudson tests run" purpose ? If we want to validate behavior, should we just relaunch tests (manually) when we have starvation issue ? I mean if it failed because of bad condition on test machine, we should retry until condition are better instead of ignoring it ?

(As I said, it's just a thought and if you still think it's better to ignore, let's go for it :) )

boaks:
FMPOV

I guess, the HIPP starvation is a result of overloading. Retries would do even more overloading. If all projects apply such a retry strategy, nothing will work.
Though it seems to be not possible, to arrange "exclusive" schedules (even on centslave1), I would just not spent too much work on these tests running on the eclipse hudson infrastructure.
Therefore I set up a HIPP on exclusive hardware by my own, and, if test failing there, it's much more effective to catch them up there (e.g. PR #325).

Issue #372, opened 2017-08-01
Observation stops after some time


peterkappelt:
Hi,

I'm observing a CoAP-Resource, with DTLS. Everything works fine, but after some minutes there are no more reactions to changes in the resource. I've reduced the problem to the following example:
<details>
<summary>Click to expand</summary>

```Java
class test{
	private static final String TRUST_STORE_PASSWORD = "rootPass";
	private static final String KEY_STORE_PASSWORD = "endPass";
	private static final String KEY_STORE_LOCATION = "certs/keyStore.jks";
	private static final String TRUST_STORE_LOCATION = "certs/trustStore.jks";
	
	private CoapClient client;
	private DTLSConnector dtlsConnector = null;
	
	public test() {
		CaliforniumLogger.initialize();
		CaliforniumLogger.setLevel(Level.FINER);
		
		ScandiumLogger.initialize();
		ScandiumLogger.setLevel(Level.FINER);
		
		client = new CoapClient();
		
		try {
			// load key store
			KeyStore keyStore = KeyStore.getInstance("JKS");
			InputStream in = getClass().getClassLoader().getResourceAsStream(KEY_STORE_LOCATION);
			keyStore.load(in, KEY_STORE_PASSWORD.toCharArray());
			in.close();

			// load trust store
			KeyStore trustStore = KeyStore.getInstance("JKS");
			in = getClass().getClassLoader().getResourceAsStream(TRUST_STORE_LOCATION);
			trustStore.load(in, TRUST_STORE_PASSWORD.toCharArray());
			in.close();

			// You can load multiple certificates if needed
			Certificate[] trustedCertificates = new Certificate[1];
			trustedCertificates[0] = trustStore.getCertificate("root");

			DtlsConnectorConfig.Builder builder = new DtlsConnectorConfig.Builder(new InetSocketAddress(0));
			builder.setPskStore(new StaticPskStore("Client_identity", "mysecret".getBytes()));
			//builder.setPskStore(new StaticPskStore("password", "sesame".getBytes()));
			builder.setIdentity((PrivateKey)keyStore.getKey("client", KEY_STORE_PASSWORD.toCharArray()),
					keyStore.getCertificateChain("client"), true);
			builder.setTrustStore(trustedCertificates);
			
			//try to fix timeouts at user
			//builder.setRetransmissionTimeout(50000);
			
			dtlsConnector = new DTLSConnector(builder.build());
			
		} catch (Exception e) {
[...]

PR #376, opened 2017-08-07
Fix client ping endpoint

boaks:
Fixing NullPointerException mentioned in #372.


sbernard31:
👍

PR #379, opened 2017-08-08
Add test for new notification during establishing the observe.

boaks:


boaks:
Not intended to be merged!

PR #380, opened 2017-08-09
use singleThreadexecutor if user requires 1 thread in DTLSConnector

sbernard31:
We currently face some CPU in production. After a long period of time the CPU increase exponentially while the traffic is constant. I try to reproduce that in a tests environment without succeed. 

But while I playing with visualVM to try to understand I see that `StripedExecutorService.SerialExecutor.scheduleNext()` is a hotspot.

So I propose a little improvement to not use it when only 1 thread is used for DTLSConnector (mainly Device use case)

```
Benchmark                                      Mode  Cnt        Score        Error  Units
StripedExecutorBenchmark.testFixedExecutor    thrpt   20  5073138,079 ± 483729,502  ops/s
StripedExecutorBenchmark.testSingleExecutor   thrpt   20  4160778,781 ± 524090,520  ops/s
StripedExecutorBenchmark.testStripedExecutor  thrpt   20  1062163,140 ± 221670,144  ops/s
```

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -72,6 +72,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
@@ -208,7 +209,7 @@ public class DTLSConnector implements Connector {
 	private RawDataChannel messageHandler;
 	private ErrorHandler errorHandler;
 	private SessionListener sessionCacheSynchronization;
-	private StripedExecutorService executor;
+	private ExecutorService executor;
 	private boolean hasInternalExecutor;
 
 	/**
@@ -346,7 +347,13 @@ public class DTLSConnector implements Connector {
 			} else {
 				threadCount = config.getConnectionThreadCount();
 			}
-			executor = new StripedExecutorService(Executors.newFixedThreadPool(threadCount, new DaemonThreadFactory("DTLS-Connection-Handler-", NamedThreadFactory.SCANDIUM_THREAD_GROUP)));
+			if (threadCount == 1) {
+				// Scheduling Striped task is costly so if user require only 1 Thread, we use a simple SingleThreadExecutor
+				executor = Executors.newSingleThreadExecutor(new DaemonThreadFactory("DTLS-Connection-Handler-", NamedThreadFactory.SCANDIUM_THREAD_GROUP));
+			} else {
+				executor = new StripedExecutorService(Executors.newFixedThreadPool(threadCount, new DaemonThreadFactory("DTLS-Connection-Handler-", NamedThreadFactory.SCANDIUM_THREAD_GROUP)));	
+			}
+			
 			this.hasInternalExecutor = true;
 		}
 		socket = new DatagramSocket(null);


sophokles73:
Not arguing against your proposed change, I would, however, like to better understand why the striped executor becomes a problem when used with a single thread only. Did you do any analysis in that direction so far?

boaks:
> After a long period of time the CPU increase exponentially while the traffic is constant. I try to reproduce that in a tests environment without succeed.

"long period of time" according the development-time or according runtime?

sbernard31:
@sophokles73 
> I would, however, like to better understand why the striped executor becomes a problem when used with a single thread only.

My investigation make me think that using StripedExecutor is not cost free (for 1 or several Threads). My idea is just that if the user want only 1 Thread to handle DTLS message, this cost is not necessary. As StripedExecutor is for handle concurrency and with 1 Thread no concurrency.

@boaks 
> "long period of time" according the development-time or according runtime?

runtime, I mean after several day in production (not determinist) the CPU consumption increase with any apparent  change in traffic.

@boaks 
> If you have a test setup for that, maybe you could recheck the hotspot removing removeEmptySerialExecutor

I will do that, I will let you know.

boaks:
> runtime, I mean after several day

that's bad ... smells like it will eat up a lot of time to fix it ...

boaks:
Just to ensure:

You now plan to run it with 1 thread and without the stripped to see, if the issue disappears?
If so, LGTM.

sbernard31:
> You now plan to run it with 1 thread and without the stripped to see, if the issue disappears?

With singleThread, I can not see it anymore in hotspot and jmh test show singleThread executor is 4x faster to schedule a task. This is the only thing I can say.

> If you have a test setup for that, maybe you could recheck the hotspot removing removeEmptySerialExecutor

A bit better without removeEmptySerialExecutor (see PatchedStripedExecutor), but not so much.
```
Benchmark                                             Mode  Cnt        Score        Error  Units
StripedExecutorBenchmark.testFixedExecutor           thrpt   20  6120422,680 ± 477349,466  ops/s
StripedExecutorBenchmark.testPatchedStripedExecutor  thrpt   20  1717593,591 ± 565597,200  ops/s
StripedExecutorBenchmark.testSingleExecutor          thrpt   20  4178708,111 ± 696056,451  ops/s
StripedExecutorBenchmark.testStripedExecutor         thrpt   20  1126006,975 ± 206222,707  ops/s
```

boaks:
> You now plan to run it with 1 thread and without the stripped to see, if the issue disappears?

let me precise my question:

> After a long period of time the CPU increase exponentially while the traffic is constant.

Means "after a couple of days running with the stripped executor". So, do you plan to check, if using a single thread without the stripped executor doesn't show that CPU increase?

boaks:
Is your StripedExecutorBenchmark available?

sbernard31:
No I don't think so. The CPU issue is just the story about why I propose this little optimisation.
But I'm pretty sure this will not solve this CPU issue and more than that in our server in production we want to use several threads to handle DTLS connection.

sbernard31:
```java
package org.eclipse.californium.benchmark;

import java.util.Random;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;

import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Level;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.TearDown;

import eu.javaspecialists.tjsn.concurrency.stripedexecutor.StripedExecutorService;
import eu.javaspecialists.tjsn.concurrency.stripedexecutor.StripedRunnable;

public class StripedExecutorBenchmark {
	
	@State(Scope.Thread)
    public static class FixedState {

        @Setup(Level.Trial)
        public void doSetup() throws InterruptedException {
        	executor = Executors.newFixedThreadPool(10);
        }
        
        
        @TearDown(Level.Trial)
        public void  tearDown() {
        	executor.shutdownNow();
        }

        public ExecutorService executor;
        public Runnable runnable;
        
        @Setup(Level.Invocation)
        public void prepare() {
            runnable = new Runnable() {
				
				@Override
				public void run() {
				}
			};
        }
    }
	
	@State(Scope.Thread)
    public static class SingleState {

        @Setup(Level.Trial)
        public void doSetup() throws InterruptedException {
        	executor = Executors.newSingleThreadExecutor();
        }
        
        
        @TearDown(Level.Trial)
        public void  tearDown() {
        	executor.shutdownNow();
        }

        public ExecutorService executor;
        public Runnable runnable;
        
        @Setup(Level.Invocation)
        public void prepare() {
            runnable = new Runnable() {
				
				@Override
				public void run() {
				}
			};
        }
    }
	
	@State(Scope.Thread)
    public static class PatchedStripedState {

        @Setup(Level.Trial)
        public void doSetup() throws InterruptedException {
        	executor = new org.eclipse.californium.benchmark.StripedExecutorService();
        }
        
        
        @TearDown(Level.Trial)
        public void  tearDown() {
        	executor.shutdownNow();
        }

        public ExecutorService executor;
        public Runnable runnable;
        public Random r = new Random();
        
        @Setup(Level.Invocation)
        public void prepare() {
            runnable = new StripedRunnable() {
				
				@Override
				public Object getStripe() {
					return r.nextInt(10);
				}
				
				@Override
				public void run() {
				}
			};
        }
    }
	
	@State(Scope.Thread)
    public static class StripedState {

        @Setup(Level.Trial)
        public void doSetup() throws InterruptedException {
        	executor = new StripedExecutorService();
        }
        
        
        @TearDown(Level.Trial)
        public void  tearDown() {
        	executor.shutdownNow();
        }

        public ExecutorService executor;
        public Runnable runnable;
        public Random r = new Random();
        
        @Setup(Level.Invocation)
        public void prepare() {
            runnable = new StripedRunnable() {
				
				@Override
				public Object getStripe() {
					return r.nextInt(10);
				}
				
				@Override
				public void run() {
				}
			};
        }
    }

    @Benchmark
    public void testSingleExecutor(SingleState state) {
    	state.executor.execute(state.runnable);
    }
    
    @Benchmark
    public void testFixedExecutor(FixedState state) {
    	state.executor.execute(state.runnable);
    }
    
    @Benchmark
    public void testStripedExecutor(StripedState state) {
    	state.executor.execute(state.runnable);
    }
    
    @Benchmark
    public void testPatchedStripedExecutor(PatchedStripedState state) {
    	state.executor.execute(state.runnable);
    }
}
```

boaks:
So I hope, I got it now.
The benchmark is only intended to show, that, if on thread is configured, the striped executor would consume unnecessary CPU time.
Therefore this PR is only intended to fix that very case.
For all other cases, the real issue behind is still unresolved.
Right?

sbernard31:
Right it's just an optimization for the 1 thread use case.

boaks:
sorry, I edited/updated my comment :-)

> But then, what's about just don't use an additional executor at all if only one thread is configured?

boaks:
To make the benchmarks fair:

```
            runnable = new StripedRunnable() {
				Object stripe = r.nextInt(10);
            	
				@Override
				public Object getStripe() {
					return stripe;
				}
				
				@Override
				public void run() {
				}
			};
```

With that, I get: 

```
StripedExecutorBenchmark.testFixedExecutor           thrpt   20   362204.120 ±   8452.734  ops/s
StripedExecutorBenchmark.testPatchedStripedExecutor  thrpt   20   397284.351 ± 289543.449  ops/s
StripedExecutorBenchmark.testSingleExecutor          thrpt   20  3666589.609 ± 390258.912  ops/s
StripedExecutorBenchmark.testStripedExecutor         thrpt   20   379899.075 ± 286783.801  ops/s
```

So the stripes penalty seems to be very similar to the `Executors.newFixedThreadPool( );`, which is also the base for the stripped.

But back, why should we add an additional executor, if we want only 1 thread?

sbernard31:
> why should we add an additional executor, if we want only 1 thread?

I didn't really think about that.
When you say "no additional executor" you mean no new thread at all. If we do that we face some concurrent task execution (because of DTLS-Retransmist-Task Thread and DTLS-Receiver Thread)

> To make the benchmarks fair:

Well spotted ! but this didn't change the result on my side : 
```
StripedExecutorBenchmark.testFixedExecutor           thrpt   20  6184897,037 ± 287306,413  ops/s
StripedExecutorBenchmark.testPatchedStripedExecutor  thrpt   20  2431119,765 ± 737786,342  ops/s
StripedExecutorBenchmark.testSingleExecutor          thrpt   20  3760446,049 ± 290996,685  ops/s
StripedExecutorBenchmark.testStripedExecutor         thrpt   20   846904,275 ± 210844,451  ops/s
```
I do not understand why our results are so different ...

I didn't imagine this PR will generate so many discussion 😆

boaks:
> When you say "no additional executor" you mean no new thread at all. If we do that we face some concurrent task execution (because of DTLS-Retransmist-Task Thread and DTLS-Receiver Thread)

But just for a "simple synchronisation" of those two using and additional Executor with all the "performance and resource" that comes with it, done for optimization, seems for me to be some conflict. But anyway, merge it. And maybe, open an issue about CPU, which triggers this PR on your side.

sbernard31:
@boaks to be clear. I don't say what you propose is bad idea. I just say I didn't think so much about that.
Replace StripedExecutor by a SingleThreadExecutor is just an easy/simple/little measured optimization without behavior changing. Maybe what you propose is better but this is not so easy to do and this will change a bit the behavior as handling connection and reading socket will not be done in parallel.

Sry about the CPU issue, I didn't want to confuse you :-(
I just want to let you know why I made this little optimization which could seems a bit silly without context. 

If you want I can open another for CPU issue but I haven't any explanation any clues for now .... and I'm not even sure if this is really a californium issue. (maybe the issue is in our proprietary code :/)

boaks:
> But anyway, merge it.

sbernard31:
It's done. Thx you both for your time :)

PR #381, opened 2017-08-09
Improve ToHex String from Scandium

sbernard31:
I'm still hunting hotspot and see that the `ByteArrayUtils.toHexString(byte[])`  is really really slow.

```
Benchmark                     Mode  Cnt        Score       Error  Units
ToHexTest.testApacheToHex    thrpt   20  2497947,633 ± 98144,759  ops/s
ToHexTest.testScandiumToHex  thrpt   20     7149,110 ±   212,437  ops/s
ToHexTest.testStackOverFlow  thrpt   20  3117261,835 ± 29545,251  ops/s
```

I was going to replace it by the apache commons codec one (same we use in Leshan) but I see that the proposed one in [stackoverflow](https://stackoverflow.com/questions/9655181/how-to-convert-a-byte-array-to-a-hex-string-in-java/9855338#9855338) is even better.

This `toHexString` method is mainly called on 
```java
DtlsCorrelationContext context = new DtlsCorrelationContext(
					session.getSessionIdentifier().toString(),
					String.valueOf(session.getReadEpoch()),
					session.getReadStateCipher());
```
in DTLSConnector (`session.getSessionIdentifier.toString()`), the others call is generally guarded by a `if LOGGER.isLoggable()`.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
@@ -586,7 +586,9 @@ public class ClientHandshaker extends Handshaker {
 				if (key != null) {
 					rawPublicKeyBytes = key.getEncoded();
 				}
-				LOGGER.log(Level.FINE, "sending CERTIFICATE message with client RawPublicKey [{0}] to server", ByteArrayUtils.toHexString(rawPublicKeyBytes));
+				if (LOGGER.isLoggable(Level.FINE)) {
+					LOGGER.log(Level.FINE, "sending CERTIFICATE message with client RawPublicKey [{0}] to server", ByteArrayUtils.toHexString(rawPublicKeyBytes));
+				}
 				clientCertificate = new CertificateMessage(rawPublicKeyBytes, session.getPeer());
 			} else {
 				X509Certificate[] clientChain = determineClientCertificateChain(certificateRequest);
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
@@ -143,27 +143,20 @@ public class ByteArrayUtils {
 	/**
 	 * Takes a byte array and returns it HEX representation.
 	 * 
-	 * @param byteArray
-	 *            the byte array.
+	 * @param bytes the byte array.
 	 * @return the HEX representation.
 	 */
-	public static String toHexString(byte[] byteArray) {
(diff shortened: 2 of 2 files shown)


boaks:
Can you please provide the benchmark as well?

sbernard31:
For this one nothing special : 

```java
    @State(Scope.Thread)
    public static class MyState {

        public byte[] payload = new byte[256];
        public Random r = new Random();
        
        @Setup(Level.Invocation)
        public void prepare() {
            r.nextBytes(payload);
        }
    }
    
    
    @Benchmark
    public void testScandiumToHex(MyState state) {
        ByteArrayUtils.toHexString(state.payload);
    }
    
    @Benchmark
    public void testApacheToHex(MyState state) {
    	Hex.encodeHex(state.payload).toString();
    }
    
    @Benchmark
    public void testStackOverFlow(MyState state) {
    	bytesToHex(state.payload);
    }
    
    
    private final static char[] hexArray = "0123456789ABCDEF".toCharArray();
    public static String bytesToHex(byte[] bytes) {
        char[] hexChars = new char[bytes.length * 2];
        for ( int j = 0; j < bytes.length; j++ ) {
            int v = bytes[j] & 0xFF;
            hexChars[j * 2] = hexArray[v >>> 4];
            hexChars[j * 2 + 1] = hexArray[v & 0x0F];
        }
        return new String(hexChars);
    }
```

sbernard31:
I create a repo this will help to share my benchmark with you : https://github.com/sbernard31/benchmark

boaks:
can we go for 

```
        char[] hexChars = new char[bytes.length * 2];
        for ( int src = 0, dest = 0; src < bytes.length; src++ ) {
            int v = bytes[src] & 0xFF;
            hexChars[dest++] = hexArray[v >>> 4];
            hexChars[dest++] = hexArray[v & 0x0F];
        }
        return new String(hexChars);
```
?

sophokles73:
we need to be careful regarding the provenance of the code. Where does it come from? Has it been copied from stackoverflow? If so, do we have consent of the original author to use the code?

boaks:
My proposal using "dest++" is my own creation.

The principal using a char array is a very old c practice. And though the "dest++" changes the stackoverflow version, it should not harm.

boaks:
But, by the way, I could also create a own PR.

sophokles73:
if we now rename the variables slightly, I would be ok with @boaks code :+1:

boaks:
OK, though I'm now the author, I prepare a PR :-).

sbernard31:
👍 
@boaks  could you cherry pick or redo the first commit too.

boaks:
done.

sbernard31:
replace by #382

Issue #168, opened 2016-12-02
Merge StringUtils into one reusable.


boaks:
Currently several StringUtils are used (at least scandium and element connector).
Merge their functionality and reuse that comon one.

boaks:
Fixed with PR #977

PR #382, opened 2017-08-10
Use faster toHexString implementation.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
@@ -13,6 +13,8 @@
  * Contributors:
  *    Matthias Kovatsch - creator and main architect
  *    Stefan Jucker - DTLS implementation
+ *    Achim Kraus (Bosch Software Innovations GmbH) - use faster toHexString
+ *                                                    implementation.
  ******************************************************************************/
 package org.eclipse.californium.scandium.util;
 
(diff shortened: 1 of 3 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
```diff
@@ -143,23 +145,21 @@
...
-				}
+			char[] bytesHexadecimal = new char[byteArray.length * 3];
+			for (int src = 0, dest = 0; src < byteArray.length; src++) {
+				int value = byteArray[src] & 0xFF;
+				bytesHexadecimal[dest++] = BIN_TO_HEX_ARRAY[value >>> 4];
+				bytesHexadecimal[dest++] = BIN_TO_HEX_ARRAY[value & 0x0F];
+				bytesHexadecimal[dest++] = ' ';
 			}
-			return builder.toString();
+			return new String(bytesHexadecimal, 0, bytesHexadecimal.length - 1);
```
sbernard31: Why not just ? new String(bytesHexadecimal)?

boaks: Because to old implementation also excludes the last space.

scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
```diff
@@ -143,23 +145,21 @@
...
-
-				if (i < byteArray.length - 1) {
-					builder.append(' ');
-				}
+			char[] bytesHexadecimal = new char[byteArray.length * 3];
+			for (int src = 0, dest = 0; src < byteArray.length; src++) {
+				int value = byteArray[src] & 0xFF;
+				bytesHexadecimal[dest++] = BIN_TO_HEX_ARRAY[value >>> 4];
+				bytesHexadecimal[dest++] = BIN_TO_HEX_ARRAY[value & 0x0F];
+				bytesHexadecimal[dest++] = ' ';
```
sbernard31: I know that before we print Hex string with ' ' and '--' for null value but am I the only one to find this strange ?

scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
```diff
@@ -143,23 +145,21 @@
...
 	public static String toHexString(byte[] byteArray) {
 
 		if (byteArray != null && byteArray.length != 0) {
-
-			StringBuilder builder = new StringBuilder(byteArray.length * 3);
-			for (int i = 0; i < byteArray.length; i++) {
-				builder.append(String.format("%02X", 0xFF & byteArray[i]));
-
-				if (i < byteArray.length - 1) {
-					builder.append(' ');
```
boaks: "- 1" => exclude last space

scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
sbernard31: sry I missed the space stuff.

scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
sbernard31: As far I as know this is only used for logging and for correlation context.
So we could remove the space and the "--" stuff ?

boaks: If correlation context is persisted, a change would invalidate already persisted data.

So, if it's for logging, why should the "--" be removed?
May be we create two variants, one for logging, one for "compact hex"?

sbernard31: > If correlation context is persisted, a change would invalidate already persisted data.

You're right. I just talk for us but it's not an issue on our side.

> So, if it's for logging, why should the "--" be removed?
May be we create two variants, one for logging, one for "compact hex"?

Personally I will not keep the spaces/-- version but I can live with that.

boaks: Let's shift that discussion into issue #168 but concentrate the next weeks on other issues.

So, would it be OK to merge?

PR #383, opened 2017-08-14
Cleanup notify exchange when replaced by newer notify.

boaks:
Complete old notification exchange.

We may also rename `Exchange.completeCurrentRequest()`, but this would trigger some other files to change. So, if we have a conclusion for a better name, I would rename then once.


boaks:
Fixes

https://hudson.eclipse.org/californium/job/1.0.x-nightly/112/

If no objection, I will go to merge this soon.

sbernard31:
LGTM

PR #384, opened 2017-08-14
remove deduplication of ACK messages

macrosak:
as it is not relevant according to the [CoAP spec](https://tools.ietf.org/html/rfc7252#section-4.5)

This change is already implemented in Californium 2: https://github.com/eclipse/californium/commit/45897ef294a7874d0fcc5b9cff7f957c6ca18141#diff-942dc0e01246d1b1fb4a246c3a09a888R282


boaks:
Thanks!

PR #385, opened 2017-08-17
Faster Cookie generation.

sbernard31:
Using profiler, I see that cookie generation is in hotspot.

I propose a faster implementation.

My result  with 40 concurrent thread : 
```
java -jar target/benchmark.jar -wi 5 -i15 -f1 -t40  Gene
Benchmark                                             Mode  Cnt        Score        Error  Units
GenerateCookieBenchmark.testEnhancedGenerateCookie   thrpt   15   463794,593 ±  90341,908  ops/s
GenerateCookieBenchmark.testEnhancedGenerateCookie2  thrpt   15  1294150,099 ± 458504,830  ops/s
GenerateCookieBenchmark.testOriginalGenerateCookie   thrpt   15   184358,824 ±  72896,625  ops/s
```

My result with 1 thread : 
```
java -jar target/benchmark.jar -wi 5 -i15 -f1 -t1  Gene 
Benchmark                                             Mode  Cnt       Score       Error  Units
GenerateCookieBenchmark.testEnhancedGenerateCookie   thrpt   15  430055,289 ±  6659,283  ops/s
GenerateCookieBenchmark.testEnhancedGenerateCookie2  thrpt   15  416250,419 ± 13385,453  ops/s
GenerateCookieBenchmark.testOriginalGenerateCookie   thrpt   15  351234,423 ±  5973,429  ops/s

```
(Benchmark are available at https://github.com/sbernard31/benchmark, I choose the EnhancedGenerateCookie2 implementation)

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
@@ -1,0 +1,99 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Institute for Pervasive Computing, ETH Zurich and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ */
+package org.eclipse.californium.scandium;
+
+import java.security.GeneralSecurityException;
+import java.security.SecureRandom;
+import java.util.concurrent.TimeUnit;
+
+import javax.crypto.Mac;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+
+import org.eclipse.californium.scandium.dtls.ClientHello;
+import org.eclipse.californium.scandium.dtls.CompressionMethod;
+import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
+
+/**
+ * Generates a cookie in such a way that they can be verified without retaining
+ * any per-client state on the server.
+ *
+ * <pre>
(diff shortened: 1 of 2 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
```diff
@@ -0,0 +1,125 @@
...
+ * <a href="http://tools.ietf.org/html/rfc6347#section-4.2.1">here</a>.
+ *
+ */
+public class CookieGenerator {
+
+	private static final long keyLifetime = TimeUnit.MINUTES.toMillis(5);
+
+	private long lastGenerationDate;
+	private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+	private Mac hmac;
```
jvermillard: you tried with an atomic reference vs read/write locks?

sbernard31: I'm not sure to see how I can use it as I must change `hmac` and `lastGenerationDate` in an atomic way.

scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
```diff
@@ -39,31 +41,56 @@
...
+		lock.writeLock().lock();
+		try {
+			// Recheck state because another thread might have acquired
+			// write lock and changed state before we did.
+			if (hmac == null) {
+				hmac = Mac.getInstance("HmacSHA256");
+				hmac.init(generateSecretKey());
+			}
+			if (isKeyExpired()) {
+				hmac.init(generateSecretKey());
```
boaks: consider "else if" (or check, if one `hmac.init` could not be left out).

sbernard31: done :)

boaks:
Using the dtls examples with

`dtlsConnector.forceResumeAllSessions();`

before each follow-up message, 1500 client, and 20 message per client, results in

before:
30000 messages in 109732 ms
after:
30000 messages in 103364 ms

So, if I didn't make a mistake,  that hotspot seems in the end to be not that fat :-) 

But anyway, it's an improvement.

PR #386, opened 2017-08-17
Use ReadWriteLock instead of sync block in InMemoryConnectionStore

sbernard31:


boaks:
Is the intention to clean-up synchronization or to speed up?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
```diff
@@ -147,105 +149,148 @@ public void onEviction(Connection staleConnection) {
...
 	@Override
-	public synchronized boolean put(final Connection connection) {
-
-		if (connection != null) {
-			return connections.put(connection.getPeerAddress(), connection);
-		} else {
-			return false;
+	public boolean put(final Connection connection) {
+		try {
+			conLock.writeLock().lock();
```
boaks: smaller scope of lock possible?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
```diff
@@ -147,105 +149,148 @@ public void onEviction(Connection staleConnection) {
...
+		
 	}
 
 	@Override
-	public synchronized Connection get(final InetSocketAddress peerAddress) {
-		return connections.get(peerAddress);
+	public Connection get(final InetSocketAddress peerAddress) {
+		try {
+			conLock.readLock().lock();
+			return connections.get(peerAddress);
```
boaks: Unfortunately, `LeastRecentlyUsedCache.get(K key)` seems not to be clearly "read only".

```
		if (key == null) {
			return null;
		}
		CacheEntry<K, V> entry = cache.get(key);
		if (entry == null) {
			return null;
		} else if (entry.isStale(expirationThreshold)) {
			cache.remove(entry.getKey());
			entry.remove();
			return null;
		} else {
			entry.recordAccess(header);
			return entry.getValue();
		}

```

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
```diff
@@ -147,105 +149,148 @@ public void onEviction(Connection staleConnection) {
...
 	}
 
-	private synchronized void removeSessionFromCache(final Connection connection) {
-		if (sessionCache != null && connection.hasEstablishedSession()) {
-			sessionCache.remove(connection.getEstablishedSession().getSessionIdentifier());
+	private void removeSessionFromCache(final Connection connection) {
+		if (sessionCache != null) {
+			try {
+				conLock.writeLock().lock();
+				if (connection.hasEstablishedSession()) {
```
boaks: smaller scope of lock possible?

sbernard31:
The idea was to avoid potential bottleneck mainly on "get" ... but you're right the get is not so read only ...
This means we should use a writelock and in this case this PR is not so useful. I close it.

(You're surely think that what I'm doing now is not so urgent and not so necessary, I'm aware that those enhancements will not change radically the performance, but when I find suspicious point I need to code and test it to confirm. And once done if it seems to me an enhancement I propose the PR,  even if it's not as faster as expected.)

Thx @boaks

PR #387, opened 2017-08-18
Do not wait client sequentially in ExampleDTLSClient

sbernard31:
When I use ExampleDTLSClient with more than 300 client, it does not finished.
 I didn't find why, like if I face a packet lost even if  I'm on localhost.

Anyway I changed the code a little to make it stop on timeout. (not 1 timeout by client)

### DIFF
--- a/demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
+++ b/demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
@@ -29,13 +29,15 @@ import java.security.cert.Certificate;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.eclipse.californium.elements.RawData;
 import org.eclipse.californium.elements.RawDataChannel;
+import org.eclipse.californium.elements.util.DaemonThreadFactory;
 import org.eclipse.californium.scandium.DTLSConnector;
 import org.eclipse.californium.scandium.ScandiumLogger;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
@@ -45,20 +47,21 @@ public class ExampleDTLSClient {
 
 	static {
 		ScandiumLogger.initialize();
-		ScandiumLogger.setLevel(Level.FINE);
+		ScandiumLogger.setLevel(Level.WARNING);
 	}
 
 	private static final int DEFAULT_PORT = 5684;
+	private static final long DEFAULT_TIMEOUT_NANOS = TimeUnit.MILLISECONDS.toNanos(10000);
 	private static final Logger LOG = Logger.getLogger(ExampleDTLSClient.class.getName());
 	private static final String TRUST_STORE_PASSWORD = "rootPass";
 	private static final String KEY_STORE_PASSWORD = "endPass";
 	private static final String KEY_STORE_LOCATION = "certs/keyStore.jks";
 	private static final String TRUST_STORE_LOCATION = "certs/trustStore.jks";
-	private static final int NB_MESSAGE = 1000;
 
-	private static CountDownLatch latch;
+	private static int maxMessages = 0;
+	private static CountDownLatch messageCounter;
+
 	private DTLSConnector dtlsConnector;
-	private final AtomicLong count = new AtomicLong();
 
 	public ExampleDTLSClient() {
 		InputStream inTrust = null;
@@ -85,20 +88,14 @@ public class ExampleDTLSClient {
 					keyStore.getCertificateChain("client"), true);
 			builder.setTrustStore(trustedCertificates);
 			builder.setEnableAddressReuse(false);
-			builder.setConnectionThreadCount(2);
+			builder.setConnectionThreadCount(1);
 			dtlsConnector = new DTLSConnector(builder.build());
 			dtlsConnector.setRawDataReceiver(new RawDataChannel() {
 
 				@Override
 				public void receiveData(RawData raw) {
-					long c = count.incrementAndGet();
-					LOG.log(Level.INFO, "Received response: {0} {1}", new Object[] { new String(raw.getBytes()), c });
-					if (c < NB_MESSAGE) {
-						dtlsConnector
-								.send(new RawData(("HELLO WORLD " + c + ".").getBytes(), raw.getInetSocketAddress()));
-					} else {
-						latch.countDown();
-						dtlsConnector.destroy();
+					if (dtlsConnector.isRunning()) {
+						receive(raw);
 					}
 				}
 			});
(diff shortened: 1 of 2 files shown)


demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
```diff
@@ -118,55 +119,72 @@ public void receiveData(RawData raw) {
...
+
+		// Start Test
+		long nanos = System.nanoTime();
 		for (ExampleDTLSClient client : clients) {
-			count += client.waitReady(100000, TimeUnit.SECONDS);
+			client.startTest(peer);
 		}
 
+		// Wait for it
+		latch.await(max * 50, TimeUnit.MILLISECONDS);
```
boaks: It seems, that you just didn't wait long enough for the "100000 seconds" timeout :-).

sbernard31: My first try :), reduce this timeout to 10 seconds but this does not works because I should wait that 10s  for each client which does not finished normally

boaks: Ah, but 50ms per client for 1000 messages, does this work?

sbernard31: You mean set 50ms for waitReady() ?

It works in a way the test finished in a reasonable time.
But this is a bit strange because first client has only 50ms but the second one can have 50ms+50ms (if the first one never finished) an so on.

And I think this is more convenient to set a deterministic timeout, I mean "I want to wait maximum for "n" minutes for this run"

By the way I finally think my smart calculation (max*50) for the timeout is pretty stupid :)
I would go for a constant instead if you're ok, something like 2 minutes ? (maybe we should add it as command line parameter)

demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
```diff
@@ -118,55 +119,72 @@ public void receiveData(RawData raw) {
...
 		nanos = System.nanoTime() - nanos;
+
+		// Count messages
+		long count = 0;
+		for (ExampleDTLSClient client : clients) {
+			count += client.nbMessageReceived();
+		}
+
+		System.out.println(count + " messages received, " + (max * NB_MESSAGE) + " expected");
```
boaks: Can we move the statistic below the stops? with logging enabled it's hard to fin the result.

sbernard31: I will do that.

demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
boaks: If some clients get stale because of dropped messages, the measured times are more or less useless, if we wait for them too long. So currently, the the short times seems to have more  advantages than disadvantages.

sbernard31: Does it means that you are ok for a constant for the timeout ? like 2 minutes ?

boaks: No, I'm OK with 50ms per client

sbernard31: Ok but on my side with 1 client 50ms is not enough ^^.
I need something like ~300 at ~600 ms for 1 client.
and something like ~8/9s for 200 clients.
That's why I find my smart calculation not so smart :p.

boaks:
@sbernard31 

I appended my idea of making the test robust against lost messages. Generally all clients use the same counter and update a timestamp. The main waits either for the counter or the timeout of the message timestamp. The behaviour would be, that even if some messages are lost, the other client would continue to sent messages and so the statistic should be OK.

demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
```diff
@@ -45,19 +47,21 @@
 
 	static {
 		ScandiumLogger.initialize();
-		ScandiumLogger.setLevel(Level.FINE);
+		ScandiumLogger.setLevel(Level.WARNING);
```
sbernard31: Good point. I take my time to set it to OFF for load testing.

demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
```diff
@@ -136,55 +148,87 @@ private void startTest(InetSocketAddress peer) {
...
 		}
-		latch = new CountDownLatch(max);
-		List<ExampleDTLSClient> clients = new ArrayList<>(max);
+		maxMessages = (messages * clients);
+		messageCounter = new CountDownLatch(maxMessages);
+		List<ExampleDTLSClient> clientList = new ArrayList<>(clients);
+		ExecutorService executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors(),
+				new DaemonThreadFactory("Aux#"));
+
+		System.out.println("Create " + clients + " DTLS example clients.");
```
sbernard31: Can we add the number of message expected by clients in this log ?

demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
```diff
@@ -136,55 +148,87 @@ private void startTest(InetSocketAddress peer) {
...
+		while (!messageCounter.await(DEFAULT_TIMEOUT_NANOS, TimeUnit.NANOSECONDS)) {
+			long current = messageCounter.getCount();
+			if (lastMessageCountDown == current && current < maxMessages) {
+				// no new messages, clients are stale
+				// adjust start time with timeout
+				nanos += DEFAULT_TIMEOUT_NANOS; 
+				break;
+			}
+			lastMessageCountDown = current;
+		}
```
sbernard31: It's clearly better way to guess the end of the test.
About statistics, it's better too, but in my tests when I have message lost I still have lower number of message by second.
I suspect that even with this attempt to fix statistic this is no so accurate. (I suppose this is because some clients can stop really soon after start and so less client in parallel means less messages exchanged)

sbernard31:
Not directly linked but why using 2 connection handler thread ?
https://github.com/eclipse/californium/pull/387/files#diff-59d933dae87c64063ebb98932bc6bf93L87

We can use only 1 as we handle only one conneciton by client ? (less threads used means possibility to create more clients because of limitation of native thread)

demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
boaks: OK, may be we add also a statistic about the number of messages each client process to see, how this is balanced. But I would go for an additional PR.

demo-apps/sc-dtls-example-client/src/main/java/org/eclipse/californium/scandium/examples/ExampleDTLSClient.java
boaks: Sure

boaks:
Yes, we can. I didn't want to add too many parameter and I change this sometimes to compare the result.

boaks:
So, still good to merge?

sbernard31:
Yep :)

PR #388, opened 2017-08-22
Use only one byte[] buffer for receiving message.

sbernard31:
Still playing with profiler.

Profiling ExampleDTLSServer againt ExampleDTLSClient (with 1000 clients/1000 message).
I see that DTLS-receiver Thread allocate 8,5 GB.
With this PR it goes down to 0,5GB.

In practice, I don't notice any performance improvement.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -180,6 +180,8 @@ public class DTLSConnector implements Connector {
 
 	/** The thread that receives messages */
 	private Worker receiver;
+	private byte[] receiverBuffer;
+	
 
 	/** The thread that sends messages */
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -491,8 +494,8 @@ public final synchronized void destroy() {
 
 	private void receiveNextDatagramFromNetwork() throws IOException {
```
boaks: I would prefer to limit the access scope of the `receiverBuffer`.
Could we move that field to the `Worker `and pass it in as argument to `receiveNextDatagramFromNetwork`?
May be we can even reuse the `DatagramPacket` and pass that to `receiveNextDatagramFromNetwork`?

sbernard31:
#391 is better :) I close this one.

PR #389, opened 2017-08-23
Use timestamp of add for deduplication.

boaks:
Fixes https://hudson.eclipse.org/californium/job/2.0.x-nightly/162/

Fixes, except the transmission of the duplicate is delayed more than TEST_EXCHANGE_LIFETIME.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/SweepDeduplicator.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/SweepDeduplicator.java
@@ -21,6 +21,7 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - use nanoTime instead of 
  *                                                    currentTimeMillis
  *    Achim Kraus (Bosch Software Innovations GmbH) - fix used milliseconds calculation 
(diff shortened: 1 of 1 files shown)


boaks:

sbernard31:
@boaks, could you explain that a little more. I didn't get the point.

boaks:
If you check the failed build, you will see, that the request was some time ago. So, if using the request time for the deduplication timeout, the deduplication doesn't apply because the second response is not longer in time. If the time of the reponse (separate CON response) is used, then the second response is in time and will be deduplicated.

sbernard31:
OK I get it.
So the `Exchange.getNanoTimestamp()` getter is not really usefull anymore ?
And the nanoTimestamp attributes seems only needed for RTT calculation which seems not really usefull.

Do you think we could remove that ?

boaks:
It will not save too much, so I wouldn't remove it now.

boaks:
FMPOV `RTT` was part of the API. SO I don't see, why it should be removed,

sbernard31:
I'm not sure to see the utility of RTT but why not.
However  for  `Exchange.getNanoTimestamp()`, it's pretty harder to see it as part of the API. It seems this was just exposed for sweep deduplicator.

PR #391, opened 2017-08-23
Reuse receive buffer and packet.

boaks:
Alternative for PR #388 


boaks:

boaks:
@sbernard31 

If you prefer your PR #388, we can still discuss that.

PR #394, opened 2017-08-23
multicast request: remove exchange only by timeout

josephguinet:
Fix issue #82 for the 1.0.x branch
Adapt the #95 pull request or KMax from master into the 1.0.x branch

Signed-off-by: joseph <joseph.guinet@ubiant.com>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
@@ -81,6 +81,7 @@ public class Matcher {
 	/* Health status output */
 	private final Level healthStatusLevel;
 	private final int healthStatusInterval; // seconds
+	private final int multicastMaxLifetime; // ms
 	
 	public Matcher(NetworkConfig config) {
 		this.started = false;
@@ -104,6 +105,9 @@ public class Matcher {
 		
 		healthStatusLevel = Level.parse(config.getString(NetworkConfig.Keys.HEALTH_STATUS_PRINT_LEVEL));
 		healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL);
+		multicastMaxLifetime = config.getInt(NetworkConfig.Keys.NON_LIFETIME) 
+				+ config.getInt(NetworkConfig.Keys.MAX_LATENCY) 
+				+ config.getInt(NetworkConfig.Keys.MAX_SERVER_RESPONSE_DELAY);
 	}
 	
 	public synchronized void start() {
@@ -139,7 +143,7 @@ public class Matcher {
 		// health status runnable is not migrated at the moment
 	}
 	
-	public void sendRequest(Exchange exchange, Request request) {
+	public void sendRequest(final Exchange exchange, Request request) {
 		
 		// ensure MID is set
 		if (request.getMID() == Message.NONE) {
@@ -173,6 +177,19 @@ public class Matcher {
 		
 		exchangesByMID.put(idByMID, exchange);
 		exchangesByToken.put(idByToken, exchange);
+		
+		if(request.isMulticast()) {
+			executor.schedule(new Runnable() {
+				@Override
+				public void run() {
+						if(exchange.getResponse() == null)
+						{
+						    exchange.getCurrentRequest().setCanceled(true);
+						}
+						exchange.setComplete();
+					}
+				}, multicastMaxLifetime, TimeUnit.MILLISECONDS);
+		}
 	}
 
 	public void sendResponse(Exchange exchange, Response response) {
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
@@ -69,6 +69,8 @@ public class NetworkConfig {
 		public static final String EXCHANGE_LIFETIME = "EXCHANGE_LIFETIME";
 		public static final String NON_LIFETIME = "NON_LIFETIME";
 		public static final String MAX_TRANSMIT_WAIT = "MAX_TRANSMIT_WAIT";
+		public static final String MAX_LATENCY = "MAX_LATENCY";
+		public static final String MAX_SERVER_RESPONSE_DELAY = "MAX_SERVER_RESPONSE_DELAY";
 		public static final String NSTART = "NSTART";
 		public static final String LEISURE = "LEISURE";
 		public static final String PROBING_RATE = "PROBING_RATE";
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
@@ -47,6 +47,8 @@ public class NetworkConfigDefaults {
 		config.setLong(NetworkConfig.Keys.EXCHANGE_LIFETIME, 247 * 1000); // ms
 		config.setLong(NetworkConfig.Keys.NON_LIFETIME, 145 * 1000); // ms
 		config.setLong(NetworkConfig.Keys.MAX_TRANSMIT_WAIT, 93 * 1000);
+		config.setInt(NetworkConfig.Keys.MAX_LATENCY, 100 * 1000); //ms
+		config.setInt(NetworkConfig.Keys.MAX_SERVER_RESPONSE_DELAY, 250 * 1000); //ms
 		config.setInt(NetworkConfig.Keys.NSTART, 1);
 		config.setInt(NetworkConfig.Keys.LEISURE, 5000);
 		config.setFloat(NetworkConfig.Keys.PROBING_RATE, 1f);
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapStack.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapStack.java
@@ -187,7 +187,7 @@ public class CoapStack {
 
 		@Override
 		public void receiveResponse(Exchange exchange, Response response) {
-			if (!response.getOptions().hasObserve())
+			if (!response.getOptions().hasObserve() && !exchange.getCurrentRequest().isMulticast())
 				exchange.setComplete();
 			if (deliverer != null) {
 				deliverer.deliverResponse(exchange, response); // notify request that response has arrived


boaks:
I have very strange feeling about the compliance of the current implementation to RFC7252, section 8, Multicast CoAP.

So, can you please explain:
- what is your use case for multicast messaging?
- what else do you use/require to test that multicast messaging?

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
```diff
@@ -69,6 +69,8 @@
 		public static final String EXCHANGE_LIFETIME = "EXCHANGE_LIFETIME";
 		public static final String NON_LIFETIME = "NON_LIFETIME";
 		public static final String MAX_TRANSMIT_WAIT = "MAX_TRANSMIT_WAIT";
+		public static final String MAX_LATENCY = "MAX_LATENCY";
+		public static final String MAX_SERVER_RESPONSE_DELAY = "MAX_SERVER_RESPONSE_DELAY";
```
boaks: Is this the "PROCESSING_DELAY" according RFC 7252, 4.8.2, page 29?

If so, I would propose to use the term of the RFC.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -173,6 +177,16 @@ public void sendRequest(Exchange exchange, Request request) {
 		
 		exchangesByMID.put(idByMID, exchange);
 		exchangesByToken.put(idByToken, exchange);
+		
+		if(request.isMulticast()) {
+			executor.schedule(new Runnable() {
+				@Override
+				public void run() {
+					exchange.getCurrentRequest().setTimedOut(true);
+					exchange.setTimedOut();
```
boaks: Timeout seems to be related to CON messages. 
According RFC 7252, 8.1, page 65, all multicast messages MUST be NON so a timeout would not apply.
Try cancel()?

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -104,6 +105,9 @@ public Matcher(NetworkConfig config) {
 		
 		healthStatusLevel = Level.parse(config.getString(NetworkConfig.Keys.HEALTH_STATUS_PRINT_LEVEL));
 		healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL);
+		multicastMaxLifetime = config.getInt(NetworkConfig.Keys.NON_LIFETIME) 
+				+ config.getInt(NetworkConfig.Keys.MAX_LATENCY) 
+				+ config.getInt(NetworkConfig.Keys.MAX_SERVER_RESPONSE_DELAY);
 	}
```
boaks: According RFC7252, 4.8.2, page 30, the `NON_LIFETIME` seems to already include the `MAX_LATENCY`.
RFC7252, 8.2, page 66, the multicastMaxLifetime considers also the `LEISURE` time.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -353,18 +367,24 @@ public Exchange receiveResponse(Response response) {
...
+			    // keep exchange open because we might receive additional responses
+			    LOGGER.log(Level.FINER, "Received response for multicast request: {0}", response);
+			}
+			else {
+				// we have received a Response matching the token of an ongoing Exchange's Request
+				// according to the CoAP spec (https://tools.ietf.org/html/rfc7252#section-4.5),
+				// message deduplication is relevant for CON and NON messages only
+				if ((response.getType() == Type.CON || response.getType() == Type.NON) &&
+						deduplicator.findPrevious(idByMID, exchange) != null) {
+					LOGGER.info("Duplicate response for open exchange: "+response);
```
boaks: The `idByMID `for CON or NON are create in the remote namespace, they include the senders address.
So if this `idByMID ` is used for deduplication, this shouldn't be in conflict with the multicast.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -353,18 +367,24 @@ public Exchange receiveResponse(Response response) {
...
+					response.setDuplicate(true);
+				} else {
+					idByMID = new KeyMID(exchange.getCurrentRequest().getMID(), null, 0);
+					exchangesByMID.remove(idByMID, exchange);
```
boaks: As far as I know, exchangesByMID are only relevant for ACK or RST. 
ACK are not used in multicast and using RST would cause RFC7252, 8.1, page 65:

> the sender MUST avoid using a
> Message ID that is also still active from this endpoint with any
> unicast endpoint that might receive the multicast message.

which would be hard to achieve when consider a server with more message load.
The 1.0 branch is using central "MID counter" for all clients, which limits the overall traffic, but grants the above.
The 2.0 branch therefore use a "per connection MID counter", so there we would need the ensure that explicitly.
(FMPOV multicast seems to be a more rare use-case then to sending more messages).

So do you use RST? If not, why do you need to change the response processing? Which issues did you observe, if the responses are processed usual?

boaks:
FMPOV,  the changes seems to cover an **other** "multicast" use case, but **not** RFC7252, 8.
For multicast according RFC7252, 8, mainly  the "setComplete()" MUST NOT be called.

If the other requested changes (processing response deduplication) are based on observed misbehaviour, please report that.

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
josephguinet: No these constants refer to RFC 7390, 2.5, page 9.
I reuse the same term and use them as suggested by sophokles73 in [https://github.com/eclipse/californium/pull/95/commits/08228a7ea7743af7ba5f91046a37b94ebd2e7e35](url)

> Instead of using the reception of a response as the trigger to finally remove the exchange from the token map, I think we should instead schedule a task for cleaning up the multicast exchange after
NON_LIFETIME + MAX_LATENCY + MAX_SERVER_RESPONSE_DELAY
as indicated by RFC 7390. This clean-up task would need to be scheduled when the original multicast request is sent out.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
josephguinet: Yes, I will try it.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
josephguinet: According RFC 7390, 2.5, page 10, the time between reuse of Token values
   used in multicast requests MUST be greater than: `NON_LIFETIME + MAX_LATENCY + MAX_SERVER_RESPONSE_DELAY`.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
josephguinet: You are right, this part isn't in conflict with the multicast.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
josephguinet: No I didn't use the RST and you are right again : with the filter in receiveResponse of CoapStack.java I didn't need to filter the multicast response again in receiveResponse of Matcher.java.

josephguinet:
My use case is some unidentified devices in an IPv6 network and I need to discover all of them. 
In order to do this : 
- All those devices subscribe to a specific address (FF03::ABBE:55E5 in my case). 
- I send a multi-cast GET request to this one.
- All the device answer in uni-cast.

In my first test, the first response was successfully proceed but for all other responses I received `org.eclipse.californium.core.network.Matcher receiveResponse INFO: Ignoring unmatchable piggy-backed response`. 
In deed, after the first response, the exchange is closed and for the next responses the matcher is unable to retrieve the corresponding request.

After some research, I found the #95 fix for my issue. But KMax wrote it for the master branch so I decide to adapt it to the 1.0.x branch while including sophokles73 's observations.

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
boaks: Reading that RFC7390 it confuses me. I don't know, why the calculation there include the MAX_LATENCY twice (explicit and implicit with NON_LIFETIME).

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
boaks: I would go for cancel and complete.

boaks:
You setup you own "multicast" connector?
If It's possible, could you share that setup? It would make it easier/faster for others (including me :-) ).

If It's not easy to share it, could you please check, if it still works for you, when removing the stuff from "receiveResponse"?

For me the RFC7390 seems to mix up the "transmission parameters" but in the end, it's just a "rough estimation" for the last response. If it's important for you to take the definition from there, I would prefer to have a reference to that definition in the javadoc of the `multicastMaxLifetime`. If the RFC7252 timing definitions would do it also, I would prefer them.

According RFC7252 the responses should use some random time to response. Do you require that in your setup?

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
boaks: More advanced:
check the exchange, if it has already at least one response. If so, just complete the exchange. If not cancel the request and complete the exchange. I think, this would be the best approach to hande the multiple responses.

boaks:
@josephguinet 

Are you still interested in this PR?

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
josephguinet: I implemented your two last recommendation and it works fine for me.

boaks:
Sounds good.
Can you push your "final" changes to this PR?
You may use a "forced push" on your branch josephguinet:multicast-handling

josephguinet:
I didn't setup my own "multicat" connector. 
In my case, I send my request to a specific IP address (coap://[FF03::ABBE:55E5]) which was previously subscribe by many devices (they used  their own implementation of coap written in C by NXP) and all of them send me back a response with their own IP address.

And yes I will push my last changes to this PR.

boaks:
Thanks in advance :-)

josephguinet:
I rebased the last modification of 1.0.x into my branch and I squashed all my commits into one and used a "forced push" to update my branch. Is it okay for you ?

californium-core/src/main/java/org/eclipse/californium/core/network/Matcher.java
```diff
@@ -173,6 +177,23 @@ public void sendRequest(Exchange exchange, Request request) {
...
+							exchange.setComplete();
+						}
+						else
+						{
+							exchange.getCurrentRequest().setCanceled(true);
+							exchange.setComplete();
+						}
+					}
+				}, multicastMaxLifetime, TimeUnit.MILLISECONDS);
+		}
```
boaks: > More advanced:
> check the exchange, if it has already at least one response. If so, just complete the exchange. If not cancel the request and complete the exchange. I think, this would be the best approach to handle the multiple responses.

I think, the if branches are exchanged according to the recommendation. 
`exchange.getResponse() == null` is the case with no response, where the request was intended to be canceled before complete the exchange.

```
if(exchange.getResponse() == null)
{
    exchange.getCurrentRequest().setCanceled(true);
}
exchange.setComplete();

```

But if it runs "for many days on many different devices", we can also use your solution.
So, do you want to adapt, or should I merge your current solution?

josephguinet: I didn't really test the case of no responses to the multicast request. So I made some tests: 
- It appears that I received a timeout notification from the RetransmissionTask of the ReliabilityLayer.
- But after the multicastMaxLifetime I still have some current request which are not canceled.

So I think the best is to implement your last recommendation.

boaks:
Thanks for your work and patience.

josephguinet:
You're welcome.

josephguinet:
And thanks for the merge.

PR #397, opened 2017-09-04
Add TCP and encryption support to plugtest.

boaks:
Prepare cf-plugtest server and client for eclipse sandbox.
Enable encryption support and experimental TCP feature.

### DIFF
--- a/demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
+++ b/demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
@@ -12,17 +12,31 @@
  * 
  * Contributors:
  *    Matthias Kovatsch - creator and main architect
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add TCP and encryption support.
  ******************************************************************************/
 package org.eclipse.californium.plugtests;
 
+import java.io.IOException;
+import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
 import java.net.SocketException;
+import java.security.GeneralSecurityException;
+import java.security.cert.Certificate;
+
+import javax.net.ssl.SSLContext;
 
 import org.eclipse.californium.core.CoapServer;
 import org.eclipse.californium.core.coap.CoAP.Type;
+import org.eclipse.californium.core.network.CoapEndpoint;
 import org.eclipse.californium.core.network.Endpoint;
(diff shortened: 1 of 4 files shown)


boaks:
Seems that I accidently formated the source. If someone thing it's import to go back to the old formating to see the changes easier, I would do it. Otherwise I think the proper format could be kept.
For coap and coaps the sandbox "californium.eclipse.org" is already up and running :-).

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
```diff
@@ -46,92 +60,203 @@
...
+			PlugtestServer server = new PlugtestServer();
+			// ETSI Plugtest environment
+			// server.addEndpoint(new CoAPEndpoint(new InetSocketAddress("::1",
+			// port)));
+			// server.addEndpoint(new CoAPEndpoint(new
+			// InetSocketAddress("127.0.0.1", port)));
+			// server.addEndpoint(new CoAPEndpoint(new
+			// InetSocketAddress("2a01:c911:0:2010::10", port)));
+			// server.addEndpoint(new CoAPEndpoint(new
+			// InetSocketAddress("10.200.1.2", port)));
```
sbernard31: Comment formating is a bit strange here.
Should we keep thoses comments ?

boaks: @mkovatsc 

Do you want to keep it, or could it be removed?

boaks:
> its a bit strange that TCPConnector properties can be set through NetworkConfig as this is up to user to create its connector

It is also strange to have `COAP_SECURE_PORT` and `HTTP_PORT` (and even more `HTTP_???`).
So I don't think, this should be "cleaned up" in this PR :-).

sbernard31:
> So I don't think, this should be "cleaned up" in this PR :-).

Totally agree.

boaks:
If no objection, I would like to merge this.

sbernard31:
👍

boaks:
So, start the milestone?

sbernard31:
yep :)
I do or you do ?

boaks:
I try it,
(and you may sort it out, if it gets broken ...)

boaks:
:-) 

Seems that the HIPP is "very stressed" today. We'll see, if it builds with the 3. try. 
Otherwise I retry it tomorrow.

PR #399, opened 2017-09-05
Move TCPConnector in its own project : element-connector-tcp

### LINKED ISSUE #119: Tls/TcpClientConnector using netty / really required?
Using netty for the server may be OK, but for the client I think it would work also without.
So are there any reasons to use netty on the **clients**?

sbernard31:
The idea is to do the same thing for `TCPconnector` than `DTLSConnector` moving it in its own project.

The main benefit is to make the `netty` dependency optional.

The main impact is in `EndpointManager`. Now it need a `setDefaultEndpoint` call for `coap+tcp` and `coaps+tcp` like for `coaps`.

This should fix #119, mainly https://github.com/eclipse/californium/issues/119#issuecomment-321240015


californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -169,18 +168,17 @@ public synchronized void setDefaultEndpoint(Endpoint newEndpoint) {
...
-					config.getInt(NetworkConfig.Keys.TCP_CONNECT_TIMEOUT),
-					config.getInt(NetworkConfig.Keys.TCP_CONNECTION_IDLE_TIMEOUT));
-			endpoint = new CoapEndpoint(connector, config);
+			LOGGER.config(
+					"Tcp endpoint must be injected via setDefaultEndpoint(Scheme, Endpoint)");
+			return null;
 		} else if (CoAP.COAP_SECURE_TCP_URI_SCHEME.equalsIgnoreCase(uriScheme)) {
-			LOGGER.config("Secure tcp endpoint must be injected via setDefaultEndpoint(Scheme, Endpoint) to use the proper credentials");
+			LOGGER.config(
+					"Secure tcp endpoint must be injected via setDefaultEndpoint(Scheme, Endpoint) to use the proper credentials");
```
boaks: If possible, could we correct the logging messages to `setDefaultEndpoint(Endpoint)` because the Scheme is read from the endpoint (and not longer passed in)? I forgot/overseen that, when I was changing the EndpointManager.

boaks:
Move the TCP stuff **NOW** to a separated package will cause much extra work on the none merge TCP PR #294 and the "tcp_experimental_features" branch. 
I'm therefore not sure, if this is the right timepoint :-). 
Are there any special issues, apart from #119 that triggers this work **NOW**?

sbernard31:
I didn't look so much #294 but the job on this PR was mainly moving files from one project to another.
(not so much modification)
So I'm not so afraid about merging this except if #294 introduce extra dependencies.

Maybe that was smarter to do that after #294 integration, but now the job is done (I mean #294 and #399)

If you want we can integrate #294 first and I will rebase my work after that.
What are missing for #294 ?

boaks:
My intention not to merge #294 (nor ask to) was, that if first want to have a defined CorrelationContext (see #344). #294 also modifies the CorrelationContext and therefore I wanted the CorrelationContext  to be "defined" before.

sbernard31:
Ok so I understand that we should integrate this PR before #294 because #294 will not be integrated soon ?
About #344, I give up to push the idea about EndpointIdentifier, so it is in your hand now. If you want to "defined" correlationContext before,  just do a summarize of your idea in #344 and I will give you my opinion.

boaks:
@sbernard31 
Though my TLS stuff is merged, I would go to "rebase" this PR. OK?

sbernard31:
OK (if too many conflicts maybe simpler to redo from scratch)

boaks:
> simpler to redo from scratch

 therefore "rebase" :-)

boaks:
See #469

sbernard31:
replaced by #469

PR #400, opened 2017-09-06
Add isEmptyWithDump() for more details in logs while testing. (2)

sbernard31:
I try to rewrite #326 without adding the `isEmptyWithDump()` method to `MessageExchangeStore`.

The idea is to make the `InMemoryMessageExchangeStore` more monitorable by exposing internal structure.
This way tests could use it for logging but  we can also imagine users could monitor it in production. E.g. by pushing periodically the number of exchanges in something like grafana.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
@@ -124,6 +124,7 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 		StringBuilder b = new StringBuilder("MessageExchangeStore contents: ");
 		b.append(exchangesByMID.size()).append(" exchanges by MID, ");
 		b.append(exchangesByToken.size()).append(" exchanges by token, ");
+		b.append(deduplicator.size()).append(" MIDs, ");
 		return b.toString();
 	}
 
@@ -163,8 +164,12 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 
 	@Override
 	public boolean isEmpty() {
-		LOGGER.finer(dumpCurrentLoadLevels());
 		return exchangesByMID.isEmpty() && exchangesByToken.isEmpty() && deduplicator.isEmpty();
+	}
+
+	@Override
+	public String toString() {
+		return dumpCurrentLoadLevels();
 	}
 
 	@Override
@@ -374,5 +379,17 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 	@Override
 	public void releaseToken(KeyToken keyToken) {
 		tokenProvider.releaseToken(keyToken);
+	}
+
+	public ConcurrentMap<KeyToken, Exchange> getExchangesByToken() {
+		return exchangesByToken;
+	}
+
+	public ConcurrentMap<KeyMID, Exchange> getExchangesByMID() {
+		return exchangesByMID;
+	}
+
+	public Deduplicator getDeduplicator() {
+		return deduplicator;
 	}
 }
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/CropRotation.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/CropRotation.java
@@ -18,6 +18,7 @@
  *    Kai Hudalla - logging
  *    Achim Kraus (Bosch Software Innovations GmbH) - make first and second
  *                                                    volatile
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add size() for test-logging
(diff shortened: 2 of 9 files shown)


boaks:
FMPOV I would prefer a documented `isEmptyWithDump()` instead of `exposing internal structure`.

I guess, also a monitoring would benefit more from an API extension then from `exposing internal structure`, because if those `internal structure` are adapted, the scope of the adaptation would get large.

So what about extending `MessageExchangeStore` with an 

`	MessageExchangeStoreInternal internal();
`

so the "internal" stuff is not accessed accidentally and add the `isEmptyWithDump()` to `MessageExchangeStoreInternal`?

sbernard31:
I will try to explain my point of view : 

The real API is `MessageExchangeStore`, the goal of this API is to let user implements its own ExchangeStore. This API should contains only what is mandatory for californium. All methods not called by `californium.core` should not be part of this API. (This means FMOV cleaning should be done)

`InMemoryExchangeStore` is just the default implementation, here we can be less strict and expose more methods. Beginner users will never face this class. Only advanced users should play with it. Advanced users want to make advanced stuff like getting access to internal structure to monitor it without the need to duplicate the class. 

So FMPOV this is not a problem to expose internal stuff and I would prefer to let it like that. If you really think adding a new class for that is necessary, I will add an internal class in `InMemoryExchangeStore` and move the getter in it. I prefer to not push in it the `isEmptyWithDump()` method (except you insist)

In general way, we should take care about what we put in `core`, the more we increase the code base, the more this is costly to maintain.

boaks:
So wait some time to get the opinion from others.

I'm not insist, that `isEmptyWithDump()` is added, but if `InMemoryMessageExchangeStore` is **just** a default implementation, it seem for me to be more object oriented to add it at least there, then the expose of internal structures. 

But we will see, if we get other opinions...

sophokles73:
My understanding of @boaks original PR is that he wanted to get some more info about the store's contents during test execution, i.e. when a test fails.
This PR, however, seems to be targeting at better supporting the (continuous) monitoring of `InMemoryMessageEchangeStore` during runtime. Both goals are worth pursuing FMPOV.
I have already given my opinion regarding the first goal in the original PR (#326).
Regarding the monitoring of the store, I would be more in favor of actively reporting any changes to the store by means of a metrics framework which abstracts away the implementation of counters etc. and how external systems can access their values. At least this seems to be the approach used by most systems today.

sbernard31:
> Regarding the monitoring of the store, I would be more in favor of actively reporting any changes

You mean by adding kind of listeners ? This is more works/code, adding getter is good first step.
JFYI on our side getter is even more convenient than listener as we have an agent which poll each minute for this kind of monitoring.

sophokles73:
> This is more works/code, adding getter is good first step.
JFYI on our side getter is even more convenient than listener as we have an agent which poll each minute for this kind of monitoring.

Actually, it is not really. Have you taken a look at e.g. http://metrics.dropwizard.io ? Once you add such a framework to the picture, you have many options getting at the metrics from the outside like using polling, active reporting, JMX etc

sbernard31:
I do not understand. Please can you explain by what you want to replace the `getter`.

sophokles73:
If you use a framework like dropwizard, then your code always (actively) reports metrics values via the API provided by dropwizard. It provides counters and stuff which you can increment, decrement etc. The way these metrics are retrieved from your component is abstracted away by dropwizard, i.e .dropwizard provides several plugins that can be used to e.g. periodically report all metrics to an influxDB or whatever or it can also expose the metrics by means of JMX or you can poll the metrics using HTTP GET calls etc. The point is that you do not need to implement any of that yourself, you just always use the dropwizard API to report the relevant values ...

sbernard31:
Thx @sophokles73. I understand what is the idea of framework like dropwizard. That was not my question ^^.

You say : 
> Regarding the monitoring of the store, I would be more in favor of actively reporting any changes to the store

My understanding was you think `getter` is not adapted to plug to this kind of framework. You seems to propose "something else" ?
I believed that this "something else" was adding `listener/event` on add/remove.

My point,  `listener` is more code/works than just add `getter` and it seems enough to plug to this kind of framework.

I think it's also ok with dropwizard.
```java
registry.register(name(ExchangeStore.class, "exchange-store"), new Gauge<Integer>() {
    @Override
    public Integer getValue() {
        return store.getExchangesByToken().getSize();
    }
});
```

sbernard31:
What should we do with that ?

sophokles73:
I agree with @sbernard31 in that IMHO we do not need to introduce a `MessageExchangeStore.isEmptyWithDump()` method. I think the approach taken in this PR is totally acceptable. Once the conflicts have been removed, we should merge it.

sbernard31:
I will try to go on this tomorrow.

sbernard31:
Replaced by #504. (which is now integrated in 2.0.x)

Issue #174, opened 2016-12-07
Make request/response correlation configurable


sbernard31:
As I explain [here][1], I think that correlate request/response by session/epoch is really too strong I recommend to just made the correlation by Identity. Having a way to do that in Californium by offering a way to implement your own correlator or by adding a new correlation mode could be a good point.

[1]:https://mailarchive.ietf.org/arch/search/?email_list=core&gbt=1&index=0vwuEnq6IRHJc70l3BTggX6p9vs

boaks:
@sbernard31 

Could you check, if the introduced `CorrelationContextMatcher` solves this issue for you?

sbernard31:
This solves partially the issue. We should now add principal to correlation context.

boaks:
A did some investigation also on my idea in #344 to move the principal and correlation context into 
`org.eclipse.californium.core.coap.Message` to make it possible to pass that information also in.  
Doing so I recognized, that the matcher could not use Message on element-connector, so I withdrawn that and extended the CorrelationContext with the Principal and InetSocketAddress and renamed it to EndpointContext. Together with an cleanup (remove the InetAddress/Port from Message and RawData), this is also a flood of changes, as some other pending stuff or PRs.

So, I think, we should first agree on the order of the work:
- separate TCP
- change logging (I'm not sure, if we should do this at all)
- introduce the "EndpointContext"

sbernard31:
It sounds a good plan.

**Just to complete it :**
- integrate #397
- release 2.0.0M5
- separate TCP
- change logging (I'm not sure, if we should do this at all)
- introduce the "EndpointContext"

**What about this other PR ?**
_#394 multicast request: remove exchange only by timeout_ : should we integrate it before ?
_#367 Bert TCP Support_ : should we integrate it before ?
_#326 Add isEmptyWithDump() for more details in logs while testing_ : we should agree on that before to merge.
_#322 Add endpoint identifier_ : probably throwed out
_#294 Add correlation context and principal to TLS connectors_ : is it part of EndpointContext ?
_#95 Exchange shouldn't be removed if it with a multicast request_ : should we integrate it before ?
_#87 Observationstore improvement_  : the problem is still here but will not be merged soon so we will see later.

boaks:
Uups, I used the "bullets" instead of numbers to leave it open, in which order the work should be done :-).

I agree with the first two.

But then I would prefer to postpone PR #399. Even if it's "mostly moving", this makes it hard to rebase. 
[...]

PR #401, opened 2017-09-07
Introduce endpoint context.

boaks:
The change flood :-)

This PR is not intended to be detailed reviewed. The idea is to provide the possibility to check, if the idea of a endpoint context would work also in the "using contexts", and as base for discussion about the request/response matching.

Though it renames a central interface from CorrelationContext to EndpointContext and move the principal and peer address into this new interface; it's really a change flood.

That modified interface is then used by the Message, provide access and the possibility to provide the available information for matching the endpoints. This would introduce the export of that EndpointContext and would require some more work there to ensure, that it could be save persisted (at least the parts the "setup" would require, e.g. "principals", if the setup relays on unique credentials).
The cleanup of the moved address in the Message causes also a large number of changes.

The intended "usual" flow could the be:
initiator of communication (e.g. lwm2m client): 
- provide address (and optional principal) to the request ans EndpointContext (may be incomplete).
- while sending, the callback provides the completed EndpointContext for the sent request.
- receiving the response also contains such an EndpointContext.
-- normal responses are first matched by their MID/token/address and then the EndpointMatcher decides, if it fulfils also the security constrains.
-- notify responses are first matched by their token and then the EndpointMatcher decides, if it fulfils also the security constrains.
- the received EndpointContext may be used for further request.

The matching of notify responses will relay on system unique tokens, but the final matching would consider the endpoint context to ensure the security constrains. 

Receiver of communication (e.g. lwm2m server):
- receiving the first incoming request contains a endpoint context.
- use that endpoint contest for sending responses or own request.

Depending on the feedback, I think providing the rename and modifying into separate (but not working) single commits should make it easier to review. 


sbernard31:
I'm ok with the main idea.

The main concern for me is about identifying notify response by token only.
I quickly resume why: 
1) if we identify notification by token, we can not use random generation because there some risk of collision and so we need to use a reserve/release way.
2) I have concern about reserve/release because this could be a source of memory leak and the main reason if you have a cluster of server you must share reserved token between instances  :scream: ... 

Did you drop your [previous idea](https://github.com/eclipse/californium/issues/344#issuecomment-314075391)?  or maybe and endpointContext should be added as identifier ?

boaks:
> Did you drop your previous idea? or maybe and endpointContext should be added as identifier ?

https://github.com/eclipse/californium/issues/174#issuecomment-327458989

If the principal and correlation context is moved into Message, the "endpoint matcher" would not be able to use the Message, because the "endpoint matcher" is in "element-connector" but Message in "core".

Therefore I updated the "idea" of collecting the "endpoint indentity candidates" into the "endpoint context".

boaks:
> I'm ok with the main idea.

Perfect. I think, reorganizing the "endpoint identity candidates" into the "endpoint context" is a good base also for future advanced ideas on the "endpoint identities".

boaks:
> The main concern for me is about identifying notify response by token only.
> I quickly resume why:
> 
>     1. if we identify notification by token, we can not use random generation because there some risk of collision and so we need to use a reserve/release way.
>     2. I have concern about reserve/release because this could be a source of memory leak and the main reason if you have a cluster of server you must share reserved token between instances 😱 ...
> 

I don't think, a "endpoint context" will block to elaborate and evolve ideas about identifying the observe-exchange for the notifies. But for **now**, it doesn't make that identifying worser but solve some security issues with new handshakes or address changes for notifies.

Acoording 2. FMPOV tokens may better be maintained by a collection with a timeout (e.g. as MID). Then the "release" will not be required. This would also help to make the proactive observation cancel using the observe token easier.

sbernard31:
> I don't think, a "endpoint context" will block to elaborate and evolve ideas about identifying the observe-exchange for the notifies

Good.

> for now, it doesn't make that identifying worser but solve some security issues with new handshakes or address changes for notifies.

Agree.

Ok, so what are the next points ?
Do you split that in several PR/Commits or should I try to review it like that ? (already tried but feel that feedback could be better if we split that)

boaks:
> Ok, so what are the next points ?

Give @sophokles73 a chance to give us his opinion (or just that he has no time to do so :-) ).
 
> Do you split that in several PR/Commits or should I try to review it like that ?

Yes, I would start after Kai's feedback. 

>  (already tried but feel that feedback could be better if we split that)
That's my feeling too :-).

I also prepared a "look ahaead PR" for leshan, to see how such a "EndpointContext could be used in the "californium-mapping-layer". If you're interessted, I would comment it and publish that PR (marked as not to be merged now).

sbernard31:
> If you're interessted, I would comment it and publish that PR (marked as not to be merged now).

Please, do it  :)

sophokles73:
Hey guys,

sorry for the long delay, I just returned yesterday to the office from my vacation ...

Moving the candidate properties of a message into the `EndpointContext` seems like a good idea to me too. I only have some limited memory of the state-of-affairs regarding the token generation question, though. So you guys need to help me out. So, what is the current thesis on how we want to generate the tokens for outbound requests?

boaks:
FMPOV, this work aims to "flexible secure" request/response/notifies and follow-up request.

It doesn't really change the current token usage, it only validates a notify-exchange (which is currently identified by the "system unique token"), more flexible either by the credentials (if DTLS resume is used) or by principal (if unique credentials are used. currently not RFC based, but discussed there).

About the token:
I'm not really sure, if we have a common way how we want to use tokens. FMPOV I would vote for a "timeout based central token management", which could be implemented based on "endpoint context" or "system wide". We may have also a "endless" for observes, which, when removed from the observation store, is changed to the usual token timeout. This could be also implemented with a "Random only maintainer". But currently I don't work on this, because the changes here would be enough for a couple of days :-).

So I start (if no objections) with a simple renaming "CorrelationContext" into "EndpointContext" commit and then fill it with the new stuff in a follow-up commit(s).

marvelous:
`AddressEndpointContext.getPeerAddress` is exactly what I need. 👍

boaks:
Merge with #419, #412, and #411.

PR #402, opened 2017-09-08
Relax timeout.

boaks:
Correct thread safe set of drops.
Use timeout scale to relax overall timeout.
Intended to reduce failing of testBlockwiseTransferToleratesLostMessages.


boaks:
This PR is introduced, because the "parent_release" job 56 reported and timeout.

sbernard31:
No objection.

Issue #406, opened 2017-09-12
DTLS session is sometimes not yet registered when onContextEstablished is called


marvelous:
Using 2.0.0-M5, I establish a connection with a server:
```
ping = RawData.outbound(payload, address, null, messageCallback, false);
connector.send(ping);
```
and in messageCallback.onContextEstablished:
```
session = connector.getSessionByAddress(address);
```
However, the returned session is sometimes null and sometimes non-null. The problem seems that `Connection.sessionEstablished` gets called after the deferred message sender, depending on the Handshaker.sessionListeners iteration order.

I guess that using a LinkedHashSet instead of a HashSet for Handshaker.sessionListeners would give an iteration order identical to the insertion order and fix this problem.

boaks:
So did you read the note in the `MessageCallback` interface?

> Note: usually this callback must be processed in a synchronous manner, because
> if it returns, the message is sent. Therefore take special care in methods called
> on this callback.

with that read, are you sure, you want to call `connector.getSessionByAddress(address)`, which has no documentation about threading?

Why do you want to access the DTLSSession? Are you interested in the `Principal`?
We currently working on make the `Principal` more accessible. So keep watching our development :-).
  
Anyway, thanks for your hint about the used `Set`, I'll create a PR to change that.

marvelous:
Indeed, it's the peerIdentity that I am trying to access here. I'll keep watching :-)

boaks:
So, can you try the PR #407, if it fixes your issue?

boaks:
And just, if you have too much time, take a look at the PR #401. It's just a preview for exporting the peer identity. May be you can give us valuable feedback, if this would work for your requirements as well.

marvelous:
Sorry, I didn't think you'd follow up here, since I wasn't using the API correctly. I'll see if I find the time to make a build, but I'm more interested in taking a look at the endpoint context.

PR #407, opened 2017-09-12
Fix order of session listener.

boaks:
Issue #406


marvelous:
I can confirm that this fix solves issue #406

PR #410, opened 2017-09-13
Fix some transparent block2 issues.

sbernard31:
I add 2 tests and 2 fix for transparent block2 issues.

1) An older block2 notification should not stop a more recent one. (issue at client side)

2) A new block2 transfer should interrupt the previous one (issue at server side)

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -217,7 +217,7 @@ public class BlockwiseLayer extends AbstractLayer {
 					// of the notify to be abandoned so that the client receives
 					// the requested response but lose the notify. 
 					clearBlock2Status(key);
-					status.completeTransfer(null);
+					status.completeOldTransfer(null);
 				}
 				
 				if (requiresBlockwise(request)) {
@@ -598,13 +598,13 @@ public class BlockwiseLayer extends AbstractLayer {
 								"discarding outdated block2 transfer {0}, current is [{1}]",
 								new Object[]{ status.getObserve(), response });
 						clearBlock2Status(key);
-						status.completeTransfer(exchange);
+						status.completeOldTransfer(exchange);
 					} else {
 						LOGGER.log(
 								Level.FINER,
 								"discarding old block2 transfer [{0}], received during ongoing block2 transfer {1}",
 								new Object[]{ response, status.getObserve() });
-						exchange.setComplete();
+						status.completeNewTranfer(exchange);
 						return;
 					}
 				}
@@ -613,7 +613,7 @@ public class BlockwiseLayer extends AbstractLayer {
 							Level.FINER,
 							"discarding outdate block2 response [{0}, {1}] received during ongoing block2 transfer {2}",
 							new Object[]{ exchange.getNotificationNumber(), response, status.getObserve() });
-					exchange.setComplete();
+					status.completeNewTranfer(exchange);
 					return;
 				}
 			}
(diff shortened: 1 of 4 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -604,7 +604,12 @@ public void receiveResponse(final Exchange exchange, final Response response) {
 								Level.FINER,
 								"discarding old block2 transfer [{0}], received during ongoing block2 transfer {1}",
 								new Object[]{ response, status.getObserve() });
-						exchange.setComplete();
+						// We don't complete older exchange 
+						// if this is the one used for more recent block2 exchange
+						if(!status.getExchange().equals(exchange))
```
sbernard31: Should we do this test for `exchange.setComplete()` below ?

boaks: I'm not sure, if introducing the usage of `Exchange.equals()`, is needed. 
I miss the implementation of equals in Exchange and until now `==` seems to work.

But I have to see the issue tomorrow more detailed.

sbernard31: agreed to  `==` instead of `equals`.

boaks: final void completeNewTransfer(Exchange newExchange) {
		Exchange exchange;
		synchronized (this) {
			exchange = this.exchange;
		}
		if (exchange != null && newExchange != exchange) {
			// complete new exchange
			newExchange.setComplete();
		}
	}

What's about add a new method instead of expose the exchange?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
```diff
@@ -433,6 +435,99 @@ public void testBlockwiseObserveAndNotificationWithoutBlockwise() throws Excepti
...
+
+		Request request = createRequest(GET, path, server);
+		request.setObserve();
+		SynchronousNotificationListener notificationListener = new SynchronousNotificationListener(request);
+		client.addNotificationListener(notificationListener);
+
+		// Send observe request
+		client.sendRequest(request);
+		// Expect observe request
+		server.expectRequest(CON, GET, path).storeMID("OBS_REQ").storeToken("OBS_TOK").go();
```
boaks: Just to say: you can use `storeBoth("ID")` and access the parts with `loadToken("ID")` or `loadMID("ID")`.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: Yes, debugging with your new test shows, that it was a bug to always complete the exchange.
But I would prefer to stick with the pattern of named "complete" in the status rather then export the exchange.

sbernard31: Ok, I will do that.

You didn't answer to : "Should we do this test for exchange.setComplete() below ?"

boaks: > Yes, debugging with your new test shows, that it was a bug to always complete the exchange. 

So, yes, we should add that test.

sbernard31: Ok thx

sbernard31:
Re-push :
- I used storeBoth/loadBoth for tests.
- I replace `getExchange` access to `completeNewTranfer` method.
- I rename `completeTransfer` in `completeOldTransfer` .
- Revert formatting in `BlockwiseServerSideTest`
- change comment about new block2 transfer.

boaks:
LGTM

PR #411, opened 2017-09-14
Extend endpoint context with inet socket address and principal.

boaks:
Cleanup RawData constructors.
Use EndpointContext and EndpointContextMatcher mocks for Matcher tests
in core.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
@@ -14,8 +14,13 @@
  *    Bosch Software Innovations GmbH - add support for correlation context to provide
  *                                      additional information to application layer for
  *                                      matching messages (fix GitHub issue #1)
+ *    Achim Kraus (Bosch Software Innovations GmbH) - extend endpoint context with
+ *                                                    inet socket address and principal
  ******************************************************************************/
 package org.eclipse.californium.elements;
+
+import java.net.InetSocketAddress;
+import java.security.Principal;
 
 /**
  * A endpoint context that explicitly supports DTLS specific properties.
@@ -29,12 +34,17 @@ public class DtlsEndpointContext extends MapBasedEndpointContext {
 	/**
 	 * Creates a new endpoint context from DTLS session parameters.
 	 * 
+	 * @param peerAddress peer address of endpoint context
+	 * @param peerIdentity peer identity of endpoint context
 	 * @param sessionId the session's ID.
 	 * @param epoch the session's current read/write epoch.
 	 * @param cipher the cipher suite of the session's current read/write state.
-	 * @throws NullPointerException if any of the params is <code>null</code>.
+	 * @throws NullPointerException if any of the params (except peerIdentity)
+	 *             is {@code null}.
 	 */
-	public DtlsEndpointContext(String sessionId, String epoch, String cipher) {
+	public DtlsEndpointContext(InetSocketAddress peerAddress, Principal peerIdentity, String sessionId, String epoch,
+			String cipher) {
+		super(peerAddress, peerIdentity);
 		if (sessionId == null) {
 			throw new NullPointerException("Session ID must not be null");
 		} else if (epoch == null) {
@@ -58,5 +68,11 @@ public class DtlsEndpointContext extends MapBasedEndpointContext {
 
 	public String getCipher() {
 		return get(KEY_CIPHER);
+	}
+
+	@Override
+	public String toString() {
+		return String.format("DTLS(%s:%d,ID:%s)", getPeerAddress().getHostString(), getPeerAddress().getPort(),
+				getSessionId());
 	}
 }
--- a/element-connector/src/main/java/org/eclipse/californium/elements/EndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/EndpointContext.java
@@ -14,15 +14,44 @@
  *    Bosch Software Innovations GmbH - add support for correlation context to provide
  *                                      additional information to application layer for
  *                                      matching messages (fix GitHub issue #1)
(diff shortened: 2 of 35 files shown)


element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
```diff
@@ -327,7 +226,7 @@ public EndpointContext getEndpointContext() {
 	 *         otherwise
 	 */
 	public boolean isSecure() {
-		return (endpointContext != null && endpointContext.get(DtlsEndpointContext.KEY_SESSION_ID) != null);
+		return (peerEndpointContext != null && peerEndpointContext.get(DtlsEndpointContext.KEY_SESSION_ID) != null);
```
sbernard31: I suspect this implementation of isSecure is not really ok or maybe this is just the name and/or javadoc which is not ok.

Now we are able to set Endpointcontext at constructor and as this context could be incomplete, there could be other condition we make this secure. (e.g. a principal is available)

I check and this is called only on CoapEndpoint.InboxImpl.receiveRequest(Request, RawData), so I don't think this is so impacting. (so probably not so important)

element-connector/src/main/java/org/eclipse/californium/elements/EndpointContext.java
```diff
@@ -14,31 +14,87 @@
...
+
+	/**
+	 * Check, if the correlation information contained, will inhibit a new
+	 * connection.
+	 * 
+	 * @return {@code true}, if no new connection could match the correlation
+	 *         context provided in this instance. {@code false}, if a new
+	 *         connection may match this correlation context.
+	 */
+	boolean inhibitNewConnection();
```
sbernard31: I do not understand the purpose of this method.

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
```diff
@@ -29,12 +34,16 @@
...
+	 * @param peerAddress peer address of endpoint context
+	 * @param peerIdentity peer identity of endpoint context
 	 * @param sessionId the session's ID.
 	 * @param epoch the session's current read/write epoch.
 	 * @param cipher the cipher suite of the session's current read/write state.
-	 * @throws NullPointerException if any of the params is <code>null</code>.
+	 * @throws NullPointerException if any of the params (except peerIdentity) is {@code null}.
 	 */
-	public DtlsEndpointContext(String sessionId, String epoch, String cipher) {
+	public DtlsEndpointContext(InetSocketAddress peerAddress, Principal peerIdentity, String sessionId, String epoch,
```
sbernard31: Should we add a toString to DTLSEndpointContext ? 
```java
@Override
public String toString() {
	return String.format("DTLS(%s:%d)", getPeerAddress().getHostString(), getPeerAddress().getPort());
}
```
Maybe with more detail ? (sessionId ?, cipher ?)

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
boaks: There was a short discussion about the isSecure() in #156.

In the end, I'm not sure, if this is still relevant. This would only make sense, if a CoapEndpoint would receive mixed request. But currently we use separate connectors and endpoints for coap or coaps.

element-connector/src/main/java/org/eclipse/californium/elements/EndpointContext.java
boaks: Sending a message using a connector sometimes "triggers" a connection of handshake. If the EndpointContext provides data in the endpoint context, which excludes new connections or handshakes to bb used succesful for sending, you can drop the message early.

E.g. if "strict" DTLS session based matching is wanted and a message for a defined dtls session is send, it would make sense to start a new handshake, because, it would not get the provided session when it gets a new session. Or if a response is sent for coap+tcp but the connection is dropped, it doesn't make sense to establish a new connection.

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
boaks: This would be in danger of overload the logging.

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
sbernard31: I continue the discussion in #156, because it's more appropriate.

element-connector/src/main/java/org/eclipse/californium/elements/EndpointContext.java
sbernard31: ok

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
sbernard31: why not for overloading logging.

But why not adding `toString` to DTLSEndpointContext knowing that all other have a custom `toString` ? (I mean a too string with just peerAddress and Principal like other endpointContext)

boaks: Uups, after changing to much, I assumed that it was already added and your comment was just about the extension. Sure, "toString()" is a that goog idea, that I thought it's already in :-).
I added also the session id but latter I would trunc this to the first characters.

element-connector/src/main/java/org/eclipse/californium/elements/RawData.java
boaks: I would postpone that discussion, because I think "isSecure()" will be better removed, and the scheme is better set using the scheme from the connector. But to check, if I have overseen a use-case, I would first try to finish the changes for the endpoint context.
The next step would be to switch to that EndpointContext in the core.

sbernard31:
So good to merge :)

PR #415, opened 2017-09-21
Improve tests

sbernard31:
Add Memory leak tests on blockwise tests.
Add a test with observe, block2 and changing IP address/port environment.

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseServerSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseServerSideTest.java
@@ -125,10 +125,15 @@ public class BlockwiseServerSideTest {
 
 	@After
 	public void shutdownEndpoints() {
-		printServerLog(serverInterceptor);
-		System.out.println();
-		client.destroy();
-		server.destroy();
+		try {
+			assertAllExchangesAreCompleted(CONFIG, exchangeStore);
+		} finally {
+			printServerLog(serverInterceptor);
+
+			System.out.println();
+			client.destroy();
+			server.destroy();
+		}
 	}
 
 	@AfterClass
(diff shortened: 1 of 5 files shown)


boaks:
This PR would introduce "some more logical copies" of the "waitForCondition" "isEmpty".

PR #326 would introduce a reusable solution for that.

sbernard31:
Ok.

So, what does it means ?
1) We merge this PR now and we must remember that changes must be done in #326. (or another PR)
2) We wait for #326 to be merged and then we adapt this one ... (but it seems discussion are not closed on #326)
3) We extract `MessageExchangeStoreTool` in another PR, then integrate this one and we will see later for #326.

boaks:
I would prefer, if you go for 3.

boaks:
Maybe, tests for `randomAccess` would help.

sbernard31:
Done.
(I can not change #326 PR as it is based on a bsinno branch)

boaks:
> Maybe, tests for randomAccess would help.

After thinking again, I guess for a blockwise block2 client completing a notify "randomAccess" doesn't apply, its only on the server side.

boaks:
LGTM.
I would like to change the way of the address change of the lockstependpoint.

sbernard31:
Ok I added your createChangedLockstepEndpoint commit, used it and fixed crappy test name (Oops :/).
It sounds good to merge.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseServerSideTest.java
```diff
@@ -798,6 +799,8 @@ public void testObserveWithBlockwiseResponse() throws Exception {
 
 		client.sendRequest(CON, GET, tok3, ++mid).path(RESOURCE_PATH).block2(2, false, 128).go();
 		client.expectResponse(ACK, CONTENT, tok3, mid).block2(2, false, 128).noOption(OBSERVE).payload(respPayload.substring(256, 290)).go();
+
+		testResource.clearObserveRelations();
```
boaks: I'm not sure, what the intent of the "clear" is.
`testResource` is "destroyed" anyway with the `server.destroy()` in `shutdownEndpoints()`. 
Or why is this required?

boaks:
LGTM.
(The intention of the `clearObserveRelations()` could also be given after merging.)

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseServerSideTest.java
sbernard31: If we don't clean observe relations the exchangeStore is not empty. (as the observe relation is still established)
Another solution could be to make the client stop the observation. (but I choose this solution as the cancel seems to not be part of the test)

PR #416, opened 2017-09-21
Use socket's reuseAddress only, if bindAddress determines a port.

boaks:
Running the scandium tests repeatedly shows on my build systems, that "reuseAddress" even causes failures with the client and server on the **same** port. This reduces that failure.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -57,6 +57,8 @@
  *                                                    and add error callback in
  *                                                    newDeferredMessageSender.
  *    Achim Kraus (Bosch Software Innovations GmbH) - reuse receive buffer and packet. 
+ *    Achim Kraus (Bosch Software Innovations GmbH) - use socket's reuseAddress only
+ *                                                    if bindAddress determines a port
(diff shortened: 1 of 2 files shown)


sbernard31:
I have clearly no idea, if this is a good fix or not.

Which kind of failure did you faced because of that ?

boaks:
As mentioned in the comment:
Running this tests repeated, causes extremly rare failures and analysing the logs shows, that server and client uses the **SAME** port!

boaks:
Generally it's configurable, this changes the default.

I would also like to use the "Direct Socket Implementation" for tests, see #422.
With that the "reuse" for tests is not longer helpful.

sbernard31:
> Running this tests repeated, causes extremly rare failures and analysing the logs shows, that server and client uses the SAME port!

Sry, I miss-understood.

> This reduces that failure.

Reduces only ? that means even with this fix there is still some failure where client and server use the same port ?

boaks:
>  that means even with this fix there is still some failure where client and server use the same port ?

No, that was not precise by me. The test still rarely fails, but for other reasons, mainly timing issues, if the CPU is not available. I still think on introduce a "relaxed timing feature" for tests running on the eclipse HIPP.

sbernard31:
Ok let's merge this PR if you think this could improve the test stability.

Issue #358, opened 2017-07-12
Blockwise feedback and exchange cleaning


sbernard31:
I see in BlockwiseLayer, there is blockTimeout.
It seems to be only for outcoming block2 and incoming block1. So this is timeout to wait for the next GET.
When this timer expired we just clean the blockstatus cache.
But should we clean the exchange (using something like exchange.setComplete) ?
And for block2, should we report this timeout with Message.onTimeout (or something else?)

boaks:
Working on #356 I got aware, that if a single block request is timed out, the outer complete request will not be aware of that. I fixed that in `Exchange.setTimedOut()`. The other "errors" need to be checked also.

If two blockwise transfers are simultaneously ongoing, I found it's not enough to clear the "blockstatus" of one, the (client-side) exchange should be also "stopped" somehow. 

Right now, I would guess, that except of the "unaware simultaneous blockwise transfer", usually the "block status" should not require to cleanup the Exchange, because the Exchange should be cleaned up on itself.

boaks:
> But should we clean the exchange (using something like exchange.setComplete) ?
> And for block2, should we report this timeout with Message.onTimeout (or something else?)

As far as I understand, the `BlockwiseStatus` is just a helper for holding or assembling the blocks. Therefore the `Exchange` cleanup is done on the Exchange mechanisms, maybe triggered when processing messages related to the resource. See #363, which addresses the timeout and the complete for several cases.

I also tried to check, if the "ExchangeStore" is empty for `ObserveServerSideTest`; it isn't. 
But without further logging information in the exchange it seems to me to be too time-consuming to fix that.

boaks:
Can we close this?

sbernard31:
I don't think so.

>  should we clean the exchange (using something like exchange.setComplete) ?

I think we should add some test about that. I mean test about block timeout.

> should we report this timeout with Message.onTimeout (or something else?)

Still no answer to this one but maybe this will be drived by tests.

sophokles73:
@sbernard31 
Achim and I are currently trying to identify issues that need to be done in 2.0.0. Is this one of them? Who is working on this?

sbernard31:
I'm working on it. #449 is related to it. #414 is probably related too.

sbernard31:
With #449 and #475, I hope this behave better now.

PR #418, opened 2017-09-21
Add blockwise test with unexpected answer.

boaks:
Introduce createChangedLockstepEndpoint() for tests with address
changes.

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
@@ -937,7 +937,39 @@ public class ObserveClientSideTest {
 		response = notificationListener.waitForResponse(1000);
 		assertResponseContainsExpectedPayload(response, respPayload);
 		printServerLog(clientInterceptor);
+	}
 
+	/**
+	 * Verifies try to establish observe relation. Response use block2 but last
+	 * block transfer does not use the right token..
+	 * 
+	 * @throws Exception if the test fails.
+	 */
+	@Test
+	public void testBlockwiseObserveUnexpectedAnswer() throws Exception {
+
+		System.out.println("Blockwise Observe with unexpected answer:");
+		respPayload = generateRandomPayload(40);
+		String path = "test";
+
+		// Established new observe relation with block2
+		Request request = createRequest(GET, path, server);
+		request.setObserve();
+		SynchronousNotificationListener notificationListener = new SynchronousNotificationListener(request);
+		client.addNotificationListener(notificationListener);
+		client.sendRequest(request);
+
+		server.expectRequest(CON, GET, path).observe(0).storeBoth("A").go();
+		server.sendResponse(ACK, CONTENT).loadBoth("A").observe(0).block2(0, true, 16).size2(respPayload.length()).payload(respPayload.substring(0, 16)).go();
+		server.expectRequest(CON, GET, path).storeBoth("B").block2(1, false, 16).go();
+		server.sendResponse(ACK, CONTENT).loadBoth("B").block2(1, true, 16).payload(respPayload.substring(16, 32)).go();
+		server.expectRequest(CON, GET, path).storeBoth("C").block2(2, false, 16).go();
+		// Send last response with bad token (with use a previous token)
+		server.sendResponse(ACK, CONTENT).loadMID("C").loadToken("B").payload(respPayload.substring(32)).go();
+
+		Response response = request.waitForResponse(1000);
+		assertNull(response);
+		printServerLog(clientInterceptor);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


boaks:
So, at least if the server (device client) answers something unexpected, the exchange get's stale.
So, we need to check, if the old transfer is ongoing, that we do trigger the retransmission of the last request. In my experience, this should be done after the `EXCHANGE_LIFETIME` to get rid of wrong deduplication.

sbernard31:
I changed this PR as `createChangedLockstepEndpoint` is now integrated in master (see #425)

boaks:
Uups, @sbernard31, sorry I saw your comment to late and pushed forced on this branch ...
sorry.

sbernard31:
Should I repush force ? 
I have changed some comment.

boaks:
feel free to do so.
But don't merge, the test will fail. 
It was just a late night idea yesterday, what might go wrong in the blockwise.

sbernard31:
Ok.
I saw the test failed :). I'm currently looking at this.

sbernard31:
Just to be sure : 

You want to check what happened for an incomplete block2 transfer ?
You add the last response with good MID/wrong token to avoid the retransmission timeout ?
(is it intended to not use block2 option)

I thought that if we want to test this correctly we should set the BLOCKWISE_STATUS_LIFETIME.
(I set it to 300ms I'm not sure this is a right value)
___________________________

So I add some tests and the use of BLOCKWISE_STATUS_LIFETIME.
(but with or without block_status_lifetime tests result are the same.)

```
Incomplete transfer with no answer to last block 
    =>  "failed : exchange not empty"
Incomplete transfer with response (with block2 + unknown token) 
    =>  "failed : exchange not empty"
Incomplete transfer with response (without block2 + unknown token) 
    =>  "failed : exchange not empty"
Incomplete transfer with response (with block2 + A token)
    =>  "failed : exchange not empty"
Incomplete transfer with response (without block2 + A token)
    =>  "failed : exchange not empty"
Incomplete transfer with response (with block2 + B token)
Incomplete transfer with response (without block2 + B token) : "failed : exchange not empty"
Incomplete transfer with response (with block2 + C token) : "failed : exchange not empty + we receive a response "
```

boaks:
I just tried to do something "unexpected" to see, if the gap may be there. And it smelled for a hit. But I'm not sure, if this has to do something with the reality.

sbernard31:
After playing a bit more with this tests, here is what I found :

```
Incomplete transfer with no answer to last block 
    =>  "failed : exchange not empty"
```
=>  this one is mainly because of bad timing configuraiton. ACK_TIMEOUT and MAX_RETRANSMIT  are too high in tests. So this is not a real issue. 

```
Incomplete transfer with response (with block2 + unknown token) 
    =>  "failed : exchange not empty"
Incomplete transfer with response (without block2 + unknown token) 
    =>  "failed : exchange not empty"
Incomplete transfer with response (with block2 + A token)
    =>  "failed : exchange not empty"
Incomplete transfer with response (without block2 + A token)
    =>  "failed : exchange not empty"
Incomplete transfer with response (with block2 + B token)
    =>  "failed : exchange not empty"
Incomplete transfer with response (without block2 + C token)
    =>  "failed : exchange not empty"
```
=> why ? because exchange is acknowledged(MID is correct), so no more retransmission and response is not valid (token is not correct, too old, without block2 option ..) so just ignored.  

```
Incomplete transfer with response (without block2 + B token) 
    =>  "failed :  we receive a response "
```
=> Here  this is just because we use same token (token "B" == token "C"), so the behavior is "normal"

sbernard31:
So the real issue is how to handle case where a block request is acknowledge but we never get any response. (Separated response or "bad message" like above)

Solutions ? 
1) launch a cleaning task. We already do that in blockwiselayer to wait next request maybe we should also do that to wait next response. (In this solution we should also "complete" exchange on clean-up #358)
2) a more generic solution, add a timeout for response in the CoAP stack. (see https://github.com/eclipse/californium/pull/74 where I argue against that idea, but I didn't take into account this block issue)

boaks:
> So the real issue is how to handle case where a block request is acknowledge but we never get any response. (Separated response or "bad message" like above)

FMPOV a response is received, but this response doesn't fit into the expectation of a "valid blockwise transfer". 

Having a ongoing blockwise transfer (BlockwiseLayer.receiveResponse, line 579), the "none blockwise, none notify" response is ignored, but therefore the blockiwse transfer is also not completed, because the request for the next block is never sent.

On the other side, you right, there may still be other cases, where it still fails ;-(, so feel free to introduce such a cleanup.

sbernard31:
Do you have any preference or argument to go for the 1srt or 2nd solution ? 
(is it normal you push force this branch ?)

boaks:
> (is it normal you push force this branch ?)

Uups, sorry.
I frequently rebase the branches of my PRs to see, if the changes still work.
And if so, I push the changes forced on the branch (because rebase requires that, AFAIK).
So I guess, I did this just with my local branch and didn't check, if there are other changes on that remote branch. 

So have I overwritten your stuff? If so, I'm very sorry.

boaks:
> Do you have any preference or argument to go for the 1srt or 2nd solution ?

No, I don't have any preferences. Please extend such a timeout on every block, not that a large latency accidentally cleansup an ongoing blockwise exchange.

sbernard31:
We can close this as relevant tests was added in #441.

PR #422, opened 2017-09-25
Use direct socket implementation for scandium.

boaks:
Support very fast reopen and didn't (rarely) drop packages.

### DIFF
--- a/scandium-core/pom.xml
+++ b/scandium-core/pom.xml
@@ -19,12 +19,14 @@
 		<dependency>
 			<groupId>${project.groupId}</groupId>
 			<artifactId>element-connector</artifactId>
+			<type>jar</type>
 		</dependency>
 		<dependency>
 			<groupId>${project.groupId}</groupId>
 			<artifactId>element-connector</artifactId>
 			<classifier>tests</classifier>
 			<type>jar</type>
+			<scope>test</scope>
 		</dependency>
 		<dependency>
 			<groupId>${project.groupId}</groupId>
@@ -53,6 +55,7 @@
 						   by means of the Scandium-logging.properties file
 						-->
 						<java.util.logging.config.file>${project.build.testOutputDirectory}/Scandium-logging.properties</java.util.logging.config.file>
+						<org.eclipse.californium.junit.socketmode>DIRECT</org.eclipse.californium.junit.socketmode>
 					</systemPropertyVariables>
 					<excludes>
 						<exclude>**/*$*</exclude>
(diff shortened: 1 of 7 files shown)


boaks:
If no objection, I would like to merge this.

sbernard31:
I'm currently looking a this, but I never look at this network rules. So I may ask stupid question. :p

boaks:
Don't hesitate to ask :-). 
We use it now for a couple of months in the core tests.

scandium-core/pom.xml
```diff
@@ -61,8 +64,8 @@
 						Running the tests in parallel may cause problems with rebinding to a an address within a test,
 						e.g. if another test running in parallel binds to that address in between.
 					-->
-					<parallel>classes</parallel>
-					<threadCount>8</threadCount>
+					<!-- parallel>classes</parallel>
+					<threadCount>8</threadCount-->
```
sbernard31: Is it normal that parallel test was deactivated ? this seems not linked to this PR ?

scandium-core/src/test/resources/Scandium-logging.properties
```diff
@@ -69,3 +69,8 @@ org.eclipse.californium.scandium.useParentHandlers = false
 org.eclipse.californium.scandium.level = INFO
 org.eclipse.californium.scandium.dtls.level = CONFIG
 org.eclipse.californium.scandium.dtls.Connection.level = INFO
+
+org.eclipse.californium.elements.handlers = java.util.logging.ConsoleHandler
+org.eclipse.californium.elements.useParentHandlers = false
+org.eclipse.californium.elements.level = INFO
+org.eclipse.californium.elements.util.DirectDatagramSocketImpl.level = FINE
```
sbernard31: is it intentional to set this log level to FINE, this make a lot of log on `mvn clean install`.

sbernard31:
Not directly linked to this PR, but here my stupid questions about NetworkRules ^^

This idea of this rules is to be able to switch to native to direct socket ?
Native is the "normal" way, direct is a kind of inmemory implementation ?
The purpose of this, having a better performance and avoid dropping packet  ?

scandium-core/pom.xml
boaks: OK, I have to check this. It was necessary for the "repeating runner", because parallel testing seems to use also a own runner. But I didn't check it for the "rule" only.

scandium-core/src/test/resources/Scandium-logging.properties
boaks: The intention is to see the exchange messages in the log. Yes this make large logs, but without it's not possible to find the cause of failures of rarely failing tests.

boaks:
> This idea of this rules is to be able to switch to native to direct socket ?

The switch is done by `org.eclipse.californium.junit.socketmode`, the rule is a filter for tests, if one would require to run only in NATIVE or DIRECT. Once the socket impl is set, it can not be removed.
See PR #241

> Native is the "normal" way, direct is a kind of inmemory implementation ?

Yes, but in mvn we use the direct implementation, so native is left when executing the test in eclipse junit test runner.

> The purpose of this, having a better performance and avoid dropping packet ?

It should control unintended dropping an reordering of messages. And provide logging.
The performance was not the reason to introduce it, but may be a side effect.

sbernard31:
Ok, so good to merge.

PR #423, opened 2017-09-26
InMemoryMessageIdProvider does not remove stale entry

sbernard31:
if `InMemoryMessageIdProvider` is full but contains stale entries, adding new entry is not allowed.

This bug comes from a miss-understanding of `remainingCapacity`.

```java
/**
 * Gets the number of entries that can be added to this cache without
 * the need for removing stale entries.
 * 
 * @return The number of entries.
*/
public final int remainingCapacity() {
```


boaks:
LGTM

PR #426, opened 2017-09-29
Move application handler call out of synchronized block.

boaks:
Though handler may call other synchronized stuff, the call is moved out
of the synchronized block.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -63,6 +63,8 @@
  *                                                    remove scheme
  *    Achim Kraus (Bosch Software Innovations GmbH) - check for cancelled retransmission
  *                                                    before sending.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - move application handler call
+ *                                                    out of synchronized block
  ******************************************************************************/
 package org.eclipse.californium.scandium;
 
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -716,8 +718,10 @@ private void processApplicationDataRecord(final Record record) {
 						// thus, the handshake seems to have been completed successfully
 						connection.handshakeCompleted(record.getPeerAddress());
 						session.markRecordAsRead(record.getEpoch(), record.getSequenceNumber());
-						// finally, forward de-crypted message to application layer
-						handleApplicationMessage(message, session);
+						// create application message, if handler is set.
+						if (messageHandler != null) {
```
sbernard31: Should we keep this test ?

boaks: If agreed, that such a connector without messageHandler doesn't make sense, we can remove it and just keep the check below.

sbernard31: Ok except, that sounds good to merge.

boaks: so remove the "if"?

sbernard31: yes.

PR #427, opened 2017-09-29
Check for cancelled retransmission before sending.

boaks:


sbernard31:
I'm not sure to get the point. Why should we do that ?
If flight is canceled, task should be canceled too, right ?

boaks:
The "task" adds a executable to the stripped-executor. 
If the task is cancel afterwards, but before the execution of the executable, stopping the sent would be a improvement.

sbernard31:
I see, I suppose the best think to do should be to cancel task in stripedExecutor but not so easy maybe even not doable.
So, I agree this is an improvement. Good to merge.

PR #428, opened 2017-09-29
Connector protocol

boaks:
Introduce protocol in Connector.
Remove isSecure from RawData.

See issue #156 


boaks:
Alternatively we may use an enumeration in Connector instead of String as protocol.

element-connector/src/test/java/org/eclipse/californium/elements/UDPConnectorTest.java
```diff
@@ -49,10 +49,8 @@ public void stop() {
 	}
 
 	@Test
-	public void testGetUriContainsCorrectSchemeAndAddress() {
-		assertThat(connector.getUri().getScheme(), is("coap"));
-		assertThat(connector.getUri().getHost(), is(connector.getAddress().getHostString()));
-		assertThat(connector.getUri().getPort(), is(connector.getAddress().getPort()));
+	public void testGetProtocol() {
+		assertThat(connector.getProtocol(), is("UDP"));
```
sbernard31: Not sure this test make so much sense

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
```diff
@@ -245,10 +245,8 @@ private static DtlsConnectorConfig newStandardConfig(InetSocketAddress bindAddre
 	 * Verifies that the connector's endpoint URI corresponds to the configuration.
 	 */
 	@Test
-	public void testGetUriContainsCorrectSchemeAndAddress() {
-		assertThat(server.getUri().getScheme(), is("coaps"));
-		assertThat(server.getUri().getHost(), is(serverEndpoint.getHostString()));
-		assertThat(server.getUri().getPort(), is(serverEndpoint.getPort()));
+	public void testGetProtocol() {
+		assertThat(server.getProtocol(), is("DTLS"));
```
sbernard31: same remarks. (if we keep it, comment is not ok)

sbernard31:
> Alternatively we may use an enumeration in Connector instead of String as protocol.

enum seems overkill for me.

Issue #425, opened 2017-09-27
EndpointContext / introduce builder instead of sub-classes?


boaks:
With the introduction of the `EndpointContext` instead of the `CorrelationContext` and moving it to `Message` the context became (more) public in the API. 

So I would like to specify it more detailed to ensure a proper an stable usage. Currently we use sub-classes (e.g. `DtlsEndpointContext`). But using theses sub-classes, makes it more hard to serialize/deserialize such an context. Therefore I think, that we should change that sub-classes into Builders to create `MapBasedEndpointContext` instead. We may add also some builders for the "none complete" contexts, as a "EndpointContext with Principal and CipherSuite", to support that usage also.

Any opinions on that?

sbernard31:
I'm not against using builder, but I'm not sure to see why this will help for serialize/deserialize ?

What do you plan for peerAddress and Principal ?  you move it in the map or you keep the separate attributes ?

boaks:
My concern was, that currently we will have a UDP, DTLS, TCP and TLS context. So deserialization must use at least though 4 sub-classes. If Builder are used and the MapBasedEndpointContext without sub-classes, it seems to be easier for me. 
I would like to keep the special fields InetSocketAddress and Principal as they are and the other fields as map of strings to strings.

sbernard31:
I suppose that you can just create MapBasedEndpointContext at deserialization and keeping UDP, DTLS, TCP and TLS context just for convenient constructor.  
It should works without the need to introduce builder.
(Just to be clear, I'm not against builder, I just don't see why it could be better. Anyway, if you think this could be more elegant go for it, it's ok for me)

boaks:
The other part of my idea (I forgot to mention) was, to make the content of the map "final".

sbernard31:
ok.

PR #430, opened 2017-10-04
Make entries map unmodifiable.

boaks:
See issue #425.
Use still sub-classes but optionally provide attributes as pair-list to
the constructor.


sbernard31:
LGTM

PR #432, opened 2017-10-06
Reuse endpoint context.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
@@ -28,6 +28,8 @@
  *                                                    value provided by peer and current write state
  *    Bosch Software Innovations GmbH - add accessors for current read/write state cipher names
  *                                      (fix GitHub issue #1)
+ *    Achim Kraus (Bosch Software Innovations GmbH) - move creation of endpoint context
+ *                                                    from DTLSConnector to DTLSSession
  ******************************************************************************/
 package org.eclipse.californium.scandium.dtls;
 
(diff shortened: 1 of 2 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
```diff
@@ -234,8 +240,26 @@ public SessionId getSessionIdentifier() {
...
+		this.connectionReadContext = null;
+		this.connectionWriteContext = null;
+	}
+
+	public synchronized EndpointContext getConnectionWriteContext() {
+		if (null == connectionWriteContext) {
+			connectionWriteContext = new DtlsEndpointContext(peer, peerIdentity, sessionIdentifier.toString(),
+					String.valueOf(writeEpoch), cipherSuite.name());
+		}
+		return connectionWriteContext;
```
sbernard31: why not just recreate it dynamically ?

boaks: FMPOV if something is frequently (re-) used, I prefer to keep it instead of dynamically recreate it. This may depend also on the amount of memory used, but though this seems to be a wrapper around already available information, that amount should not be an issue.

sbernard31: Ok, personally I found the code a bit more complex and I'm not sure this optimization is really needed, but if you think this is better  let's go like this.

boaks: @sophokles73 
Do you have also an opinion on this?

sbernard31: (Just to be clear it's not a strong opinion)

boaks:
No more reuse, but moving the creation of the endpoint context into the DTLSSession.

PR #435, opened 2017-10-12
Introduce NatUtil and standalone demo and test NAT.

boaks:
Adjust secure-integration-tests (cf-secure) to use that NatUtil.


sbernard31:
LGTM

Issue #429, opened 2017-10-03
Memory leak in MessageExchangeStore


sbernard31:
It seems that memoryExchangeStore leaks.
Analyzing heap dump show that the leak is in  ExchangeByToken map and mainly concern observe request with block2.

The strange thing is most of exchanges have a response set, so it seems there are "complete" but not removed from the map... 

Here some old discussion about that :
https://dev.eclipse.org/mhonarc/lists/cf-dev/msg01404.html

#418 this PR aime to reproduce it via Junit.

sbernard31:
Waiting a possible workaround is to use a MessageExchangeStore with expirable cache instead of classic map. (like guava or cache2k)

<details>
  <summary> <code>ExpirableMessageExchangeStore.java</code> using guava (click to see code)</summary>

```java
import java.net.InetSocketAddress;
import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;
import java.util.Map.Entry;
import java.util.concurrent.Executors;
import java.util.concurrent.ScheduledExecutorService;
import java.util.concurrent.ScheduledFuture;
import java.util.concurrent.TimeUnit;
import java.util.logging.Level;
import java.util.logging.Logger;

import org.eclipse.californium.core.coap.Message;
import org.eclipse.californium.core.coap.Request;
import org.eclipse.californium.core.network.Exchange;
import org.eclipse.californium.core.network.Exchange.KeyMID;
import org.eclipse.californium.core.network.Exchange.KeyToken;
import org.eclipse.californium.core.network.InMemoryMessageIdProvider;
import org.eclipse.californium.core.network.InMemoryRandomTokenProvider;
import org.eclipse.californium.core.network.MessageExchangeStore;
import org.eclipse.californium.core.network.MessageIdProvider;
import org.eclipse.californium.core.network.TokenProvider;
import org.eclipse.californium.core.network.config.NetworkConfig;
import org.eclipse.californium.core.network.deduplication.Deduplicator;
import org.eclipse.californium.core.network.deduplication.DeduplicatorFactory;
import org.eclipse.californium.elements.util.DaemonThreadFactory;

import com.google.common.cache.Cache;
import com.google.common.cache.CacheBuilder;

/**
 * A copy of {@code InMemoryMessageExchangeStore} with expiration for {@code Exchange} map.
 *
 */
public class ExpirableMessageExchangeStore implements MessageExchangeStore {

	private static final long DEFAULT_EXPIRATION =  60L * 30; // in seconds
[...]

PR #437, opened 2017-10-12
 Add a way to define response timeout on request. 

sbernard31:
The idea is to add timeout for response on request.
I reject this idea by the past https://github.com/eclipse/californium/pull/74 to keep the coapstack as simple as possible,  but finally using it to resolve memory blockwise issue (#418) seems an elegant way.

Here is how this behave :
>Define a timeout value to wait the response. Zero or negative value means no timeout. 
By default no timeout is used, which means user should cancel request manually when it will not expect response anymore. (not doing this may cause memory leak) 
The time is measured before the request is sent. So you should set a time large enough to allow DTLS handshake or CoAP retransmission or transparent block-wise transfer. 

For transparent blockwise, `BLOCKWISE_STATUS_LIFETIME` is used

>The maximum amount of time (in milliseconds) allowed between transfers of individual blocks in a blockwise transfer before the blockwise transfer state is discarded. 

I had to modify some timing in tests, I hope this will not increase hudson/jenkins tests instability...

ResponseTimeoutLayer implementation is really simple. If this causes performance issue, we will be able to change the implementation easily. 
(Not directly linked but it is so bad that we cannot change layer implementation more easily, this could help to check if blockwiselayer leaks in tests)

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -248,6 +248,7 @@ public class BlockwiseLayer extends AbstractLayer {
 			status = getOutboundBlock1Status(key, exchange, request);
 
 			final Request block = status.getNextRequestBlock();
+			block.setResponseTimeout(blockTimeout);
 
 			block.addMessageObserver(new MessageObserverAdapter() {
 
@@ -735,6 +736,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		int nextNum = status.getCurrentNum() + currentSize / newSize;
 		LOGGER.log(Level.FINE, "sending next Block1 num={0}", nextNum);
 		Request nextBlock = status.getNextRequestBlock(nextNum, newSzx);
+		nextBlock.setResponseTimeout(blockTimeout);
 		// we use the same token to ease traceability
 		nextBlock.setToken(response.getToken());
 		addBlock1CleanUpObserver(nextBlock, key);
@@ -817,6 +819,7 @@ public class BlockwiseLayer extends AbstractLayer {
 						Request request = exchange.getRequest();
 
 						Request block = new Request(request.getCode());
+						block.setResponseTimeout(blockTimeout);
 						// do not enforce CON, since NON could make sense over SMS or similar transports
 						block.setType(request.getType());
 						block.setDestinationContext(response.getSourceContext());
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
@@ -99,6 +99,7 @@ public class CoapUdpStack extends BaseCoapStack {
 				new ExchangeCleanupLayer(),
 				new ObserveLayer(config),
 				new BlockwiseLayer(config),
+				new ResponseTimeoutLayer(),
 				reliabilityLayer };
 
 		setLayers(layers);
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ResponseTimeoutLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ResponseTimeoutLayer.java
@@ -1,0 +1,60 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Institute for Pervasive Computing, ETH Zurich and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ */
+package org.eclipse.californium.core.network.stack;
+
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.eclipse.californium.core.coap.MessageObserverAdapter;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.coap.Response;
+import org.eclipse.californium.core.network.Exchange;
+
+/**
+ * Handle response timeout for request. It Should mark request as timeout if responseTimeout expired.
+ * 
+ * @see Request#getResponseTimeout()
+ */
+public class ResponseTimeoutLayer extends AbstractLayer {
+	
+	@Override
+	public void sendRequest(final Exchange exchange,final Request request) {
+		if (request.getResponseTimeout() > 0) {
+			
+			// Schedule task at timeout
+			final ScheduledFuture<?> task = executor.schedule(new Runnable() {
+				@Override
+				public void run() {
+					exchange.setTimedOut(request);
+				}
+			}, request.getResponseTimeout(), TimeUnit.MILLISECONDS);
+
+			// Cancel task if response received or request failed before.
+			request.addMessageObserver(new MessageObserverAdapter() {
+
+				@Override
+				public void onResponse(Response response) {
+					task.cancel(false);
+				}
+				
+				@Override
+				protected void failed() {
+					task.cancel(false);
+				}
+			});
+		}
+		
+		lower().sendRequest(exchange, request);
+	}
+}
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
@@ -126,6 +126,7 @@ public class ObserveServerSideTest {
 	public void setupClient() throws Exception {
 
 		client = createLockstepEndpoint(serverAddress);
+		testObsResource.clearObserveRelations();
 	}
 
 	@After
@@ -436,6 +437,40 @@ public class ObserveServerSideTest {
 
 		Thread.sleep(ACK_TIMEOUT + 100);
 		assertThat("Resource has not removed observe relation", testObsResource.getObserverCount(), is(0));
+	}
+	
+	/**
+	 * Test incomplete block2 notification (missing request)
+	 * 
+	 * @throws Exception
+	 */
+	@Test
+	public void testIncompleteBlock2Notification() throws Exception {
+		System.out.println("Observe with blockwise");
+		respPayload = generateRandomPayload(32);
+		byte[] tok = generateNextToken();
+
+		// Establish observe relation
+		respType = null; // first type is normal ACK
+		client.sendRequest(CON, GET, tok, ++mid).path(RESOURCE_PATH).observe(0).go();
+		client.expectResponse(ACK, CONTENT, tok, mid).storeObserve("A").storeETag("tag").payload(respPayload).go();
+		Assert.assertEquals("Resource has not added relation:", 1, testObsResource.getObserverCount());
+		serverInterceptor.log(System.lineSeparator() + "Observe relation established");
+
+		// First notification
+		respType = CON;
+		testObsResource.change(generateRandomPayload(80));
+		serverInterceptor.log(System.lineSeparator() + "   === changed ===");
+		client.expectResponse().type(CON).code(CONTENT).token(tok).storeMID("MID").checkObs("A", "B").storeETag("tag")
+				.block2(0, true, 32).size2(respPayload.length()).payload(respPayload, 0, 32).go();
+		client.sendEmpty(ACK).loadMID("MID").go();
+
+		// Get remaining blocks
+		byte[] tok3 = generateNextToken();
+		client.sendRequest(CON, GET, tok3, ++mid).path(RESOURCE_PATH).loadETag("tag").block2(1, false, 32).go();
+		client.expectResponse(ACK, CONTENT, tok3, mid).block2(1, true, 32).payload(respPayload, 32, 64).go();
+		// we don't send last request, @after should check is there is
+		// no leak.
 	}
 
 	// All tests are made with this resource
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
@@ -137,6 +137,9 @@ public class Request extends Message {
 	/** Contextual information about this request */
 	private Map<String, String> userContext;
 
+	/** maximum time to wait for a response to this request */
+	private long responseTimeout;
+
 	/**
 	 * Creates a request of type {@code CON} for a CoAP code.
 	 * 
@@ -729,6 +732,36 @@ public class Request extends Message {
 				notifyAll();
 			}
 		}
+	}
+
+	/**
+	 * <p>
+	 * Define a timeout value to wait the response. Zero or negative value means
+	 * no timeout.
+	 * </p>
+	 * <p>
+	 * By default no timeout is used, which means user should cancel request
+	 * manually when it will not expect response anymore. (not doing this may
+	 * cause memory leak)
+	 * </p>
+	 * <p>
+	 * The time is measured before the request is sent. So you should set a time
+	 * large enough to allow DTLS handshake or CoAP retransmission or
+	 * transparent block-wise transfer.
+	 * </p>
+	 * 
+	 * @param time maximum time to wait in milliseconds (zero or negative value
+	 *            means no timeout)
+	 */
+	public void setResponseTimeout(long time) {
+		responseTimeout = Math.max(0, time);
+	}
+	
+	/**
+	 * The maximum time to wait for a response in milliseconds (Zero means no timeout)
+	 */
+	public long getResponseTimeout() {
+		return responseTimeout;
 	}
 
 	/**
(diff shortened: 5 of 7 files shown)


boaks:
Did you check, if this PR fixes #429 ?

sbernard31:
I can check this easily :/
But all the "crappy" tests you erased :p in #418 seems to pass now.

boaks:
Sorry again for #418 :-)

sbernard31:
no problem ;)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
```diff
@@ -99,6 +99,7 @@ public CoapUdpStack(final NetworkConfig config, final Outbox outbox) {
 				new ExchangeCleanupLayer(),
 				new ObserveLayer(config),
 				new BlockwiseLayer(config),
+				new ResponseTimeoutLayer(),
```
boaks: If this is placed below the blockwise layer, I would guess, this will then apply to the "single block requests". Right?

sbernard31: That's right.

We could imagine that we do not need this for "single block request" of transparent block transfer, because there is already a response timeout for the "parent" request.

But for block2 notification, there is no "parent" request.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -248,6 +248,7 @@ private Request startBlockwiseUpload(final Exchange exchange, final Request requ
 			status = getOutboundBlock1Status(key, exchange, request);
 
 			final Request block = status.getNextRequestBlock();
+			block.setResponseTimeout(blockTimeout);
```
boaks: If that's a "single block request", why should that large timeout be used?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ResponseTimeoutLayer.java
```diff
@@ -0,0 +1,60 @@
...
+ * Handle response timeout for request. It Should mark request as timeout if responseTimeout expired.
+ * 
+ * @see Request#getResponseTimeout()
+ */
+public class ResponseTimeoutLayer extends AbstractLayer {
+	
+	@Override
+	public void sendRequest(final Exchange exchange,final Request request) {
+		if (request.getResponseTimeout() > 0) {
+
```
boaks: I'm not sure, but is this done for each single block request?

(I guess, I have to checkout the branch and do some tests with logging ...)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: What do you mean by large timeout ? you talk about the parent request timeout ? or you just found that `blocktTimeout` is a too large value ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ResponseTimeoutLayer.java
sbernard31: Yes for each single block.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: FMPOV the intention of the `blockTimeout` is different and therefore the value seems for me to be rather high.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
boaks: OK.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
```diff
@@ -437,6 +438,40 @@ public void testQuickChangeAndTimeout() throws Exception {
...
+		client.expectResponse().type(CON).code(CONTENT).token(tok).storeMID("MID").checkObs("A", "B").storeETag("tag")
+				.block2(0, true, 32).size2(respPayload.length()).payload(respPayload, 0, 32).go();
+		client.sendEmpty(ACK).loadMID("MID").go();
+
+		// Get remaining blocks
+		byte[] tok3 = generateNextToken();
+		client.sendRequest(CON, GET, tok3, ++mid).path(RESOURCE_PATH).loadETag("tag").block2(1, false, 32).go();
+		client.expectResponse(ACK, CONTENT, tok3, mid).block2(1, true, 32).payload(respPayload, 32, 64).go();
+		// we don't send last request, @after should check is there is
+		// no leak.
```
sbernard31: We don't check if exchangestore is empty in @after. I fixed it on my side but I don't "push force" it for now as I don't want to disturb @boaks review.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ResponseTimeoutLayer.java
boaks: Though "onResponse()" seems to be only called at the end of a blockwise transfere, I'm not sure if the resulting behaviour is the intended on. 

The blockwise cleanup (server side) is triggered again by each block (and cancel the old schedule)

This response timeout is set upon the first block request and on the following bock requests, it will fire for the first block request regardless of (slowly) incoming follow up blocks. And the response timeout handling for the follow up block request may fire, because they never get an onResponse.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: The default value for blockTimeout is 30s
An exchange with piggypack response is MAX_TRANSMIT_WAIT => 97s.
The [spec](https://tools.ietf.org/html/rfc7959#section-2.4) say : 
_"Clients that want to retrieve all blocks of a resource SHOULD strive to do so without undue delay. Servers can fully expect to be free to discard any cached state after a period of EXCHANGE_LIFETIME ([RFC7252], Section 4.8.2) after the last access to the state, however, there is no requirement to always keep the state for as long."_
(same thing for [block1](https://tools.ietf.org/html/rfc7959#section-2.5))
EXCHANGE_LIFETIME is about 247s

Before this PR, `blockTimeout` was about server which does not receive request in time.
Using it  for a client which does not receive response seems not so bad.
We could imagine to add a new configuration entry, but I found this a bit overkill and add complexity for users (maybe the price of flexibility...)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ResponseTimeoutLayer.java
sbernard31: > Though "onResponse()" seems to be only called at the end of a blockwise transfere, I'm not sure if the resulting behaviour is the intended on.

Good catch. For me setResponse should be called on each block request.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: > The default value for blockTimeout is 30s

I see also a value of 5 minutes in NetworkConfigDefaults.

sbernard31: I looked at the BlockwiseLayer class ... 
It's a trap for developer :p
I suppose we should set the same value for both (ideally only one constant)

We should choose between 30s and 5min (or something else).

No matter which value we choose, does it make sense to you to use the same value for missing request and missing response for transparent block transfer ?

boaks: FMPOV:
I would stick with the value of the resonseTimeout from the Request as timeout for single blocks (last pending request).

boaks: And with the 5 minutes for the state cleanup, but not for the response timeout.

sbernard31: > I would stick with the value of the resonseTimeout from the Request as timeout for single blocks (last pending request).

I didn't understand.

Could you argue why you consider time between request and response should be different than time between response and request ?

boaks: FMPOV, the blocktime out is for more a caching time on the server side. The server can't trigger the client to send the next block request. Their is no specification saying, that the next block request must be within a time.

For the reponse timeout this seems to be slightly different. Yes with a separate response there will also be no specified timeout, but as user of the API I would prefer, that the value I had provided is used.

sbernard31: > Their is no specification saying, that the next block request must be within a time.

As the spec not saying, that a response must be within a time.

>  I would prefer, that the value I had provided is used.

The API could say that the value for block response timeout is  BLOCKWISE_STATUS_LIFETIME. So the timeout will be value you provide.

But ok let's explore the other solution. Just focus on transparent block transfer.
If I well understand you prefer to always use the "parent" request timeout for each block ?
E.g. I send a GET request with response timeout set to 5minutes, each block will have a 5minutes response timeout ? that's what you mean ?

boaks: Do you want to used the "blocktime", because you usually don't want to provide the response timeout?

boaks: I ment, provide as response time at the request not as provided blocktime in the configration.

boaks: > If I well understand you prefer to always use the "parent" request timeout for each block ?

yes.

> E.g. I send a GET request with response timeout set to 5minutes, each block will have a 5minutes response timeout ? that's what you mean ?

yes, but though it is a single block request, I would recommend a smaller value (but that's up to the caller).

sbernard31: > Do you want to used the "blocktime", because you usually don't want to provide the response timeout?

I wanted to let the possibility to not set timeout to keep a kind of backward compatibility, but this is not mandatory.

Ok let's continue with this example.
User send a GET request and set its timeout to 5minutes. 
In my mind that's mean he didn't want to wait more than 5 minutes for his response.
User can not know if transparent block mode will be used. (This is up to the server)
If for each block request we reuse the same timer than the original/parent GET. This timer will never be used as the original/parent timer will alway be reached first.

In that case why using response timeout for block... but as I explained above for block notification there is no "parent/original" request. So we need to set a response timeout for each block.
So we can do a special case for this one. What could it be ?
- Using the timeout set on the observe request for each block ? maybe a bit strange. We should choose a default value if there is no timeout on this observe request.
- or, using BLOCKWISE_STATUS_LIFETIME just for notification.
- or new config parameters.

Not sure all of this will be clearer for users.
FMPOV,  this is clearer something like "all transparent block exchange should take at least LOCKWISE_STATUS_LIFETIME".

boaks: >  I wanted to let the possibility to not set timeout to keep a kind of backward compatibility, but this is not mandatory.

My question was imprecise: though the blocktime is always used as response time, even if 0 was used for the request, I thought that's your intention.

sbernard31: In the current implementation, for block request we always use BLOCKWISE_STATUS_LIFETIME even if parent/original request as a response timeout set to 0.

boaks: Assuming unintended transparent blockwise transfer together with a response time out will fail.

Either you aware of that and provide a proper value, or you will be faced too many response timeouts.

If we want a "reliable" function, then the response timeout must be the timeout of a single response, and yes , maybe changing the original expectation. But that's an issue of this transparent blockwise mode at all.

And I would go for only one scheduled timeout task and refresh that on every block, as the blockwise cleanup does also.

boaks: > in that case why using response timeout for block... but as I explained above for block notification there is no "parent/original" request. So we need to set a response timeout for each block.

When we create the requests for the followup transfer, we have the clone of the original request. We may require to serialize/persist the response timeout as well.

boaks: Sorry, I still get the strong feeling that using the blocktime is intended as always response timeout for blockwise transfer. 
If so, (because using the response time of the request may be not that easy,) I think using the 1. approach in your comment

https://github.com/eclipse/californium/pull/418#issuecomment-335461350

launch a cleaning task for blockwise exchanges would be better.

boaks: > In the current implementation, for transparent block request we always use BLOCKWISE_STATUS_LIFETIME even if parent/original request as a response timeout set to 0.

I read your comment too late :-). But with that, I think the cleaner task will do a better job.

sbernard31: why ... ? it just do the same thing without the benefits of adding an optional timeout response for all request... 

By the way, I ask before which solution you prefer and you say you have no preference... now you say you prefer the other one without any clear arguments (I mean if there is arguments I can understand but I don't see any) ... 

Sry I loose a bit my temper, but I feel this is more and more difficult to work on this project...

boaks: My answer in

https://github.com/eclipse/californium/pull/418#issuecomment-335710910

includes "Please extend such a timeout on every block, not that a large latency accidentally cleansup an ongoing blockwise exchange."

sbernard31: I do not understand go for 1. solution.

boaks:
FMPOV:

This PR targets the memory leaks, see issue #429. And in the discussion, if a timer base "garbage collector" would be a solution, I had no preference on the solution, except that the timer should be triggered again by each block (see https://github.com/eclipse/californium/pull/418#issuecomment-335710910).

My first question on this PR was, if this solves the leak! And this would be the most valuable information. Without answer to that main question, I started trying to understand the PR.

The PR introduces a response timeout, which is documented for blockwise with:

> The time is measured before the request is sent. So you should set a time large enough to allow DTLS handshake or CoAP retransmission or transparent block-wise transfer.

So it simply ignores my recommendation! Discussing that definition of a “request timeout”, you argument to define it in that "complete timeout" way (and NOT use a per single block extending definition). I'm OK with your arguments, but in my experience this approach with a fixed timeout and potential unaware “much longer” blockwise operation, doesn’t work.

Analyzing deeper, I found, that it is not implemented in the way, the javadoc pretends to. For blockwise the “response timeout” is ignored and always replaced by the “blockwise timeout”. Your also right, there are also good reasons to use that time, but for me, this seems to be more a “implementation limitation based” solution, than comply to the provided javadoc.
 
And so my sum-up was, the “response timeout” approach may cause too many surprises with all the pitfalls above. So a cleanup task implementation (close to the already implemented server side one) seems for me **NOW** (after that discussion) to be straighter forward. 

Your last argument, why not having both, is hard to follow up for me. The pitfalls for the "response timeout" will cause too frequently to fail, so I don’t see too much benefit on it.

The current implementation would also require some work according the “onResponse”, this also make me believe, that using a “cleanup task” (close to the server side solution) will cause less trouble, than the “response timeout”.

sbernard31:
About your [comment]( https://github.com/eclipse/californium/pull/418#issuecomment-335710910) you say : _"accidentally cleansup an ongoing blockwise exchange."_
For me if you define a maximum time to wait before to get your response this is not "accidentally cleansup". It's an expected clean-up.

> My first question on this PR was, if this solves the leak! And this would be the most valuable information. Without answer to that main question, I started trying to understand the PR.

I answer that I can not certify it. But I write tests which failed before and succeed with this PR. What do you want more ? 

> So it simply ignores my recommendation! 

As explain below it does not. This is more about communication problem. There is always communication problem. Not native language does not help. We should consider to try to make more simple and explicit sentence.

> I'm OK with your arguments, but in my experience this approach with a fixed timeout and potential unaware “much longer” blockwise operation, doesn’t work.

Strange that we are able to define a maximum message size for block, but it's impossible to define an approximative (optional) timeout for a message ?

> Analyzing deeper, I found, that it is not implemented in the way, the javadoc pretends to. 

... ... please re-read https://github.com/eclipse/californium/pull/437#issue-264994613 ... ....

> but for me, this seems to be more a “implementation limitation based” solution, than comply to the provided javadoc.

Of course, I try to make the best trade-off between technical constraints and usability ... if the javadoc is not well enough it can be improved. What is the point ?

> And so my sum-up was, the “response timeout” approach may cause too many surprises with all the pitfalls above.

if you set a timeout (this is optional), your request timeout after the given time. All the transparent block request should be answered in BLOCKWISE_STATUS_LIFETIME delay. Where is the surprise ?

> The pitfalls for the "response timeout" will cause too frequently to fail, so I don’t see too much benefit on it.

Too frequently seems to me clearly excessive. But just to let you know currently if you don't want leaks in californium, if there is no timeout feature, you should create your own timeout then use "request.cancel". So you will do exactly the same thing... so with the cleanup way we go to the exact same behavior.

> The current implementation would also require some work according the “onResponse”, this also make me believe, that using a “cleanup task” (close to the server side solution) will cause less trouble, than the “response timeout”.

Just adding a request.setResponse on block request which should be done even if we don't use it to be compliant with the messageObserver API...

--------------------------------------------
So I can see any really persuasive argument here, just a doubt that is possible to define a good response timeout.
As I already experience the way you do when you "don't like" something. I rethink a lot about this solution to see if it's really the good one and so if it's worst to defend it and I found a real "con" argument. The API is not respected for block1 because the "original/request" response does not go through the ResponseTimeoutLayer. I think this is a real issue.

So I will go for the cleanup task. And I will probably add a response timeout Mechanism in Leshan.

boaks:
> And I will probably add a response timeout Mechanism in Leshan.

Isn't it already there in Leshan?

sbernard31:
Just for sync send, I will add it for async send.

sbernard31:
#441 was chosen. We don't need this PR anymore

Issue #172, opened 2016-12-07
HttpTranslator hardcoded path


nxSwe:
HttpTranslator has the following member:

`public static final Properties HTTP_TRANSLATION_PROPERTIES = new MappingProperties("Proxy.properties");`
And then it automatically writes to that file when using the class. Sometimes this is not preferable as the user may not have the rights to write there.

An idea is to make the HttpTranslator methods non-static, and that the class instead had constructors/builder methods to either use the default configuration or to point out a path/filename where to get/write the file.

sophokles73:
Sounds good :-) Would you care to provide a PR?

boaks:
Your PR is merge, so this should be fixed.

PR #438, opened 2017-10-12
Make MappingProperties file configurable.

nxSwe:
Issue #172.
Made HttpTranslator methods nonstatic and added constructors.

Signed-off-by: Björn Elmers <bjorn.elmers@nexusgroup.com>

### DIFF
--- a/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpStack.java
+++ b/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpStack.java
@@ -313,7 +313,7 @@ public class HttpStack {
 
 			try {
 				// translate the coap response in an http response
-				HttpTranslator.getHttpResponse(httpRequest, coapResponse, httpResponse);
+				new HttpTranslator().getHttpResponse(httpRequest, coapResponse, httpResponse);
 
 //				if (Bench_Help.DO_LOG) 
 					LOGGER.finer("Outgoing http response: " + httpResponse.getStatusLine());
@@ -460,7 +460,7 @@ public class HttpStack {
 
 				try {
 					// translate the request in a valid coap request
-					Request coapRequest = HttpTranslator.getCoapRequest(httpRequest, localResource, proxyingEnabled);
+					Request coapRequest = new HttpTranslator().getCoapRequest(httpRequest, localResource, proxyingEnabled);
 //					if (Bench_Help.DO_LOG) 
 						LOGGER.info("Received HTTP request and translate to "+coapRequest);
 
--- a/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpTranslator.java
+++ b/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpTranslator.java
@@ -92,7 +92,8 @@ public final class HttpTranslator {
 	 * Property file containing the mappings between coap messages and http
 	 * messages.
 	 */
-	public static final Properties HTTP_TRANSLATION_PROPERTIES = new MappingProperties("Proxy.properties");
+	public static final MappingProperties DEFAULT_HTTP_TRANSLATION_PROPERTIES = new MappingProperties("Proxy.properties");
+	private Properties httpTranslationProperties;
 
 	// Error constants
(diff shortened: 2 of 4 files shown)


californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpStack.java
```diff
@@ -313,7 +313,7 @@ public void run() {
 
 			try {
 				// translate the coap response in an http response
-				HttpTranslator.getHttpResponse(httpRequest, coapResponse, httpResponse);
+				new HttpTranslator().getHttpResponse(httpRequest, coapResponse, httpResponse);
```
boaks: Maybe I wrong, but doesn't `new HttpTranslator()` always load the properties file?

nxSwe: Yes that's true. Do you think we should avoid that? 

Do you have any suggestion of how to make it possible to use another properties file without always loading the properties file?

boaks: I'm not sure, if this would downgrade the performance, but maybe that's academical.

A alternative approach would be check the File.lastModified() in intervals and load it, when that changed, and replace the used properties than by the complete read ones.

nxSwe: WIth that approach, it will still not be possible to use another file than the default Proxy.properties. Right? And at least from my perspective that is the objective.

boaks: >it will still not be possible to use another file than the default Proxy.properties, right?

Do you mean, changing that filename while running? Can you point me to the part of your PR which did this?

As fas, as I understood you PR, you just load the same filename over and over.

nxSwe: Yes that is correct. That is also what is described by issue #172.

The HttpTranslator has 2 constructors. One that uses the default Proxy.properties file, and one that has the filename as a parameter.

boaks: So, where do you want to trigger such filename change?

nxSwe: In my CoAP proxy implementation. Currently all my endpoints has their own content type logic. MappingProperties could take care of this for me if the HttpTranslator was able to use another file than the hardcoded one.

boaks: OK, so therefore I can't see, how it is used. 

I see two ways to continue:
- performance test. If there is no penalty in loading the same file over and over, OK
- change the code in a way, that other users are not forced to use your "always loading" solution. Maybe adding a replaceable "HttpTranslator" Factory with a reusing default implementation and you can replace that with your reloading implementation in your proxy implementation.

nxSwe: What do you think of a solution where the default HttpTranslator constructor uses MappingProperties.std instead of creating a new one? That would mean that ProxyMapping.properties would be used instead of Proxy.properties.

That would mean changing the default constructor to:
	public HttpTranslator() {
		httpTranslationProperties = MappingProperties.std;
	}

In the current implementation I do not understand why both ProxyMapping.properties and Proxy.properties exists when they look the same.

What do you think?

boaks: Sounds good.
I would prefer to remove the "MappingProperties.std" (I will do that later, so just leave it currently as it is) and instead use a additional static properties in the HttpTranslator and reuse them as httpTranslationProperties when using the default constructor HttpTranslator().

nxSwe: Great! I've pushed a new commit. Do you want me to squash the commits or do you take care of it at merge?

boaks: LGTM.
I would enjoy it, if you squash :-).

boaks:
Thanks for your work and patience.

PR #441, opened 2017-10-17
Clean up transparent block exchange if no response received

### LINKED ISSUE #429: Memory leak in MessageExchangeStore

sbernard31:
This aims to fix #429 (at least partially)
tests of #418 are also rebased on this branch. They could be used for more "checking". (#418 is not intended to be merged)

See #437 for more details about this choice of implementation.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
@@ -48,6 +48,7 @@ final class Block1BlockwiseStatus extends BlockwiseStatus {
 	static Block1BlockwiseStatus forOutboundRequest(final Exchange exchange, final Request request, final int preferredBlockSize) {
 		Block1BlockwiseStatus status = new Block1BlockwiseStatus(0, request.getOptions().getContentFormat());
 		status.request = request;
+		status.exchange = exchange;
 		status.setCurrentSzx(BlockOption.size2Szx(preferredBlockSize));
 		return status;
 	}
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
@@ -110,7 +110,8 @@ public class BlockwiseClientSideTest {
 				.setInt(NetworkConfig.Keys.ACK_TIMEOUT, ACK_TIMEOUT_IN_MS)
 				.setInt(NetworkConfig.Keys.ACK_RANDOM_FACTOR, 1)
 				.setInt(NetworkConfig.Keys.MAX_RETRANSMIT, 2)
-				.setInt(NetworkConfig.Keys.ACK_TIMEOUT_SCALE, 1);
+				.setInt(NetworkConfig.Keys.ACK_TIMEOUT_SCALE, 1)
+				.setInt(NetworkConfig.Keys.BLOCKWISE_STATUS_LIFETIME,300);
 	}
 
 	@Before
@@ -345,7 +346,7 @@ public class BlockwiseClientSideTest {
 		//server.sendResponse(ACK, CONTENT).loadBoth("B").block2(1, true, 64).payload(respPayload, 64, 128).go();
 		// give client a chance to repeat
 		int timeout = config.getInt(NetworkConfig.Keys.ACK_TIMEOUT, ACK_TIMEOUT_IN_MS);
-		Thread.sleep(timeout * 2);
+		Thread.sleep((long) (timeout*1.25));
 		// repeat GET 1
 		server.expectRequest(CON, GET, path).sameBoth("B").block2(1, false, 64).go();
 		server.sendResponse(ACK, CONTENT).loadBoth("B").block2(1, true, 64).payload(respPayload, 64, 128).go();
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
@@ -109,7 +109,8 @@ public class BlockwiseClientSideTest {
 				.setLong(NetworkConfig.Keys.EXCHANGE_LIFETIME, TEST_EXCHANGE_LIFETIME)
 				.setInt(NetworkConfig.Keys.ACK_TIMEOUT, ACK_TIMEOUT_IN_MS)
 				.setInt(NetworkConfig.Keys.ACK_RANDOM_FACTOR, 1)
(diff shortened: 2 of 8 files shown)


californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
```diff
@@ -84,9 +84,9 @@
 	private static final int MAX_RESOURCE_BODY_SIZE = 1024;
 	private static final int RESPONSE_TIMEOUT_IN_MS = 1000;
 	private static final int ERROR_TIMEOUT_IN_MS = 500;
-	// client retransmits after 200 ms
```
boaks: Any reason, why the comment is changed?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
```diff
@@ -437,6 +451,40 @@ public void testQuickChangeAndTimeout() throws Exception {
...
+	public void testIncompleteBlock2Notification() throws Exception {
+		System.out.println("Observe with blockwise");
+		respPayload = generateRandomPayload(32);
+		byte[] tok = generateNextToken();
+
+		// Establish observe relation
+		respType = null; // first type is normal ACK
+		client.sendRequest(CON, GET, tok, ++mid).path(RESOURCE_PATH).observe(0).go();
+		client.expectResponse(ACK, CONTENT, tok, mid).storeObserve("A").storeETag("tag").payload(respPayload).go();
+		Assert.assertEquals("Resource has not added relation:", 1, testObsResource.getObserverCount());
```
boaks: I'm not sure, if we want to use a more uniform way, but if, I would have expected 

`assertThat("Resource has not added relation:", testObsResource.getObserverCount(), is(1)); `

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
```diff
@@ -32,6 +32,7 @@
 final class Block1BlockwiseStatus extends BlockwiseStatus {
 
 	private Request request;
+	private Exchange exchange;
```
boaks: Any reason, why the common stuff (`exchange` and `timeoutCurrentTranfer`) of `Block1BlockwiseStatus` and `Block2BlockwiseStatus` is not moved to the base class `BlockwiseStatus`?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
```diff
@@ -493,7 +494,7 @@ public void testBlockwiseObserveNotInterruptedByOdlerNotificationWithoutBlockwis
 		server.sendResponse(CON, CONTENT).loadToken("OBS").observe(0).mid(++mid).payload(notifyPayload).go();
 		server.expectEmpty(ACK, mid).go();
 		// Check this one is discard.
-		Response response = request.waitForResponse(1000);
+		Response response = request.waitForResponse(150);
```
boaks: My experience with "small" timeouts and a "undeterministic" CPU allocation on a shared Buildserver are not that good.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
sbernard31: I will move it in `BlockwiseStatus`.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
sbernard31: I know what do you propose ?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
sbernard31: All the tests in this class is written using :
Assert.assertEquals("Resource has not added relation:", 1, testObsResource.getObserverCount());

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
sbernard31: Just a mistake.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveServerSideTest.java
boaks: OK.

(Except of 2, which use the assertThat. Rise the question, which one should be used for future tests. But that's sure out of scope this PR.)

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
boaks: keep the 1000 (wasting some maven build time :-) ).

FMPOV, the check, if a response is accidentally received with a timeout, which seems to be too short for "shared build system", would not really check/ensure that.

sbernard31:
I would like to use the same value for default `BLOCKWISE_STATUS_LIFETIME` in `BlockwiseLayer` and `NetworkConfigDefaults`. 5min, 30s or something else ?

boaks:
I would prefer the 5min. 30s seems for me too short.

sbernard31:
I remember that 5 minutes seemed too high for you, but ok for 5min.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
sbernard31: That means increase the `blockwise_life_status` to more than 1s and so made all the test more than 1s longer this is really want you want ?

boaks: Yes.

Alternatively,  I (or you) creates a new branch and configure the unit test to execute that test repeated (using the RepeatingTestRunner) and set the "hudson_debug" job on it. So we will see, if that timing may cause an issue.

sbernard31:
Is it better now ?

boaks:
LGTM.

PR #443, opened 2017-10-18
Add license in markdown.

boaks:
The license are already provided as html, but github community web page
seems to prefer markdown.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,230 @@
+# Dual License:
+
+- Eclipse Distribution License - v 1.0
+https://www.eclipse.org/org/documents/edl-v10.php
+
+or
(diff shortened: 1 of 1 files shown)


sophokles73:
Hm, looks kind of repetitious and redundant, doesn't it? 
I would rather have *either* LICENSE *or* the two html files bot probably not both. However, if we go for LICENSE then it should probably be plain text and not markdown. WDYT?

boaks:
The markdown was initiated by your e-mail, saying 

"In the end this just seems to result in a LICENSE.md file in the root folder"

Therefore I used markdown. But sure, we can use plain text.
And yes, if that LICENSE is agreed, we can remove the html files.

boaks:
Done as plain text.

PR #446, opened 2017-10-18
Introduce CredentialsUtil for simplified credentials management.

boaks:
Make certificates optional.

### DIFF
--- a/demo-apps/cf-secure/src/main/java/org/eclipse/californium/examples/CredentialsUtil.java
+++ b/demo-apps/cf-secure/src/main/java/org/eclipse/californium/examples/CredentialsUtil.java
@@ -1,0 +1,180 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial implementation
+ ******************************************************************************/
+
+package org.eclipse.californium.examples;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.security.cert.Certificate;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.eclipse.californium.elements.util.SslContextUtil;
+import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
+import org.eclipse.californium.scandium.dtls.pskstore.InMemoryPskStore;
+
+/**
+ * Credentials utility for setup DTLS credentials.
+ */
+public class CredentialsUtil {
+
+	/**
+	 * Credentials mode.
+	 */
+	public enum Mode {
+		/**
+		 * Preshared secret keys
+		 */
+		PSK,
+		/**
+		 * Raw public key certificates
+		 */
+		RPK,
+		/**
+		 * X.509 certificates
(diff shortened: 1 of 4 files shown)


demo-apps/cf-secure/src/main/java/org/eclipse/californium/examples/CredentialsUtil.java
```diff
@@ -0,0 +1,203 @@
...
+		if (psk && config.getIncompleteConfig().getPskStore() == null) {
+			// Pre-shared secret keys
+			InMemoryPskStore pskStore = new InMemoryPskStore();
+			pskStore.setKey(PSK_IDENTITY, PSK_SECRET);
+			config.setPskStore(pskStore);
+		}
+		int x509 = indexOf(Mode.X509, modes);
+		int rpk = indexOf(Mode.RPK, modes);
+
+		if (0 <= x509 || 0 <= rpk) {
```
sbernard31: I do not understand why you always put constant in first.
I know this is logically  the same but it seems more readable to put the variable in first. 
I mean you want check if `x509` greater or equals to `0`, not if `0` is lesser or equals than `x509` ...
Almost all the code was written like that and if I remember well Kai already ask you for that.

demo-apps/cf-secure/src/main/java/org/eclipse/californium/examples/CredentialsUtil.java
```diff
@@ -0,0 +1,203 @@
...
+	 * 
+	 * @param mode mode to check for
+	 * @param modes modes to check
+	 * @return {@code true}, if mode is contained in modes, or {@code false}, if
+	 *         not
+	 * @throws NullPointerException if one of the provided parameters are
+	 *             {@code null}
+	 */
+	public static boolean contains(Mode mode, Mode... modes) {
+		return 0 <= indexOf(mode, modes);
```
sbernard31: why not using Arrays.asList().containt()

demo-apps/cf-secure/src/main/java/org/eclipse/californium/examples/CredentialsUtil.java
```diff
@@ -0,0 +1,203 @@
...
+	 * 
+	 * @param config DTLS configuration builder. May be already initialized with
+	 *            PskStore.
+	 * @param certificateAlias alias for certificate to load as credentials.
+	 * @param modes list of supported mode. If a RPK is in the list before X509,
+	 *            the RPK is setup as preferred
+	 */
+	public static void setupCredentials(DtlsConnectorConfig.Builder config, String certificateAlias, Mode... modes) {
+
+		boolean psk = 0 <= indexOf(Mode.PSK, modes);
```
sbernard31: why not using Arrays.asList().indexOf

demo-apps/cf-secure/src/main/java/org/eclipse/californium/examples/CredentialsUtil.java
boaks: That's a habit from coding C with MISRA :-).

sophokles73: I have to agree with @sbernard31, I would also very much prefer if you could drop this habit of yours from the *good old days* (at least while working on Cf) ;-)

demo-apps/cf-secure/src/main/java/org/eclipse/californium/examples/CredentialsUtil.java
```diff
@@ -0,0 +1,180 @@
...
+	/**
+	 * Parse arguments to modes.
+	 * 
+	 * @param args arguments
+	 * @return array of modes. if the arguments are empty,
+	 *         {@link #DEFAULT_MODES} is returned.
+	 */
+	public static List<Mode> parse(String[] args) {
+		int len = args.length;
+		if (0 == len) {
```
sbernard31: Still remains a MISRA's memory :) (this is one is less confusing)

I think the use of  `len` variable is not really needed. 
Unless you eventually reuse it below in `new ArrayList<>(len)`, personnaly I prefer to use args.length in the first `if`

PR #448, opened 2017-10-20
Added code to verify raw public keys during the handshake

### LINKED ISSUE #301: Add method to Scandium for setting trusted RPKs
When doing a raw-public-key handshake according to RFC 7250 the raw public key of the other peer is said to be verified out-of-band. 

I would like to have a feature in Scandium where I can add a list of trusted raw public keys, from which I will accept DTLS connections (and reject others).

LudwigSeitz:
Fixes #301 

Signed-off-by: Ludwig Seitz <ludwig.seitz@ri.se>

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/InMemoryRpkTrustStore.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/InMemoryRpkTrustStore.java
@@ -1,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2017 RISE SICS.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Ludwig Seitz (RISE SICS) - initial creation
+ ******************************************************************************/
+package org.eclipse.californium.scandium.auth;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * A raw public key store that stores the trusted keys in memory.
+ * 
+ * @author Ludwig Seitz
+ *
+ */
+public class InMemoryRpkTrustStore implements TrustedRpkStore {
+
+	private final Set<RawPublicKeyIdentity> trustedRPKs;
+
+	/**
+	 * Constructor.
+	 * 
+	 * @param trustedRPKS the list of raw public key identities that are trusted
+	 */
+	public InMemoryRpkTrustStore(Set<RawPublicKeyIdentity> trustedRPKS) {
+		this.trustedRPKs = new HashSet<>();
+		this.trustedRPKs.addAll(trustedRPKS);
+	}
+
+	@Override
+	public boolean isTrusted(RawPublicKeyIdentity id) {
+		return this.trustedRPKs.contains(id);
+	}
+
+}
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -18,6 +18,7 @@
  *                                               configuration
  *    Kai Hudalla (Bosch Software Innovations GmbH) - fix bug 483559
  *    Achim Kraus (Bosch Software Innovations GmbH) - add enable address reuse
+ *    Ludwig Seitz (RISE SICS) - Added support for raw public key validation
  *******************************************************************************/
 
 package org.eclipse.californium.scandium.config;
@@ -31,6 +32,8 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
+import org.eclipse.californium.scandium.auth.TrustAllRpks;
+import org.eclipse.californium.scandium.auth.TrustedRpkStore;
 import org.eclipse.californium.scandium.dtls.ServerNameResolver;
 import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
 import org.eclipse.californium.scandium.dtls.pskstore.PskStore;
@@ -104,6 +107,9 @@ public final class DtlsConnectorConfig {
 
 	/** the supported cipher suites in order of preference */
 	private CipherSuite[] supportedCipherSuites;
+	
+	/** default is trust all RPKs **/
+	private TrustedRpkStore trustedRPKs = new TrustAllRpks();
 
 	private int outboundMessageBufferSize = 100000;
 
@@ -317,6 +323,14 @@ public final class DtlsConnectorConfig {
 	 */
 	public long getStaleConnectionThreshold() {
 		return staleConnectionThreshold;
+	}
+	
+	/**
+	 * @return The trust store for raw public keys verified out-of-band for
+	 *         DTLS-RPK handshakes
+	 */
+	public TrustedRpkStore getRpkTrustStore() {
+		return trustedRPKs;
 	}
 
 	/**
@@ -816,6 +830,19 @@ public final class DtlsConnectorConfig {
 			}
 
 			config.supportedCipherSuites = ciphers.toArray(new CipherSuite[0]);
+		}
+	
+
+		/**
+		 * Sets the store for trusted raw public keys.
+		 * 
+		 * @param store the trust store
+		 */
+		public void setRpkTrustStore(TrustedRpkStore store) {
+			if (store == null) {
+				throw new IllegalStateException("Must provide a non-null trust store");
+			}
+			config.trustedRPKs = store;
 		}
 	}
 }
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateMessage.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateMessage.java
@@ -18,6 +18,7 @@
  *    Kai Hudalla (Bosch Software Innovations GmbH) - add accessor for peer address
  *    Kai Hudalla (Bosch Software Innovations GmbH) - improve handling of empty messages
  *    Kai Hudalla (Bosch Software Innovations GmbH) - fix 477074 (erroneous encoding of RPK)
+ *    Ludwig Seitz (RISE SICS) - Moved certificate validation to Handhaker
  ******************************************************************************/
 package org.eclipse.californium.scandium.dtls;
 
@@ -27,21 +28,16 @@ import java.security.GeneralSecurityException;
 import java.security.KeyFactory;
 import java.security.PublicKey;
 import java.security.cert.CertPath;
-import java.security.cert.CertPathValidator;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
-import java.security.cert.PKIXParameters;
-import java.security.cert.TrustAnchor;
 import java.security.cert.X509Certificate;
 import java.security.spec.EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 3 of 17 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -104,6 +107,8 @@
 
 	/** the supported cipher suites in order of preference */
 	private CipherSuite[] supportedCipherSuites;
+	
+	private Set<RawPublicKeyIdentity> trustedRPKs;
```
boaks: Using a `Set<RawPublicKeyIdentity>` seems to be much less flexible than the interface `PskStore`, which may be either implemented with a simple "InMemory" implementation or something else (as long , as it is fast enough).

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -817,5 +832,15 @@ private void determineCipherSuitesFromConfig() {
...
 		}
+	
+
+		public void setTrustedRPKs(Set<RawPublicKeyIdentity> rpks) {
+		    if (rpks == null || rpks.isEmpty()) {
+		        throw new IllegalStateException("Must provide a non-null, non-empty"
+		                + " set of trusted raw public keys"); 
+		    }
+		    config.trustedRPKs = new HashSet<>();
+		    config.trustedRPKs.addAll(rpks);
```
boaks: If this is intended to be "unmodifiable", please use `Collections.unmodifiableSet()`. Then all consumer may just use it without cloning the outer `Set`.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
sbernard31: +1, I also think adding an interface will be better. Maybe with a default implementation which trust anyone. (isTrusted return always true)

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: FMPOV: a interface would be really a better solution. Then `Collections.unmodifiableSet()` would be obsolete.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateMessage.java
```diff
@@ -286,7 +287,8 @@ public CertPath getCertificateChain() {
 	 * 
 	 * @throws HandshakeException if any of the checks fails
 	 */
-	public void verifyCertificate(X509Certificate[] trustedCertificates) throws HandshakeException {
+	public void verifyCertificate(X509Certificate[] trustedCertificates,
```
boaks: OK, I see the "collection/array" solution was also chosen for x.509.
Maybe, we add a "common" trust interface with two different methods for x509 and rpk.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateMessage.java
```diff
@@ -308,6 +310,17 @@ public void verifyCertificate(X509Certificate[] trustedCertificates) throws Hand
 				AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.BAD_CERTIFICATE, getPeer());
 				throw new HandshakeException("Certificate chain could not be validated", alert);
 			}
+		} if (rawPublicKeyBytes != null) {
+		    RawPublicKeyIdentity rpk = new RawPublicKeyIdentity(getPublicKey());
+		    for (RawPublicKeyIdentity trusted : trustedRPKs) {
```
boaks: Did you check, if "contains" works? If the RPKs are already included in a Set, this may have a chance to work.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateMessage.java
sbernard31: I didn't think to this so much but intuitively I will go for separate interface for RPK and x509.
With maybe a default class which implement both.

LudwigSeitz:
Thank you for your comments. I will unfortunately not be able to do much more work on this before November.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateMessage.java
boaks: FMPOV: there are two places, where the one or two interfaces occurs:
- the implementation, but if this implements both, it seems quite similar
- the  verifyCertificate(X509validator, RPKvalidator) then be called with the same "multi validator" for both arguments.

for the first, I agree, but for the second, where we the pass in the same twice, I'm not sure, if the separate interface is more intuitive.

boaks: For two interfaces:
Maybe we better move the verifyCertificate from CertificateMessage to the Handshaker, which should then use the right interface to validate the certPath or the RPK.

sbernard31: Thinking a bit more about that and as we have a specific interface for psk, it make sense to me to have 2 others specific interfaces for RPK and x509.

Finally, a default class which implements both interface make no so much sense.
I suppose the default one for RPK will be "trust-all" and no default one for certificates.
Both could have a static implementations which take list of arrays/list of trusted elements (rpk/certificate).
Builder.setTrustStore() could be a shortcut to create this static store for certificate.

boaks: As I wrote: going for two interfaces, I would move the verifyCertificate from CertificateMessage to the Handshaker.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
LudwigSeitz: Ok. I'm thinking of placing this in the auth package, or do you think it should go somewhere else?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateMessage.java
LudwigSeitz: How do you suggest to transfer the necessary parameters from CertificateMessage?
Currently verifyCertificate() depends on:
1.  CertificateMessage.certPath
2.  AbstractMessage.getPeer()
3. CertificateMessage. rawPublicKeyBytes
4. CertificateMessage.getPublicKey()

boaks: 1. `CertificateMessage.getCertificateChain()` returns the `certPath`, if available.
2. the peer seems only be required for alert message. The Handshaker already creates such alert messages using the peer from the session, see `Handshaker.wrapMessage`
3. if `CertificateMessage.getCertificateChain()` returns `null`, `CertificateMessage.getPublicKey()` will return also `null`, if `rawPublicKeyBytes` is not set. So only 
4. `CertificateMessage.getPublicKey()` is required.

FMPOV, if we use two interfaces for validating the different certificate types, it seems for me to be not that natural to provide both to one method. And the stuff inside that `CertificateMessage.verifyCertificate()` doesn't depend on too much internals. Therefore I would prefer to move it into the Handshaker. But that's just my opinion.

LudwigSeitz: Ok I've tried that option out and codewise it works. However CertificateMessageTest is seriously broken by that change and needs to be reworked in concert with HandshakerTest

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateMessage.java
LudwigSeitz: Yes that should work (since RawPublicKeyIdentity implments equals()).

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/InMemoryRpkTrustStore.java
```diff
@@ -0,0 +1,20 @@
+package org.eclipse.californium.scandium.auth;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class InMemoryRpkTrustStore implements TrustedRpkStore {
+
+    private Set<RawPublicKeyIdentity> trustedRPKs;
```
boaks: consider to use final

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -120,8 +121,10 @@
 	protected final ServerNameResolver serverNameResolver;
 	protected ServerNames indicatedServerNames;
 	protected SignatureAndHashAlgorithm negotiatedSignatureAndHashAlgorithm;
-
-
+ 
+    /** The trusted raw public keys */
+    protected TrustedRpkStore rpkStore;
```
boaks: If that is moved to the base class `Handshaker`, `verifyCertificate` would only require the `CertificateMessage` as parameter (the `rootCertificates` are already there).

boaks: And consider to use final.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -910,4 +916,66 @@ final boolean isChangeCipherSpecMessageExpected() {
...
+     * 
+     * @param trustedCertificates the list of trusted root CAs
+	 * @param message 
+     * 
+     * @throws HandshakeException if any of the checks fails
+     */
+    public void verifyCertificate(X509Certificate[] trustedCertificates, 
+            TrustedRpkStore rpkStore, CertificateMessage message) 
+                    throws HandshakeException {
+        if (message.getCertificateChain() != null) {
```
boaks: If `rpkStore` is moved to `Handshaker`, the parameter list could be reduced to the`CertificateMessage`.

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/InMemoryRpkTrustStore.java
LudwigSeitz: ok

LudwigSeitz: new version pushed

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
LudwigSeitz: new version pushed

boaks:
The implementation looks now good for me :-).

If you use the eclipse IDE, would it be possible, that you apply the "eclipse-formatter-profile.xml" to your changes using source->format element (please not the complete files, this may change much more :-) ).
And could you rebase and squash your work?

LudwigSeitz:
Will do tomorrow.

boaks:
One more wish. Could you add a license header?

> /*******************************************************************************
>  * Copyright (c) 2017 "person or company" and others.
>  *
>  * All rights reserved. This program and the accompanying materials
>  * are made available under the terms of the Eclipse Public License v1.0
>  * and Eclipse Distribution License v1.0 which accompany this distribution.
>  *
>  * The Eclipse Public License is available at
>  *    http://www.eclipse.org/legal/epl-v10.html
>  * and the Eclipse Distribution License is available at
>  *    http://www.eclipse.org/org/documents/edl-v10.html.
>  *
>  * Contributors:
>  *    person (company) - initial creation
>  ******************************************************************************/
>  

Replace "person" and "company" with what ever fits for you.
Sometimes I add also comments to that header, if I change a file. But this is not required.

LudwigSeitz:
For the formatter profile, I've got Eclipse and Eclipse 2.1 which one should I use?

boaks:
There is a "eclipse-formatter-profile.xml" in the parent (californium root). Please import that into the Eclipse IDE, and use it then in the californium projects (e.g. scandium) as project specific formatter..

boaks:
Thanks! 

LGTM. After merging I would like to move your new RpkStores from `org.eclipse.californium.scandium.auth` to `org.eclipse.californium.scandium.dtls.rpkstore` (similar to the psk-store), if you have no objection.

LudwigSeitz:
Do move to any place that seems reasonable. Can you check if the squash and rebase worked as expected? It's the first time I did that.

sbernard31:
Good improvement.

(Ok to move to rpkstore package)

boaks:
>  Can you check if the squash and rebase worked as expected? It's the first time I did that.

I would have update the master (pull from origin repo), before rebasing.
Do you still want to spend time in rebase it again with an update master?
The idea of that would be, that you also check, that the unit tests are still running.
The "no conflicts" above is only about textual changes, not about the logic :-)

LudwigSeitz:
I did check `HandshakerTest` and `CertificateMessageTest` and they still work, but I agree that it would be good to run all tests to see if any unexpected side-effects show.

boaks:
The parent commit of your branch is 803b03bc9193134e3d289abce8f55cdb662c4acd
But the current master changed to 4b52a3daf7fd1869f1fd68886151e0ad5ad8756d

I have no doubt, that the tests are still good. 
But the idea of update the master and rebase your branch to the updated master is to ensure,
that the tests are actually working.

LudwigSeitz:
Was that what you suggested?

boaks:
You rebased the master on your "rpkVerify" branch.

boaks:
Or you merged the master on your branch. Anyway, I will sort it out.

boaks:
I rebased your commit 

https://github.com/eclipse/californium/pull/448/commits/293d1e98e625f35b045154c84c7277e89669c36b

to master 4b52a3daf7fd1869f1fd68886151e0ad5ad8756d

and pushed it to the master. So this PR is merged, even if it seems to be just closed.

Issue #439, opened 2017-10-13
Make layer stack of CoAP endpoint customizable


sbernard31:
My first need is to have access to BlockwiseLayer to check memory leak.
But I think this could be useful to be able to customize the coapstack of an endpoint.

I imagine to add protected methods  `createUdpStack` and `createTcpStack` which could be overridden to use a custom stack. Why not a parameters to constructors ? It's possible but we begin to have lot of parameters in constructors and a lot of constructors in CoapEndpoint.

The second point is to make easier to choose layer implementation in udp/tcp stack.
Two solutions :
1) add also protected "create" method for each layer
2) just pass each layer as arguments (null == default one)

This is optional because without that, it is still possible to create your own stack with your own layers  but I think you will often change only one layer.

Any opinion ?

boaks:
FMPOV we currently debugging and want to check different ideas. For that it seems a good deal. 
But I'm not sure, if I understand, how you intend to do it.

sbernard31:
For the first point, something like this : 

```java
CoapEndpoint ep = new CoAPEndpoint() {
       protected CoapStack createUdpStack(NetworkConfig config, Outbox outbox){
               return new MyCustomUdpStack(config,outbox);
       }
}
```

boaks:
Maybe, this feature could be added with issue #198 
(Or at least a clean up of the constructors :-) ).

sbernard31:
If we add a builder, we will need something like a CoapStackFactory as coapstack constructor need some arguments (config and outbox). Maybe somethink like we do in https://github.com/eclipse/leshan/issues/372#issuecomment-325592364 ?

(I hope the fact that the leshan example use coapendpoint is not so confusing :/)

danielhqv:
This sounds exactly like what I'm interested in as well, but in my case it's for simulation purposes on the client. I would like to inject my own custom BlockwiseLayer to be able to simulate block transfer latency and other unexpected behavior in a controlled way.

boaks:
That feature is already merged to 2.0.x (see PR #449, commit 5bf7e90e3da839e4758b1235ad721e502fd3258f in 2.0.x)

@sbernard31 
Is it possible, that you close this issue?

boaks:
https://github.com/eclipse/californium/issues/439#issuecomment-340249511

The CoAP stack is, as far as I understand it, not intended to delay message processing. I'm not sure, if that will cause more issues than expected.
[...]

PR #449, opened 2017-10-20
Enhance tests about BlockwiseLayer 

sbernard31:
The main idea is to be able to check if blockwiseLayer is empty or not.

This allow me to find more issue in BlockwiseLayer and fix it.

First commit is about adding a way to create custom coap stack ( see #439). A Builder was not implemented because I'm currently focus on fixing leak and custom coap stack is just a "bonus" which help me to achieve this goal).

I advice to review this PR commit by commit.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
@@ -87,7 +87,28 @@ public class CoapUdpStack extends BaseCoapStack {
 	 */
 	public CoapUdpStack(final NetworkConfig config, final Outbox outbox) {
 		super(outbox);
+		Layer layers[] = new Layer[] {
+				createExchangeCleanupLayer(config),
+				createObserveLayer(config),
+				createBlockwiseLayer(config),
+				createReliabilityLayer(config)};
 
+		setLayers(layers);
+	}
+
+	protected Layer createExchangeCleanupLayer(NetworkConfig config) {
+		return new ExchangeCleanupLayer();
+	}
+
+	protected Layer createObserveLayer(NetworkConfig config) {
+		return new ObserveLayer(config);
+	}
+
+	protected Layer createBlockwiseLayer(NetworkConfig config) {
+		return new BlockwiseLayer(config);
+	}
+
+	protected Layer createReliabilityLayer(NetworkConfig config) {
 		ReliabilityLayer reliabilityLayer;
 		if (config.getBoolean(NetworkConfig.Keys.USE_CONGESTION_CONTROL) == true) {
 			reliabilityLayer = CongestionControlLayer.newImplementation(config);
@@ -95,15 +116,6 @@ public class CoapUdpStack extends BaseCoapStack {
 		} else {
 			reliabilityLayer = new ReliabilityLayer(config);
 		}
-
-		Layer layers[] = new Layer[] {
-				new ExchangeCleanupLayer(),
-				new ObserveLayer(config),
-				new BlockwiseLayer(config),
-				reliabilityLayer };
-
-		setLayers(layers);
-
-		// make sure the endpoint sets a MessageDeliverer
+		return reliabilityLayer;
 	}
 }
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
@@ -408,7 +408,7 @@ public class Exchange {
 		// forward timeout to message
 		message.setTimedOut(true);
 		Request request = this.request;
-		if (currentRequest == message && request != message) {
+		if (request != null && currentRequest == message && request != message  ) {
 			// forward timeout to request
 			request.setTimedOut(true);
 		}
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
@@ -68,6 +68,7 @@ final class Block1BlockwiseStatus extends BlockwiseStatus {
 			bufferSize = block.getOptions().getSize1();
 		}
 		Block1BlockwiseStatus status = new Block1BlockwiseStatus(bufferSize, contentFormat);
+		status.exchange = exchange;
 		status.setFirst(block);
 		return status;
 	}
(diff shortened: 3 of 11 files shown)


boaks:
I would like to propose to add the first commit (CoapEndpoint) fast, so that a clean up of the CoapEndpoint constructors is possible.

The exposing of the layers is something, I don't think it is a good solution. I would prefer a "isEmpty()" for the and the layers, with the stack accumulating the isEmpty over the layers. But it's a discussion pretty close to the "isEmptyWithDump()" #400 and #326. Either add "special test code" or "expose internal structure" for testing.

For some other changes this seems, that the already implemented calling sequences through matcher, exchange and message, are pretty complex. I'm not sure, if this PR adds some more complexity or reduce that. To be frank, my hope was, that the "blockwise clean up task" will make the clean up of the calling sequences unnecessary. I still hope for that :-). 
After some other opinion about the exposing of the layers, I can't say too much about the other fixes. If they help to fix the leaks, go for it (I have no better idea).

sbernard31:
> I would like to propose to add the first commit (CoapEndpoint) fast, so that a clean up of the CoapEndpoint constructors is possible.

Please do it (fast forward of 2.0.x should do the job)

> The exposing of the layers is something, I don't think it is a good solution. I would prefer a "isEmpty()" for the and the layers, with the stack accumulating the isEmpty over the layers. But it's a discussion pretty close to the "isEmptyWithDump()" #400 and #326. Either add "special test code" or "expose internal structure" for testing.

IsEmpty does not allow to check by layer... If you really don't want to expose this. I will go for create a specific udp stack for testing. But not sure this will make the code clearer.

> For some other changes this seems, that the already implemented calling sequences through matcher, exchange and message, are pretty complex.

I agree, I really dislike this complexity. 
There is so many strange stuff like : 
 - request vs currentrequest
 - redundant state between request and exchange (like timeout)
 - the "exchange complete" concept
 - request event which could change exchange state, changing exchange state which raise request event.

> I'm not sure, if this PR adds some more complexity or reduce that

This PR change almost nothing about complexity. See changes on real code, this is almost nothing ... This is mainly test enhancing ! 

I try to focus on tests because : 
- it ensure the code works as we expect (even if it is not as elegant as we wish) 
- it helps me to find issue and fix it.
- it could help to go for a rewriting of matcher/exchange/blockwiselayer. (maybe in cf 3.0.0)

sbernard31:
What is the status of this PR ?

boaks:
Apart from the exposing of the inner data structures:

> If they help to fix the leaks, go for it (I have no better idea). 

About the exposing: 

```
+		client = new CoapEndpoint(new InetSocketAddress(InetAddress.getLoopbackAddress(), 0), config,
+				clientExchangeStore) {
+
+			@Override
+			protected CoapStack createUdpStack(NetworkConfig config, Outbox outbox) {
+				CoapUdpStack coapUdpStack = new CoapUdpStack(config, outbox);
+				blockwiseLayer = coapUdpStack.getBlockwiseLayer();
+				return coapUdpStack;
+			}
+		};

```

I'm not sure, if a approach with protected getLayers in BaseCoapStack and derive a test-stack which searchs for the blockwise layer will be that more complex than the above. But in the end, you have your opinion, and I have my opinion about the right design. At the eclipseCon I asked the others to also put there opinion on that design (test methods in interface vs. exposing internal data structures in the implementation). But until now, they didn't fin the time to answer :-).

sbernard31:
> At the eclipseCon I asked the others to also put there opinion on that design

Where should they give their answer ? is there a list of arguments for one or another solution ? 

> Apart from the exposing of the inner data structures:

So this PR is blocked because of that ?

boaks:
> Where should they give their answer ? is there a list of arguments for one or another solution ?

Where ever they want. But I guess, they are too busy.

> So this PR is blocked because of that ?

I can't block anything, I'm only one committer of 5.

So ask @mkovatsc , @sophokles73 or @jvermillard to approve this PR.
Or other developer, which are not californium committer e.g. @msangoi to comment/agree on your proposed design to expose internal data structure at implementation level).

boaks:
@sbernard31 

Or just label it with "merged soon - if no objection" and merge it on monday, if noone has a objection :-)

sbernard31:
Objection is already here ;)

boaks:
> Objection is already here ;)

That's just a "1 vs. 1" :-). 
So, if no-one else state his/her objection, you are not blocked by a 50/50, if you feel, that this is important for you.

sbernard31:
Ok let's go with this kind of organization.

But I feel that we both are the most involved at this moment in this project. Others committers are not so much involved anymore.
So we will face the 50/50 state relatively often. I prefer if we aim consensus and in case we don't find it, be clear about the level of divergence:
- just a little/cosmetics divergence. (I prefer like that)
- a real disagreement but non blocking. (I really don't like but I can live with it)
- a blocking one. (No way!)

To go back to the PR, firstly you talk about an isEmpty() method then you talk about protected getter ?
I would like to access to BlockwiseLayer in tests.
So protected method could be ok. Another solution is to add protected "create" method for each layer ? (like we do for coapstack : https://github.com/eclipse/californium/pull/449/commits/5bf7e90e3da839e4758b1235ad721e502fd3258f)
Maybe it makes more sense ?

boaks:
> a real disagreement but non blocking. (I really don't like but I can live with it)

The "isEmpty()" approach is my personal preferred one :-). I think, it's better to have a  small (mainly) test method within a class then to expose the classes internal. But there is no consensus at this.

The "raw idea" was than, that you use the `createUdpStack` rather then expose the class internal generally. But you don't agree. So, I don't want to block you longer,more then two weeks chances for others, to give also their opinion :-).

sbernard31:
Ok but maybe adding protected "create" method for each layer don't expose internal structure so much and allow with a little bit of java gymnastics to access to internal structure ? kind of compromise ?

boaks:
My idea was something as in branch "add_testcoapstack" :-).

After that, I think, that even a TestCoapEndpoint would encapsulate more of the testing stuff.
But now it's weekend, so more on Monday :-).

boaks:
The branch add_testcoapstack now also contains a CoapTestEndpoint, which encapsulates the data structure. That would be my idea about the extension for leak testing.

sbernard31:
Is it more acceptable ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapUdpStack.java
```diff
@@ -86,23 +86,35 @@
...
+		Layer layers[] = new Layer[] {
+				createExchangeCleanupLayer(config),
+				createObserveLayer(config),
+				createBlockwiseLayer(config),
+				createReliabilityLayer(config)};
+
+		setLayers(layers);
+	}
+
+	protected Layer createExchangeCleanupLayer(NetworkConfig config) {
```
boaks: FMPOV, my solution with `getLayer()` instead of this set of `creates` is more flexible.

sbernard31: Why I prefer protected `creates` instead of protected `getLayer` :

1) This allow to easily change  implementation of one layer.
2) This allow to easily change order of layer (in the constructor).
3) This allow to create public getter without the "not so elegant" for+instance loop.

I don't really see the case where this is more flexible. Maybe when we add a new Layer ?
But I suppose if you add a new layer in CoapUdpStack, it's obvious you should use the same `create` pattern. 

Maybe I missed something ?

PR #452, opened 2017-10-24
Re-create executor if the current one is shutdown

ivivanov-bg:


californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -715,13 +705,34 @@ private void reject(Message message) {
 	 * @param task the task
 	 */
 	private void runInProtocolStage(final Runnable task) {
-		executor.execute(new Runnable() {
-			public void run() {
-				try {
-					task.run();
-				} catch (Throwable t) {
-					LOGGER.log(Level.SEVERE, "Exception in protocol stage thread: "+t.getMessage(), t);
+
```
boaks: I'm not sure, that, if a endpoint is shutdown, the executor should be restarted. If someone depends on using it's own executor, this may also cause troubles. 

So, I would prefer, if you write, how you use californium, that a shutdown of the executor causes issues.

ivivanov-bg: Please take a look at this Issue:
https://github.com/eclipse/smarthome/issues/3867

Basically we extend ```CoapClient```, then create an Endpoint using ```public CoapEndpoint(Connector connector, NetworkConfig config)```

and then - issue asynchronous get and put request

boaks: So, how did you ensure, that the coap endpoint is used proper?
If you shut it down, why do you still use it for get()? 
It looks pretty much for a life-cylce problem on the usage rather the a issue in californium.

boaks: See

https://github.com/eclipse/smarthome/issues/3867#issuecomment-338983118

kaikreuzer: I agree with @boaks - if the client is shut down, this instance must not be used anymore. Re-initializing it automatically feels pretty wrong.

ivivanov-bg: OK - changed it a little bit - the initial tests were with only this change and there were no exceptions

boaks:
I would prefer to know the root-cause for the executor to being shutdown instead of fixing the symptom by restarting it.

if it's not intended by the smarthome life-cycle, then I would prefer, that you "temporary/experimental/for debugging" add a stacktrace log at the destroyed of the `EndpointObserver` in `CoapEndpoint.start()` 

```
public void destroyed(Endpoint endpoint) {
executor.shutdown();
}

```
and then we see, why this is called.

boaks:
As written above, I would prefer to analyse the cause of the `shutdown` instead of restarting the executor to fix the symptom. Yesterday I was able to build the smarthome and I have also a tradfri, so I hope, I find some time over the weekend (but may be the weekend after the next :-) ) .
   
As proposed in https://github.com/eclipse/smarthome/issues/3867#issuecomment-339461064, if you cant wait, I would propose you use 

```
ScheduledExecutorService executor = 
   Executors.newSingleThreadScheduledExecutor(new Utils.DaemonThreadFactory());
CoapEndpoint.setExecutor(executor);

``` 

in the smarthome tradfri binding. This should not shutdown the executor on destroying the `CoapEndpoint`, because the `EndpointObserver` is not used.

kaikreuzer:
I assume this PR should be closed, right?

>  I would prefer, that you "temporary/experimental/for debugging" add a stacktrace log at the destroyed of the EndpointObserver in CoapEndpoint.start()

@boaks Is this something we can do in ESH or would we need to created a patched Californium lib for it?

boaks:
If the approach to set the Executor in the tradfri binding works for you, closing this PR would be great.
But I'm not sure, if this will really fix the issue, though the root cause seems for me still being unknown.

The executor is, as fas as I understand the `Executors.newSingleThreadScheduledExecutor()`, only shutdown, if `Executor.shutdown()` is called. So I'm still wondering, why the CoapEndpoint is destroyed.
If it's not the case, that the CoapEndpoint is destroyed, I'm pretty uncertain, what caused the executor to shutdown. Exceptions/Errors, which are thrown in the jobs, will no do that.

Finally I was able to both, build ESH from scratch and install it via OOMPF. But I didn't made it to really use it with my tradfri to see an debug the errors on my own. I still plan to do it :-).

> Is this something we can do in ESH or would we need to created a patched Californium lib for it?   

I would really enjoy to know the cause of the executors shutdown!
If https://github.com/eclipse/smarthome/issues/4479 works, this will be a hint, that `CoapEndpoint.destroy()` is called. For that, a "stacktrace" of that destroy (debugger or logger) would be great to see the cause for that `destroy()`. 

> TradfriGatewayHandler 
> @Override
>     public void setStatus(ThingStatus status, ThingStatusDetail statusDetail) {
>         // are we still connected at all?
>         if (endPoint != null) {
>             updateStatus(status, statusDetail);
>             if (dtlsConnector != null && status == ThingStatus.OFFLINE) {
>                 try {
>                     dtlsConnector.stop();
>                     dtlsConnector.start();
>                 } catch (IOException e) {
>                     logger.debug("Error restarting the DTLS connector: {}", e.getMessage());
>                 }
>             }
>         }
> }
> 

And stuff like, that makes me belief, that the real root cause is still hidden by several restarts, which just fixes symptoms :-).

kaikreuzer:
> If the approach to set the Executor in the tradfri binding works for you, closing this PR would be great.

I don't think that the one depends on the other. Don't we agree that the suggested change in this PR is in any case a bad idea (always creating a new one, even if it had been deliberately shut down)?

> But I'm not sure, if this will really fix the issue, though the root cause seems for me still being unknown.

That's why I would not want to apply that workaround, but rather add helpful debugging that can lead us the way to find the root cause of the issue.

boaks:
> Don't we agree that the suggested change in this PR is in any case a bad idea (always creating a new one, even if it had been deliberately shut down)?

Yes. I just use such PRs also as marker/reminder for me, that there may be work left.

> That's why I would not want to apply that workaround, but rather add helpful debugging that can lead us the way to find the root cause of the issue.

OK, so either check, if `CoapEndpoint.destroy()` is called (accidentally?) and provide a stacktrace, to see who's calling it, or wait until I can spent the time on running ESH with trafri binding on my own (but this will take some time).

kaikreuzer:
Ok, thanks - I have created https://github.com/eclipse/smarthome/pull/4489, so that soon all users can enable the debug logging and help finding out from where exactly the endpoint is destroyed.

boaks:
As far as I understand it on the EclipseCon 2018, SmartHome has currently no left issue with californium 1.0

kaikreuzer:
Thanks @boaks, I'm ok with closing this PR.
Wrt "no left issues", I am not completely sure as we still experience reconnection issues - we just could not boil this down to a concrete bug. If you are interested, have a look at https://github.com/eclipse/smarthome/issues/6065#issuecomment-427137919.

Issue #451, opened 2017-10-23
observation getting removed after reaching retransmission limit


charybr:
Scenario:
1. Retransmission count set to 8. (MAX_RETRANSMIT=8).
2. Get observe is active (Token=9953ef769546f8bb) and data is being received in blockwise transfers.
3. At 11:49:38, for one blockwise transfer (Token=dddaf89f20959df1), getting block-1 started failing.
4. Retries started happening in exponential backoff manner between 11:49:38 to 12:08:43 for blockwise transfer with token dddaf89f20959df1.
5. Between 11:49 to 12:08, there were 6 new data notification from same server (Get-obs Token=9953ef769546f8bb). Blockwise transfer happened successfully.
6. At 12:08:43, for blockwise transfer with token dddaf89f20959df1 reached retransmission limit and observation with token 9953ef769546f8bb was removed. 
7. MaxAge was 1800, get-obs with same token was re-registered and data started coming after a gap of 30 mins.

Questions:
1. Because of one blockwise failure, observe got deleted (though other blockwise transfers were successful). Since get-obs got removed, there was no data for 30mins till re-register. Is this fine?
2. Should we disable transparent blockwise i.e set MAX_RESOURCE_BODY_SIZE to 0 ? So that only one blockwise is active per-getobs.

boaks:
Are both sides, client and server, californium? 
Is this bug repeatable? Do you have log?
What makes me really wondering: if the communication works (5. 6 new data notification ...), why is the get in 3. still not responded?

charybr:
Thanks! for responding.

Client side is californium. Server side it is coap stack on tinyos.
We are using Californium 2.0.0-snapshot.

Californium log file attached.
[californium.0.0.log.zip](https://github.com/eclipse/californium/files/1410738/californium.0.0.log.zip)

Max retransmit exceeded for blockwise transfer with token dddaf89f20959df1 between 11:49 to 12:08. During this time there were 7 new data notification for which  blockwise transfer was successful.
1. MID= 4175, Token=aaa15a5093a430c1
2. MID= 4179, Token=39242e3bab4aeb69
3. MID= 4185, Token=46b1d8acaf352f27
4. MID= 4191, Token=f847e911711af4a0
5. MID= 4197, Token=dc78447df42b9702
6. MID= 4203, Token=345163e46fc3a5c7
7. MID= 4209, Token=1e283ad9848f258c

[...]

PR #454, opened 2017-10-25
Fix 451

boaks:
#451

Add unit test for timedout blockwise notify (also correcting a config type).
Ignore timeout for blockwise notifies.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
@@ -40,6 +40,8 @@
  *                                                    observe option.
  *    Achim Kraus (Bosch Software Innovations GmbH) - replace parameter EndpointContext 
  *                                                    by EndpointContext of response.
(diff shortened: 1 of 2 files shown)


boaks:
@sbernard31 
I would like to merge this, but this will raise a conflict with your PR #449.
The conflict should be easy to solve.

sbernard31:
```java
    @Override
    public void onCancel() {
        failed();
    }

    @Override
    protected void failed() {
         observationStore.remove(request.getToken());
    }
```
On notification I think we don't want to remove observation or release token at all ?

boaks:
> On notification I think we don't want to remove observation or release token at all ?

I don't know (or at least, I'm very unsure). Therefore I only changed the case, which caused an error.

sbernard31:
E.g. on a block2 notification, the client try to send a GET with block2 option  but it fails onSendError()
This will cancel the observation. I don't think this is what we want.

boaks:
My only intention is to fix #451. And the scope of this PR is exactly that. issue 
For additional fixes, feel free to add them.

sbernard31:
OK ... so LGTM ...

Issue #455, opened 2017-10-26
InMemoryConnectionStore's sessionEstablished method isn't thread safe and could corrupt the session cache


lhotari:
The method [`org.eclipse.californium.scandium.dtls.InMemoryConnectionStore#sessionEstablished`](https://github.com/eclipse/californium/blob/e788d077f73e4794522910719981351604ac52e3/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java#L255-L260) should be `synchronized`. The current code can corrupt the `HashMap` used in [`org.eclipse.californium.scandium.dtls.InMemorySessionCache`](https://github.com/eclipse/californium/blob/cd942792ce2d3e534dbefb288e4c35265e5ab9a2/scandium-core/src/test/java/org/eclipse/californium/scandium/dtls/InMemorySessionCache.java#L30). The corruption of HashMap's datastructures [could cause an infinite loop and 100% CPU usage](https://stackoverflow.com/questions/17070184/hashmap-stuck-on-get).

boaks:
```
	@Override
	public void sessionEstablished(final Handshaker handshaker, final DTLSSession establishedSession) throws HandshakeException {
		if (sessionCache != null) {
			// put current connection state to second level cache
			sessionCache.put(establishedSession);
		}
	}
```
So `sessionEstablished` calls `put` of the session cache.

```
	/**
	 * Creates a DTLS connector for a given set of configuration options.
	 * 
	 * @param configuration The configuration options.
	 * @param sessionCache An (optional) cache for <code>DTLSSession</code> objects that can be used for
	 *       persisting and/or sharing of session state among multiple instances of <code>DTLSConnector</code>.
	 *       Whenever a handshake with a client is finished the negotiated session is put to this cache.
	 *       Similarly, whenever a client wants to perform an abbreviated handshake based on an existing session
	 *       the connection store will try to retrieve the session from this cache if it is
	 *       not available from the connection store's in-memory (first-level) cache.
	 * @throws NullPointerException if the configuration is <code>null</code>.
	 */
	public DTLSConnector(final DtlsConnectorConfig configuration, final SessionCache sessionCache) {

```

According the javadoc, the session cache may be shared, so `synchronize` at the `ConnectionStore` may fail either. FMPOV the javadoc for the `SessionCache` should mention the required threadsafe implementation.

The `InMemorySessionCache` is included in the test classes, so it's intention is for testing only. 
But we can implement it there.

[...]

PR #456, opened 2017-10-26
Add javadoc about thread-safe requirement to SessionCache.

boaks:
Implementation test InMemorySessionCache thread-safe.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SessionCache.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SessionCache.java
@@ -12,14 +12,17 @@
  * 
  * Contributors:
  *    Bosch Software Innovations GmbH - Initial creation
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add javadoc about required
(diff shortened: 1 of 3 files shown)


boaks:
Proposed fix issue #455

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SessionCache.java
```diff
@@ -12,14 +12,17 @@
...
  ******************************************************************************/
 package org.eclipse.californium.scandium.dtls;
 
 /**
  * A second level cache for current connection state of DTLS sessions.
  * <p>
- * Connection state can be put to the cache and later retrieved by the
- * DTLS session's ID.
+ * Connection state can be put to the cache and later retrieved by the DTLS
+ * session's ID. Though the cache is intended to be used concurrently, it must
```
sophokles73: IMHO this should read: `The provided cache is required to be thread-safe because it will be accessed concurrently.`

boaks:
Corrected according proposal.

boaks:
If no objection, I would merge this on Monday.

PR #458, opened 2017-10-27
Fix skip test builds.

boaks:
Builds with -Dmaven.test.skip=true failed. Therefore introduce profile
as proposed workaround in
https://issues.apache.org/jira/browse/MJAR-138.
Change to use type "test-jar" instead of "jar" for test only libraries.

### DIFF
--- a/californium-core/pom.xml
+++ b/californium-core/pom.xml
@@ -28,15 +28,31 @@
 		<dependency>
 			<groupId>${project.groupId}</groupId>
 			<artifactId>element-connector</artifactId>
-			<classifier>tests</classifier>
-			<type>jar</type>
-		</dependency>
-		<dependency>
-			<groupId>${project.groupId}</groupId>
-			<artifactId>element-connector</artifactId>
-			<type>jar</type>
 		</dependency>
 	</dependencies>
+
+	<profiles>
+		<profile>
+			<!-- maven compile would try to resolve test dependencies, 
+				even if tests are skipped. Therefore include this 
+				test dependency only, if tests are enabled -->
+			<id>tests</id>
+			<activation>
+				<property>
+					<name>maven.test.skip</name>
+					<value>!true</value>
+				</property>
+			</activation>
+			<dependencies>
+				<dependency>
+					<groupId>${project.groupId}</groupId>
+					<artifactId>element-connector</artifactId>
+					<classifier>tests</classifier>
+					<type>test-jar</type>
+				</dependency>
+			</dependencies>
+		</profile>
+	</profiles>
 
 	<build>
 		<plugins>
@@ -136,10 +152,10 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-jar-plugin</artifactId>
-				<version>3.0.2</version>
 				<executions>
 					<execution>
 						<goals>
+							<!-- export CoapNetworkRule for tests -->
 							<goal>test-jar</goal>
 						</goals>
 					</execution>
--- a/element-connector/pom.xml
+++ b/element-connector/pom.xml
@@ -65,10 +65,10 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-jar-plugin</artifactId>
-				<version>3.0.2</version>
 				<executions>
 					<execution>
 						<goals>
+							<!-- export NetworkRule and RepeatingTestRunner for tests -->
 							<goal>test-jar</goal>
 						</goals>
 					</execution>
--- a/pom.xml
+++ b/pom.xml
@@ -153,8 +153,9 @@
 				<groupId>${project.groupId}</groupId>
 				<artifactId>element-connector</artifactId>
 				<version>${project.version}</version>
-				<scope>test</scope>
 				<classifier>tests</classifier>
+				<scope>test</scope>
+				<type>test-jar</type>
 			</dependency>
 			<dependency>
 				<groupId>${project.groupId}</groupId>
@@ -257,6 +258,11 @@
 							</goals>
 						</execution>
 					</executions>
+				</plugin>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-jar-plugin</artifactId>
+					<version>3.0.2</version>
 				</plugin>
 			</plugins>
 		</pluginManagement>
--- a/demo-apps/cf-secure/pom.xml
+++ b/demo-apps/cf-secure/pom.xml
@@ -28,47 +28,60 @@
 		<dependency>
 			<groupId>${project.groupId}</groupId>
 			<artifactId>element-connector</artifactId>
-			<classifier>tests</classifier>
-            <type>jar</type>
-		</dependency>
-		<dependency>
-			<groupId>${project.groupId}</groupId>
-			<artifactId>element-connector</artifactId>
-            <type>jar</type>
 		</dependency>
 		<dependency>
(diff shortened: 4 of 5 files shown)


pom.xml
```diff
@@ -153,8 +153,9 @@
 				<groupId>${project.groupId}</groupId>
 				<artifactId>element-connector</artifactId>
 				<version>${project.version}</version>
-				<scope>test</scope>
 				<classifier>tests</classifier>
+				<scope>test</scope>
```
sophokles73: this doesn't look right to me. The managed dependency should probably cover the standard artifact which is required for compiling. FMPOV this should be:
````
<dependency>
 <groupId>${project.groupId}</groupId>
 <artifactId>element-connector</artifactId>
 <version>${project.version}</version>
</dependency>
````
If a submodule requires the *test-jar* for executing tests, it should declare this dependency explicitly (but this should not be the default).

File: element-connector/pom.xml [hunk 2]
```diff
@@ -65,7 +65,6 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-jar-plugin</artifactId>
-				<version>3.0.2</version>
```
sophokles73: if we remove the version here, we should declare it in the top level POM in  `build/pluginManagement`. Otherwise we might use different versions of the plugin in different builds.

californium-core/pom.xml
```diff
@@ -136,7 +149,6 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-jar-plugin</artifactId>
-				<version>3.0.2</version>
```
sophokles73: f we remove the version here, we should declare it in the top level POM in build/pluginManagement. Otherwise we might use different versions of the plugin in different builds.

boaks: OK

pom.xml
boaks: OK, my interpretation of dependency management is, that if several modules (but maybe not all) have a dependency to something (shared dependency), that dependency is managed in the parent's POM. Therefore junit or hamcrest are also managed in the parents POM.

The test-jars are some special cases. Usually test-stuff is not used outside. But as mentioned in #238 and #241 the element-connector contains some "test only" stuff, which is also used by test stuff outside of the element-connector. currently core and scandium use it, therefore I put it in the parent POM to manage that shared dependency.

boaks:
The maven-jar-plugin version is now in the parent POM.

@sophokles73 
Is there a reason, why shared test-jars should not be managed in the parents POM, but the other test only dependencies e.g. junit, should  be managed there? I would like to have this bugfix merged, but currently I don't know, why you insist in special handling of the test-jars.

sophokles73:
> I would like to have this bugfix merged, but currently I don't know, why you insist in special handling of the test-jars.

I do not insist on special handling of the test-jars, my concern is that if someone declares a dependency to element-connector without explicitly specifying that it is *not* the test-jar he is interested in, then maven will resolve the test-jar because that is the managed dependency. I assume that almost all modules depend on `element-connector` and therefore declare a dependency on it ...

boaks:
> if someone declares a dependency to element-connector without explicitly specifying that it is not the test-jar he is interested in, then maven will resolve the test-jar because that is the managed dependency.

OK, I 
- removed the californium in the .m2/repo
- removed the explicit test-jar dependency in scandium (leaving only the general element-connector)
- and execute `mvn -Dmaven.test.skip=true clean install -pl element-connector,scandium-core`

This still results in a successful build without creating a test-jar. So it seems, that the classifier is used for the managed dependency.

https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

> NOTE: In two of these dependency references, we had to specify the <type/> element. This is because the minimal set of information for matching a dependency reference against a dependencyManagement section is actually {groupId, artifactId, type, classifier}. In many cases, these dependencies will refer to jar artifacts with no classifier. This allows us to shorthand the identity set to {groupId, artifactId}, since the default for the type field is jar, and the default classifier is null.

So using the classifier and type test-jar should eliminate the unintended dependency to the test-jar.

sophokles73:
meaning?

boaks:
According to the maven doc and my experiment, a "plain dependency" (no classifier nor type) will not create accidentally a dependency on a test-jar. Therefore I think, we can stick with "shared dependencies go into the dependency management in the parent POM", also for the test-jar. Your 

> concern is that if someone declares a dependency to element-connector without explicitly specifying that it is not the test-jar he is interested in, then maven will resolve the test-jar because that is the managed dependency.

turns out to be not the case.

pom.xml
```diff
@@ -259,6 +259,18 @@
...
 					</executions>
 				</plugin>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-jar-plugin</artifactId>
+					<version>3.0.2</version>
+					<executions>
+						<execution>
+							<goals>
+								<goal>test-jar</goal>
```
sophokles73: does this mean that we build a test-jar for each module? Wouldn't it be sufficient to add the configuration to the modules that actually want to build a test-jar?

sophokles73:
> turns out to be not the case.

I see.

pom.xml
boaks: I hope not, as it is in the pluginManagement section. 
And I don't see the scandium test-jar after building with tests.

boaks: > Wouldn't it be sufficient to add the configuration to the modules that actually want to build a test-jar?

Sure, could the executions be removed from the parent POM and be copied twice (or where ever we will create a test-jar). But I'm not sure, why this copies are intended to have advantages.

sbernard31:
Silly question but what do you mean by build failed ?
I just test  : mvn clean install  -Dmaven.test.skip=true  on branch 2.0.x and it works for me.
(using maven 3.5.0)

boaks:
I tried to build californium after I removed it from the local maven repo (really clean build).
In that case, it fails. If you build it prior with test, you can skip them.

sbernard31:
So why do we want to support this particular use case ?
I mean when we use  -Dmaven.test.skip=true on a "empty repository".
Knowing that using -DskipTests instead of -Dmaven.test.skip=true will work ?

boaks:
Because, when using eclipse IDE with "skip test" on a clean repo fails :-).
I agree, that this is more a "comfort issue". But though this cleanup of the test-jar dependency fix it and don't bring disadvantages (at least, in my opinion), I don't see, why we should not correct it.

sbernard31:
The only disadvantages is "pom readability". ( more code often means more complexity)

> Because, when using eclipse IDE with "skip test" on a clean repo fails :-).

I don't use eclipse for maven but there is no way to tell to eclipse to use -DskipTests ?

Anyway, if you guys think this is worth it. Go for it :)

boaks:
My concern was, that others are more confused about "skip test" failing with eclipse, than even looking into the POM. But I will put some comments into the "profile" to explain the cause of the complexity.

pom.xml
boaks: So, though I would add some comments about the extra "profile", should we go for copies of the goal test-jar or should we keep it without copies?

sophokles73: My personal preference would be to add the executions only to those modules that actually want to build a test-jar. Doing so is actually not very common anyways (at least I haven't been working on any other project where this has been done). To me it makes it much more transparent where and why we want/need to create test-jars (assuming that we document the execution configuration properly in the modules' POM files).

boaks: OK, I will mode the goal test-jar

boaks:
Comments added and test-jar's again generated explizit.

Issue #442, opened 2017-10-17
Can I run "cf-secure" on Android?


Jimyeol:
I want to run "cf-secure example" by running Android local server. Is it possible to change "JKS" to "BKS"?

boaks:
Yes, if you implement it :-).

Note: if you use PSK only, the 

`builder.setPskStore(new StaticPskStore("Client_identity", "secretPSK".getBytes()));`

defines all required credentials and you can just remove the certificate stuff (no key nor trust store is really required for that).

boaks:
```
DtlsConnectorConfig.Builder builder = new DtlsConnectorConfig.Builder();
builder.setPskStore(new StaticPskStore("password", "sesame".getBytes()));
DTLSConnector dtlsConnector = new DTLSConnector(builder.build());

```

For the client, this would be enough to setup the PSK mode.

```
DtlsConnectorConfig.Builder builder = new DtlsConnectorConfig.Builder();
builder.setAddress(new InetSocketAddress(DTLS_PORT));
InMemoryPskStore pskStore = new InMemoryPskStore();
pskStore.setKey("password", "sesame".getBytes());
builder.setPskStore(pskStore);
DTLSConnector dtlsConnector = new DTLSConnector(builder.build());
```

And that for the server.

boaks:
So, did this work for you?

Jimyeol:
No :( 
Server is running. However, client does not receive a response. -> "No Response"

`@Override
    public void onCreate() {
        this.server = new CoapServer();
        server.add(new CoapResource("secure") {
            @Override
            public void handleGET(CoapExchange exchange) {
                exchange.respond(CoAP.ResponseCode.CONTENT, "hello security");
            }
        });

        DtlsConnectorConfig.Builder builder = new DtlsConnectorConfig.Builder();
        builder.setAddress(new InetSocketAddress(DTLS_PORT));
        InMemoryPskStore pskStore = new InMemoryPskStore();
        pskStore.setKey("password", "sesame".getBytes());
        builder.setPskStore(pskStore);
        dtlsConnector = new DTLSConnector(builder.build());

        server.addEndpoint(new CoapEndpoint(dtlsConnector, NetworkConfig.getStandard()));
        server.start();
    }`

Server


` public void clickGet(View view) {
        DtlsConnectorConfig.Builder builder = new DtlsConnectorConfig.Builder();
        builder.setPskStore(new StaticPskStore("password", "sesame".getBytes()));
[...]

PR #459, opened 2017-10-29
Fix broken mac clone

boaks:
Bugfix for android, which MacSpi reports a CloneNotSupportedException.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
@@ -10,12 +10,17 @@
  * and the Eclipse Distribution License is available at
  *    http://www.eclipse.org/org/documents/edl-v10.html.
  *
+ * Contributors:
+ *    Simon Bernard (Sierra Wireless)               - initial implementation
+ *    Achim Kraus (Bosch Software Innovations GmbH) - recreate hmac, if clone fails.
+ *                                                    Support for android (4.4),
+ *                                                    which fails to clone hmac.
+ *                                                    Use nanoTime instead of 
+ *                                                    currentTimeMillis.
  */
 package org.eclipse.californium.scandium;
 
 import java.security.GeneralSecurityException;
-import java.security.InvalidKeyException;
-import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -41,10 +46,27 @@ import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
  */
 public class CookieGenerator {
 
-	private static final long keyLifetime = TimeUnit.MINUTES.toMillis(5);
+	/**
+	 * Key lifetime in nanos.
+	 */
+	private static final long KEY_LIFE_TIME = TimeUnit.MINUTES.toNanos(5);
 
-	private long lastGenerationDate;
+	/**
+	 * Nanos of next key generation.
+	 */
+	private long nextKeyGenerationNanos;
+	/**
+	 * Last generated key.
+	 */
+	private SecretKeySpec lastSecretKey;
+	/**
+	 * Lock to protect {@link #hmac}, {@link #lastSecretKey} and
+	 * {@link #nextKeyGenerationNanos}
+	 */
 	private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+	/**
+	 * Mac initialized with {@link #lastSecretKey}, if clonable.
+	 */
 	private Mac hmac;
 
 	// attributes used for random byte generation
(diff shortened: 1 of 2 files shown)


boaks:
Fix for #442.

scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
```diff
@@ -55,10 +79,17 @@
...
-			if (hmac != null && !isKeyExpired()) {
-				return (Mac) hmac.clone();
+			if (lastSecretKey != null && !isKeyExpired()) {
+				if (hmac != null) {
+					return (Mac) hmac.clone();
+				} else {
+					Mac hmac = Mac.getInstance("HmacSHA256");
+					hmac.init(lastSecretKey);
+					return hmac;
+				}
```
sbernard31: What is the purpose of this new code ?

If clone is not supported each new hmac generation will result in `this.hmac=null`. 

Is it for avoid to generate hmac in write lock ?

boaks: In the write lock, the generateSecretKey() is called, in the read lock not.
So this is intended to create a hmac with the lastSecretKey instead of generate a new one.

sbernard31: Ok I get it.

What about the idea to have 2 cookie generators implementation and if this one failed because of clone exception we switch to a simple one (without all of this)

(If we stay with one class, getHMAC and generateCookie should not raise CloneNotSupportedException)

boaks: I thought of that. But the performance penalty of this PR should be academical. And the changes in this PR are very minimal compared to two generators.

sophokles73:
Is there a particular reason why we are *caching* the MAC instance at all? Why don't we simply create a new instance in `getHMAC`?

boaks:
PR #385 

commit 4b0cfe82a541e44ab3d326caeabfbefe700c9965

sophokles73:
thanks for explaining

scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
```diff
@@ -76,20 +107,39 @@ private Mac getHMAC() throws NoSuchAlgorithmException, CloneNotSupportedExceptio
...
+	/**
+	 * check, if {@link #lastSecretKey} has expired.
+	 * 
+	 * MUST be called in read or write scope of {@link #lock}!
+	 * 
+	 * @return {@code true}, if key is expired, {@code false} otherwise.
+	 */
 	private boolean isKeyExpired() {
-		return System.currentTimeMillis() - lastGenerationDate > keyLifetime;
+		return System.nanoTime() - lastGenerationNanos > keyLifetime;
```
sophokles73: while we are at counting nanoseconds ...
we should not store the time we created the key but should calculate the time of expiration once only when we create the key and then simply compare `System.nanoTime()` to the expiration time. That way we can save the recurring calculation of the expiration time ...

boaks: Just to say: my intention was to fix the issue in android. I had considered to just revert the previous optimization, but I assumed that this approach would cause less discussion then such a revert.
On that I was really wrong :-).

boaks: I use nanoTime instead of currentTimeMillis, because nanoTime is not affected by system time adjustments. So dependencies on date time currentTimeMillis is the proper, dependencies on time periodes, nanoTime is proper. The resulution is not the concern.

boaks:
Added more comments.
Reduce time calculations.
Fix race condition when gemerating key.

scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
```diff
@@ -41,55 +46,125 @@
...
+	 * check, if {@link #lastSecretKey} has expired.
+	 * 
+	 * MUST be called in read or write scope of {@link #lock}!
+	 * 
+	 * @return {@code true}, if key is expired, {@code false} otherwise.
+	 */
 	private boolean isKeyExpired() {
-		return System.currentTimeMillis() - lastGenerationDate > keyLifetime;
+		// consider sign wrap in longs (very optimistic about the uptime :-) )
+		return (System.nanoTime() - nextKeyGenerationNanos) >= 0;
```
sophokles73: I have no clue if the JVM actually does what you have written explicitly when you want to check
`System.nanoTime() >= nextKeyGenerationNanos`
or if the comparison can be done more efficiently at the bit level. But this seems to be an academic discussion ;-)

boaks: Javadoc of nanoTime recommends this :-).

https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#nanoTime()

> To compare two nanoTime values
>
> long t0 = System.nanoTime();
> ...
> long t1 = System.nanoTime();
>
> one should use t1 - t0 < 0, not t1 < t0, because of the possibility of numerical overflow.

sbernard31:
So complicated ...
I don't understand why 
`(System.nanoTime() - nextKeyGenerationNanos) >= 0;`
is really better than 
`System.nanoTime() - lastGenerationNanos > keyLifetime`

PR #460, opened 2017-10-29
Configurable sslcontextutil

boaks:
Improve support for android.
Note: old android doesn't support ecc cipher suites and therefore fail using californium demo certs!

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/util/SslContextUtil.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/util/SslContextUtil.java
@@ -12,6 +12,9 @@
  * 
  * Contributors:
  *    Bosch Software Innovations GmbH - initial implementation. 
+ *    Achim Kraus (Bosch Software Innovations GmbH) - introduce configurable 
+ *                                                    key store type and 
+ *                                                    InputStreamFactory. 
  ******************************************************************************/
 package org.eclipse.californium.elements.util;
 
@@ -29,6 +32,8 @@ import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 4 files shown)


boaks:
Enhancement for #442

element-connector/src/main/java/org/eclipse/californium/elements/util/SslContextUtil.java
```diff
@@ -524,6 +740,40 @@ public static SSLContext createSSLContext(String alias, PrivateKey privateKey, X
...
+
+	/**
+	 * Input stream factory for classpath resources.
+	 */
+	private static class ClassLoaderInputStreamFactory implements InputStreamFactory {
+
+		@Override
+		public InputStream create(String uri) throws IOException {
+			String resource = uri.substring(CLASSPATH_SCHEME.length());
+			InputStream inStream = SslContextUtil.class.getClassLoader().getResourceAsStream(resource);
```
sbernard31: any reason to use :
` SslContextUtil.class.getClassLoader().getResourceAsStream(resource)`
instead of 
` SslContextUtil.class.getResourceAsStream(resource)`
?

element-connector/src/test/java/org/eclipse/californium/elements/util/SslContextUtilConfigurationTest.java
```diff
@@ -0,0 +1,161 @@
...
+	public static final String CUSTOM_ENDING = ".cks";
+	public static final String CUSTOM_TYPE = "CKS";
+
+	private TestInputStreamFactory testFactory;
+
+	@Before
+	public void init() {
+		SslContextUtil.configureDefaults();
+		testFactory = new TestInputStreamFactory();
+		testFactory.stream = SslContextUtil.class.getClassLoader().getResourceAsStream("certs/keyStore.jks");
```
sbernard31: same here

element-connector/src/main/java/org/eclipse/californium/elements/util/SslContextUtil.java
boaks: What should be the advantage of that? Do you expect some certificate store located in the class package? If not, the class loader is the better choice.

sbernard31: It seems just lighter but you right the behavior is not exactly the same.

Any reason to use the class SslContextUtil class loader ? It is better than  `ClassLoader.getSystemResourceAsStream()`

boaks: Mainly, because it reuses the working DtlsTestTools.loadKeyStore().
If you feel, that the system classloader is a better choice. feel free to change AND test it (in the different environments e.g. OSGi and Android).

boaks: By the way this PU doesn't change the class loader. It changes the possible keystore type from fixed to configurable and the reading to an extendable configurable solution to enable android to plugin a AssetManger based solution.

boaks: Just to add: If you prefer to use the system classloader, you may just configure your own InputStreamFactory, which uses that instead. I prefer that classloader I used.

sbernard31: That was just curiosity, personally I didn't know when to use one or another and I just asked in case you know.

sbernard31:
no objection

PR #461, opened 2017-10-30
Polish option javadocs

### LINKED ISSUE #450: Polish javadocs for OptionSet and Option classes
As agreed in #440 javadocs for particular classes in question should get polished to reflect more closely how specification defines use of options as framework works relatively close on particular details what gets exposed to a user.
Creating this issue for PR to come.

jvalkeal:
- Rewrite OptionSet and Option javadocs for option usage.
- Fixes #450 

This is an attempt to add a little more javadocs to ease users to understand how these classes can be used. I made this PR atop of `2.0.x`.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Option.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Option.java
@@ -44,8 +44,24 @@ import java.util.Arrays;
  * UnSafe = (onum &amp; 2);
  * NoCacheKey = ((onum &amp; 0x1e) == 0x1c);
  * </pre></blockquote><hr>
- * 
- * CoAP defines several option numbers {@link OptionNumberRegistry}.
+ *
+ * {@code CoAP} defines several option numbers detailed in {@link OptionNumberRegistry}.
+ * <p>
+ * Class variables {@code number} and {@code value} directly maps to {@code CoAP} request
+ * and response options as is. In {@code CoAP} specification {@code number} is
+ * an option header key as {@code int} and {@code value} is represented as
+ * a raw byte array {@code byte[]}. User must be careful when using {@code value} directly
+ * as depending on actual option, {@code value} may be {@code empty}, {@code opaque},
+ * {@code uint} or {@code string}. For example, for {@code uint} the number 0 is represented
+ * with an empty option value (a zero-length sequence of bytes) and the number 1 by a single
+ * byte with the numerical value of 1 (bit combination 00000001 in most significant bit first
+ * notation). A recipient MUST be prepared to process values with leading zero bytes.
+ * <p>
+ * {@code Option} has helper methods, namely {@link #getIntegerValue()} and
+ * {@link #toValueString()} taking into account actual option type and how it
+ * may be represented in a native {@code value}.
+ *
+ * @see OptionSet
  */
 public class Option implements Comparable<Option> {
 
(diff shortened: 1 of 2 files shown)


jvalkeal:
Right ok, I'll try to figure out this eclipse `ip-validation` thing if it's needed?

jvalkeal:
Ok, signed that ECA and force pushed this PR with Signed-off-by commit.

californium-core/src/main/java/org/eclipse/californium/core/coap/Option.java
```diff
@@ -44,8 +44,24 @@
...
+ * an option header key as {@code int} and {@code value} is represented as
+ * a raw byte array {@code byte[]}. User must be careful when using {@code value} directly
+ * as depending on actual option, {@code value} may be {@code empty}, {@code opaque},
+ * {@code uint} or {@code string}. For example, for {@code uint} the number 0 is represented
+ * with an empty option value (a zero-length sequence of bytes) and the number 1 by a single
+ * byte with the numerical value of 1 (bit combination 00000001 in most significant bit first
+ * notation). A recipient MUST be prepared to process values with leading zero bytes.
+ * <p>
+ * {@code Option} has helper methods, namely {@link #getIntegerValue()} and
+ * {@link #toValueString()} taking into account actual option type and how it
```
boaks: `toValueString()` is more a textual description (maybe sued for logging). For string options, `getStringValue()` is the right one to access the value with the proper type. So I would prefer to have `getStringValue()` mentioned here instead of the `toValueString()`.

californium-core/src/main/java/org/eclipse/californium/core/coap/Option.java
```diff
@@ -169,7 +185,9 @@ public String getStringValue() {
 	}
 	
 	/**
-	 * Gets the option value as integer.
+	 * Gets the option value as integer. Handles cases where {@code value}
+	 * contains leading 0's or a case where {@code value} is empty which
+	 * returns 0.
```
boaks: Though this applies also to `getLongValue()`, I would prefer to have it add also there.

californium-core/src/main/java/org/eclipse/californium/core/coap/Option.java
```diff
@@ -319,8 +337,12 @@ public String toString() {
...
-	 * 
+	 * Renders the option value as string. Takes into account of option type,
+	 * thus giving more accurate representation of an option {@code value}.
+	 * Formats {@code value} as integer or string if so defined in
+	 * {@link OptionNumberRegistry}. In case of option {@code value} is just
+	 * an opaque byte array, formats this value as hex string.
+	 *
 	 * @return the option value as string
 	 */
 	public String toValueString() {
```
boaks: I think `toDescription()` would be a better name. But this may be proposed after that PR.

californium-core/src/main/java/org/eclipse/californium/core/coap/Option.java
jvalkeal: Good point, I'll add something there.

boaks:
Hi Janne,
though we want to fix now the issues in the options, would it be ok to merge your polish and copy the documentation from Integer to Long on our own?

jvalkeal:
@boaks my apologies for not updating this PR as I promised. Life seriously got in a way and couldn't find any time to get on it. Thanks for taking care of it!

PR #466, opened 2017-11-03
Add hooks for customizing ServerMessageDeliverer by extension

### LINKED ISSUE #318: Why is ServerMessageDeliverer final?
Hello,

I just noticed that the ServerMessageDeliverer (SMD) class was changed to final in some (not so recent) update. Is there any particular reason for doing so?

I am currently using an older version where I subclass SMD to implement access control on incoming messages (on recommendation by Matthias btw.), having to re-implement the SMD instead of subclassing it is quite inconvenient.

sophokles73:
Provide methods that subclasses can override to safely customize message
deliverer behavior.

fixes #318

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/server/ServerMessageDeliverer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/server/ServerMessageDeliverer.java
@@ -22,6 +22,7 @@
 package org.eclipse.californium.core.server;
 
 import java.net.InetSocketAddress;
+import java.util.Deque;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.Executor;
@@ -42,7 +43,7 @@ import org.eclipse.californium.core.server.resources.Resource;
  * The ServerMessageDeliverer delivers requests to corresponding resources and
  * responses to corresponding requests.
  */
-public final class ServerMessageDeliverer implements MessageDeliverer {
+public class ServerMessageDeliverer implements MessageDeliverer {
 
 	private static final Logger LOGGER = Logger.getLogger(ServerMessageDeliverer.class.getCanonicalName());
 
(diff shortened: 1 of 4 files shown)


californium-core/src/main/java/org/eclipse/californium/core/server/ServerMessageDeliverer.java
```diff
@@ -144,8 +179,8 @@ private void checkForObserveOption(final Exchange exchange, final Resource resou
 	 * @param list the path as list of resource names
 	 * @return the resource or null if not found
 	 */
-	private Resource findResource(final List<String> list) {
-		LinkedList<String> path = new LinkedList<String>(list);
+	protected final Resource findResource(final List<String> list) {
+		Deque<String> path = new LinkedList<String>(list);
```
boaks: Why do you use the base `Deque` instead of the `LinkedList`? Seems to be only locally used, so I don't see the real benefit.

californium-core/src/main/java/org/eclipse/californium/core/server/ServerMessageDeliverer.java
```diff
@@ -154,19 +189,56 @@ private Resource findResource(final List<String> list) {
...
 			throw new NullPointerException("Response must not be null");
 		} else if (exchange == null) {
 			throw new NullPointerException("Exchange must not be null");
 		} else if (exchange.getRequest() == null) {
 			throw new IllegalArgumentException("Exchange does not contain request");
 		} else {
-			exchange.getRequest().setResponse(response);
+			boolean processed = preDeliverResponse(exchange, response);
+			if (!processed && exchange.getRequest().getResponse() == null) {
+				exchange.getRequest().setResponse(response);
```
boaks: `getResponse() == null`

I'm not sure, if this additional test is valid for notifies. According `Request.setResponse(Response)`, the intention related to notifies seems not to mach that check. (Currently this seem just to be broken, that notify responses are delivered to the request, but this depends more on the shallow clone of the observe request, when notifies are processed.)

californium-core/src/main/java/org/eclipse/californium/core/server/ServerMessageDeliverer.java
sophokles73: FMPOV this is just a matter of style. We usually use the abstract (interface) type for declaring variables instead of using the concrete type.

californium-core/src/main/java/org/eclipse/californium/core/server/ServerMessageDeliverer.java
sophokles73: I am not sure I understand your point. Are you saying that for notifications (which are responses to an original request), the response is not (or should not be) delivered to the request?

boaks: The PR added an new check in line 217, the `getResponse() == null`. So the first impression of that would be, that **NO** newer `Responses` will be set, because a old `Response` will cause that new check to fail.
Currently in 2.0.x, notifies are processed with "shallow clones" of the original `Request`, which also removes a already received response, and therefore the new check didn't fail.
But the downside of the current solution in 2.0.x is, that observe/notify using the `Request.getResponse()` on the original request is broken  and so the cf-plugtest-checker fails also.

FMPOV, I would prefer **not** to manifest that "current 2.0.x way" of processing notifies by adding this check in line 217!

sophokles73: Now I understand :-) I agree, we should then simply remove the `getResponse() == null` check and rely on the returned boolean only. WDYT?

boaks: That was exactly what I want.

sophokles73: what about the `deliverRequest` method? Should we omit the check for the existence of the response there as well? IMHO we do not have the same potential problem there as we have with notifications, do we? However, for consistency reasons we might be inclined to check for the returned boolean only as well. WDYT?

boaks: I just don't know, why this check was added.  

Without transparent blockwise, I would guess, that `UdpMatcher.receiveRequest()` will either create a new `Exchange` (therefore without `Response`), or mark the `Request` as duplicate, but then the `ReliabilityLayer` will not forward it to the other layers and it would therefore not be passed to `ServerMessageDeliverer`.

So the check in line 89 would only check, what ever happens in `preDeliverRequest`.
Therefore I don't see any conflict (and also no benefit) in that check.

sophokles73: > I just don't know, why this check was added.

It is just for making sure that the request has not been processed already. This is just a precaution to double-check that the `preDeliverRequest` method has not already created a response albeit having returned `false`.

boaks: FMPOV remove that "precaution to double-check". It's easy to do it right in preDeliverRequest, and, if a not expected result is delivered, that would cause the implementer to do it proper.

PR #468, opened 2017-11-06
Reduce CLIENT_HELLO, if TLS_ECDHE_ECDSA is not used.

boaks:
Don't send the EC extensions, if no TLS_ECDHE_ECDSA cipher suite is used.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -19,6 +19,8 @@
  *    Kai Hudalla (Bosch Software Innovations GmbH) - fix bug 483559
  *    Achim Kraus (Bosch Software Innovations GmbH) - add enable address reuse
  *    Ludwig Seitz (RISE SICS) - Added support for raw public key validation
+ *    Achim Kraus (Bosch Software Innovations GmbH) - include trustedRPKs in
+ *                                                    determineCipherSuitesFromConfig
  *******************************************************************************/
 
 package org.eclipse.californium.scandium.config;
@@ -865,9 +867,6 @@ public final class DtlsConnectorConfig {
 			if (config.trustStore == null) {
 				config.trustStore = new X509Certificate[0];
 			}
-			if (config.trustedRPKs == null) {
-				config.trustedRPKs = new TrustAllRpks();
-			}
 			if (config.earlyStopRetransmission == null) {
 				config.earlyStopRetransmission = true;
 			}
@@ -892,16 +891,22 @@ public final class DtlsConnectorConfig {
 			if (config.staleConnectionThreshold == null) {
 				config.staleConnectionThreshold = DEFAULT_STALE_CONNECTION_TRESHOLD;
 			}
-			if (config.getSupportedCipherSuites().length == 0) {
+			if (config.supportedCipherSuites == null || config.supportedCipherSuites.length == 0) {
 				determineCipherSuitesFromConfig();
+			}
+			if (config.trustedRPKs == null) {
+				// must be set after determineCipherSuitesFromConfig(),
+				// otherwise this would be interpreted for client only
+				// as ECDHE_ECDSA support!
+				config.trustedRPKs = new TrustAllRpks();
 			}
 
 			// check cipher consistency
-			if (config.getSupportedCipherSuites().length == 0) {
+			if (config.supportedCipherSuites == null || config.supportedCipherSuites.length == 0) {
 				throw new IllegalStateException("Supported cipher suites must be set either " +
 						"explicitly or implicitly by means of setting the identity or PSK store");
 			}
-			for (CipherSuite suite : config.getSupportedCipherSuites()) {
+			for (CipherSuite suite : config.supportedCipherSuites) {
 				switch (suite) {
 				case TLS_PSK_WITH_AES_128_CCM_8:
 				case TLS_PSK_WITH_AES_128_CBC_SHA256:
@@ -943,7 +948,13 @@ public final class DtlsConnectorConfig {
 			// user has not explicitly set cipher suites
 			// try to guess his intentions from properties he has set
 			List<CipherSuite> ciphers = new ArrayList<>();
-			if (isConfiguredWithKeyPair()) {
+			boolean certificates = isConfiguredWithKeyPair();
+			if (!certificates && clientOnly) {
+				certificates = config.trustedRPKs != null
+						|| (config.trustStore != null && config.trustStore.length > 0);
+			}
+
+			if (certificates) {
 				ciphers.add(CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8);
 				ciphers.add(CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256);
 			}
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
@@ -30,6 +30,8 @@
  *                                                    supportedServerCertificateTypes
  *    Ludwig Seitz (RISE SICS) - Updated calls to verifyCertificate() after refactoring
  *    Bosch Software Innovations GmbH - migrate to SLF4J
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add certificate types only,
+ *                                                    if certificates are used
  ******************************************************************************/
 package org.eclipse.californium.scandium.dtls;
 
@@ -41,6 +43,7 @@ import java.security.cert.CertPath;
 import java.security.cert.X509Certificate;
 import java.security.interfaces.ECPublicKey;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -84,18 +87,18 @@ public class ClientHandshaker extends Handshaker {
 	protected ClientHello clientHello = null;
 
 	/** the preferred cipher suites ordered by preference */
-	private final CipherSuite[] preferredCipherSuites;
+	private final List<CipherSuite> preferredCipherSuites;
 
 	protected Integer maxFragmentLengthCode;
 
 	/**
 	 * The certificate types this server supports for client authentication.
 	 */
-	protected List<CertificateType> supportedClientCertificateTypes;
+	protected final List<CertificateType> supportedClientCertificateTypes;
 	/**
 	 * The certificate types this server supports for server authentication.
 	 */
-	protected List<CertificateType> supportedServerCertificateTypes;
+	protected final List<CertificateType> supportedServerCertificateTypes;
 
 	/*
 	 * Store all the message which can possibly be sent by the server. We need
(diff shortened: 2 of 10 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -154,22 +156,25 @@ public ClientHandshaker(DTLSSession session, RecordLayer recordLayer, SessionLis
...
-			if (certificateChain == null || certificateChain.length == 0) {
-				this.supportedClientCertificateTypes.add(CertificateType.RAW_PUBLIC_KEY);
-			} else if (config.isSendRawKey()) {
-				this.supportedClientCertificateTypes.add(CertificateType.RAW_PUBLIC_KEY);
-				this.supportedClientCertificateTypes.add(CertificateType.X_509);
-			} else {
-				this.supportedClientCertificateTypes.add(CertificateType.X_509);
-				this.supportedClientCertificateTypes.add(CertificateType.RAW_PUBLIC_KEY);
+
+		if (config.useEcdheEcdsa()) {
```
sophokles73: So, the idea here is that we only need to indicate preference for any particular types of certificates if we are supporting a cipher suite that requires the exchange of one or more certificates, right?
If so, then I would rather rename the `config.useEcdheEcdsa` method to something that better reflects the purpose. There are more cipher suites that require the exchange of certificates, in particular the ones that we do not support yet which are based on RSA but which have been asked about in the past already. What about something like `isCipherRequiringCertificateExchangeSupported`?

sophokles73: BTW the CLIENT_HELLO also includes extensions that are only relevant for ECC based ciphers. I think we should make inclusion of those also dependent on the support for ECC based ciphers ...

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -363,8 +364,20 @@ public TrustedRpkStore getRpkTrustStore() {
...
 
 	/**
-	 * A helper for creating instances of <code>DtlsConnectorConfig</code>
-	 * based on the builder pattern.
+	 * Check, if a "TLS_ECDHE_ECDSA_WITH_AES_???" cipher suite is used.
+	 * 
+	 * @return {@code true}, if such a cipher suite is configured,
+	 *         {@code false}, otherwise.
+	 */
+	public boolean useEcdheEcdsa() {
```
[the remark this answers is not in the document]

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -363,8 +364,20 @@ public TrustedRpkStore getRpkTrustStore() {
...
+		return supportedCipherSuites != null
+				&& (supportedCipherSuites.contains(CipherSuite.TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8)
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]
boaks: Why should we stay with that array, if it's always copied or created?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
boaks: Uups, I just forgot to pick that commit. I will do that in the evening.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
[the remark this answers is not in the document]
sophokles73: because you then do not change the code in so many places. We need to copy the collection as well when passing values in and out via the getters ...

boaks:
Just to mention:
The current limitation is, that certificates and ECC are tightly coupled. So none ECC certificates (e.g. RSA) would require some changes.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHello.java
```diff
@@ -148,33 +150,41 @@ private ClientHello(ProtocolVersion version, SecureRandom secureRandom, SessionI
...
-			CertificateTypeExtension clientCertificateType = new ClientCertificateTypeExtension(true);
-			for (CertificateType certificateType : supportedClientCertificateTypes) {
-				clientCertificateType.addCertificateType(certificateType);
+		
+		boolean supportedClientCertificate = supportedClientCertificateTypes != null
+				&& !supportedClientCertificateTypes.isEmpty();
+		boolean supportedServerCertificate = supportedServerCertificateTypes != null
+				&& !supportedServerCertificateTypes.isEmpty();
+
+		if (supportedClientCertificate || supportedServerCertificate) {
```
sophokles73: IMHO we should make the inclusion of the extensions dependent on whether the supported cipher suites require a certificate exchange and/or whether they are ECC based.
For that, we should include the preferred cipher suites in the constructor instead of having a separate setter for adding them individually. There are not many places where `ClientHello.addCipherSuite()` is used ...

boaks:
Apart from the fact, that CipherSuite currently only support ECC-based certificates, the PR should now use the cipher suites to decide, which extension are included.

boaks:
@sophokles73 
Though the scope of this PR is to remove unrequired extensions in the CLIENT_HELLO, it's done.
Add other certificate based cipher suites is just out of the scope of this PR.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -152,25 +155,31 @@ public ClientHandshaker(DTLSSession session, RecordLayer recordLayer, SessionLis
...
-				this.supportedClientCertificateTypes.add(CertificateType.RAW_PUBLIC_KEY);
-				this.supportedClientCertificateTypes.add(CertificateType.X_509);
-			} else {
-				this.supportedClientCertificateTypes.add(CertificateType.X_509);
-				this.supportedClientCertificateTypes.add(CertificateType.RAW_PUBLIC_KEY);
+
+		// Currently scandium offers only ECC-based cipher suite, 
+		// but no other certificate-based CiperSuites.
+		// Correct this check, if other certificates are supported
+		if (CipherSuite.containsEccBasedCipherSuite(preferredCipherSuites)) {
```
sophokles73: I do not think that this check should be based on the fact that an ECC based cipher suite is supported but based on the fact if any of the supported cipher suites require a certificate being exchanged. IMHO  we can use the existing `CipherSuite.requiresServerCertificateMessage()` method for doing such a check...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHello.java
```diff
@@ -98,46 +102,48 @@
...
-	 *            the list of certificate types supported by the client
-	 * @param supportedServerCertificateTypes
-	 *            the list of certificate types supported by the server
-	 * @param peerAddress
-	 *            the IP address and port of the peer this message has been
-	 *            received from or should be sent to
+	 * @param version the protocol version to use
+	 * @param secureRandom a function to use for creating random values included
+	 *            in the message
+	 * @param cipherSuites the list of the supported cipher suites
```
sophokles73: please rename to `preferredCipherSuites` to better reflect the purpose of the cipher suites and adapt the description accordingly

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHello.java
```diff
@@ -98,46 +102,48 @@
...
-		addCipherSuite(session.getWriteState().getCipherSuite());
+		this(version, secureRandom, session.getSessionIdentifier(),
+				Arrays.asList(session.getCipherSuite()), supportedClientCertificateTypes,
+				supportedServerCertificateTypes, session.getPeer());
 		addCompressionMethod(session.getWriteState().getCompressionMethod());
 	}
 
-	private ClientHello(ProtocolVersion version, SecureRandom secureRandom, SessionId sessionId, List<CertificateType> supportedClientCertificateTypes,
+	private ClientHello(ProtocolVersion version, SecureRandom secureRandom, SessionId sessionId,
+			List<CipherSuite> cipherSuites, List<CertificateType> supportedClientCertificateTypes,
```
sophokles73: please rename `cipherSuites` to `preferredCipherSuites`

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHello.java
```diff
@@ -148,16 +154,20 @@ private ClientHello(ProtocolVersion version, SecureRandom secureRandom, SessionI
...
 
-		// the supported groups
-		// TODO make list of supported groups configurable
-		SupportedGroup[] supportedGroups = SupportedGroup.getPreferredGroups().toArray(new SupportedGroup[]{});
-		this.extensions.addExtension(new SupportedEllipticCurvesExtension(supportedGroups));
-
-		// the supported point formats
-		List<ECPointFormat> formats = Arrays.asList(ECPointFormat.UNCOMPRESSED);
-		this.extensions.addExtension(new SupportedPointFormatsExtension(formats));
+		if (CipherSuite.containsEccBasedCipherSuite(cipherSuites)) {
```
sophokles73: In this place, the check for the support of an ECC based cipher suite is exactly the right one FMPOV :+1:

boaks:
@sophokles73 

Sorry, seems that you still want to fix more issues and enlarge the scope of PR in to no-man's land!

I write and write, that currently scandium will not be able to differentiate between ECC-based an only certificate based cipher suites, but you tend to ignore me.

sophokles73:
I am sorry if I hurt your feelings but I cannot really see why we shouldn't add five more lines of code in order to solve the issue in a more generic way while we (you) are already on it. I would agree that we should not extend the scope of a PR *into no man's land* but I have a hard time seeming that I am urging you to do so. I would really appreciate if you would re-consider and re-open this PR so that we can successfully merge it in no time ...

boaks:
You don't hurt my feelings, but it seems to be easier to wait until the cipher suite management is extended to support those other certificates. Including tests of those other certificates, I guess, this is more than 5 lines, at least much more time then just typing those 5 lines.
So my preferred way was to document/state the limitation clearly and leave the extension of the cipher suite management to those, who require that and therefore are willing to spent the adequate time into that issue.

sophokles73:
We are in total agreement that we should not extend this PR to *cipher suite management* and we should also not use this PR to add support for RSA based cipher suites. The *five lines* of code would, however, make it easier to later add such functionality because we would already account for any cipher suite requiring certificate exchange when creating the CLIENT_HELLO, wouldn't you agree?

sophokles73:
Here's my peace offering ;-)
We reopen the PR, merge it as-is, and I do the little changes I am so keen on myself. WDYT?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHello.java
boaks: FMPOV `preferredCipherSuites` is the WRONG term! According RFC5246, 7.4.1.2, CLIENT_HELLO, the cipher suites are just ORDERED according the preference. 

>The cipher suite list, passed from the client to the server in the
> ClientHello message, contains the combinations of cryptographic
> algorithms supported by the client in order of the client’s
> preference (favorite choice first).
> ...
> The server will select a cipher
> suite or, if no acceptable choices are presented, return a handshake
> failure alert and close the connection.

So the client do not prefer the cipher suites, it only supports those in the list.

PR #469, opened 2017-11-07
Split TCP implementation into own package

boaks:
Move TCP implementation into package element-connector-tcp to remove netty dependency from package element-connector. Then remove TCP dependency from californium-core.


sbernard31:
LGTM

sophokles73:
What about the `TcpEndpoint*` and `TlsEndpoint*` classes in *element-connector*? Shouldn't they also be moved to *element-connector-tcp*?

boaks:
No, the idea was not to do a "perfect separation"; the TCP stuff in californium-core also still exists.
The idea was just to move the netty dependency from mandatory into optional to reduce the size of the resulting jars.

sophokles73:
Ok, I see.

sbernard31:
This is the same level of separation than with DTLS and coaps

Issue #436, opened 2017-10-12
Wrong use of java.util.logging.Logger is causing extensive GC and performance degradation 


martinmine:
In many places around the code-base in this framework, logging is done as following:
```Java
LOGGER.fine("some message with a parameter: " + parameterObject);
```
The message will not get written unless the logger is configured to do so. However, the issue is that the `.toString()` function on `parameterObject` is called at each time this line is called. This happens with for example message objects where each message gets duplicated as their `toString` is called before calling e.g. a logger's `fine` method, for then to disregard the message as it is not going to print anything anyways. 
Instead, these parameters can be provided as arguments to the logger functions. For example:

```Java
LOGGER.log(Level.FINE, "some message with a parameter: {0}, and another one: {1}", 
    new Object[]{parameterObject, anotherParameterObject});
```

Furthermore, there are places where a `StringBuilder` is used to build a log message surrounded with am if-statement checking whether the log level for this class allows logging. This can be completely avoided using the method above.

Looking at the log function's implementation, it will first check whether logging is allowed, _then if it is_, format the message and print it (or write to file, or whatever). The reason I am submitting this issue instead of a PR is that I want to make you guys aware of this issue, as I get the impression you are not aware of the drawbacks of not using log formatting correctly. Neither do I have time to go through the code-base and fix this. This problem became apparent to me from profiling the framework where it turned out that it spent a large amount of time in the `.toString()` method.

martinmine:
@boaks It looks like most of this is fixed in the 2.x release, at least from what I can see :+1:
That being said, this issue is still on the 1.x branch though.

boaks:
Just to mention, currently not too many people can spend their time in developing for californium. 
Therefore I'm not sure, if such "inherited issues"could be cleaned up.

martinmine:
@boaks fair enough. I will go ahead and close this as it seems to be fixed in the master branch and the upcoming release of 2.x.

PR #474, opened 2017-11-12
Performance improvements to Californium-proxy

martinmine:
With this PR, the stability and throughput of the californium-proxy gets significant improved: 

- Proper use of log formatting to avoid the performance problem I have discussed in issue https://github.com/eclipse/californium/issues/436.
- For some strange reason, for each request that goes through the proxy (towards a CoAP server) creates two threads: One worker thread that will perform the CoAP request, and one that will send the HTTP response back to the original requester. The second one will wait for the first one to complete. On top of it all, these two threads are created for each time there is a request. Californium has an async/nonblocking API where there is also a heavy worker thread model already in place, which renders this way of creating threads (for what I assume is either before they got the async support with the message handler in place, or before the threading model was introduce) for each incoming request useless. 
- Outgoing HTTP requests are now using the Apache HttpAsyncClient.
- Prevent race conditions where concurrent HTTP requests may use the same Endpoint for CoAP requests where the message body might get inter-mixed with other concurrent messages. This is solved by using a pool of EndpointManagers across requests.

To benchmark the impact of these performance optimizations, I send as many requests through a http/coap/http pipeline for 5 minutes towards a local nginx server:
```
HTTP requestster -- <HTTP> -- Californium -- <CoAP> -- Californium -- <HTTP> -- nginx
```

With the current master, I get an average request throughput of 2.7ms (std.dev 2.5ms), with the changes propsed in this PR: 2.1ms (std.dev 1.01ms). This can be further improved by scaling down the threading model in Californium. Currently, each request is captured in a networking thread, which is then queued to a "protocol stage thread" which will do the actual packet processing. Messages sent back are also running on their own thread. This leads to each message being passed through 4 different threads (2 additional in the current master) which introduce delays where a message is picked up by the next thread, and limits cache-hit rate for threads. Some of these threads could have been taken care of at a lower level, both simplifying application design and performance. However, this goes beyond the scope of this PR.

One thing I'm not sure about is whether the commit 8cd7ffd97c3e076ba7197e3432b3b07f01380dc9 would impact whether this PR gets merged or not. To simplify things, I use CompleteableFuture<T>, which is a Java 8 feature. This required me to add target 1.8 to pom.xml. Not sure if Californium in general is targeting Java 8 yet or if this was kind of a mistake, as Java 7 is kind of ancient. 

I have also signed off the commits in this PR (which I am still severely confused why I have to do, considering I don't need to do this for any other GH OSS project).

### DIFF
--- a/californium-proxy/pom.xml
+++ b/californium-proxy/pom.xml
@@ -42,8 +42,8 @@
 		</dependency>
 		<dependency>
 			<groupId>org.apache.httpcomponents</groupId>
-			<artifactId>httpclient</artifactId>
-			<version>${httpcomponents.version}</version>
+			<artifactId>httpasyncclient</artifactId>
+			<version>4.1.3</version>
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
--- a/californium-proxy/pom.xml
+++ b/californium-proxy/pom.xml
@@ -73,6 +73,14 @@
 					</instructions>
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<source>1.8</source>
+					<target>1.8</target>
+				</configuration>
+			</plugin>
 		</plugins>
 	</build>
 </project>
--- a/californium-proxy/src/main/java/org/eclipse/californium/proxy/CoapTranslator.java
+++ b/californium-proxy/src/main/java/org/eclipse/californium/proxy/CoapTranslator.java
@@ -21,6 +21,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.eclipse.californium.core.coap.CoAP.Code;
@@ -91,10 +92,10 @@ public final class CoapTranslator {
 			serverUri = new URI(proxyUriString);
 			// set after options have been copied from incomingRequest
 		} catch (UnsupportedEncodingException e) {
-			LOGGER.warning("UTF-8 do not support this encoding: " + e);
+			LOGGER.log(Level.WARNING, "UTF-8 do not support this encoding: {0}", e);
 			throw new TranslationException("UTF-8 do not support this encoding", e);
 		} catch (URISyntaxException e) {
-			LOGGER.warning("Cannot translate the server uri" + e);
+			LOGGER.log(Level.WARNING, "Cannot translate the server uri {0}", e);
 			throw new TranslationException("Cannot translate the server uri", e);
 		}
 
--- a/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpClientFactory.java
+++ b/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpClientFactory.java
@@ -33,11 +33,11 @@ import org.apache.http.protocol.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-public class HttpClientPool {
+public class HttpClientFactory {
 	private static final int KEEP_ALIVE = 5000;
-	private static final Logger LOGGER = Logger.getLogger(HttpClientPool.class.getName());
+	private static final Logger LOGGER = Logger.getLogger(HttpClientFactory.class.getName());
 
-	private HttpClientPool() {
+	private HttpClientFactory() {
 	}
 
 	public static CloseableHttpAsyncClient createClient() {
--- a/californium-proxy/src/main/java/org/eclipse/californium/proxy/EndPointManagerPool.java
+++ b/californium-proxy/src/main/java/org/eclipse/californium/proxy/EndPointManagerPool.java
@@ -1,0 +1,61 @@
+package org.eclipse.californium.proxy;
+
+import org.eclipse.californium.core.network.EndpointManager;
+
+import java.util.ArrayDeque;
+import java.util.Queue;
+import java.util.logging.Logger;
+
+/**
+ * A pool of EndpointManagers to avoid concurrency issues across concurrent requests.
+ */
+public class EndPointManagerPool {
+	private static final int INIT_SIZE = 10;
+	private static final Queue<EndpointManager> managers = initManagerPool(INIT_SIZE);
+
+	private static final Logger LOGGER = Logger.getLogger(EndPointManagerPool.class.getName());
+
+	private static Queue<EndpointManager> initManagerPool(final int size) {
+		final Queue<EndpointManager> clients = new ArrayDeque<>(size);
+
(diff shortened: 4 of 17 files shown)


californium-proxy/pom.xml
```diff
@@ -73,6 +73,14 @@
 					</instructions>
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<source>1.8</source>
```
boaks: We currently try to support 1.7.
So would this work with 1.7 as well?

californium-proxy/src/main/java/org/eclipse/californium/proxy/CoapTranslator.java
```diff
@@ -91,10 +92,10 @@ public static Request getRequest(final Request incomingRequest) throws Translati
 			serverUri = new URI(proxyUriString);
 			// set after options have been copied from incomingRequest
 		} catch (UnsupportedEncodingException e) {
-			LOGGER.warning("UTF-8 do not support this encoding: " + e);
+			LOGGER.log(Level.WARNING, "UTF-8 do not support this encoding: {0}", e);
```
boaks: We changed to slf4j, sorry ;-).

californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpClientFactory.java
```diff
@@ -0,0 +1,102 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Institute for Pervasive Computing, ETH Zurich and others.
```
boaks: Greate that you use a license. But, if you'r not working for the ETH, I woudl prefer to have either you name, or if you did it at work, the name of your company.

boaks:
Thanks a lot, greate work!
But, thought we use slf4j in the meantime, would it be possible to change the logging to use that?
And would it be possible, to use only java 1.7? Because the rest of californium is also using that 1.7, I would appreciate, that your PR also complies to that,

boaks:
> I have also signed off the commits in this PR (which I am still severely confused why I have to do, considering I don't need to do this for any other GH OSS project).

Californium is Eclipse project, an Eclipse requires that to ensure, that you're the author.

boaks:
> as Java 7 is kind of ancient.

Others request java 6 compatbility :-).

I'm not sure, if java 8 would bring that benefits for californium, but currently it seems, that java 7 does it stable and very widespread, also on embedded boxes.

boaks:
And, smaller PRs have also the advantage, that they may get merged faster :-).

I would recomment, if you still can spent the time, start with changing the logging to slf4j.

boaks:
Uups, I'm so focused on 2.0.x that I have overseen, this PR is against master.
So let me frankly say, if it would be merged to master, it would be somehow lost,
currently only a 2.0.0 release is planned. 
Would it be possible, that you shift your work to 2.0.x?
(The logging changes for slf4j are already done there :-).

martinmine:
@boaks Thanks for your replies! Regarding the license, I added it there as I saw it recurring throughout the source files. I saw more activity on the master branch than the 2.0.x branch, so I targeted master instead. Perhaps it would be nice to add something about this to the contributors file? :) 
Regarding the Java 7 targeting: fine, not a big problem. I will try and refactor it to not use the CompleteableFuture or backport the class somehow. 
The slf4j won't be a big issue, as the log changes is in its own commit, it is mostly a question of rebasing. 

I have a quite tight time schedule out this week, I will try and address these issues within the next coming days.

boaks:
> I saw more activity on the master branch than the 2.0.x branch, so I targeted master instead. Perhaps it would be nice to add something about this to the contributors file? :)

Yes, we should do that, sorry.

> The slf4j won't be a big issue, as the log changes is in its own commit, it is mostly a question of rebasing.

slf4j is already done in the 2.0.x, so hopefully the rebase doesn't report too much conflicts.

martinmine:
@boaks Thanks! I will close this PR for now and re-open a new one towards the 2.0.x branch once I have fixed the issues we discussed here.

vikram919:
@martinmine 
I tried to clone your pull request and run the example from project cf-proxy `ExampleCrossProxy.java` but there are some problems with the pom.xml of project cf-proxy which depends on califronium-proxy project and It throws below exception:
`Nov 17, 2017 11:00:30 AM org.eclipse.californium.core.network.config.NetworkConfig load
INFO: loading properties from file *******\git\californium_proxy\demo-apps\cf-proxy\Californium.properties
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/http/conn/SchemePortResolver
	at org.eclipse.californium.proxy.HttpClientFactory.createClient(HttpClientFactory.java:45)
	at org.eclipse.californium.proxy.resources.ProxyHttpClientResource.<clinit>(ProxyHttpClientResource.java:44)
	at org.eclipse.californium.examples.ExampleCrossProxy.<init>(ExampleCrossProxy.java:51)
	at org.eclipse.californium.examples.ExampleCrossProxy.main(ExampleCrossProxy.java:85)
Caused by: java.lang.ClassNotFoundException: org.apache.http.conn.SchemePortResolver
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 4 more`
would be nice if you can fix this so that I can take some performance statistics with the changes you have done.

martinmine:
@vikram919 yeah, it is unable to add the required dependencies (possibly due to OSGi). I currently have my hands full, but I will be able to look at it either this weekend or next week. If you want to get around this, you can add the dependencies to your project directly. (httpasyncclient should be enough). Also, if you are interested in how it affects the 2.x branch you can have a look at this branch: https://github.com/martinmine/californium/commits/proxy-optimizations-2.0.x-upstream Note that it also adds some more dependencies for e.g. backporting of the CompletableFuture<T> class so you will also have to add this manually. 

While we are at it, does anyone know how the `httpcore-nio` dependency is added in pom.xml in such a way that I do not have to include `httpcore-nio` in my test project? Adding other dependencies to pom.xml does not seem to get included when I build other projects against it without adding them explicitly to the parent project. Any ideas for this? (Would save me some time when I re-visit this issue later)

martinmine:
@vikram919 See #479

vikram919:
@martinmine 
Thanks for your information, after adding the dependency it worked for now, I will let you know If I know the solution for above question.
One question as you described the test setup:

 `HTTP requestster -- <HTTP> -- Californium -- <CoAP> -- Californium -- <HTTP> -- nginx`

 as from the example we can acheieve 

`HTTP requester --- <HTTP> --Californium Server`

how  can we establsih?

`Coap requester ----<CoAP>----HTTP Server`

Issue #414, opened 2017-09-20
Add event on request to know where it is removed from exchangeStore.


sbernard31:
We can be aware of several events on a message (see [MessageObserver](https://github.com/eclipse/californium/blob/2.0.x/californium-core/src/main/java/org/eclipse/californium/core/coap/MessageObserver.java))

Unfortunately there is no way to be aware if the exchange related to this message was complete.
I'm not sure if this make sense.

**My real problem :**
In Leshan we have a class which store all the requests we send to be able to cancel it if a device deregister. (see https://github.com/eclipse/leshan/blob/master/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/CaliforniumLwM2mRequestSender.java#L142)
The idea is to avoid retransmission after deregistration.

We clear this map when we receive the ack or a response, on cancel, on failed.
But it seems this maps leaks, I suspect that we could face situation where the exchange will be complete and no event was raised through `MessageObserver`.

Maybe a better approach is to clean this map after MESSAGE_EXCHANGE_LIFETIME.

boaks:
Feel free to extend the MessageObserver.

boaks:
> Maybe a better approach is to clean this map after MESSAGE_EXCHANGE_LIFETIME.

If blockwise is not used ...

Otherwise we should also solve issue #316

sbernard31:
I add some tests #475 and all seems ok. So I will close this one. We will see if this re-happen.

PR #475, opened 2017-11-13
Ensure that all requests finished with an event.

sbernard31:
This aims to check if there are issue related to #414.

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
@@ -19,10 +19,16 @@ package org.eclipse.californium.core.test;
 import static org.junit.Assert.assertTrue;
 
 import java.net.InetSocketAddress;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
```diff
@@ -175,5 +186,65 @@ public CoapUdpTestStack getStack() {
...
+
+		public void registerRequest(final Request request) {
+			requests.add(request);
+
+			request.addMessageObserver(new MessageObserverAdapter() {
+
+				@Override
+				public void onCancel() {
+					requests.remove(request);
+					System.out.println(request + " cancelled");
```
boaks: I know, that the test frequently use `System.out` or `err`. should we keep this or change it to the logger?

boaks:
Aside from the question about the logging: LGTM

californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
sbernard31: Oups sry, I forgot to remove this before to push :/

PR #479, opened 2017-11-17
Performance improvements to Californium-proxy

martinmine:
This is another submission of PR #474 which targets the 2.0.x branch instead of the master branch in addition to support Java versions prior to Java 8. I will paste the original text from #474 here:

With this PR, the stability and throughput of the californium-proxy gets significantly improved: 

- Proper use of log formatting to avoid the performance problem I have discussed in issue https://github.com/eclipse/californium/issues/436.
- For some strange reason, for each request that goes through the proxy (towards a CoAP server) creates two threads: One worker thread that will perform the CoAP request, and one that will send the HTTP response back to the original requester. The second one will wait for the first one to complete. On top of it all, these two threads are created for each time there is a request. Californium has an async/nonblocking API where there is also a heavy worker thread model already in place, which renders this way of creating threads (for what I assume is either before they got the async support with the message handler in place, or before the threading model was introduce) for each incoming request useless. 
- Outgoing HTTP requests are now using the Apache HttpAsyncClient.
- Prevent race conditions where concurrent HTTP requests may use the same Endpoint for CoAP requests where the message body might get inter-mixed with other concurrent messages. This is solved by using a pool of EndpointManagers across requests.

To benchmark the impact of these performance optimizations, I send as many requests through a http/coap/http pipeline for 5 minutes towards a local nginx server:
```
HTTP requestster -- <HTTP> -- Californium -- <CoAP> -- Californium -- <HTTP> -- nginx
```

With the current master, I get an average request throughput of 2.7ms (std.dev 2.5ms), with the changes propsed in this PR: 2.1ms (std.dev 1.01ms). This can be further improved by scaling down the threading model in Californium. Currently, each request is captured in a networking thread, which is then queued to a "protocol stage thread" which will do the actual packet processing. Messages sent back are also running on their own thread. This leads to each message being passed through 4 different threads (2 additional in the current master) which introduce delays where a message is picked up by the next thread, and limits cache-hit rate for threads. Some of these threads could have been taken care of at a lower level, both simplifying application design and performance. However, this goes beyond the scope of this PR.

### DIFF
--- a/californium-proxy/pom.xml
+++ b/californium-proxy/pom.xml
@@ -17,9 +17,7 @@
 	<description>Cross-proxy module</description>
 
 	<properties>
-		<httpcomponents.version>4.2.5</httpcomponents.version>
-		<httpcomponents.version.lowerbound>4.2</httpcomponents.version.lowerbound>
-		<httpcomponents.version.upperbound>5</httpcomponents.version.upperbound>
+		<httpasyncclient.version>4.1.2</httpasyncclient.version>
 	</properties>
 
 	<dependencies>
@@ -35,14 +33,8 @@
 		</dependency>
 		<dependency>
 			<groupId>org.apache.httpcomponents</groupId>
-			<artifactId>httpcore-nio</artifactId>
-			<version>${httpcomponents.version}</version>
-			<scope>compile</scope>
-		</dependency>
-		<dependency>
-			<groupId>org.apache.httpcomponents</groupId>
-			<artifactId>httpclient</artifactId>
-			<version>${httpcomponents.version}</version>
+			<artifactId>httpasyncclient</artifactId>
+			<version>${httpasyncclient.version}</version>
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
@@ -73,7 +65,6 @@
 							org.eclipse.californium.proxy.resources
 						</Export-Package>
 						<Import-Package>
-							org.apache.http*; version="[${httpcomponents.version.lowerbound},${httpcomponents.version.upperbound})",
 							*
 						</Import-Package>
 						<Bundle-SymbolicName>${project.groupId}.proxy</Bundle-SymbolicName>
--- a/californium-proxy/src/main/java/org/eclipse/californium/compat/CompletableFuture.java
+++ b/californium-proxy/src/main/java/org/eclipse/californium/compat/CompletableFuture.java
@@ -1,0 +1,56 @@
+/*******************************************************************************
+ * Copyright (c) 2017 NTNU Gjøvik and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *    Martin Storø Nyfløtt (NTNU Gjøvik) - performance improvements to HTTP cross-proxy
+ ******************************************************************************/
+
+package org.eclipse.californium.compat;
+
+/**
+ * Backport of the class introduced in Java java.util.concurrent.CompletableFuture<T>.
+ * @param <T> Type of the result the future shall complete with.
+ */
+public class CompletableFuture<T> {
+    private T result;
+    private Consumer<T> consumer;
+
+    /**
+     * Sets the result of the action completed with the future.
+     * The result will be executed on the consumer registered on this future.
+     * @param result Result of the future.
+     */
+    public void complete(T result) {
+        synchronized (this) {
+            this.result = result;
+
+            if (consumer != null) {
+                consumer.accept(result);
+            }
+        }
+    }
+
+    /**
+     * Sets the consumer of the future.
+     * @param consumer Will receive a notification one the result of the
+     *                 future has been set.
+     */
+    public void thenAccept(Consumer<T> consumer) {
+        synchronized (this) {
+            this.consumer = consumer;
+
+            if (result != null) {
+                consumer.accept(result);
+            }
+        }
+    }
+}
--- a/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpClientFactory.java
+++ b/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpClientFactory.java
@@ -1,0 +1,93 @@
+/*******************************************************************************
+ * Copyright (c) 2017 NTNU Gjøvik and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *    Martin Storø Nyfløtt (NTNU Gjøvik) - performance improvements to HTTP cross-proxy
+ ******************************************************************************/
+package org.eclipse.californium.proxy;
+
+import org.apache.http.HttpResponse;
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.client.protocol.RequestAcceptEncoding;
+import org.apache.http.client.protocol.ResponseContentEncoding;
+import org.apache.http.impl.client.DefaultConnectionKeepAliveStrategy;
+import org.apache.http.impl.nio.client.CloseableHttpAsyncClient;
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
+import org.apache.http.impl.nio.conn.PoolingNHttpClientConnectionManager;
+import org.apache.http.impl.nio.reactor.DefaultConnectingIOReactor;
+import org.apache.http.nio.reactor.ConnectingIOReactor;
+import org.apache.http.nio.reactor.IOReactorException;
+import org.apache.http.protocol.*;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class HttpClientFactory {
+	private static final int KEEP_ALIVE = 5000;
+	private static final Logger LOGGER = Logger.getLogger(HttpClientFactory.class.getName());
+
+	private HttpClientFactory() {
+	}
+
+	public static CloseableHttpAsyncClient createClient() {
+		try {
+			final CloseableHttpAsyncClient client = HttpAsyncClientBuilder.create()
+					.disableCookieManagement()
+					.setDefaultRequestConfig(createCustomRequestConfig())
+					.setConnectionManager(createPoolingConnManager())
+					.addInterceptorFirst(new RequestAcceptEncoding())
+					.addInterceptorFirst(new RequestConnControl())
+					// .addInterceptorFirst(new RequestContent())
+					.addInterceptorFirst(new RequestDate())
+					.addInterceptorFirst(new RequestExpectContinue())
+					.addInterceptorFirst(new RequestTargetHost())
+					.addInterceptorFirst(new RequestUserAgent())
+					.addInterceptorFirst(new ResponseContentEncoding())
+					.setKeepAliveStrategy(new DefaultConnectionKeepAliveStrategy() {
+						@Override
+						public long getKeepAliveDuration(HttpResponse response, HttpContext context) {
+							long keepAlive = super.getKeepAliveDuration(response, context);
+							if (keepAlive == -1) {
+								// Keep connections alive if a keep-alive value
+								// has not be explicitly set by the server
+								keepAlive = KEEP_ALIVE;
+							}
+							return keepAlive;
+						}
+
+					})
+					.build();
+			client.start();
+			return client;
+		} catch (IOReactorException e) {
+			LOGGER.log(Level.SEVERE, e.getMessage(), e);
+			return null;
+		}
+	}
+
+	private static RequestConfig createCustomRequestConfig() {
+		return RequestConfig.custom()
+				.setConnectionRequestTimeout(5000)
+				.setConnectTimeout(1000)
+				.setSocketTimeout(500).build();
+	}
+
+	private static PoolingNHttpClientConnectionManager createPoolingConnManager() throws IOReactorException {
+		ConnectingIOReactor ioReactor = new DefaultConnectingIOReactor();
+
+		PoolingNHttpClientConnectionManager cm = new PoolingNHttpClientConnectionManager(ioReactor);
+		cm.setMaxTotal(50);
+		cm.setDefaultMaxPerRoute(50);
+
+		return cm;
+	}
+}
--- a/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpTranslator.java
+++ b/californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpTranslator.java
@@ -233,7 +233,7 @@ public final class HttpTranslator {
 				try {
 					optionNumber = Integer.parseInt(optionCodeString.trim());
 				} catch (Exception e) {
-					LOGGER.warn("Problems in the parsing: " + e.getMessage());
+					LOGGER.warn("Problems in the parsing", e);
 					// ignore the option if not recognized
 					continue;
 				}
@@ -283,7 +283,7 @@ public final class HttpTranslator {
 							try {
 								maxAge = Integer.parseInt(headerValue.substring(index + 1).trim());
 							} catch (NumberFormatException e) {
-								LOGGER.warn("Cannot convert cache control in max-age option");
+								LOGGER.warn("Cannot convert cache control in max-age option", e);
 								continue;
 							}
 						}
@@ -317,7 +317,7 @@ public final class HttpTranslator {
 				// Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
 				// This cannot be parsed into a single CoAP Option and yields a
 				// NumberFormatException
-				LOGGER.warn("Could not parse header line "+header);
+				LOGGER.warn("Could not parse header line {}", header);
 			}
 		} // while (headerIterator.hasNext())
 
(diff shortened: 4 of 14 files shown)


martinmine:
Before this gets merged: Both me and @vikram919 have had some issues regarding the dependencies. Im building the Californium framework by running `mvn clean install`, however by doing so the produced JARs does not contain the dependencies that is required to use the framework, and these does then have to be added explicitly together with whatever project is then referencing the JARs. Is this intentional? I have a strong feeling I am missing something here.

vikram919:
@martinmine 
Thanks for your information, after adding the dependency it worked for now, I will let you know If I know the solution for above question.
One question as you described the test setup:

 `HTTP requestster -- <HTTP> -- Californium -- <CoAP> -- Californium -- <HTTP> -- nginx`

 as from the example we can acheieve 

`HTTP requester --- <HTTP> --Californium Server`

how  can we establsih?

`Coap requester ----<CoAP>----HTTP Server`

californium-proxy/pom.xml
```diff
@@ -41,10 +43,20 @@
...
 			<groupId>org.apache.httpcomponents</groupId>
-			<artifactId>httpclient</artifactId>
-			<version>${httpcomponents.version}</version>
+			<artifactId>httpasyncclient</artifactId>
+			<version>${httpasyncclient.version}</version>
 			<scope>compile</scope>
 		</dependency>
+		<dependency>
+			<groupId>net.sourceforge.streamsupport</groupId>
+			<artifactId>streamsupport</artifactId>
```
boaks: GPL! 
I'm not sure, if you read the ECA carefully, but I guess using GPL is not possible :-)

boaks:
Thanks!
But including GPL dependency is, as fa as I know, not possible for a eclipse project.
(I hope, I haven't overseen this in your first PR.)

For the "httpasyncclient" we (the committers) must apply a CQ at eclipse for that.
I'm out of the office until 27.11. but will work on your PR, when I'm back.

martinmine:
@boaks Thanks, sorry, I was not aware of this regarding the licenses.
On another note, I just re-visited the benchmarks on the 2.x branch.
Against the commit before the changes (850f6f3d1fc33f73405685e114200acebb64d14f): Avg: 3.47ms, stdDev: 1.09ms, N: 86399
Against the changes in this PR: Avg: 1.42ms, stdDev: 1ms, N: 210529

boaks:
Very promising!

boaks:
@sophokles73 
Maybe, you can already ask eclipse about the "httpasyncclient" (CQ)

martinmine:
@vikram919 I wrote my setup based on the proxy example in the repo. Requests are then made towards the coap2http resource. There is also one small change I made in CoapTranslator.java to make it work (for some reason it wasn't setting the proxy URIs on the requests:
```diff
diff --git a/californium-proxy/src/main/java/org/eclipse/californium/proxy/CoapTranslator.java b/californium-proxy/src/main/java/org/eclipse/californium/proxy/CoapTranslator.java
index 6ef83f6e..4d81c92a 100755
--- a/californium-proxy/src/main/java/org/eclipse/californium/proxy/CoapTranslator.java
+++ b/californium-proxy/src/main/java/org/eclipse/californium/proxy/CoapTranslator.java
@@ -90,6 +90,7 @@ public final class CoapTranslator {
 
 		// get the uri address from the proxy-uri option
 		URI serverUri;
+		String proxyUriForward = null;
 		try {
 			/*
 			 * The new draft (14) only allows one proxy-uri option. Thus, this
@@ -97,6 +98,13 @@ public final class CoapTranslator {
 			 */
 			String proxyUriString = URLDecoder.decode(
 					incomingRequest.getOptions().getProxyUri(), "UTF-8");
+
+			final String proxyEndpoint = "coap2http/";
+			final int proxyEndpointPos = proxyUriString.indexOf(proxyEndpoint);
+			if (proxyEndpointPos > 0) {
+				proxyUriForward = proxyUriString.substring(proxyEndpointPos + proxyEndpoint.length());
+				proxyUriString = proxyUriString.substring(0, proxyEndpointPos + proxyEndpoint.length());
+			}
 			serverUri = new URI(proxyUriString);
 		} catch (UnsupportedEncodingException e) {
 			LOGGER.warn("UTF-8 do not support this encoding: {}", e);
@@ -121,6 +129,11 @@ public final class CoapTranslator {
 		options.removeBlock2();
 		options.clearUriPath();
 		options.clearUriQuery();
+
+		if (proxyUriForward != null && proxyUriForward.length() > 0) {
+			options.setProxyUri(proxyUriForward);
+		}
+
 		outgoingRequest.setOptions(options);
 		
 		// set the proxy-uri as the outgoing uri
```

californium-proxy/pom.xml
martinmine: I've read through it briefly (and now, yet again) and to my understanding they are talking about including source code, not referencing through maven. This is more discussed here: https://www.eclipse.org/legal/eplfaq.php#USEINANOTHER What do you think?

martinmine: Thanks for making me aware of this, I'll try and find a workaround :)

martinmine:
@boaks I've rebased and re-implemented the CompletableFuture class under the compat package. Should be good from the GPL license now :)

vikram919:
Hi @martinmine 
Thanks for the help with the code.
I have created a test case:
`Coap Client` -->` Californium<coap2coap>` -->`Californium<coap2http>` --> `HTTP local Server`
 and calculated the average after exchanging requests 1000 times it is 8.432ms
I have not tried with the same test case as yours.
I will put it here as soon as I have them.

martinmine:
@vikram919 Do you also have some warmup period/request in your test? What was the standard deviation in your result? In my setup, I run as many requests as possible through the system for 10 minutes on one single thread requesting a HTML document and one warmup request.

martinmine:
Btw, what system are you testing it on? (CPU/OS/memory)

vikram919:
> Do you also have some warmup period/request in your test? What was the standard deviation in your result?

SD is 3.105ms, I see at the beginning it started with 16ms (initially).

> Btw, what system are you testing it on? (CPU/OS/memory)

CPU: Intel 4600M, i7 2.90GHz processor
16GB RAM Windows OS

californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpTranslator.java
```diff
@@ -815,10 +815,10 @@ public HttpRequest getHttpRequest(Request coapRequest) throws TranslationExcepti
 					coapRequest.getOptions().getProxyUri(), "UTF-8");
 			proxyUri = new URI(proxyUriString);
 		} catch (UnsupportedEncodingException e) {
-			LOGGER.warn("UTF-8 do not support this encoding: " + e);
+			LOGGER.warn("UTF-8 do not support this encoding: {}", e);
```
sophokles73: either remove placeholder or use `e.getMessage()`

californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpTranslator.java
```diff
@@ -815,10 +815,10 @@ public HttpRequest getHttpRequest(Request coapRequest) throws TranslationExcepti
...
 			throw new TranslationException("UTF-8 do not support this encoding", e);
 		} catch (URISyntaxException e) {
-			LOGGER.warn("Cannot translate the server uri" + e);
+			LOGGER.warn("Cannot translate the server uri {}", e);
```
sophokles73: either remove placeholder or use `e.getMessage()`

californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpTranslator.java
```diff
@@ -893,7 +893,7 @@ public void getHttpResponse(HttpRequest httpRequest, Response coapResponse, Http
 		try {
 			httpCode = Integer.parseInt(httpCodeString.trim());
 		} catch (NumberFormatException e) {
-			LOGGER.warn("Cannot convert the coap code in http status code" + e);
+			LOGGER.warn("Cannot convert the coap code in http status code {}", e);
```
sophokles73: either remove placeholder or use `e.getMessage()`

californium-proxy/src/main/java/org/eclipse/californium/proxy/ProxyHttpServer.java
```diff
@@ -106,9 +106,10 @@ public void sendResponse(Response response) {
 				try {
 					request.setResponse(response);
 					responseProduced(request, response);
-					httpStack.doSendResponse(request, response);
-				} catch (IOException e) {
-					LOGGER.warn("Exception while responding to Http request", e);
+					context.handleRequestForwarding(response);
+					LOGGER.info("HTTP returned {}", response);
+				} catch (Exception e) {
+					LOGGER.warn("Exception while responding to Http request {}", e);
```
sophokles73: either remove placeholder or use `e.getMessage()`

californium-proxy/src/main/java/org/eclipse/californium/proxy/ProxyHttpServer.java
```diff
@@ -106,9 +106,10 @@ public void sendResponse(Response response) {
 				try {
 					request.setResponse(response);
 					responseProduced(request, response);
-					httpStack.doSendResponse(request, response);
-				} catch (IOException e) {
-					LOGGER.warn("Exception while responding to Http request", e);
+					context.handleRequestForwarding(response);
+					LOGGER.info("HTTP returned {}", response);
+				} catch (Exception e) {
```
sophokles73: please catch specific exceptions only

californium-proxy/src/main/java/org/eclipse/californium/proxy/resources/ProxyCoapClientResource.java
```diff
@@ -46,67 +50,103 @@ public ProxyCoapClientResource(String name) {
...
-			LOGGER.info("ProxyCoapClient received CoAP request and sends a copy to CoAP target");
-			outgoingRequest.send();
-
-			// accept the request sending a separate response to avoid the
-			// timeout in the requesting client
-			LOGGER.debug("Acknowledge message sent");
+			LOGGER.debug("Sending proxied CoAP request.");
+
+			if (outgoingRequest.getDestination() == null)
+					throw new NullPointerException("Destination is null");
```
sophokles73: please add brackets

californium-proxy/src/main/java/org/eclipse/californium/proxy/resources/ProxyCoapClientResource.java
```diff
@@ -46,67 +50,103 @@ public ProxyCoapClientResource(String name) {
...
+			if (outgoingRequest.getDestinationPort() == 0)
+				throw new NullPointerException("Destination port is 0");
```
sophokles73: please add brackets

californium-proxy/pom.xml
```diff
@@ -18,6 +18,7 @@
 
 	<properties>
 		<httpcomponents.version>4.2.5</httpcomponents.version>
+		<httpasyncclient.version>4.1.3</httpasyncclient.version>
```
sophokles73: in order for this to work properly we should also update http-core etc accordingly. IMHO since the dependency on the async client pulls in everything else needed from  HttpComponents, we should remove the explicit dependency on httpcore-nio and also remove the `httpcomponents.version` property.

vikram919:
@martinmine 
As @sophokles73 mentioned,
This dependency is alone okay, it works now without including it again in cf-proxy pom.xml.
we can remove the rest.

```

<dependency>
	<groupId>org.apache.httpcomponents</groupId>
	<artifactId>httpasyncclient</artifactId>
	<version>4.1.3</version>
	<scope>compile</scope>
</dependency>

```

martinmine:
@vikram919 @sophokles73 Thanks for your comments, I will address them shortly.

californium-proxy/src/main/java/org/eclipse/californium/proxy/HttpClientFactory.java
```diff
@@ -0,0 +1,103 @@
...
+				.setConnectionRequestTimeout(5000)
+				.setConnectTimeout(1000)
+				.setSocketTimeout(500).build();
+	}
+
+	/**
+	 * Optional function to pre-initialize the manager instead of lazy-loading.
+	 */
+	public static void init() {
+		LOGGER.info("http client pool ready");
```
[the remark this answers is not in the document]
martinmine: Sorry, this was left there when the class used lazy loading for some other strategy, now it serves no purpose. Thanks for discovering this, I will remove this function.

californium-proxy/src/main/java/org/eclipse/californium/compat/CompletableFuture.java
```diff
@@ -0,0 +1,59 @@
...
+ *    Martin Storø Nyfløtt - performance improvements to HTTP cross-proxy
+ ******************************************************************************/
+
+package org.eclipse.californium.compat;
+
+/**
+ * Backport of the class introduced in Java java.util.concurrent.CompletableFuture<T>.
+ * @param <T> Type of the result the future shall complete with.
+ */
+public class CompletableFuture<T> {
```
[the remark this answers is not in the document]
martinmine: Nope, I just wrote it from the top of my head. It is a very basic straight-forward implementation.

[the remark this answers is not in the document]
sophokles73: ok, that's good :-)

martinmine:
@sophokles73 Maybe you should add the thing about the copyright headers to the contributors.md file as well? :) Regarding the exceptions, I highly appreciate (as a developer when using Californium) that I get a more appropriate message out rather than just for example "Null pointer exception". A stack trace is soooo much more useful if no other exception is thrown. For future work, I would suggest that you go by the convention of providing the Throwable object to the logger so it can print the stack trace (if any).

vikram919:
@martinmine 
I created a test setup without changing the code just only with the pull request,
`Coap client`<--`<coap2http>`-->`HTTP local Server` - Average 0,236708 milli seconds for a each request-response
`Http client`<--`<HttpProxy>`-->`Coap local Server` - Average 0,865336 milli seconds

This time I run the tests using a single thread sending 1000 get requets and calculating delay on every response  in approx nano time precision with the same machine mentioned above
Thank you for the good work!

californium-proxy/pom.xml
```diff
@@ -17,9 +17,7 @@
 	<description>Cross-proxy module</description>
 
 	<properties>
-		<httpcomponents.version>4.2.5</httpcomponents.version>
-		<httpcomponents.version.lowerbound>4.2</httpcomponents.version.lowerbound>
-		<httpcomponents.version.upperbound>5</httpcomponents.version.upperbound>
+		<httpasyncclient.version>4.1.3</httpasyncclient.version>
```
sophokles73: ok, I have done some research regarding already approved CQs for HttpComponents artifacts at Eclipse. It seems that using version 4.1.2 of the async client will induce the least effort regarding dependencies while still being sufficiently up to date. So please change the version to 4.1.2.

californium-proxy/src/main/java/org/eclipse/californium/compat/CompletableFuture.java
```diff
@@ -0,0 +1,56 @@
+/*******************************************************************************
+ * Copyright (c) 2017 NTNU Gjøvik and others.
```
sophokles73: this looks good :+1:

californium-proxy/pom.xml
sophokles73: we will then need to make a lot of additions to the maven bundle plugin configuration in order to specify the required versions of packages to import. Sadly but truly it seems that the arbitrary HttpComponents artifacts do not contain separate *sub-trees* of packages which would allow us to use some reasonable wild cards but instead contain a wild mix of complementary packages at all levels of the package hierarchy (sigh). I will take take of that once this PR is integrated, though, so you do not need to bother ...

martinmine: Done.

martinmine:
@vikram919 Thanks for your investigation, did you run it without the changes from this PR as well?

vikram919:
@martinmine yes, I run it without changing anything in the current pull request.

martinmine:
@vikram919 Sorry, if I was unclear. Did you run your benchmark on the head of the branch `eclipse:2.0.x` so that we know how much of a performance boost this PR gives? According to your results, we only know how it performs with the changes proposed in this PR.

martinmine:
@vikram919 Thats awesome! Thanks a lot for looking into this!
@sophokles73 Great!

sophokles73:
@martinmine can you please squash your changes into one or two commits, e.g. one adapting the logging and the other implementing the improvements?

martinmine:
@sophokles73 Done. I squashed it into one commit as I had commits between log-related changes that I didn't want to start mess with. Hope this is fine.

sophokles73:
@martinmine 
thanks for squashing, but you should fix the commit message which now contains all messages of the squashed commits including the "signed-off-by" footer. You should at least remove the duplicate footers so that there is only one single "signed-off-by" at the bottom. Maybe you could also arrange the changes in a bullet point list? Or include some excerpts from the excellent description you created for the PR?

martinmine:
@sophokles73 Thanks. Do you think it looks better now?

sophokles73:
Excellent :+1:

vikram919:
@martinmine 
Did you try using observe option over coap2coap proxy?

martinmine:
@vikram919 are you talking about this? https://github.com/martinmine/californium/blob/8c0fbe3cf92d99a3e6e3385471fdd662b41edea7/californium-proxy/src/main/java/org/eclipse/californium/proxy/resources/ProxyCoapClientResource.java#L78 Otherwise, I use the proxy example in the current master.

vikram919:
@martinmine 
I mean,
Have you tested the eclipse/californium with below scenario:
`coap client `(sends GET request with observe option )-----`<coap2coap>`------`coap server` (sending notifications when the resource is changed)

martinmine:
@vikram919 Nope. The goal of my project was to proxy HTTP requests using CoAP, so I have not looked into that.

martinmine:
@vikram919 But thanks, you gave me some ideas for some more things I can discuss in my thesis :)

vikram919:
@martinmine 
some more information,
If the application uses proxy,
Coap server must explicitly define max age option for the response , If the max age option is not defined it will assume the default value MAX_AGE = 60 seconds.
**As per RFC 7641, Section 3.3.1:**
If the proxy receives a response from coap server before MAX_AGE, its simply does not assume it as the current state of the resource.
I have just now tested the above definition californium coap2coap proxy neglects any notification from server, if received before MAX_AGE option.
Good Luck with your Thesis :-)

martinmine:
@vikram919 I see. One problem might be how we use the endpoint pooling to avoid the race condition I mentioned earlier. If you want to discuss this more, we can take it somewhere else as it goes beyond the scope of this PR (maybe on Twitter or whatever else you can use for IM?) 
And thanks!

vikram919:
We close this here. you found me already,please check your twitter for further discussions

PR #480, opened 2017-11-17
Rtt

boaks:
Improve RTT handling.
Use Long to detect, if its not set.
Preserve the original response in blockwise, if it's for the original request.
Add rtt to the pretty printing.


sophokles73:
LGTM

PR #490, opened 2017-12-02
Update readme and contribution.

boaks:
Seems to be frequently "requested" :-).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,11 +39,16 @@ You can find all the details in the [Contributing via Git](http://wiki.eclipse.o
 
 ## Making your Changes
 
+1. Consider to start with creating an issue on GitHub to discuss your plans and get the proper startup information.
 1. Fork the repository on GitHub
-1. Create a new branch for your changes
-1. Make your changes
+1. Create a new branch for your changes based on the 2.0.x branch. 
+   Please note: work based on other branches without prior discussion in an issue, may be in vain.
+1. If you use the eclipse IDE, please import our prefer formatter `eclipse-formatter-profile.xml` from the californium parent-folder and apply it to your changes (only :-) ).
+1. Make your changes 
 1. Make sure you include test cases for non-trivial features/changes
 1. Make sure the test suite runs successfully after you made your changes
+1. If new files are created, provide a proper license header
+   (see license_header_template.txt and copy the adjusted javadoc comment to the top of your java file)
 1. Commit your changes into the branch you created in step 2
 1. Use descriptive and meaningful commit messages
 1. If you have a lot of commits squash them into a single commit
--- a/README.md
+++ b/README.md
@@ -30,18 +30,20 @@ to your `pom.xml` (without the dots):
     <dependency>
             <groupId>org.eclipse.californium</groupId>
             <artifactId>californium-core</artifactId>
-            <version>1.0.4</version>
+            <version>1.0.6</version>
     </dependency>
     ...
   </dependencies>
   ...
 ```
 
-If you want to give the most recent milestone a try, use version `2.0.0-M2` instead.
+If you want to give the most recent milestone a try, use version `2.0.0-M6` instead.
 
-##### Current Master Builds
+Please Note: in the meantime, a lot of new bugfixes and improvements has been made in branch 2.0.x, which had affected the API and could not be back-ported to the other branches. Even if 2.0.x is still under development, we recommend to focus on using the 2.0.x branch, especially for larger changes.
 
-You can also be bold and try out the most recent build from `master`.
+##### Current Builds
+
+You can also be bold and try out the most recent build from `2.0.x`.
 However, we are not publishing those to Maven Central but to Californium's project repository at Eclipse only.
 You will therefore need to add the Eclipse Repository to your `pom.xml` first:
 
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -39,11 +39,15 @@ You can find all the details in the [Contributing via Git](http://wiki.eclipse.o
...
 1. Fork the repository on GitHub
-1. Create a new branch for your changes
-1. Make your changes
+1. Create a new branch for your changes based on the 2.0.x branch. 
+   Please note: work based on other branches without prior discussion in an issue, may be in vain.
+1. If you use the eclipse IDE, please import our prefer formatter `eclipse-formatter-profile.xml` from the californium parent-folder and apply it to your changes (only :-) ).
+1. Make your changes 
 1. Make sure you include test cases for non-trivial features/changes
 1. Make sure the test suite runs successfully after you made your changes
+1. If new files are created, provide a proper license header
```
sophokles73: we should also provide some guidance as to how the license header should look like.
Maybe something like:
`If new files are created, provide a proper license header. For that purpose, copy the license header from an existing source file and adapt it to reflect the current year and your organization/name. Please also make sure that the list of contributors only contains a single entry (representing your initial work on the file) and does not contain the entries from the file that you copied the header from.`

README.md
```diff
@@ -30,18 +30,20 @@ to your `pom.xml` (without the dots):
...
 
-If you want to give the most recent milestone a try, use version `2.0.0-M2` instead.
+If you want to give the most recent milestone a try, use version `2.0.0-M6` instead.
+
+Please Note: in the meantime, a lot of new bugfixes and improvements has been made in branch 2.0.x, which had affected the API and could not be back-ported to the other branches. Even if 2.0.x is still under development, we recommend to focus on using the 2.0.x branch, especially for larger changes.
 
 ##### Current Master Builds
 
-You can also be bold and try out the most recent build from `master`.
+You can also be bold and try out the most recent build from `master` or `2.0.x`.
```
sophokles73: I think we shouldn't encourage people to use `master` at all since we do not plan to do a 1.1 release anyway, right?
FMPOV we should replace `master` with `2.0.x` and adapt the heading to `Nightly Builds`

README.md
```diff
@@ -56,7 +58,7 @@ You will therefore need to add the Eclipse Repository to your `pom.xml` first:
     ...
   </repositories>
 ```
-You can then simply depend on `1.1.0-SNAPSHOT`.
+You can then simply depend on `1.1.0-SNAPSHOT` or `2.0.0-SNAPSHOT`.
```
sophokles73: again, we should endorse the 2.0.0-SNAPSHOTs  only, I think.

sophokles73:
@boaks 
Thanks for doing this :+1:  was on my bucket list for quite some time already :-)

README.md
boaks: Maybe we even better switch master to 2.0.x.

boaks:
So the main question would be, change master to 2.0.x or change the recommendations to 2.0.x. 
My idea was, to fast adjust the readme and contribution, and, if decided to move the master to 2.0.x or not, clean it up again :-).

CONTRIBUTING.md
boaks: Maybe we provide a file with just such a header.

sbernard31:
FMPOV, we should go to 1.0.x branch for maintenance release and master for development release (which should be the 2.0.x branch as we don't plan to release a 1.1 version)
We should remove 1.1 branch to limit confusion.

I would also encourage to not use exact version number in README, CONTRIBUTION and other documentation. Because we need to update this each release. Maybe we can go with something like : 
use the [last 2.0.x version](https://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.eclipse.californium%22%20%26%26%20v%3A2.0.*)
use the [last 1.0.x version](https://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.eclipse.californium%22%20%26%26%20v%3A1.0.*)

boaks:
so for the question is still, 
1. set master on 2.0.x and 
2. then correct the readme/contribution
or
1. fast correct the readme/contribution 
2. set master on 2.0.x and 
3. then correct the readme/contribution finally.

I created this PR, because I got aware, that some people would require that update faster then we switch the master to 2.0.x.

sbernard31:
Sry, I just given my 2 cents about future branches.

I'm not again the "fast update" first.

license_header_template.txt
```diff
@@ -0,0 +1,17 @@
+/*******************************************************************************
+ * Copyright (c) <year (e.g. 2017) <(company) name> and others.
```
sophokles73: there seems to be missing a closing `>` after the year.

sophokles73:
IMHO we should keep the current branch layout and simply indicate in the readme which of the branches *receives the most love* at the moment, i.e. which is the branch that leads to the next release.
This means we keep master *as is* and simply continue to work on 2.0.x

boaks:
But too many are just not reading and use "master". Therefore I would vote for moving master to 2.0.x, maybe not now, but in near future.

boaks:
Typo corrected and 2.0.x liked :-)

sbernard31:
> But too many are just not reading and use "master". Therefore I would vote for moving master to 2.0.x, maybe not now, but in near future.

I agree with @boaks

PR #493, opened 2017-12-04
Adjust endpoint manager to reduced factory function.

boaks:
Return IllegalStateException instead of null.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
@@ -24,6 +24,9 @@
  *    Joe Magerramov (Amazon Web Services) - CoAP over TCP support.
  *    Achim Kraus (Bosch Software Innovations GmbH) - add getDefaultEndpoint(String scheme)
  *    Bosch Software Innovations GmbH - migrate to SLF4J
+ *    Achim Kraus (Bosch Software Innovations GmbH) - getDefaultEndpoint throws 
+ *                                                    IllegalStateException
+ *                                                    instead of returning null
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -47,15 +50,15 @@ import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.server.MessageDeliverer;
 
 /**
- * A factory for {@link Endpoint}s that can be used by clients for sending
+ * A manager and coap-factory for {@link Endpoint}s that can be used by clients for sending
  * outbound CoAP requests.
  *
  * The EndpointManager contains the default endpoint for coap (on port 5683) and
  * endpoint for other schemes (coaps, CoAP over DTLS, coap+tcp, COAP over TCP,
- * coaps+tcp, COAPS over TCP,). When an application serves only as client but
- * not server it can just use the default endpoint to send requests. When the
- * application sends a request by calling {@link Request#send()} the send method
- * sends itself over the default endpoint.
+ * coaps+tcp, COAPS over TCP,), if these additional endpoints are previously set. 
+ * When an application serves only as client but not as server it can just use the
+ * default endpoint to send requests. When the application sends a request by calling
+ * {@link Request#send()} the send method sends itself over the default endpoint.
  * <p>
  * To make a server listen for requests on the default endpoint, call
  * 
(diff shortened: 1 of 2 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -121,7 +126,22 @@ public synchronized Endpoint getDefaultEndpoint(String uriScheme) {
...
+			} else if (CoAP.COAP_SECURE_TCP_URI_SCHEME.equalsIgnoreCase(uriScheme)) {
+				throw new IllegalStateException("URI scheme " + uriScheme + " requires a previous set connector!");
+			} else {
+				endpoint = new CoapEndpoint();
+			}
+			try {
+				endpoint.start();
+				LOGGER.info("created implicit endpoint {} for {}", new Object[] { endpoint.getUri(), uriScheme });
+			} catch (IOException e) {
+				LOGGER.error("could not create {} endpoint", uriScheme, e);
```
sophokles73: shouldn't we throw `IllegalStateException` here as well? It looks like we are putting the *unstartable* endpoint to the map so that it will be returned to clients on subsequent invocations ...

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
```diff
@@ -98,13 +101,15 @@ public static EndpointManager getEndpointManager() {
...
+	 * Get default endpoint for provided scheme. 
+	 * 
+	 * If not available, try to create one, if the provide uriScheme is simple "coap".
 	 * 
 	 * @param uriScheme scheme to select endpoint
-	 * @return the default endpoint for the provided scheme, or null, if not
-	 *         available
-	 * @throws IllegalArgumentException if uriScheme is not supported
+	 * @return the default endpoint for the provided scheme
+	 * @throws IllegalArgumentException if uriScheme is not generally not supported
```
sophokles73: `IllegalArgumentException` seems to not be thrown at all

boaks: public synchronized Endpoint getDefaultEndpoint(String uriScheme) {
		// default endpoints
		if (null == uriScheme) {
			uriScheme = CoAP.COAP_URI_SCHEME;
		}
		if (!CoAP.isSupportedScheme(uriScheme)) {
			throw new IllegalArgumentException("URI scheme " + uriScheme + " not supported!");
		}
		uriScheme = uriScheme.toLowerCase();
		Endpoint endpoint = endpoints.get(uriScheme);
		if (null == endpoint) {
			if (CoAP.COAP_SECURE_URI_SCHEME.equalsIgnoreCase(uriScheme)) {
				throw new IllegalStateException("URI scheme " + uriScheme + " requires a previous set connector!");
			} else if (CoAP.COAP_TCP_URI_SCHEME.equalsIgnoreCase(uriScheme)) {
				throw new IllegalStateException("URI scheme " + uriScheme + " requires a previous set connector!");
			} else if (CoAP.COAP_SECURE_TCP_URI_SCHEME.equalsIgnoreCase(uriScheme)) {
				throw new IllegalStateException("URI scheme " + uriScheme + " requires a previous set connector!");
			} else {
				endpoint = new CoapEndpoint();
			}
			try {
				endpoint.start();
				LOGGER.info("created implicit endpoint {} for {}", new Object[] { endpoint.getUri(), uriScheme });
			} catch (IOException e) {
				LOGGER.error("could not create {} endpoint", uriScheme, e);
			}
			endpoints.put(uriScheme, endpoint);
		}
		return endpoint;
	}

If the scheme is not supported at all, you get such a IllegalArgumentException.

californium-core/src/main/java/org/eclipse/californium/core/network/EndpointManager.java
boaks: Though the factory-nature is "gone" in the meantime, I received irritating NullPointerExceptions later. But I didn't receive such a IOException. Therefore I created this PR to indicate the cause of actual errors and didn't touch the other stuff.

boaks: @sophokles73 
So, should we really encapsulate the IOException (which never seems to occur), or can we go with this PR "as it is"?

boaks:
I the IOException should also be encapsulated, we may do this in an other PR.

PR #496, opened 2017-12-08
Update android app.

boaks:
Update to usage with android studio 3.0.1.
Remove IDE generated files from git repository.
Update gradle build to new versions. 
Provide x509 certificates in PKCS12.


boaks:
Though this changes ".gitignore" and removes "superfluous" android studio files, it's troublesome to change to this "cleanup_android" branch or back. It only affects android hopefully), so I would like to merge it soon.

Issue #198, opened 2017-01-24
Introduce a CoapEndpointBuilder


boaks:
CoapEndpoint has become a class with plenties of constructors.
Maybe it's better to convert that into a Builder pattern.

vikram919:
I have introduced CoapEndpointBuilder class based on builder pattern inside CoapEndpoint class.
Now we can use only one CoapEndpoint constructor instead many constructors and code looks more readable and customizable than before with less ambiguity.

PR #497, opened 2017-12-08
Introduce CoapEndpointBuilder and deprecate the constructors.

boaks:
The deprecated constructors are intended to be removed before or after
M7.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -45,6 +45,8 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - use EndpointContext
  *    Achim Kraus (Bosch Software Innovations GmbH) - use connectors protocol
  *    Bosch Software Innovations GmbH - migrate to SLF4J
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add Builder and deprecate 
+ *                                                    constructors
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -57,8 +59,6 @@ import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import org.eclipse.californium.core.coap.CoAP;
 import org.eclipse.californium.core.coap.CoAP.Type;
@@ -95,6 +95,8 @@ import org.eclipse.californium.elements.RawData;
 import org.eclipse.californium.elements.RawDataChannel;
 import org.eclipse.californium.elements.UDPConnector;
 import org.eclipse.californium.elements.util.DaemonThreadFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
@@ -199,6 +201,7 @@ public class CoapEndpoint implements Endpoint {
 	 * <p>
 	 * The endpoint will bind to all network interfaces and listen on an ephemeral port.
 	 */
+	@Deprecated
 	public CoapEndpoint() {
 		this(0);
 	}
@@ -210,6 +213,7 @@ public class CoapEndpoint implements Endpoint {
 	 *
 	 * @param port The port to listen on.
 	 */
+	@Deprecated
 	public CoapEndpoint(final int port) {
 		this(new InetSocketAddress(port));
 	}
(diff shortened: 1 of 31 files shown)


boaks:
See issue #198

boaks:
In the meantime I checked that for usage with leshan. There are several (not too many) places, where the code must be updated to use the builder, but it seems for me to be more straight forward to use this builder instead of searching for a constructor with a proper combination of parameters.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -989,4 +1009,271 @@ protected CoapStack createUdpStack(NetworkConfig config, Outbox outbox) {
...
+		 * Create {@link CoapEndpoint} using the provided parameter or defaults.
+		 * 
+		 * @return new endpoint
+		 * @throws IllegalArgumentException if applyConfiguration is
+		 *             {@code true}, but the connector doesn't implement
+		 *             {@link ConnectorConfigurator}
+		 */
+		public CoapEndpoint build() {
+			if (config == null) {
+				config = NetworkConfig.getStandard();
```
sbernard31: I really don't like this idea of standard config with a shared static instance. I already faced several issue about that.

I don't know if this has big impact but I would prefer to go with a new `NetworkConfig()` by default (which uses same default value without this shared instance and file creation)

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -304,37 +315,66 @@ public CoapEndpoint(final InetSocketAddress address, final NetworkConfig config,
...
+		// when remove the deprecated constructors, this checks and defaults maybe removed also
+		MessageExchangeStore localExchangeStore = (null != exchangeStore) ? exchangeStore
+				: new InMemoryMessageExchangeStore(config);
 		ObservationStore observationStore = (null != store) ? store : new InMemoryObservationStore();
 		if (null == endpointContextMatcher) {
 			endpointContextMatcher = EndpointContextMatcherFactory.create(connector, config);
 		}
+
+		// keep for subclasses
+		if (applyConfiguration) {
```
sbernard31: I would move this to `builder.build()` and so limit this `applyConfiguration` to the builder.

boaks: The pain of such a move would be that the idea of sub-classing the CoapEndpoint would then not longer benefit from that feature. For the CoapTestEndpoint we could go for that, so, should we limit the sub-classing in the javadoc?

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
boaks: I'm not sure about the issues. I would guess, they are caused by "parallel initialization" or "adjust the configuration within the code". I already tried to isolate the usage of that `getStandard()` much more (see PR #502). 

But currently I don't think we should change the behaviour here too fundamentally (at least for now, so not within this PR).

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
sbernard31: A possibility : 
1. move this code in the builder.
2. add builder static utility  function which set UDPConnector from config and which is used by the builder.
3. use this utility function in CoapTestEndpoint.

The idea is to well separate responsibility. Now we have a builder, it is responsible to build endpoint. So Endpoint constructor should be simple. By the way I will remove null checking + default value from CoapEndpoint constructor.

This just a point of view, not a big deal if you don't share it and you prefer to keep the code like this ;)

boaks: That solution looks like the previous one (current one, without this PR).

FMPOV, it seems to be more self-documented to have a "applyConfiguration" switch in the parameter list of the constructor than using a separate method.

The builder pattern and sub-classing doesn't really match, that's the main cause of this complexity.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
sbernard31: > I would guess, they are caused by "parallel initialization" or "adjust the configuration within the code".

You assume right. If I remember well the issue was 2 tests using NetworkConfig.getStandard() with different values launched in parallel.

In Leshan we don't use it any more.

boaks:
Sorry, I forgot to squash.

PR #498, opened 2017-12-12
Add auto-resume for NAT support.

boaks:
Triggers session resume, if no messages are exchanged for a configurable period of time.


sbernard31:
a bit late but LGTM :p

Issue #473, opened 2017-11-10
NPE in ClientHandshaker.receivedHelloVerifyRequest


kaikreuzer:
I just had this exception in my logs, using Californium 1.0.6:

```
Nov 10, 2017 2:52:52 PM org.eclipse.californium.scandium.DTLSConnector receiveNextDatagramFromNetwork
INFO: Unexpected error occurred while processing record from peer [/192.168.0.177:5684]
java.lang.NullPointerException
	at org.eclipse.californium.scandium.dtls.ClientHandshaker.receivedHelloVerifyRequest(ClientHandshaker.java:343)
	at org.eclipse.californium.scandium.dtls.ClientHandshaker.doProcessMessage(ClientHandshaker.java:218)
	at org.eclipse.californium.scandium.dtls.Handshaker.processMessage(Handshaker.java:410)
	at org.eclipse.californium.scandium.DTLSConnector.processOngoingHandshakeMessage(DTLSConnector.java:796)
	at org.eclipse.californium.scandium.DTLSConnector.processDecryptedHandshakeMessage(DTLSConnector.java:783)
	at org.eclipse.californium.scandium.DTLSConnector.processHandshakeRecordWithConnection(DTLSConnector.java:767)
	at org.eclipse.californium.scandium.DTLSConnector.processHandshakeRecord(DTLSConnector.java:703)
	at org.eclipse.californium.scandium.DTLSConnector.receiveNextDatagramFromNetwork(DTLSConnector.java:438)
	at org.eclipse.californium.scandium.DTLSConnector.access$200(DTLSConnector.java:105)
	at org.eclipse.californium.scandium.DTLSConnector$2.doWork(DTLSConnector.java:331)
	at org.eclipse.californium.scandium.DTLSConnector$Worker.run(DTLSConnector.java:1448)
```

I cannot really give you code on how to reproduce it, but it certainly looks like a bug as an NPE should happen in no situation here.
Checking the code base, it also seems that 2.x still looks the same, so although I am using 1.0.6 I would assume that the issue still exists.

sophokles73:
Hmm, it's really hard to imagine what causes this problem. As you already pointed out, there is basically no way that any of the objects at the indicated line of code is `null`.

kaikreuzer:
There must be SOME way... I have seen it twice already now.

sophokles73:
what is the scenario in which you use Cf?

boaks:
Race condition. I would guess, there are stop/start close to that exception.
[...]

PR #499, opened 2017-12-13
Fix issue 473

boaks:
The first commit uses the stripped executor to fix the race condition.
The seconds removes also the not longer required send worker.
Fix issue #473

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -81,14 +81,14 @@ import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1218,26 +1206,37 @@ public final void send(final RawData msg) {
...
-
-				@Override
-				public Object getStripe() {
-					return msg.getEndpointContext().getPeerAddress();
-				}
-
-				@Override
-				public void run() {
-					if (running.get()) {
+			if (0 <= pendingOutboundMessages.decrementAndGet()) {
```
sbernard31: MISRA ghost come back !

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1412,66 +1414,78 @@ private void sendFlight(DTLSFlight flight) {
...
+				throw new IOException("Socket closed.");
 			}
-		} else {
-			LOGGER.debug("Socket [{}] is closed, discarding packet ...", config.getAddress());
-			throw new IOException("Socket closed.");
+		} catch (InterruptedException e) {
+			// this means that the worker thread for sending
+			// outbound messages has been interrupted, most
+			// probably because the connector is shutting down
+			Thread.currentThread().interrupt();
```
sbernard31: I'm not sure to understand the purpose of this ?

boaks:
Release 1.0.6:

```
protected DTLSFlight receivedHelloVerifyRequest(HelloVerifyRequest message) throws HandshakeException {
   clientHello.setCookie(message.getCookie()); // line 343

```

I suspect the `clientHello` to be `null`. That's set by `getStartHandshakeMessage`.
So receiving "Hello Verify" during execution of `sendNextMessageOverNetwork`, after the handshaker was created, but before the `getStartHandshakeMessage` was called may have caused the `NullPointerException`. 

To fix that, I use the striped executor for the "sending" as well.
And doing so, I got aware, that the `sender` Worker and the `outboundMessages` seems to be just wasting resources. Therefore I added the second commit to remove that stuff also.

sbernard31:
>So receiving "Hello Verify" during execution of sendNextMessageOverNetwork, after the handshaker was created, but before the getStartHandshakeMessage was called may have caused the NullPointerException.

It make senses.
And in general way I think its a good thing that there is no concurrent call for a same peer.

About worker, for the reason I expose above, I don't know if this is worst or not in term of performance. but it seems obvious this will consume less memory without worker and `BlockingQueue`, so it sounds reasonnable to merge this PR.

(Could you bust the MISRA ghost before :))

boaks:
> Create 500 DTLS example clients, expect to send 500000 messages overall.
> 
> messages per s
> 
> without send worker
> 28538
> 30246
> 30165
> 
> with send worker
> 28431
> 29349
> 30712

So there is no real difference.

sbernard31:
I suspect that performance "issue" could come with large payload and lot of different foreign peers on the same connector.

But I don't think this is worthy go deeper now.

PR #501, opened 2017-12-15
Create unique tokens with server/system scope.

boaks:
Remove address usage in KeyToken.
Use address in upd endpoint context matcher.

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
@@ -13,6 +13,7 @@
  * Contributors:
  *    Bosch Software Innovations - initial creation
  *    Achim Kraus (Bosch Software Innovations GmbH) - reduce external dependency
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add address ignoring matcher
  ******************************************************************************/
 package org.eclipse.californium.core.test;
 
@@ -39,6 +40,8 @@ import org.eclipse.californium.core.network.stack.CoapStack;
 import org.eclipse.californium.core.network.stack.CoapUdpStack;
 import org.eclipse.californium.core.network.stack.Layer;
 import org.eclipse.californium.core.observe.InMemoryObservationStore;
+import org.eclipse.californium.elements.EndpointContext;
+import org.eclipse.californium.elements.EndpointContextMatcher;
 import org.eclipse.californium.elements.UDPConnector;
 import org.eclipse.californium.elements.UdpEndpointContextMatcher;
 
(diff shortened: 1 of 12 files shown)


californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
```diff
@@ -237,9 +248,46 @@ public void onTimeout() {
...
+		/**
+		 * {@inheritDoc}
+		 * 
+		 * Depending on {@link #ignoreAddress}, the address is ignored for
+		 * matching.
+		 */
+		@Override
+		public boolean isResponseRelatedToRequest(EndpointContext requestContext, EndpointContext responseContext) {
+			if (ignoreAddress) {
+				return isToBeSent(requestContext, responseContext);
```
sbernard31: Just a detail : I would use `internalMatch` instead of `isToBeSent`. (and so change internalMatch to protected)

This make me read again `internalMatch` and
 -  I still have difficulty to understand the `inhibitNewConnection`
 - is it possible to have an availableContext == null ?

boaks: inhibitNewConnection is generally for cases, where a special EndpointContext is passed in (e.g. with DTLS session) but currently no "connection" is associated and therefore the connectionContext is null. 
For the UDPConnector this seems to be not too relevant, but you will get an error, if you use a EndpointContext which inhibitNewConnection returns true (e.g. with DTLS session).

sbernard31: The main use case of `inhibitNewConnection` is here right ?
```java
if (session == null) {
	if (!checkOutboundEndpointContext(message, null)) {
		return;
	}
	// no session with peer established yet, create new empty session &
	// start handshake
```

But I still not understand what is the concrete use case of this :/

So this is the case where you send a message and you want to explicitly forbid a new connection by passing a requestedEndpointContext with inhibitNewConnection ? The message will be send only if this is the same connection ?

boaks: If you provide a DTLS Session ID in the EndpointContext, but you don't have already a session for that target, depending on the EndpointContextMatcher, it is useless to start an handshake, because this could only end in a different EndpointContext! 

The PrincipalEndpointContextMatcher doesn't use the inhibitNewConnection, because there it's possible to match even with a new handshake.
The Matchers derived from KeySetEndpointContextMatcher will use inhibitNewConnection, because here a new handshake could not end with the same session.

sbernard31: So this is a kind of optimization ? and this is strongly dependent of contextMatcher policy ?
Maybe `inhibitNewConnection` is more linked to contextMatcher than to endpointContext itself ? Could it make sense to move `inhibitNewConnection` in context matcher instead ?

boaks: It depends on the data contained in the EndpointContext and its usage in the EndpointContextMatcher. Therefore the inhibitNewConnection reflects the data, and the Matcher, if its used.

sbernard31: What I mean : this seems more understandable to me if we have somethink like this:

`EndpointContextMatcher.canCreateNewConnection(EndpointContext requestedContext)`

boaks:
replaced by #520

PR #503, opened 2017-12-18
Protect critical handshaker section.

boaks:
Fix issue #473


sbernard31:
Is it intended this aim branch 1.0.x ?
Is this a specific fix for 1.0.x as strippedExecutor does not exist in 1.0.x ?

boaks:
Yes, it's for 1.0.x, because the issue was reported for 1.0.6.
And also yes, because the StrippedExecutor is missing there.

Issue #500, opened 2017-12-14
token uniqueness (branch 2.0.x)


boaks:
We discussed this topic already in other issues, with long histories. I hope my sum-up therefore contains the important information. If not, feel free to add that.

Currently I see four variants:
1. Create token unique in the scope of an endpoint
1. Create token unique in the scope of the system
1. Create token random (and assume uniqueness in the scope of an endpoint)
1. Create token random (and assume uniqueness in the scope of the system)

The random variant will not require to do the housekeeping, the none-random will exclude extremely rare cases of token clashes (or allow to use smaller tokens, if wanted). 

Currently a mixture is implemented: create with 1. but partially used as 2 (ObservationStore), so in facto it's 4 :-).

To limit the changes, I would go for a clear 2, and take the house-keeping, mainly because I think, right now the housekeeping must be done anyway in the observe- and exchange-stores. 
Using approach 2 alone (system unique tokens) will not be enough, see issue #173. Therefore, an additional verification would be required for the request/response matching. Though the EndpointContextMatcher does this, I belief, this approach of handling tokens will be the one with the fewest changes. So I would go to prepare a PR as base for a further discussion.

Using approach 3 seems for me currently to result in some more changes, but feel free to provide an alternative PR with that.

sbernard31:
The 2. variant force us to go for a reserve/release pattern. Housekeeping is one of the "con" but not the only one. In cluster mode this will force to go for a shared tokenprovider.

The issue with 3. is that making token unique for a given endpoint bring us to define what is an endpoint identifier, we partially address it with endpointContext.

Using approach 3. solves this is issue in a more stateless way.
So I still feel that 3. is the better one. I agree this will make probably more changes. But I also know that housekeeping and race condition hunting is timeconsuming too.

About proposing an alternative PR for 3, I could go for this only if there is chances that was accepted. My last experience on this topic discourages me a bit. Lot of time to code, lot of time to argue for nothing :'(

boaks:
> The 2. variant force us to go for a reserve/release pattern.

FMPOV it allows us to `keep` not force us to `go`. 
To get rid of the housekeeping, it must be removed from the observe- and exchange-stores also. Or what are your plans for that stores?

I also thought about making variant 2 and 3 both possible with the interfaces and allow different implementations. 

[...]

PR #506, opened 2017-12-19
Fix token with endpoint

boaks:
As base for discussion of issue #500 

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
@@ -32,6 +32,7 @@
  *                                                    completeCurrentRequest.
  *    Achim Kraus (Bosch Software Innovations GmbH) - rename CorrelationContext to
  *                                                    EndpointContext.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - use endpoint identifier for KeyToken
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -748,23 +749,21 @@ public class Exchange {
 	 */
 	public static final class KeyToken {
 
-		private static final int MAX_PORT_NO = (1 << 16) - 1;
 		private final byte[] token;
-		private final byte[] address;
-		private final int port;
+		private final byte[] identity;
 		private final int hash;
 
-		private KeyToken(byte[] token, byte[] address, int port) {
+		private KeyToken(final byte[] token, final byte[] identity) {
 			if (token == null) {
 				throw new NullPointerException("token bytes must not be null");
-			} else if (address == null) {
-				throw new NullPointerException("address must not be null");
-			} else if (port < 0 || port > MAX_PORT_NO) {
-				throw new IllegalArgumentException("port must be a 16 bit unsigned int");
 			}
 			this.token = Arrays.copyOf(token, token.length);
-			this.address = address;
-			this.port = port;
+			if (identity == null) {
+				this.identity = null;
+			}
+			else {
+				this.identity = Arrays.copyOf(identity, identity.length);
+			}
 			this.hash = createHash();
 		}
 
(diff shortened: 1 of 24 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
```diff
@@ -748,23 +749,21 @@ public static KeyMID fromOutboundMessage(Message message) {
...
 
-		private static final int MAX_PORT_NO = (1 << 16) - 1;
 		private final byte[] token;
-		private final byte[] address;
-		private final int port;
+		private final byte[] identity;
 		private final int hash;
 
-		private KeyToken(byte[] token, byte[] address, int port) {
+		private KeyToken(final byte[] token, final byte[] identity) {
```
sbernard31: Details : I don't really understand the benefits of using final keyword for parameters. (this make the signature heavier)

californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
```diff
@@ -804,23 +789,22 @@ public static KeyToken fromOutboundMessage(final Message message) {
 		 * @throws NullPointerException if token or address is {@code null}
 		 * @throws IllegalArgumentException if port &lt; 0 or port &gt; 65535.
 		 */
-		public static KeyToken fromValues(byte[] token, byte[] address, int port) {
-			return new KeyToken(token, address, port);
+		public static KeyToken fromValues(final byte[] token, final byte[] identity) {
```
sbernard31: same here

element-connector/src/main/java/org/eclipse/californium/elements/PrincipalEndpointContextMatcher.java
```diff
@@ -34,6 +36,15 @@ public String getName() {
 		return "principal correlation";
 	}
 
+	@Override
+        public byte[] getEndpointIdentifier(EndpointContext endpointContext) {
+                Principal principal = endpointContext.getPeerIdentity();
+                if (principal == null) {
+                        throw new IllegalArgumentException("principal must be provided!");
+                }
+                return principal.getName().getBytes();
```
sbernard31: I rethink about that this night :p and principal.getName().getBytes() seems not enough to me.
Identifier should contains the type too.  I mean a PreSharedKeyIdentity and RawPublicKeyIdentity are not the same even if both have the same "name".

boaks: Therefore the matcher will report that violation in `isResponseRelatedToRequest`.

Maybe I move the special `UserInfo` processing than to `isToBeSent`.

sbernard31: > Therefore the matcher will report that violation in isResponseRelatedToRequest.

But there is still the issue when we register outbound request.  (in `UdpMatcher.sendRequest`),  or when we register observation (not yet implemented).
I imagine there is poor chance this kind of collision happens but in the meantime this is easy to avoid this just by adding one byte for the type in endpoint identifier.

boaks: RPK has a special prefix ("ni:///sha-256;"), x509 uses a structured DN.
So I guess, if someone uses a PSK identity, which, matches one of the others, he will also find other ways to croc the system :-).

sbernard31: > he will also find other ways to croc the system :-).

I'm not sure, usurp identity seems just super easy in that case...
You are a customer of a iot platform. You sniff traffic and get public key of public certificate of devices. 
Now you register a new device and you choose PSK mode and set the identity with information you get before and you could have same endpoint identifier than devices you sniffed. So all the security is on token generation.

Anyway, if you think this must not be in californium, we are still able to implement this only in Leshan. That's the good point with the way this is implemented :)

boaks: > Now you register a new device and you choose PSK mode and set the identity with information

FMPOV, it's not possible to support "free PSK identity" selection, because this may cause ambiguous identities. So, if the identities are checked anyway, it's easy to filter out the RPK header and check, that no DN pattern is used.

By the way, AFAIK, leshan could not completely reconstruct the Principal from the registry (without extend it more), therefore I implemented a `Lwm2mEndpointContextMatcher`. From the LWM2M TS, it's clear that only 3 kind of principals are supported and so it's easy to add the prefix at the level of that matcher.

But I found an other small leak, if we use PSK with ServerNameIndication, I guess the Identity name must be extended by that ServerName.

boaks:
Replaced by #518

PR #509, opened 2017-12-22
Add simple benchmark to extended plugtest.

boaks:
Add BenchmarkClient for coap and coaps tests.


boaks:
Though only demo-apps is changed, it should not be that big deal.
The basic motivation was to provide a tool to ensure, that introducing the "Token" object doesn't downgrade the performance.

PR #510, opened 2017-12-22
Introduce Token and KeyToken.

boaks:
Introduce Token and KeyToken.

Change API of MessageExchangeStore and ObservationStore accordingly.
To limit the scope of the changes, the KeyTokenFactory is not used but
TODOs are added.
Replace byte array token by Token.

That are the base changes for issue #500.
Based on that, the implementations for variant 2 or 3 will (hopefully)  be easier and isolate the required changes for the different approaches.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -34,12 +34,12 @@
  *    Bosch Software Innovations GmbH - migrate to SLF4J
  *    Achim Kraus (Bosch Software Innovations GmbH) - get default timeout from configuration
  *                                                    of effective endpoint
+ *    Achim Kraus (Bosch Software Innovations GmbH) - replace byte array token by Token
  ******************************************************************************/
 package org.eclipse.californium.core;
 
 import java.io.IOException;
 import java.net.URI;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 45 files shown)


sbernard31:
I'm a bit confused with KeyToken and Token. 
In my head :
- Token is a byte array include in CoAP request and response.
- KeyToken is the key used to match request and response. (KeyToken includes a Token + an endpoint identifier)

But in this PR a Token is KeyToken.
Another confusing point, the KeyTokenFactory takes a Token (which is a KeyToken) to create a KeyToken.
Maybe this is just a naming issue. (or this will be clearer when I will see more code)

KeyTokenFactory also take an endpointContext where I expect the endpoint Identifier return by previous `EndpointContextMatcher`(in #506) but maybe you plan to move or delegate the endpoint identifier creation in KeyTokenFactory ?

boaks:
I just had not enough time to completely work out a solution :-).

Yes, a `Token` is based on the byte array and I wrapped it into a class to prevent to much byte array copies.
A `KeyToken` is a "Token based Key" used to match request and response.
If it's only based on the Token itself or contains additional information as the endpoint identifier, depends then on the `KeyTokenFactory` interface implementation. The basic relation of the `KeyToken` and `Token` is therefore the `getToken()` in the `KeyToken`.
I moved the factory outside the `EndpointContextMatcher`, to isolate it at this point in time more from californium. Currently I'm not sure, if I have enough time to provide a solution triggered by `onContextEstablished` (as my first experiments showed some issues in cancel observation). Though alternatives either would come with a "connect" or only work on the "DTLS server side", I think it is more clear to move such extensions outside of californium and only provide the interface to do so, if the usage is done in such "server only" context. But we will see ...

(The Token implements KeyToken is at this point in time to prevent the create "not required object instanzes".)

boaks:
Withdrawn in favour for #521.

PR #515, opened 2018-01-08
Use daemon thread for dtls receiver.

boaks:


boaks:
I'm not sure, if the none daemon receiver thread was intended. I guess not.

PR #518, opened 2018-01-11
KeyTokenFactory variant 1

boaks:
This variant requires a pre-filled endpoint context, but keeps the most logic as it is.
Only limited useful for "dtls server only usage" or when introducing a "connect".


boaks:
Obsoletes #506

sbernard31:
I will look at this and #519  tomorrow :)

boaks:
Withdrawn in favour for #521.

PR #519, opened 2018-01-11
KeyTokenFactory variant 2

boaks:
This variant create key tokens only after onContextEstablished().
More general, but more changes.
Issue for cancel observes (see comment on issue #500) still not solved. 


boaks:
Withdrawn in favour for #521.

sbernard31:
bye

Bye bye endpoint identifier, I will never forget you :'(

PR #520, opened 2018-01-12
Plain token

boaks:
This variant introduces only a Token and no KeyToken.


boaks:
Withdrawn in favour for #521.

PR #527, opened 2018-01-18
remove etag when request next block2 payload.

rkimsb2:
When we request next block of blockwise payload, it is FIRST TIME request of that block. which means we have no contents of requesting block.

Signed-off-by: Rokwoon Kim <k862390@gmail.com>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -850,10 +850,6 @@ public class BlockwiseLayer extends AbstractLayer {
 						// copy options
 						block.setOptions(new OptionSet(request.getOptions()));
 						block.getOptions().setBlock2(newSzx, false, nextNum);
-						if (response.getOptions().getETagCount() > 0) {
-							// use ETag provided by peer
-							block.getOptions().addETag(response.getOptions().getETags().get(0));
-						}
 
 						// make sure NOT to use Observe for block retrieval
 						block.getOptions().removeObserve();
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
@@ -195,9 +195,9 @@ public class BlockwiseClientSideTest {
 		server.expectRequest(CON, GET, path).storeBoth("A").go();
 		server.sendResponse(ACK, CONTENT).loadBoth("A").block2(0, true, 128).size2(respPayload.length())
 			.etag(etag).payload(respPayload.substring(0, 128)).go();
-		server.expectRequest(CON, GET, path).storeBoth("B").block2(1, false, 128).hasEtag(etag).go();
+		server.expectRequest(CON, GET, path).storeBoth("B").block2(1, false, 128).go();
 		server.sendResponse(ACK, CONTENT).loadBoth("B").block2(1, true, 128).etag(etag).payload(respPayload.substring(128, 256)).go();
-		server.expectRequest(CON, GET, path).storeBoth("C").block2(2, false, 128).hasEtag(etag).go();
+		server.expectRequest(CON, GET, path).storeBoth("C").block2(2, false, 128).go();
 		server.sendResponse(ACK, CONTENT).loadBoth("C").block2(2, false, 128).etag(etag).payload(respPayload.substring(256, 300)).go();
 
 		Response response = request.waitForResponse(RESPONSE_TIMEOUT_IN_MS);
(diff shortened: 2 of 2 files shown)


boaks:
Do you have any issue with the ETAG (of the last response) in the request for the next block?
Which implementation do you use as CoAP server?

rkimsb2:
i tested on coap://coap.me:5683/large with CoapClient.newGet().
and when it request second block, server responsed with 2.03 VALID.

rkimsb2:
And i think even blockwise transfer sequence is also satisfy coap protocol rules.
According to [RFC 7252 Section 5.10.6.2](https://tools.ietf.org/html/rfc7252#section-5.10.6.2), GET request with etag option is working to check entity-tag.
And on [RFC 7959 Section 2.4](https://tools.ietf.org/html/rfc7959#section-2.4) only defines how to works when received response with block2 option and etag.
Beyond these rules, GET request with block2 and etag looks also work as checking entity-tag.

boaks:
So, yes, good catch :-).

When it was implemented, someone wants to make it possible, to transfer the same resource with changed content in different transfers, therefore the ETAG is misused as selector/identifier for the different resource values (see KeyUri). But the specification clearly states.

https://tools.ietf.org/html/rfc7959#section-2.4

> The Block2 Option provides no way for a single endpoint to perform
> multiple concurrently proceeding block-wise response payload transfer
> (e.g., GET) operations to the same resource.  This is rarely a
> requirement, but as a workaround, a client may vary the cache key
> (e.g., by using one of several URIs accessing resources with the same
> semantics, or by varying a proxy-safe elective option).

Using that ETAG selector approach then rises a conflict with

https://tools.ietf.org/html/rfc7252#section-5.10.6.2. 

> A server can issue a 2.03 Valid response (Section 5.9.1.3) in place
> of a 2.05 Content response if one of the ETags given is the entity-
> tag for the current representation, i.e., is valid; the 2.03 Valid
> response then echoes this specific ETag in a response option.

So, if californium uses a ETAG in the request to select the resource version,
a compliant server **CAN** answer with 2.03, and therefore breaks the transfer.
 
FMPOV we now need to know, if someone is using this "none compliant feature".
If so, then we should add a configuration value to enable this feature (and have it disabled by default?).
If nobody is using it, we could clean up the ETAG stuff even in the KeyUri.

So, @all any opinion on that?

sbernard31:
We are not using this ETag feature on our side.

As I understand the blockwise [RFC 7959](https://tools.ietf.org/html/rfc7959#section-2.4).
ETAG should only be added by server to Response :

_"In the latter case, the Block2 Option SHOULD be used in
   conjunction with the ETag Option ([RFC7252], Section 5.10.6), to
   ensure that the blocks being reassembled are from the same version of
   the representation: The server SHOULD include an ETag Option in each
   response.  If an ETag Option is available, the client, when
   reassembling the representation from the blocks being exchanged, MUST
   compare ETag Options.  If the ETag Options do not match in a GET
   transfer, the requester has the option of attempting to retrieve
   fresh values for the blocks it retrieved first."_

ETag is only to ensure we does not mixed several values of a same resource.
Not to be able to do several transfer in parallel.

I think the way we use it is out of the spec, so I feel we should remove it.

What could make sense in californium : 
1. For transparent block transfer, we don't need Etag at server side as we "cache" the value to transfer via block2.
2. At client side, we should use ETag to be sure we don't mixed value.

I think this is what we do currently.

Am I right ?
(Maybe for 1. even with cache this could make sense to add Etag option ?)

boaks:
> 1. 
FMPOV a cache can't solve the issue a value change would cause. When will you update the cached value?

> 2.
Sure, but that's the function of ETAG in the response, so it must be removed in the request, as this PR does it.

> I think this is what we do currently.

Currently we use a ETAG in the request to include it in the KeyUri and use it to select different versions of the same resource.

sbernard31:
> 1.

The cache limits the issue. 
Imagine this scenario without ETAG: 
```
request block 0 -->
<-- return block 0
request block 1 -->
<-- return block 1 
=== value changed ===
request --> block 2
<-- return block2
```
With cache at server side : no problem. (californium implementation)
Without cache at server side  : mixed value at client side. (This is what  happened with wakaama last time I looked at this). In this case ETag help to detect mixed block.

Maybe there is more complicate scenario where ETAG helps, but as the spec say caching the response minimize issue.
_"To minimize the resulting inefficiency, the server MAY cache the current value of a
   representation for an ongoing sequence of requests."_

------------
Anyway I'm probably a bit out of topic.

So short answer and as I said before : **adding ETAG on the request seems to me an error for block mode** so I agree to remove this from request.

boaks:
I'm not sure, if "happy path example" explains, when the cached response should be replaced by a new one. 
But also anyway, just wait a little, if nobody is using that feature, we could remove it and clean-up the KeyUri also.

sophokles73:
When I introduced the ETAG in the blockwise GET requests, the idea was to be able to clearly identify the *ongoing* blockwise transfer that the request belongs to and thus the particular version of the resource representation at the start of the transfer (which is copied to the transfer *cache*). Note that the *transparent blockwise transfer* implicitly creates a *tag* for the resource representation by caching it in the `Block2BlockwiseStatus`.

RFC 7959 doesn't seem to preclude ETAGs being used in block2 requests at all, it just says that the server SHOULD include them. So it seemed reasonable to also include them in the requests and, consequently, in the URI key. However, I wasn't aware at that time that the server could respond with a 2.03 instead of 2.05 so FMPOV we should remove the ETAGs from the block2 requests.

boaks:
OK, thanks!
That was also my understanding of the implementation idea.
So, just wait a couple of days, if someone claims to using it (then go for configurable), if not then we remove it completely.

PR #528, opened 2018-01-18
Introduce integration tests module.

boaks:
Move test from demo-apps/cf-secure into the new integration tests
module. Add tests to ensure, that responses/notifies are dropped on the
server side.

Verify, issue #104 is fixed.


sbernard31:
I agree with creating a new module for integration tests.
LGTM !

Issue #339, opened 2017-06-28
Remove ErrorHandler in scandium


sbernard31:
Now we can get Handshake exception via `MessageCallback.onError`, I think we can remove `ErrorHandler`. See [discussion](https://github.com/eclipse/californium/pull/334#discussion_r124233339).

sbernard31:
I would like to remove `ErrorHandler`, this should not be a big change. Is there any objection ?

sophokles73:
I have no problem with removing the `ErrorHandler`. I agree that `MessageCallback.onError()` is the better approach in that it provides more information about the context in which the ALERT has occurred.

So, now (before we release 2.0.0) would be a good opportunity to get rid of it because it will break existing (but probably mostly unused) API. If you find the time, please go on. Otherwise, I would volunteer to remove it as well, given that I haven't done a lot on Cf recently ;-)

sbernard31:
I will not have time this week, maybe the next one. So if you want you can go on it.

sophokles73:
I'll give it a try ...

sophokles73:
The more I think about the less convinced I am that we should actually remove the `ErrorHandler`. I agree that we should not use it for the same purpose as the `MessageCallback.onError()` method. However, it provides a way to get notified about a peer closing a session/connection which we would have no other means for if we removed the `ErrorHandler`. That said, I would like to rename it to just `AlertHandler`, because that is its sole purpose, and update the JavaDocs accordingly.
In #530 I had started to move the handler to the `org.eclipse.californium.scandium.dtls` package and use the handler kind of *internally* only. However, I now believe that it would be best to keep it in `org.eclipse.californium.scandium` and to keep registering on the `DTLSConnector` instead of the `Connection`. This is because application code will probably never (need to) get a grip on a particular connection/session but will usually need to set up and configure the `DTLSConnector` anyway. So it seems reasonable to me to also set a handler on the connector if an application is interested in notifications that a peer has closed a connection.

sbernard31:
I'm not sure to see the interest about knowing we receive an alert.
I don't know how to use it for now, except maybe for monitoring but in this case logging seems enough to me.

Currently, my feeling is that this is not useful.

But I will not fight to much for this.

@boaks, maybe an opinion about this ?

sophokles73:
At least it seems to be helpful for implementing test cases :-)

sbernard31:
> At least it seems to be helpful for implementing test cases :-)

[...]

PR #530, opened 2018-01-19
[#339] Replace ErrorHandler from DTLSConnector with AlertHandler.

sophokles73:
It looks like the original `ErrorHandler` can be put to good use at the `Connection` level in order to support cases where you want to react to ALERT messages that have been received outside the scope of a handshake or a request (for which we can use the `MessageCallback`).

### DIFF
--- a/scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
+++ b/scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2016 Bosch Software Innovations GmbH and others.
+ * Copyright (c) 2015, 2018 Bosch Software Innovations GmbH and others.
  * 
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -65,6 +65,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.californium.elements.AddressEndpointContext;
+import org.eclipse.californium.elements.EndpointContext;
+import org.eclipse.californium.elements.MessageCallback;
 import org.eclipse.californium.elements.RawData;
 import org.eclipse.californium.elements.RawDataChannel;
 import org.eclipse.californium.elements.util.SimpleMessageCallback;
@@ -243,7 +245,7 @@ public class DTLSConnectorTest {
 		}
 		serverConnectionStore.clear();
 		serverRawDataChannel.setProcessor(serverRawDataProcessor);
-		server.setErrorHandler(null);
+		server.setAlertHandler(null);
 	}
 
 	private static DtlsConnectorConfig newStandardConfig(InetSocketAddress bindAddress) throws Exception {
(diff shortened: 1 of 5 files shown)


boaks:
Can you describe your intention, how to use this `AlertHandler` outside of scandium?

FMPOV only providing a "own" `ResumptionSupportingConnectionStore` would make it possible to use it. Is that the intention?

sbernard31:
My idea was rather to remove ErrorHandler instead of replace it by something else. 🤔 

I'm not sure to see benefits of AlertHandler compared to ErrorHandler ?

sophokles73:
My intention was not so much to provide an alternative to `ErrorHandler` to application developers. However, the `MessageCallback` does not help with detecting a session/connection being closed due to an ALERT message received from the peer *after* the session has been established, i.e. after the handshake has completed. We do have a test case in `DTLSConnectorTest` which verifies that a connection is closed and removed from the store once a fatal alert is received and I did not want to simply drop that test case. But given that I haven't been involved very much over the last months, I might not be aware of another, more desirable way of detecting the fatal alert?

sbernard31:
If this is for internal use only, should we just check that the connection is removed from the store ?
(and so no need to Alert/ErrorHandler)

sophokles73:
> If this is for internal use only, should we just check that the connection is removed from the store ?
(and so no need to Alert/ErrorHandler)

Yes, it is used internally only (at the moment).

> If the issue if about to know when should check that.

That is indeed the issue.

> I think a Thread.sleep could be acceptable, but I could understand that @boaks which try to make the build more stable on hudson/jenkins will not like this so much.
The other solution using a kind of polling strategy like in org.eclipse.californium.TestTools.waitForCondition

I do not really see the problem with having the `AlertHandler` on `Connection`. The problem with the `ErrorHandler` mainly was that it would have been another (worse) way of doing what the `MessageCallback` should now be used for. I agree that we should not have it for that purpose. But with `AlertHandler` we are talking about a different purpose, which BTW the `MessageHandler` cannot be used for. I also do not see why we should fall back to polling or sleeping when we can have a nice and clean notification mechanism instead.

sophokles73:
I have updated the PR according to my latest comment on #339.

I would like to merge it this way soon if nobody else has any strong opinion against it.

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
```diff
@@ -1468,22 +1477,31 @@ public void testConnectorAbortsHandshakeOnUnknownPskIdentity() throws Exception
...
-				assertThat(peerAddress, is(serverEndpoint));
+			public void onSent() {
 			}
-		});
-		client.start();
-		SimpleMessageCallback callback = new SimpleMessageCallback();
+
+			@Override
+			public void onError(Throwable error) {
+				if (HandshakeException.class.isInstance(error)) {
```
boaks: Is `error instanceof HandshakeException` out of style?

sophokles73: No, not out of style :-) I just got used to the style we are using in Hono. But I will happily change it to the style you seem to have gotten accustomed to. However, I would have actually expected some sympathy from you based on your ongoing usage of e.g.
```
if (0 == variable) {
...
```
right? ;-)

sophokles73: I have updated the PR accordingly ...

PR #532, opened 2018-01-22
Fix plugchecker

boaks:
Collection of fixes for the plugchecker.

Add support for secure and TCP connections.
Use LinkFormat to fix parsing of attributes with multiple values. 
Test catalog sort results and doesn't load inner test classes.
Add block2 option of last block to assembled response in transparent
blockwise. Forward response to original request.


californium-core/src/main/java/org/eclipse/californium/core/observe/ObservationUtil.java
```diff
@@ -48,6 +52,12 @@ public static Observation shallowClone(Observation observation) {
 		clonedRequest.setOptions(request.getOptions());
 		clonedRequest.setPayload(request.getPayload());
 		clonedRequest.setUserContext(request.getUserContext());
+		clonedRequest.addMessageObserver(new MessageObserverAdapter() {
+			@Override
+			public void onResponse(Response response) {
+				request.setResponse(response);
+			}
+		});
```
sbernard31: Why do we need this ?
This will make we will have different behavior between a persistent store and the InMemoryObservationStore.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -881,7 +895,7 @@ private void handleBlock2Response(final Exchange exchange, final Response respon
 								status.getBlockCount());
 						Response assembled = new Response(response.getCode());
 						status.assembleReceivedMessage(assembled);
-
+						assembled.getOptions().setBlock2(response.getOptions().getBlock2());
```
sbernard31: Why we need to add this block2 option on the assembled one ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -268,7 +275,12 @@ public void onReadyToSend() {
 					// original request so that at the end of the
 					// blockwise transfer the Matcher can correctly
 					// close the overall exchange
-					request.setToken(block.getToken());
+					if (request.getToken() == null) {
```
sbernard31: Which case do we try to  support here ?

boaks: The case, where the token is supplied by the application. I don't think, that overwriting makes sense, if the token is already set.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: I added it, because the "plugchecker" assumes it. 
FMPOV, test- and analysis-clients benefit from that information.
On the other side, I don't think others are mixed up by not having pure "transparent blockwise".
Alternatively, we can adjust the test to not use "transparent mode", but this would require much more work.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: OK

californium-core/src/main/java/org/eclipse/californium/core/observe/ObservationUtil.java
sbernard31: > With the introduction of the "clustered notify", unfortunately the behaviour of
Request.waitForResponse() was also changed, but the consequences of that change have never been cleaned up in the code and documentation. It would have been great, if that change to "clustered notify" would have been worked out with more details also for the "plain coap" usage.

I take this for me. This is surely deserved :( ... Introduce this new "clustered notifiy" was not easy and have huge impact on the code + very hard to explain. We expected it could be long  that's why we do that in a 2.0.0 experimental branch. When you start to involve in the project you maybe expect that this part of the code was finished but that was not the case. We always been a bit sceptical about  CoAPClient API and we don't use it, so it's harder to us to detect impact. We even start discussion about the need to have 2 API in californium.
But if you found issues about that and you thought that had to be fixed by us, you should have to let us know.  I'm not sure if this is the sense of you remarks :/

> The CoapClient seems to get downside of this implementation (potentially inefficient large number of NotificationListener) and never benefits from clustering, because clustering would break too much of the API of the CoapClient, or do you see that different?

I agree.  (about NotificationListener I'm not sure this is a real problem, do you experiment issue about that?)

> So there are different ways to go:

Why I do not like the idea to have 2 different behavior ? because that's means tests will check only one behavior.

The initial idea was to go with a waitForResponse which only return response (not notification). We forgot to change the javadoc. CoapClient was changed (with NotificationListener) with that idea in mind.

boaks: > I agree. (about NotificationListener I'm not sure this is a real problem, do you experiment issue about that?)

I guess this depends just on the numbers of `NotificationListener` which are registered at a CoapEndpoint.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: I feel strange that the "transparent mode" returns a complete response with block parameter :/

I understand this is the simple way to fix plugchecker, I just played a bit with it and I'm not sure to see the purpose of this plugchecker ? Do we still need it ? What does it bring more than our unit tests ?

boaks: Yes :-).

1. it covers the interoperability test defined for 
    "ETSI IoT CoAP Plugtests, London, UK, 7--9 Mar 2014."
    (OK, the cf-plugtest-client does it similar)

2. it has detected the issues with coaps and blockwise 
    (missing transfer of scheme, ends up in a lot of "intermediary block" requests)

sbernard31: 1. I see that in the javadoc, but does we plan to reuse it ?

2. I understand the more we run code, the more we have chance to detect issue but this is not necessary a good argument to keep that tests on this form. I mean the fact we do not detect this error before means that we should maybe improve our Junit tests instead of counting on plugchecker ?

Anyway, if I well understand :
- plug-client seems to check if a server could be use for checking a client.
- plug-check seems to be used to check a server. 
- plug-server seems to be used to check a client.
(names are strange :p)

So I look at the code of plug-check, especially tests which failed about this line. I'm not sure checking if the block 2 option is present or not is a good way to check we do a block transfer correctly. 
So if we want to "fix" plugchecker, I prefer  :
 a. we just remove the block2 test options ... this is not really less precise than before.
 b. if we really want to check block transfer we need to do the test without transparent block mode.

boaks: > 1. I see that in the javadoc, but does we plan to reuse it ?

The eclipse sandbox offers exactly that "plugtest". 

Currently the plugtest client is used.
> Dec 2017:
> 334 different hosts, 2157 tests
> Jan 2018 (until now):
> 213 different hosts, 1155 tests 

But FMPOV, both plugtest clients are very useful.
There are other possibilities for the two remaining issues:
- block2 option in transparent response
- notify in in `Request.waitForResponse`

As said above, I chose these fixes to keep the efforts limited. But If you feel, that's a wrong priority, I would go for adjust the plugchecker in a first step to support coaps and postpone the fixes.

> that we should maybe improve our Junit tests instead of counting on plugchecker ?

From my personal experience, I disagree. I don't see, that the unit test have a constructed coverage and I don't see, that a constructed coverage is possible with the current number of committers. So I would simple prefer to provide "high level test tools", as the plugchecker, to get aware of issues, which then could be put into a unit test (as I did for the coaps blockwise in #534).

sbernard31: JUnit could be used for High level tests, but ok even if I'm not really convinced, let's keep plug projects.

I would prefer to not return the block2 option for transparent block mode, if you agree.
So what do we do ? a. or b. or something else ?

californium-core/src/main/java/org/eclipse/californium/core/observe/ObservationUtil.java
sbernard31: Any opinion about just changing the javadoc ?

boaks: Sure, if the notifies are not longer supported, then the javadoc should be changed.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: Currently I isolate the changes from the "fixes". For the notifies I add a NotificationListener, which forwards the response again to the origin request, but just for these test. For the blockwise I'm searching for an alternative solution.

boaks:
Use work-arounds for notification processing and blockwise transfer checks.

sbernard31:
LGTM

Issue #484, opened 2017-11-28
Add a TrustedCertificateStore like PskStore and TrustedRpkStore


sbernard31:

sbernard31:
@rkimsb2 I begin to implement this in #533 but we are not totally sure about its usefulness.
I see you +1 this issue, could you explain your use case or why you think this is a good idea ?

rkimsb2:
When client bootstrapping from bootstrap server, there's possibility to change security policy of dtls. If it happened, secured endpoint also have to update it. For now, the way of doing this is only reboot secured endpoint. But rebooting endpoint is some danger for unable to response, and lost messages on UDPConnector's message queue.

boaks:
Does PR #602 solve this issue?

sbernard31:
I think it does.

rkimsb2:
And the point of the comment  i wrote in 26 Jan was changing endpoint's certificate of it's 'own'. It has quite different meaning with this issue.

sbernard31:
So you are more interested by something like #536.
We can close thie one.

PR #533, opened 2018-01-23
Add a TrustedCertificateStore like PskStore and TrustedRpkStore

sbernard31:
This aims to implements #484.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -36,6 +36,8 @@ import java.util.Arrays;
 import java.util.List;
 
 import org.eclipse.californium.scandium.dtls.ServerNameResolver;
+import org.eclipse.californium.scandium.dtls.certstore.StaticCertificateTrustStore;
+import org.eclipse.californium.scandium.dtls.certstore.TrustedCertificateStore;
 import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
 import org.eclipse.californium.scandium.dtls.pskstore.PskStore;
 import org.eclipse.californium.scandium.dtls.rpkstore.TrustAllRpks;
@@ -72,7 +74,6 @@ public final class DtlsConnectorConfig {
 	private static final String EC_ALGORITHM_NAME = "EC";
 
 	private InetSocketAddress address;
-	private X509Certificate[] trustStore;
 
 	/**
 	 * Experimental feature : Stop retransmission at message receipt
@@ -121,6 +122,10 @@ public final class DtlsConnectorConfig {
 
 	/** default is trust all RPKs **/
 	private TrustedRpkStore trustedRPKs;
+
+	/** store for all trusted certificated **/
+	private TrustedCertificateStore trustedCertificates;
+
 
 	private Integer outboundMessageBufferSize;
 
@@ -299,8 +304,19 @@ public final class DtlsConnectorConfig {
 	 * 
 	 * @return the root certificates
 	 */
+	@Deprecated
 	public X509Certificate[] getTrustStore() {
-		return trustStore;
+		return trustedCertificates.getTrustedCertificate();
+	}
+
+	/**
+	 * Gets the trusted root certificates to use when verifying
+	 * a peer's certificate during authentication.
+	 * 
+	 * @return the store which contains the root certificates
+	 */
+	public TrustedCertificateStore getCertificateTrustStore() {
+		return trustedCertificates;
 	}
 
 	/**
@@ -747,11 +763,12 @@ public final class DtlsConnectorConfig {
 		 * @throws NullPointerException if the given array is <code>null</code>
 		 * @throws IllegalArgumentException if the array contains a non-X.509 certificate
 		 */
+		@Deprecated
 		public Builder setTrustStore(Certificate[] trustedCerts) {
 			if (trustedCerts == null) {
 				throw new NullPointerException("Trust store must not be null");
 			} else {
-				config.trustStore = toX509Certificates(trustedCerts);
+				config.trustedCertificates = new StaticCertificateTrustStore(toX509Certificates(trustedCerts));
 				return this;
 			}
 		}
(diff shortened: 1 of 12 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -299,8 +304,19 @@ public PublicKey getPublicKey() {
 	 * 
 	 * @return the root certificates
 	 */
+	@Deprecated
```
sbernard31: I just deprecate all method, but I think we should remove it instead.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -166,7 +166,7 @@ public ClientHandshaker(DTLSSession session, RecordLayer recordLayer, SessionLis
 
 			// we always support receiving a RawPublicKey from the server
 			this.supportedServerCertificateTypes.add(CertificateType.RAW_PUBLIC_KEY);
-			if (rootCertificates != null && rootCertificates.length > 0) {
+			if (!certificateStore.isEmpty()) {
```
sbernard31: I just translate the previous code.
But I doubt this is the good way to do now the store is dynamic.
Maybe we should get  certificates from credentials store only once, then store it in the Handshaker and always use the same.
This sounds better but it will consume more memory if there is lot of Handshakers and a lot of Certificates store in a persistent store without cache.

(if we do that I remove the isEmpty from the store)

boaks:
No objection, but it seems not to cover the issues:
- select credentials based on address (or new alias)
- select validating function (exact match or `CertPathValidator.validate()`)

So, is the intended benefit to make a custom store implementation possible?

sbernard31:
It's just a simple change to make the root certificate dynamic. So yes the idea is to let the possibility to implement custom dynamic store.
(So it does not aims the 2 issues you mention)

@boaks , Any comments on my own comments 🤔  ?

Anyway, I will explore your way for "select credentials based on address (or new alias)" before to integrate this PR.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: I would remove it now.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
boaks: I'm not sure, how dynamic the store is intended to be.
What should happen, if a trusted certificate is removed during an ongoing handshake?

sbernard31: We look if the store is empty to express if we support X509 certificate.
Then later we re-ask to the store the trusted certificates to verify foreign peer certificate.
E.g. :
1. We check the store, it is not empty so we ask for X509 certificate.
2. Later we validate foreign peer certificate,  at this moment  the store is now so we can not validate it.

It's not a big deal,  just a bit strange.

The other points maybe more problematic is about 
https://github.com/eclipse/californium/pull/533/files#diff-5da9149ee37db8bcc492b03743f00dfeR620
This time we will send list of acceptable certificate_authorities then later when we receive the client certificate we could face a situation where the right authorities is not in the store anymore.

boaks:
I simply didn't understand the use-case you want to support!
As you pointed out, there may be several issues when updating the trusts during an handshake, but I'm wondering, why this should be done.
If we copy the trusts at the begin of the handshake calling `X509Certificate[] getTrustedCertificate()` once and reuse those trusts, it seems to be easier.
For session resumption,  I don't know, if we should reuse the initial list or get a new one.

sbernard31:
> I simply didn't understand the use-case you want to support!

Have a dynamic store means being able to change truststore without restart connector or delegate it to another service or plug it on filesystem but I agree this is really hypothetical.

> If we copy the trusts at the begin of the handshake calling X509Certificate[] getTrustedCertificate() once and reuse those trusts, it seems to be easier.

We agree.

> For session resumption, I don't know, if we should reuse the initial list or get a new one.

It seems to me that TrustedCertificated (and rpkStore) should not be needed for resumption. If I am right, that's means that we face a bad design with Handshaker hierarchy. I begin to play with your idea of CredentialStore and I have exactly the same problem. (I mean I need to pass parameter which seems not useful resumption)

boaks:
@sbernard31 

Do you intend this PR to be part of 2.0.0-M8? I would like to test the release-job using 2.0.0-M8 begin next week.

sbernard31:
No, I don't.
I'm not even sure if this should be part of the 2.0.0 version.
I try to explore the CredentialStore which seems to fit a more concrete use case, but I have difficulty to find time on this. We deployed yesterday Leshan 1.0.0-M5 (with Cf 2.0.0-M6) and we still face ExchangeMemory leak. I try to investigate.

boaks:
Thanks for the information!

About the leak: 
I'm not sure, if `BaseMatcher.cancelObserve()` (completing the exchange) in parallel the send the "next block request" may cause the problem. Therefore I would check exchanges for completeness  in `OutboxImpl.sendRequest` (additional to `request.isCanceled()`) and complete such a exchange again.
I planned to extend the BenchmarkClient to verify that, but currently it's hard to find enough time to do so.

sbernard31:
> I'm not sure, if BaseMatcher.cancelObserve() (completing the exchange) in parallel the send the "next block request" may cause the problem.

In our use case and if I didn't missed something we don't use BaseMatcher.cancelObserve().

Last time I checked in details what was stucked in the exchangeStore. that was exchanges with complete observe response (I mean block2 was used, we get all block but the exchange seems to be still in the map)

boaks:
To solve the issue described in #549 the `TrustedCertificateStore` should also offer a `isTrusted` method. With that, someone could set a "TrustAllCertificates" implementation, which returns a empty set of certificates and always true on `isTrusted`.

sbernard31:
replaced by #602.

PR #534, opened 2018-01-23
Fix coaps blockwise

boaks:
Fix several rare blockwise issues.

Copy token and mid for error responses, copy scheme to assembled
blockwise payload (fix using coaps).


boaks:
@sbernard31 

I split the blockwise fixes into this PR, so that #532 could be postponed for now and be discussed with more time.

sbernard31:
LGTM

PR #536, opened 2018-01-29
Add a credentials store

sbernard31:
Add a credentials store to be able to change credentials depending on foreign peer.

This PR is not really polished, this is version with minimal changing.
This aims to start more technical discussion about it.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -138,6 +138,7 @@ import org.eclipse.californium.scandium.dtls.SessionCache;
 import org.eclipse.californium.scandium.dtls.SessionListener;
 import org.eclipse.californium.scandium.dtls.SessionTicket;
 import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
+import org.eclipse.californium.scandium.dtls.credentialsstore.CredentialsConfiguration;
 import org.eclipse.californium.scandium.util.ByteArrayUtils;
 
 import eu.javaspecialists.tjsn.concurrency.stripedexecutor.StripedExecutorService;
@@ -1097,8 +1098,10 @@ public class DTLSConnector implements Connector {
 		DTLSSession newSession = new DTLSSession(record.getPeerAddress(), false, record.getSequenceNumber());
 		// initialize handshaker based on CLIENT_HELLO (this accounts
 		// for the case that multiple cookie exchanges have taken place)
+		CredentialsConfiguration credConfig = config.getCredentialsStore()
+				.getCredentialsConfiguration(newSession.getPeer());
 		Handshaker handshaker = new ServerHandshaker(clientHello.getMessageSeq(), newSession,
-				getRecordLayerForPeer(peerConnection), peerConnection, config, maximumTransmissionUnit);
+				getRecordLayerForPeer(peerConnection), peerConnection, config, credConfig, maximumTransmissionUnit);
 		addSessionCacheSynchronization(handshaker);
 		handshaker.processMessage(record);
 	}
@@ -1129,8 +1132,10 @@ public class DTLSConnector implements Connector {
 			final DTLSSession sessionToResume = new DTLSSession(clientHello.getSessionId(), record.getPeerAddress(),
 					ticket, record.getSequenceNumber());
 
+			CredentialsConfiguration credConfig = config.getCredentialsStore()
+					.getCredentialsConfiguration(sessionToResume.getPeer());
 			final Handshaker handshaker = new ResumingServerHandshaker(clientHello.getMessageSeq(), sessionToResume,
-					getRecordLayerForPeer(peerConnection), peerConnection, config, maximumTransmissionUnit);
+					getRecordLayerForPeer(peerConnection), peerConnection, config, credConfig, maximumTransmissionUnit);
 			addSessionCacheSynchronization(handshaker);
 
 			if (previousConnection.hasEstablishedSession()) {
@@ -1290,11 +1295,13 @@ public class DTLSConnector implements Connector {
 			}
 			// no session with peer established yet, create new empty session &
 			// start handshake
+			CredentialsConfiguration credConfig = config.getCredentialsStore().getCredentialsConfiguration(peerAddress);
 			Handshaker handshaker = new ClientHandshaker(
 					DTLSSession.newClientSession(peerAddress, message.getEndpointContext().getVirtualHost()),
 					getRecordLayerForPeer(connection),
 					connection,
 					config,
+					credConfig,
 					maximumTransmissionUnit);
 			addSessionCacheSynchronization(handshaker);
 			handshaker.addSessionListener(newDeferredMessageSender(message));
@@ -1309,8 +1316,10 @@ public class DTLSConnector implements Connector {
 			Connection newConnection = new Connection(peerAddress, config.getAutoResumptionTimeoutMillis());
 			terminateConnection(connection, null, null);
 			connectionStore.put(newConnection);
-			Handshaker handshaker = new ResumingClientHandshaker(resumableSession,
-					getRecordLayerForPeer(newConnection), newConnection, config, maximumTransmissionUnit);
+			CredentialsConfiguration credConfig = config.getCredentialsStore()
+					.getCredentialsConfiguration(session.getPeer());
+			Handshaker handshaker = new ResumingClientHandshaker(resumableSession, getRecordLayerForPeer(newConnection),
+					newConnection, config, credConfig, maximumTransmissionUnit);
 			addSessionCacheSynchronization(handshaker);
 			handshaker.addSessionListener(newDeferredMessageSender(message));
 			handshaker.startHandshake();
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -685,6 +685,17 @@ public final class DtlsConnectorConfig {
 		}
 
 		/**
+		 * Sets credentials Store use to be able to have different credentials depending on foreign peers
+		 * @param credStore
+		 *            the credentials store
+		 * @return this builder for command chaining
+		 */
+		public Builder setCredentialsStore(CredentialsStore credStore) {
+			config.credentialsStore = credStore;
+			return this;
+		}
+
+		/**
 		 * Sets the connector's identifying properties by means of a private
 		 * and public key pair.
 		 * <p>
@@ -971,9 +982,6 @@ public final class DtlsConnectorConfig {
 			if (config.clientAuthenticationRequired == null) {
 				config.clientAuthenticationRequired = true;
 			}
-			if (config.sendRawKey == null) {
-				config.sendRawKey = true;
-			}
 			if (config.outboundMessageBufferSize == null) {
 				config.outboundMessageBufferSize = 100000;
 			}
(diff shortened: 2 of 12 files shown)


boaks:
FMPOV, the leak issue in #429 consumes all my time I can invest in californium for release 2.0.0.
So I'm not able to work on this, sorry.

sbernard31:
I understand. Does it means that I should not even rely on a review from you ?

boaks:
Yes, in fact, the plan was to release 2.0.0 next week. 
So this will go to 2.1. and will be discussed after 2.0.0.

rkimsb2:
+1

sbernard31:
I rebased it on 2.0.x, I will start an experimental branch in leshan to test if this makes the job.

sbernard31:
I added support of RPK to Leshan client thx to this PR
See https://github.com/eclipse/leshan/pull/538 as example of usage.

My first feeling is pretty good.

@boaks, I would really appreciate if we could consider to integrate this for the 2.0.0 release. (Not for the 2.0.0-M11 of course)

Next step support of x509.

boaks:
FMPOV, the reuse of a client-side-connector for multiple destinations, comes with the risk, that it fails, if the servers are potentially co-locate on the same server-side-endpoint. This PR would not change the possibility to do so, it makes it just easier to do and so more probable.

Maybe server-co-location is not too much used, but, on the other side, I'm not sure, how many different destinations will be used by one client. If it's only a small number, then there are also alternatives.
One alternative would be consider to optimize the Connectors to be used as "Set". This would eliminate that nasty "co-location" pitfall.
  
In a first step, I have reused the DTLS timer to be also the executor, if only 1 thread is used (I will update PR #699 tomorrow). This will lower the resources per DTLS connector.  
A second step would be, to prepare the connectors to share executors (may be just for sending, or, using nio/async io, also for receiving).

sbernard31:
I'm sorry :-/ but I can not understand the server-co-location issue ? are you talking about virtualhost(SNI) ?

boaks:
> I'm sorry :-/ but I can not understand the server-co-location issue ? are you talking about virtualhost(SNI) ?

Yes, "connecting" to two servers but ends up at the same server-endpoint. Basically this may lead to different issues before, and would fail even earlier. But if it somehow succeeds, then you will not be able to decrypt the messages, if your connected to that endpoint twice.

sbernard31:
If this is the only issue, we could start with a first version of `CredentialsStore` which is not compatible with SNI ?
(I can add a comment and/or raise an exception if a custom `CredentialsStore` is used and SNI is enabled in the DtlsConnectorConfig builder)

We will be able to  add SNI support later, like we did for PSKStore.

boaks:
> If this is the only issue, we could start with a first version of CredentialsStore which is not compatible with SNI ?

That's a misunderstanding! If that "virtual"/"co-located" stuff is intended to be supported, then this PR will bring not too much benefits at all, but may aggravate the pitfalls.

> Maybe server-co-location is not too much used, but, on the other side, I'm not sure, how many different destinations will be used by one client. If it's only a small number, then there are also alternatives.

So, FMPOV, the question is: how many destinations are intended for one client? I think this number is low enough, to go for the save variant of using different client-endpoints for different destinations. And so I would consider to optimize the use of a connectors instead.
This may also be more useful for https://github.com/eclipse/leshan/issues/491

sbernard31:
Could you explain which problems this will bring exactly ?
Currently, I only see the benefits which is that this simplify a lot the implementation of RPK and x509 support for Leshan client.

boaks:
client c-ep1  ---- handshake --> server1 (actually s-ep_common), establish session 1
client c-ep1  ---- handshake --> server2 (also actually s-ep_common), establish session 2

client c-ep1 ---- request --> server 1 (s-ep-common)
client c-ep1 <-- response -- server 1  (s-ep-common)

now the client doesn't know, which session it has to sue to decrypt the response. its identified by the "s-ep-common" but this is the same as for server 2 / session 2.

With TLS, this works, because the second handshake will use it's own connection.
Therefore, if DTLS is intended to be used with virtual server, the client need a own endpoint / connector to communicate with servers.

sbernard31:
So this is the case where, we are using SNI and, the client tries to talk to 2 different servers behind the same IP/port.
(Still in draft but [connection ID](https://tools.ietf.org/html/draft-ietf-tls-dtls-connection-id-01) should help)

I don't know if this is a common use case.
Do you have this kind of use case on your side ?

So, I think this still makes sense to provide an API like this even if this does not support this very particular use case. This API could also be used to change credentials dynamically without restarted endpoint.

By the way, I already tried to implement the "1 endpoint by server" solution a long time ago (https://github.com/eclipse/leshan/pull/450) and you oriented me towards this kind of solution (https://github.com/eclipse/leshan/pull/450#issuecomment-352011524) and now I find this very 
elegant. :-)

boaks:
> By the way, I already tried to implement the "1 endpoint by server" solution a long time ago (eclipse/leshan#450) and you oriented me towards this kind of solution (eclipse/leshan#450 (comment)) and now I find this very
> elegant. :-)
> 

Investigating in issue #677 I got aware, that the "virtual" stuff requires different connections, for both
DTLS (SNI) and in CoAP (URI-HOST, see https://tools.ietf.org/html/rfc7252#section-5.10.1). You can't communicate from one client endpoint to two server endpoint, which are virtual, but co-located on the same host. The client would assume communication artefacts (MID or MAC), which are unique per endpoint and so co-locating the physical-endpoint of virtual-endpoints will fail. SNI or URI-HOST are just helpers to overcome the pain caused by virtual hosts. The real issue would be the use of different DNS entries pointing to the same host. So If you can't ensure, that this DNS stuff is done, your communication may fail. And therefore, since one or two weeks, I started to think, that on the client-side we need different connectors as different endpoints for different destinations.   

> (Still in draft but connection ID should help)
It will not help for DTLS 1.2, because the server side will not use a cid (page 7, figure 4). 
So the client can't use a cid to distinguish both sessions.
 
> I don't know if this is a common use case.
> Do you have this kind of use case on your side ?

No, but though the "virtual" stuff is included in DTLS (SNI) and in CoAP (URI-HOST), and the real pain is caused by DNS entries to the same host, I think this requires either require an explicit exclusion or the "multi connector" approach.   

But you ask questions without answering my question:
How many destinations are intended for a client?

I still guess, this are only a few connections.

And so my conclusion is:

Optimize the use of multiple connectors instead!
In a first step, I have reused the DTLS timer to be also the executor, if only 1 thread is used (I will update PR #699 tomorrow). This will lower the resources per DTLS connector.
A second step would be, to prepare the connectors to share executors (may be just for sending, or, using nio/async io, also for receiving).
And this may also be more useful for eclipse/leshan#491

sbernard31:
> But you ask questions without answering my question:
How many destinations are intended for a client?

Minimum 1 device management server and probably 1 bootstrap server, I think in real world not so much more (even if the lwm2m spec allow up to 65535 server for 1 client)

> Optimize the use of multiple connectors instead!

I'm not really interested in optimizing resource for now but mainly by implementing RPK, x509 at lwm2m client side (with bootstrap of course). This PR helps to do that in an elegant way with the only pitfall that it does not work for co-located server which sounds to be a really particular use case.

> In a first step, I have reused the DTLS timer to be also the executor, if only 1 thread is used (I will update PR #699 tomorrow). This will lower the resources per DTLS connector.
A second step would be, to prepare the connectors to share executors (may be just for sending, or, using nio/async io, also for receiving).
And this may also be more useful for eclipse/leshan#491

This PR does not prevent all of this.

boaks:
> This PR does not prevent all of this.

Sure, but as I already wrote above: 
If that "virtual"/"co-located" stuff is intended to be supported, then this PR will bring not too much benefits at all, but may aggravate the pitfalls.

rkimsb2:
Could I join this issue?

As the disscusion above all, also I didn't find any clues for recognizing server name on exchanging message after handshaking.
But then, Does SNI extension has meaning for DTLS?

First of all, with TLS, this will be work as like @boaks mentioned.
But Dtls doesn't working because of the problem as mentioned above.

Second of all, also I didn't find any clues that the port(endpoint) should be separated when DTLS with SNI extension used.

IMHO, the DTLS spec doesn't explain then SNI extention enoughly.
Could we ask to IETF about this issue?

boaks:
> Could I join this issue?

Sure, your welcome :-).

> But then, Does SNI extension has meaning for DTLS? 

That's also, what I ask myself. AFAIK the intention to support SNI was to make the (PSK) credentials tenant aware. But with #677 and URI-HOST option, which also targets the "virtual server", I'm not longer sure, if this limitation to tenants is really good. 

> But DTLS doesn't working because of the problem as mentioned above.  

The point is: **it works, if the client uses different endpoints for different destinations!**
And that's true for both DTLS and CoAP, the client must use different endpoints!
And therefore, just since one or two weeks, I changed my own view of using CoAP and DTLS.
And that's now the main topic in this discussion, because @sbernard31 invested a lot of time,
but for me, in the meantime, this goes into a wrong direction with pitfalls, if such "virtual" stuff is then used. And as I wrote above, FMPOV SNI and URI-HOST are just helpers to use "virtual host", so what must be clear, if a client should use only one endpoint for different destinations, is that "virtual hosts" are not used at all!

> IMHO, the DTLS spec doesn't explain the SNI extension enoughly.
Could we ask to IETF about this issue?

Sure, but sometimes it takes longer to get answers. :-).
So, do you want to write to the IETF core-mailing-list?

And what's you impression, go for these changes here in this PR and risk, that it fails, if "virtual host" are getting used?
Or work on improving to use sets of connectors more efficiently?
Or try to wait for an answer?

rkimsb2:
What I wanted to talk about is that if SNI extension is for both of TLS and DTLS, it also work as same way on both of it. but actually it does not. On TLS, it can used by single endpoint with multiple virtual server. And on DTLS, it must be separate different endpoint. This is what I feel weird and want to check to IETF. Also because of it's undefinition of RFC docs.

rkimsb2:
Am I have a wrong point of view?
is SNI actually work as same way on TLS and DTLS?

boaks:
> On TLS, it can used by single endpoint with multiple virtual server. 

No, that's a misunderstanding! 
If you connect to multiple servers using TLS, you need multiple TLS connections!
That includes, that the client implicitly create a new endpoint for each connection (endpoint := ip-address+port). The californium TlsClientConnector hides this, but in fact, it alos opens a TLS connection for each destination and therefore creates a new endpoint.

boaks:
> Am I have a wrong point of view?
> is SNI actually work as same way on TLS and DTLS?

FMPOV, it works in the same way, but requires the client to use explicit different endpoints, what is implicit done in the same way for TLS.

rkimsb2:
Thank you @boaks. Now I got the point.

boaks:
OK, so what should now be the next steps?

Write to the IETF core list? 
Or do we agree, that id "virtual hosts" are used, coap clients must use explicit different endpoint?

Should this PR be reviewed anyway? It doesn't introduce this issue nor does it prevent a solution. But it may aggravate this misuse. So add clear warnings will be one action, which would be required for this PR.

rkimsb2:
> Write to the IETF core list?
No, I think this is so far from focus for now.
> Or do we agree, that id "virtual hosts" are used, coap clients must use explicit different endpoint?
Yes. I think the first concept(multiple server for single endpoint) was wrong point.

> Should this PR be reviewed anyway?
I think it's No. 

And then, what my first focus about this issue was reusing endpoint when the connection has changed to another server.
But now it seems that this work should be done by shutdown old endpoint and create new one, including security policy.
These works are all out of boundary of endpoint's works.

boaks:
> The limitation is: the client must not talk to several servers which are behind the same "real host".

Exactly.

> do not hesitate to share your thought as soon as possible

My point here is, that "share" mostly mean "time consuming discussions", and, though the hono-coap-adapter is behind my timelines, I had to focus on that work.

By the way, if your workmates think it's a good solution, I have no general objection, 
I just think, it doesn't really pay off. And as I wrote, as long as a warning comment is added to the stuff in californium, you may use it for the leshan client, if you think it's a good idea.

rkimsb2:
To @sbernard31 

Actually, I made some private branch of old version leshan(2.0.0-M3).  In case of client, i have to make it as possible to connect multiple server, and modified it myself.

If this issue is closed in any of way, and multiple registration is possible, I think I can move to latest version.

Currently, my private branch shutdown endpoint every time for updating security policy.
This is my current revision of my usage.

sbernard31:
I revive the 1 endpoint by server solution for Leshan, so I think we can close this PR.

PR #539, opened 2018-02-01
Observe benchmark

boaks:
Add test for reverse observe. 
ExtendedTestServer is extended with the ReverseObserve resource, which registers a observation on the multiple client of the BenchmarkTest.
The leak in the message exchange store seems to be related to race conditions. Without additionally logging it seems not to appear. Using

> coap://localhost:5783/reverse-observe?obs=25&res=feed&rlen=400 50 3 x 500

result (on my machine) in:

> 09:54:49.027 INFO [ReverseObserve]: 0 observes, 0 by peers
> 09:54:49.028 INFO [ReverseObserve]: 15063 notifies overalls
> 09:54:49.388 TRACE [InMemoryMessageExchangeStore]: MessageExchangeStore contents: 8 exchanges by MID, 8 exchanges by token, 
> 09:54:49.404 TRACE [InMemoryMessageExchangeStore]: MessageExchangeStore contents: 0 exchanges by MID, 0 exchanges by token, 
> 09:54:49.669 TRACE [InMemoryMessageExchangeStore]: MessageExchangeStore contents: 0 exchanges by MID, 0 exchanges by token, 
> 09:54:49.732 TRACE [InMemoryMessageExchangeStore]: MessageExchangeStore contents: 0 exchanges by MID, 0 exchanges by token, 

The one with the 8 exchanges is the used one, the other stores are from unused endpoints.
(I will create a additional PR improve the logging, which makes it easier to see, what happens :-).

Signed-off-by: Achim Kraus <achim.kraus@bosch-si.com>

PR #540, opened 2018-02-01
Enable status logging triggered by first put.

boaks:
Add logging for put exchanges and exchange dumps for status.

Together with PR #539 it shows (on my machine) the leaks in the message exchange store.

> 10:21:11.532 INFO [InMemoryMessageExchangeStore]: MessageExchangeStore contents: 13 exchanges by MID, 54 exchanges by token, 0 MIDs, 
> 10:21:11.532 INFO [InMemoryMessageExchangeStore]:   KeyMID[52028, [7f000001]:50599], complete true, retransmission 0, + CON-GET    MID=51913, Token=[8f269d7bfbda3946], OptionSet={"Observe":0, "Uri-Path":"feed", "Uri-Query":"rlen=400"}, no payload, null
> ...
> 10:21:11.532 INFO [InMemoryMessageExchangeStore]:   Token=[7084e27ba473dfb9], complete true, retransmission 0, org Token=[2138d80b39e0571b], CON-GET    MID=31064, Token=[f44c5c812c6b1cb4], OptionSet={"Uri-Path":"feed", "Uri-Query":"rlen=400", "Block2":"(szx=2/64, m=false, num=4)"}, no payload, null
> 

So it shows, that exchange doesn't contain anymore the KeyMID or Token.
 
### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
@@ -30,14 +30,17 @@
  *                                                    of provider. Remove not longer
  *                                                    required releaseToken.
  *    Achim Kraus (Bosch Software Innovations GmbH) - replace byte array token by Token
+ *    Achim Kraus (Bosch Software Innovations GmbH) - start status logging with first
+ *                                                    stored exchange. 
+ *                                                    Add exchange dump to status.
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executors;
@@ -46,7 +49,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import org.eclipse.californium.core.coap.BlockOption;
 import org.eclipse.californium.core.coap.Message;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Token;
@@ -58,7 +61,6 @@ import org.eclipse.californium.elements.util.DaemonThreadFactory;
 
 /**
  * A {@code MessageExchangeStore} that manages all exchanges in local memory.
- *
  */
 public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 
@@ -67,6 +69,7 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 	private final ConcurrentMap<KeyMID, Exchange> exchangesByMID = new ConcurrentHashMap<>();
 	// for outgoing
 	private final ConcurrentMap<Token, Exchange> exchangesByToken = new ConcurrentHashMap<>();
+	private volatile boolean enableStatus;
 
 	private final NetworkConfig config;
 	private final TokenGenerator tokenGenerator;
@@ -84,7 +87,7 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 	 */
 	public InMemoryMessageExchangeStore(final NetworkConfig config) {
 		this(config, new RandomTokenGenerator(config));
-		LOGGER.info("using default TokenProvider {}", RandomTokenGenerator.class.getName());
+		LOGGER.debug("using default TokenProvider {}", RandomTokenGenerator.class.getName());
 	}
 
 	/**
@@ -111,14 +114,16 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 		final int healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL, 60); // seconds
 		// this is a useful health metric
 		// that could later be exported to some kind of monitoring interface
-		if (LOGGER.isTraceEnabled()) {
+		if (healthStatusInterval > 0 && LOGGER.isInfoEnabled()) {
 			this.scheduler = Executors
 					.newSingleThreadScheduledExecutor(new DaemonThreadFactory("MessageExchangeStore"));
 			statusLogger = scheduler.scheduleAtFixedRate(new Runnable() {
 
 				@Override
 				public void run() {
-					LOGGER.trace(dumpCurrentLoadLevels());
+					if (enableStatus) {
+						dump(5);
+					}
 				}
 			}, healthStatusInterval, healthStatusInterval, TimeUnit.SECONDS);
 		}
@@ -128,6 +133,7 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 		StringBuilder b = new StringBuilder("MessageExchangeStore contents: ");
 		b.append(exchangesByMID.size()).append(" exchanges by MID, ");
 		b.append(exchangesByToken.size()).append(" exchanges by token, ");
+		b.append(deduplicator.size()).append(" MIDs, ");
 		return b.toString();
 	}
 
@@ -192,33 +198,39 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 
 	private int registerWithMessageId(final Exchange exchange, final Message message) {
 
+		enableStatus = true;
 		int mid = message.getMID();
 		if (Message.NONE == mid) {
 			mid = assignMessageId(message);
 			if (Message.NONE != mid) {
 				KeyMID key = KeyMID.fromOutboundMessage(message);
 				if (exchangesByMID.putIfAbsent(key, exchange) != null) {
-					throw new IllegalArgumentException(String
-							.format("automatic message ID [%d] already in use, cannot register exchange", message.getMID()));
+					throw new IllegalArgumentException(String.format(
+							"generated mid [%d] already in use, cannot register exchange", message.getMID()));
 				}
+				LOGGER.debug("exchange added with generated mid {}, {}", key, message);
 			}
 		} else {
-			Exchange existingExchange = exchangesByMID.putIfAbsent(KeyMID.fromOutboundMessage(message), exchange);
+			KeyMID key = KeyMID.fromOutboundMessage(message);
+			Exchange existingExchange = exchangesByMID.putIfAbsent(key, exchange);
 			if (existingExchange != null) {
 				if (existingExchange != exchange) {
 					throw new IllegalArgumentException(String
-							.format("message ID [%d] already in use, cannot register exchange", message.getMID()));
+							.format("mid [%d] already in use, cannot register exchange", message.getMID()));
 				} else if (exchange.getFailedTransmissionCount() == 0) {
 					throw new IllegalArgumentException(String.format(
-							"message with already registered ID [%d] is not a re-transmission, cannot register exchange",
+							"message with already registered mid [%d] is not a re-transmission, cannot register exchange",
 							message.getMID()));
 				}
+			} else {
+				LOGGER.debug("exchange added with {}, {}", key, message);
 			}
 		}
 		return mid;
 	}
 
 	private void registerWithToken(final Exchange exchange) {
+		enableStatus = true;
 		Request request = exchange.getCurrentRequest();
 		Token token = request.getToken();
 		if (token == null) {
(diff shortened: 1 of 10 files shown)


boaks:
@sbernard31 
@sophokles73 

Though we discussed already about that dump, but in combination with "isEmptyWithDump", I would prefer to get some feedback. If you still insist on a "external test only dump" implementation, I would close this PR, but I would also not care, if you could reproduce the test logs. So it's just up to you.

The isEmptyWithDump could stay in the test stuff and would then call the dump().

sbernard31:
Sry, but I'm not sure to understand : "If you still insist on a "external test only dump" implementation, I would close this PR, but I would also not care, if you could reproduce the test logs."

boaks:
As far as I remember, you didn't want the `dump` and `dumpExchanges` to be included in  `InMemoryMessageExchangeStore`, and therefore you moved them into the test stuff. 
If this is still the case, this PR could not be merged. 
Not merging is no big thing, but I will use it local for debugging. If so, my logs will be different from yours.
So:
Should the dumps stay in test-only?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -192,18 +199,21 @@ public int assignMessageId(final Message message) {
 
 	private int registerWithMessageId(final Exchange exchange, final Message message) {
 
+		enableStatus = true;
```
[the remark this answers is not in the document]
boaks: Yes, currently, without more sophisticated arguments for the demo's, too many "empty endpoints" are created. This will keep these silent.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -213,29 +223,43 @@ private int registerWithMessageId(final Exchange exchange, final Message message
...
 				return;
 			}
-			if (exchangesByToken.put(token, exchange) != exchange) {
-				if (!(exchange.getFailedTransmissionCount() > 0 || request.getOptions().hasBlock1()
-						|| request.getOptions().hasBlock2() || request.getOptions().hasObserve())) {
+			Exchange previous = exchangesByToken.put(token, exchange);
+			if (previous == null) {
+				BlockOption block2 = request.getOptions().getBlock2();
+				if (block2 != null && block2.getNum() > 1) {
+					LOGGER.warn("New exchange for ongoing blockwise transfer: {}, block {}", token, block2.getNum());
```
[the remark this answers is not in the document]
boaks: Hm, yes that's overdone. It's much more for debugging the current leaks. I will add a TODO to remove that afterwards.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -358,6 +381,52 @@ public synchronized void stop() {
 		}
 	}
 
+	/**
+	 * Dump exchanges of stores.
+	 * 
+	 * @param logMaxExchanges maximum number of exchanges to include in dump.
+	 */
+	public void dump(int logMaxExchanges) {
```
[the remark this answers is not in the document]
boaks: Sure, when the PR is agreed, the clean up will be added to this PR.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -192,18 +199,21 @@ public int assignMessageId(final Message message) {
...
 			mid = assignMessageId(message);
 			if (Message.NONE != mid) {
 				KeyMID key = KeyMID.fromOutboundMessage(message);
 				if (exchangesByMID.putIfAbsent(key, exchange) != null) {
-					throw new IllegalArgumentException(String
-							.format("automatic message ID [%d] already in use, cannot register exchange", message.getMID()));
+					throw new IllegalArgumentException(String.format(
+							"automatic message ID [%d] already in use, cannot register exchange", message.getMID()));
 				}
+				LOGGER.debug("exchange added with generated mid {}, {}", key, message);
```
sophokles73: this feels more like TRACE but I do not really know the context in which you need this info ...

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -213,29 +223,43 @@ private int registerWithMessageId(final Exchange exchange, final Message message
...
 	private void registerWithToken(final Exchange exchange) {
+		enableStatus = true;
 		Request request = exchange.getCurrentRequest();
 		Token token = request.getToken();
 		if (token == null) {
 			do {
 				token = tokenGenerator.createToken(false);
 				request.setToken(token);
 			} while (exchangesByToken.putIfAbsent(token, exchange) != null);
+			LOGGER.debug("exchange added with gnerated token {}, {}", token, request);
```
sophokles73: spelling, TRACE?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -213,29 +223,43 @@ private int registerWithMessageId(final Exchange exchange, final Message message
...
 				return;
 			}
-			if (exchangesByToken.put(token, exchange) != exchange) {
-				if (!(exchange.getFailedTransmissionCount() > 0 || request.getOptions().hasBlock1()
-						|| request.getOptions().hasBlock2() || request.getOptions().hasObserve())) {
+			Exchange previous = exchangesByToken.put(token, exchange);
+			if (previous == null) {
+				BlockOption block2 = request.getOptions().getBlock2();
+				if (block2 != null) {
+					LOGGER.debug("block2 exchange for block {} add with token {}", block2.getNum(), token);
```
sophokles73: TRACE?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -358,6 +381,52 @@ public synchronized void stop() {
 		}
 	}
 
+	/**
+	 * Dump exchanges of stores.
+	 * 
+	 * @param logMaxExchanges maximum number of exchanges to include in dump.
+	 */
+	public void dump(int logMaxExchanges) {
+		if (LOGGER.isInfoEnabled()) {
```
sophokles73: Who will be calling this method and in what context?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -358,6 +381,52 @@ public synchronized void stop() {
...
+	}
+
+	/**
+	 * Dump collection of exchange entries.
+	 * 
+	 * @param logMaxExchanges maximum number of exchanges to include in dump.
+	 * @param exchangeEntries collection with exchanges entries
+	 */
+	private <K> void dumpExchanges(int logMaxExchanges, Set<Entry<K, Exchange>> exchangeEntries) {
+		for (Entry<K, Exchange> exchangeEntry : exchangeEntries) {
```
sophokles73: shouldn't this be guarded by a check for the log level as well?

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
```diff
@@ -36,8 +44,33 @@
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(InMemoryObservationStore.class.getName());
 	private final ConcurrentMap<Token, Observation> map = new ConcurrentHashMap<>();
+	private volatile boolean enableStatus;
 
-	public InMemoryObservationStore() {
+	public InMemoryObservationStore(NetworkConfig config) {
+		
+		if (LOGGER.isInfoEnabled()) {
```
sophokles73: INFO would be one of the adequate production log levels FMPOV. How do you prevent the excessive logging in this case? Maybe a flag in NetworkConfig for enabling/disabling this would be more appropriate than just guarding by log level?

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
```diff
@@ -90,8 +125,11 @@ public Observation get(Token token) {
 	@Override
 	public void remove(Token token) {
 		if (token != null) {
-			map.remove(token);
-			LOGGER.debug("removed observation for token {}", token);
+			if (map.remove(token) != null) {
+				LOGGER.debug("removed observation for token {}", token);
```
sophokles73: TRACE?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: Searching for exchange leaks, I felt it was debugging :-),

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: The health status.
And unit tests.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: The method is not intended to be called outside of the dump().

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
boaks: We may use HEALTH_STATUS_INTERVAL of 0 to disable it.

boaks: Or use a different LOGGER?

boaks:
The general questions seems:

Is debugging intended to maintain extra logging changes in separate branches and frequently merge them in (which comes with more work), or could it be added to the "release branches" and speed up debugging.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -111,14 +115,16 @@ private void startStatusLogging() {
 		final int healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL, 60); // seconds
 		// this is a useful health metric
 		// that could later be exported to some kind of monitoring interface
-		if (LOGGER.isTraceEnabled()) {
+		if (healthStatusInterval > 0 && LOGGER.isInfoEnabled()) {
```
sophokles73: is `NetworkConfig.Keys.HEALTH_STATUS_INTERVAL` used in other places as well to control debugging output? If so, is this check for > 0 compatible with the usage in the other places?

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
```diff
@@ -201,7 +201,7 @@ public static void setDefaults(final NetworkConfig config) {
 		config.setInt(Keys.HTTP_CACHE_RESPONSE_MAX_AGE, 86400);
 		config.setInt(Keys.HTTP_CACHE_SIZE, 32);
 
-		config.setInt(Keys.HEALTH_STATUS_INTERVAL, 60); // s
+		config.setInt(Keys.HEALTH_STATUS_INTERVAL, 60); // s, 0 for ddisable
```
sophokles73: spelling

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
```diff
@@ -36,8 +44,33 @@
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(InMemoryObservationStore.class.getName());
 	private final ConcurrentMap<Token, Observation> map = new ConcurrentHashMap<>();
+	private volatile boolean enableStatus;
 
-	public InMemoryObservationStore() {
+	public InMemoryObservationStore(NetworkConfig config) {
+		int healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL, 60); // seconds
+		
+		if (healthStatusInterval > 0 && LOGGER.isInfoEnabled()) {
```
sophokles73: this is what I meant above regarding the *compatible use of the check for > 0*. However, it seems that we either want to dump the info everywhere or nowhere at all, right?

sophokles73:
> Is debugging intended to maintain extra logging changes in separate branches and frequently merge them in (which comes with more work), or could it be added to the "release branches" and speed up debugging.

I am not sure if I understand the question. Are you asking whether we can/should merge this PR to 2.0.x or keep it on a separate branch? If that is indeed the question, I would rather have it on 2.0.x. FMPOV the only important thing is that users can enable/disable the additional logging e.g. by means of configuration properties.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: Currently it's used for the health_status in 
InMemoryMessageExchangeStore
InMemoryObservationStore
And in the demo-app ExtendedTestServer for the status of the "reverse observes".
So disable it everywhere with 0 seems possible without unwanted side effects.

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
boaks: Yes, that's my current intention. This logging is just a util for the "main work", the debugging of the message exchange leak.

boaks:
> I would rather have it on 2.0.x.

OK, I prefer this also.

> FMPOV the only important thing is that users can enable/disable the additional logging e.g. by means of configuration properties.

Should now be possible with HEALTH_STATUS_INTERVAL 0

test stuff also cleaned up

PR #542, opened 2018-02-01
Fix resend con response

boaks:
Add test and fix.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
@@ -30,6 +30,7 @@
  *                                                    Exchange.setTimedOut()
  *    Achim Kraus (Bosch Software Innovations GmbH) - correct timeout calculation
  *    Bosch Software Innovations GmbH - migrate to SLF4J
+ *    Achim Kraus (Bosch Software Innovations GmbH) - fix resend current response 
  ******************************************************************************/
 package org.eclipse.californium.core.network.stack;
 
@@ -77,7 +78,7 @@ public class ReliabilityLayer extends AbstractLayer {
 		max_retransmit = config.getInt(NetworkConfig.Keys.MAX_RETRANSMIT);
 
 		LOGGER.info("ReliabilityLayer uses ACK_TIMEOUT={}, ACK_RANDOM_FACTOR={}, and ACK_TIMEOUT_SCALE={}",
-				new Object[]{ack_timeout, ack_random_factor, ack_timeout_scale});
+				new Object[] { ack_timeout, ack_random_factor, ack_timeout_scale });
 	}
 
 	/**
@@ -135,7 +136,7 @@ public class ReliabilityLayer extends AbstractLayer {
 			}
 
 			LOGGER.trace("switched response message type from {} to {} (request was {})",
-					new Object[] {respType, response.getType(), reqType });
+					new Object[] { respType, response.getType(), reqType });
 
 		} else if (respType == Type.ACK || respType == Type.RST) {
 			response.setMID(exchange.getCurrentRequest().getMID());
@@ -196,10 +197,33 @@ public class ReliabilityLayer extends AbstractLayer {
 
 		if (request.isDuplicate()) {
 			// Request is a duplicate, so resend ACK, RST or response
-			if (exchange.getCurrentResponse() != null) {
+			Response currentResponse = exchange.getCurrentResponse();
+			if (currentResponse != null) {
+				if (currentResponse.getType() == Type.NON || currentResponse.getType() == Type.CON) {
+					// separate response
+					if (request.isConfirmable()) {
+						// resend ACK,
+						// comply to RFC 7252, 4.2, cross-layer behavior
+						EmptyMessage ack = EmptyMessage.newACK(request);
+						sendEmptyMessage(exchange, ack);
+					}
+					if (currentResponse.isConfirmable()) {
+						// retransmission cycle
+						int failedCount;
+						synchronized (exchange) {
+							failedCount = exchange.getFailedTransmissionCount() + 1;
+							exchange.setFailedTransmissionCount(failedCount);
+						}
+						LOGGER.debug("Request Duplicate: retransmit response, failed: {}, response: {}",
+								new Object[] { failedCount, currentResponse });
+						currentResponse.retransmitting();
+						sendResponse(exchange, currentResponse);
+						return;
+					}
+				}
 				LOGGER.debug("respond with the current response to the duplicate request");
 				// Do not restart retransmission cycle
-				lower().sendResponse(exchange, exchange.getCurrentResponse());
+				lower().sendResponse(exchange, currentResponse);
 
 			} else if (exchange.getCurrentRequest().isAcknowledged()) {
 				LOGGER.debug("duplicate request was acknowledged but no response computed yet. Retransmit ACK");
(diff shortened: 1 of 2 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -196,10 +202,15 @@ public void receiveRequest(final Exchange exchange, final Request request) {
...
 		if (request.isDuplicate()) {
 			// Request is a duplicate, so resend ACK, RST or response
-			if (exchange.getCurrentResponse() != null) {
+			Response currentResponse = exchange.getCurrentResponse();
+			if (currentResponse != null) {
 				LOGGER.debug("respond with the current response to the duplicate request");
+				if (currentResponse.isConfirmable()) {
+					// Simulate retransmission cycle, CON responses requires this
+					exchange.setFailedTransmissionCount(1);
+				}
```
sbernard31: The [RFC-7252](https://tools.ietf.org/html/rfc7252#section-4.2) say :

_"A CoAP endpoint that sent a Confirmable message MAY give up in
   attempting to obtain an ACK even before the MAX_RETRANSMIT counter
   value is reached.  For example, the application has canceled the
   request as it no longer needs a response, or there is some other
   indication that the CON message did arrive.  In particular, a CoAP
   request message may have elicited a separate response, in which case
   it is clear to the requester that only the ACK was lost and a
   retransmission of the request would serve no purpose.  **However, a
   responder MUST NOT in turn rely on this cross-layer behavior from a
   requester**, i.e., it MUST retain the state to create the ACK for the
   request, if needed, even if a Confirmable response was already
   acknowledged by the requester."_

So 2 questions : 
1. For separated response, does it means that we should send a ACK before ?
2.  Should we even consider to just send the ACK and rely on CON retransmission for the response ?

boaks: During tests a exception occurs and therefore I just suppressed that. 
To comply to that, some more work must be done :-).

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ServerDeduplicationTest.java
```diff
@@ -98,32 +120,85 @@ public void destroyClient() {
...
+	 * @throws Exception if the test fails.
+	 */
+	@Test
+	public void testServerSeparateRespondsToDuplicateRequest() throws Exception {
+
+		Token token = Token.fromProvider(new byte[] { 0x00, 0x01 });
+		int mid = 2345;
+
+		client.sendRequest(CON, POST, token, mid).path(resourceName).go();
+		// server will send response but response is lost
```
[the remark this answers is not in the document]
boaks: Your right, the repeat is caused by the lost ACK.

boaks:
Requested changes done.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -196,10 +202,17 @@ public void receiveRequest(final Exchange exchange, final Request request) {
 
 		if (request.isDuplicate()) {
 			// Request is a duplicate, so resend ACK, RST or response
-			if (exchange.getCurrentResponse() != null) {
+			Response currentResponse = exchange.getCurrentResponse();
+			if (currentResponse != null) {
 				LOGGER.debug("respond with the current response to the duplicate request");
+				if (currentResponse.isConfirmable()) {
+					// Simulate retransmission cycle, CON responses requires this
+					exchange.setFailedTransmissionCount(1);
```
sbernard31: I'm not sure to understand this line ? Why setting it to `1` ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -196,10 +202,17 @@ public void receiveRequest(final Exchange exchange, final Request request) {
 
 		if (request.isDuplicate()) {
 			// Request is a duplicate, so resend ACK, RST or response
-			if (exchange.getCurrentResponse() != null) {
+			Response currentResponse = exchange.getCurrentResponse();
+			if (currentResponse != null) {
 				LOGGER.debug("respond with the current response to the duplicate request");
+				if (currentResponse.isConfirmable()) {
```
sbernard31: I think we should send the ACK all separated responses (non only for CON) ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
boaks:
````
java.lang.IllegalArgumentException: message with already registered ID [41845] is not a re-transmission, cannot register exchange
	at org.eclipse.californium.core.network.InMemoryMessageExchangeStore.registerWithMessageId(InMemoryMessageExchangeStore.java:212)
	at org.eclipse.californium.core.network.InMemoryMessageExchangeStore.registerOutboundResponse(InMemoryMessageExchangeStore.java:323)
	at org.eclipse.californium.core.network.UdpMatcher.sendResponse(UdpMatcher.java:144)
	at org.eclipse.californium.core.network.CoapEndpoint$OutboxImpl.sendResponse(CoapEndpoint.java:694)
	at org.eclipse.californium.core.network.stack.BaseCoapStack$StackBottomAdapter.sendResponse(BaseCoapStack.java:197)
	at org.eclipse.californium.core.network.stack.ReliabilityLayer.receiveRequest(ReliabilityLayer.java:215)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveRequest(AbstractLayer.java:81)
	at org.eclipse.californium.core.network.stack.BaseCoapStack.receiveRequest(BaseCoapStack.java:99)
````

Running tests for the leak issue #429, I got this exception and the initial intention of this PR was to fix exactly that exception. I documented that therefore in the header

>    Achim Kraus (Bosch Software Innovations GmbH) - set retransmission for
>                                                    CON responses sent again for
>                                                    duplicated requests.
>                                                    Changed to pass check in 
>                                                    InMemoryMessageExchangeStore
>                                                    registerWithMessageId

**Comment by boaks** [comment]
Fix also for NON.

File: californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java [hunk 5]
```diff
@@ -196,10 +197,33 @@ public void receiveRequest(final Exchange exchange, final Request request) {
...
+						// retransmission cycle
+						int failedCount;
+						synchronized (exchange) {
+							failedCount = exchange.getFailedTransmissionCount() + 1;
+							exchange.setFailedTransmissionCount(failedCount);
+						}
+						LOGGER.debug("Request Duplicate: retransmit response, failed: {}, response: {}",
+								new Object[] { failedCount, currentResponse });
+						currentResponse.retransmitting();
+						sendResponse(exchange, currentResponse);
```
sbernard31: Using `sendResponse` instead of `lower().sendResponse()` is about preparing retransmission right ?

I asked my self if this could not go to some kind of race condition.
I mean we can face 2 preparing retransmission task at the same time. (because of classic CON response retransmission mechanism + reception of duplicate request)

Maybe not re-sending the response for CON response and just rely on the CON response retransmission mechanism is safer.

This is just an idea and if you think this is better like this. This PR sounds good to me.

boaks: Even without that, some of my logs contains a failed retransmission counter with values much larger the 30 :-).

boaks:
The race-conditions will be solved in an separate PR.

PR #538, opened 2018-01-31
 Stop cleaning task when we clear a block (fix for #537) 

sbernard31:
This PR aims to fix #537 (It should also fix same kind of error for block2)

PR #543, opened 2018-02-01
Fix deadlock introduce by #538

sbernard31:


californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -1041,27 +1041,32 @@ private Block2BlockwiseStatus getBlock2Status(final KeyUri key) {
...
 		synchronized (block1Transfers) {
-			Block1BlockwiseStatus removedTracker = block1Transfers.remove(key);
-			if (removedTracker != null) {
-				LOGGER.debug("removing block1 tracker [{}], block1 transfers still in progress: {}",
-						new Object[] { key, block1Transfers.size() });
-				removedTracker.setComplete(true);
-			}
-			return removedTracker;
+			removedTracker = block1Transfers.remove(key);
+			size = block1Transfers.size();
```
sbernard31: I hesitated to enclose this line with a if LOGGER.isDebugEnabled() but the size() is just an access to an attribute so I'm no sure if this is worth.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -1041,27 +1041,32 @@ private Block2BlockwiseStatus getBlock2Status(final KeyUri key) {
...
-						new Object[] { key, block2Transfers.size() });
-				removedTracker.setComplete(true);
-			}
-			return removedTracker;
+			removedTracker = block2Transfers.remove(key);
+			size = block2Transfers.size();
+		}
+		if (removedTracker != null) {
+			LOGGER.debug("removing block2 tracker [{}], block2 transfers still in progress: {}",
+					new Object[] { key, size });
```
boaks: Can we remove the `new Object[]`? It seems not to be any longer required.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: Don't make it too complex. So leave it, as it is here.

boaks:
Except of the new Object[], LGTM

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: I will remove the `Object[]`.

By the way do you understand the log message ?
First part ok : _removing block2 tracker [{}]_
Second part : _block2 transfers still in progress: {size}_ sounds a bit strange to me ...

boaks: replace "still" with "left"?

PR #544, opened 2018-02-02
Redesign cipher implementation 

boaks:
to improve performance and heap consumption.


boaks:
Depending on the payload size, it speeds up to more than 10% (400 bytes payload)

sbernard31:
I didn't look at this in details.
Anyway, limiting byte array instantiation could not be a bad idea and tests should check this new implementation is OK.

PR #545, opened 2018-02-05
Redesign message exchange

boaks:
Add ExchangeObserver remove to ensure message exchange
house-keeping. This only a very early stage. The clean up will follow.
Testing with blockwise notifes showed, that there are race conditions.
I tried to limit the defect using the striped executor. I hope I can provide
also a early preview for that in the afternoon. 

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Message.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Message.java
@@ -775,6 +775,13 @@ public abstract class Message {
 		}
 	}
 
+	public void onComplete() {
+		LOGGER.trace("Message completed {}", this);
+		for (MessageObserver handler : getMessageObservers()) {
+			handler.onComplete();
+		}
+	}
+
 	/**
 	 * Checks if this message is a duplicate.
 	 *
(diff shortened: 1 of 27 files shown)


sbernard31:
I tried to look at this but it was hard to follow.
I see lot of little changes and so more deeper one, it's hard to me to be able to make good feedback, but maybe this is not intended to be review now.

I will try to look at #546 to see if this make all of this clearer.

boaks:
FMPOV the current implementation of the "current request/current complete" is confusing. Therefore I want to encapsulate the exchange house-keeping in the exchange/exchangeobserver.
The observe only removes the exchange, and the setter (current request/response) will detect, if a previous message exchange store registration must be removed using that exchange observer.  
The basic idea and principal, when something is kept or removed is added as javadoc in Exchange:

> If the exchange represents a "blockwise" transfer and if the transparent mode
> is used, the exchange keeps also the (original) request and use the current
> request for transfer the blocks. A request not using observe use the same
> token for easier tracking. A request using observe keeps the origin request
> with the origin token in store, but use a different token for the transfer of
> the left blocks. This enables to catch new notifies while a transfer is
> ongoing.
>  

So in the end will contain even more changes, though it cleans up a "mess" :-).

boaks:
Yes, some more changes, but FMPOV more clear responsibilities. 
I'm through it. Tomorrow more on PR #546

boaks:
Included in PR #551.

PR #550, opened 2018-02-12
Striped exchange

boaks:
Introduce striped exchange executor.


boaks:
Includes PR #545

boaks:
There are still issues in the BlockwiseLayer and there will be a follow up PR to address them also.

boaks:
Included in PR #551

Issue #548, opened 2018-02-08
Storing client X509CertPath chain on TlsEndpointContext


AlexITC:
* branch: 2.0.x
* commit: 05b12ed8

Currently, on [NettyContextUtils#buildEndpointContext](https://github.com/eclipse/californium/blob/2.0.x/element-connector-tcp/src/main/java/org/eclipse/californium/elements/tcp/NettyContextUtils.java#L63), the `X500Principal` is being extracted, it would be useful to have the whole certificate chain, this could be retrieved from `SSLSession#getPeerCertificateChain()`.

I have a working code for this and I wonder if you are willing to merge it.

On `NettyContextUtils` we do this instead of storing the `X500Principal`:
```java
try {
  X509Certificate[] peerCertificateChain = sslSession.getPeerCertificateChain();
  sslSession.getPeerCertificateChain()
  if (peerCertificateChain != null && peerCertificateChain.length != 0) {
    java.security.cert.CertPath javaCertPath = toJavaCertPath(peerCertificateChain);
    principal = new X509CertPath(javaCertPath);
  }
} catch (Exception e) {
  // do nothing
}
```

And we'll need some helper methods:

```java
static java.security.cert.CertPath toJavaCertPath(javax.security.cert.X509Certificate[] javaxCertificatePath) throws Exception {
  List<java.security.cert.Certificate> javaCertificatePath = new LinkedList<>();
  for (javax.security.cert.X509Certificate javaxCertificate : javaxCertificatePath) {
    java.security.cert.Certificate javaCertificate = toJavaCertificate(javaxCertificate);
    javaCertificatePath.add(javaCertificate);
  }

  CertificateFactory cf = CertificateFactory.getInstance("X.509");
  return cf.generateCertPath(javaCertificatePath);
}

static java.security.cert.Certificate toJavaCertificate(javax.security.cert.X509Certificate javaxCertificate) throws Exception {
  byte[] encoded = javaxCertificate.getEncoded();
  ByteArrayInputStream bis = new ByteArrayInputStream(encoded);
  java.security.cert.CertificateFactory cf = java.security.cert.CertificateFactory.getInstance("X.509");
  return cf.generateCertificate(bis);
}
```

Of course, we'll need to update the tests.

Having this on the 2.0.x branch would avoid us to fork the project just to add this change and it might be useful for more people using TCP.

boaks:
>  it would be useful 

Why? 
To have "unique principal types" across DTLS/TLS? 
[...]

PR #554, opened 2018-02-18
Update NettyContextUtils to store the peer certificate chain

AlexITC:
While using TLS, we try to get the peer certificate chain (X509CertPath) or
fallback to get the peer principal (X500Principal).

### DIFF
--- a/element-connector-tcp/src/main/java/org/eclipse/californium/elements/tcp/NettyContextUtils.java
+++ b/element-connector-tcp/src/main/java/org/eclipse/californium/elements/tcp/NettyContextUtils.java
@@ -20,8 +20,11 @@
  ******************************************************************************/
 package org.eclipse.californium.elements.tcp;
 
+import java.io.ByteArrayInputStream;
 import java.net.InetSocketAddress;
 import java.security.Principal;
+import java.util.List;
+import java.util.LinkedList;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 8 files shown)


AlexITC:
@boaks Here are the changes that we talked on #548, I have some hard feeling on moving the `X509CertPath` to the `element-connector` module because there are similar classes that still live in `scandium`, maybe should we also move them? or creating a new module for these classes only might be the right move.

What do you think?

boaks:
> I have some hard feeling on moving the X509CertPath to the element-connector module because there are similar classes that still live in scandium, maybe should we also move them? or creating a new module for these classes only might be the right move.

I have the same feelings, but as written frequently, I want to concentrate first on issue #429 .

Your approach serialize an deserialize the `certificate[]`. I would prefer a implementation similar to `CertificateMessage.setCertificateChain` and there for move that to X509CertPath (e.g. as 
`X509CertPath fromCertificatesChain(Certificate[] certificateChain) ` and `CertPath generateCertPath(Certificate[] certificateChain)` to also use it then from the `CertificateMessage`.

So, this smells for me, that there is more "common encryption" functionallity, and therefore, after issue #429, I guess we should discuss, if this should go to an own module.

boaks:
@AlexITC 
maybe you can check branch "certpath" where I added the above mentioned changes if it works for you.

AlexITC:
great, thanks!

boaks:
The idea to move the scandium `auth` package (or currently only the `X509CertPath` class) was from my side, because in my opinion a dependency to scandium only for reuse of the `Principal` implementation seems for to be not that good.

sbernard31:
I'm also not so comfortable with the idea of making element-connector-tcp depend on scandium

AlexITC:
as I remember, moving the `X509CertPath` was a requirement to avoid circular dependencies.

boaks:
@sophokles73 

Do you refuse this PR? 
Otherwise I would like to got for merging it.

boaks:
Just to be precise, I want to merge branch 'certpath'. 
So I create a PR for that.

boaks:
OK, so first merge this, and the create a PR to certpath :-).

PR #557, opened 2018-02-21
Fix scandium exceptions during tests

boaks:
Running the scandium unit tests causes unintended exceptions.
One commit is solving this by ensuring, that the server also finish the handshake.
Analysing the exceptions shows, that an other test has also a gap.
See comments in the commits.


scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
```diff
@@ -1054,9 +1062,11 @@ public void testResumeFinishedMessageRetransmission() throws Exception {
 			}
 
 			// Ensure handshake is successfully done
-			assertTrue("handshake failed",
+			assertTrue("client handshake failed",
 					sessionListener.waitForSessionEstablished(MAX_TIME_TO_WAIT_SECS, TimeUnit.SECONDS));
```
sbernard31: should we rename sessionListener in clientSessionListener?

PR #565, opened 2018-02-27
Set message's bytes in DataParser

bgiori:
This commit sets the bytes of the message getting parsed by DataParser.
Previously, responses would not contain the bytes of the RawData
response even if the exchange was successful.

Signed-off-by: Brian Giori <briangiori@gmail.com>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataParser.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataParser.java
@@ -58,22 +58,29 @@ public abstract class DataParser {
 	 */
 	public final Message parseMessage(final byte[] msg) {
 
-		String message = "illegal message code";
+		String errorMsg = "illegal message code";
 		DatagramReader reader = new DatagramReader(msg);
 		MessageHeader header = parseHeader(reader);
 		try {
+			Message message = null;
 			if (CoAP.isRequest(header.getCode())) {
-				return parseMessage(reader, header, new Request(CoAP.Code.valueOf(header.getCode())));
+				message = parseMessage(reader, header, new Request(CoAP.Code.valueOf(header.getCode())));
 			} else if (CoAP.isResponse(header.getCode())) {
-				return parseMessage(reader, header, new Response(CoAP.ResponseCode.valueOf(header.getCode())));
+				message = parseMessage(reader, header, new Response(CoAP.ResponseCode.valueOf(header.getCode())));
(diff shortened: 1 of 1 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataParser.java
```diff
@@ -58,22 +58,29 @@ public final Message parseMessage(final RawData raw) {
...
-				return parseMessage(reader, header, new Response(CoAP.ResponseCode.valueOf(header.getCode())));
+				message = parseMessage(reader, header, new Response(CoAP.ResponseCode.valueOf(header.getCode())));
 			} else if (CoAP.isEmptyMessage(header.getCode())) {
-				return parseMessage(reader, header, new EmptyMessage(header.getType()));
+				message = parseMessage(reader, header, new EmptyMessage(header.getType()));
+			}
+
+			// Set the message's bytes and return the message
+			if (message != null) {
+				message.setBytes(msg);
```
boaks: What is the intended use of the original byte[] from the Datagram?

bgiori: I'm implementing the [Newt Manager protocol](https://mynewt.apache.org/latest/newtmgr/command_list/newtmgr_conn/), a management protocol for devices running [Apache Mynewt](http://mynewt.apache.org/) and [Zephyr](https://www.zephyrproject.org/). Without getting into the gory details, Newt Manager allows multiple transport schemes which includes both CoAP and non CoAP messages. 

I have a method which builds a Newt Manager response object from the bytes in the response based on scheme. Because the Newt Manager library is independent from Californium I do some simple CoAP parsing without DataParser. If the message uses CoAP, I parse the Response Code to determine whether the CoAP exchange was a success then parse the payload. 

Granted, I don't have to do it this way. I thought it was cleaner for transporter (BLE or UDP) to just dump the response bytes into the Newt Manager response builder regardless of transport or scheme. Also, I originally thought it was a bug that getBytes() always returned null, but now that I think about it more, there are few other use cases for the original datagram byte[]. 

If you think that this is just not useful enough to adopt feel free to close the PR. I implemented a custom endpoint where I can get the bytes from the RawData object and set the bytes in the response (I just had an urge to put it in the proper place). 

Great project by the way :)

boaks: Thanks!

Issue #567, opened 2018-03-04
StandardCharsets crashes on Android API < 19.


Thomas-Vos:
This library uses `java.nio.charset.StandardCharsets` in multiple places. This class was added in Android API 19, so it won't work on older versions. If I use Californium on Android API <19 it crashes with the following error:

    java.lang.NoClassDefFoundError: java.nio.charset.StandardCharsets

I manually copied the `java.nio.charset.StandardCharsets` file into my project and then Californium works without any issues.

Changing all usages of this Java class to `Charset.forName(...)` instead of the static constants will fix this crash for Android API <19. Is this something that could be changed in Californium?

See the following for all usages of this class: https://github.com/eclipse/californium/search?q=StandardCharsets

boaks:
> I manually copied the java.nio.charset.StandardCharsets file into my project and then Californium works without any issues.

Did you check "plain" coap, or also with "DTLS" coaps?

My oldest device to test is API 19, and frequently we run into "encryption" issues, which hard to solve and would be much harder to solve without test-device.

Thomas-Vos:
@boaks I tested it with DTLS. I'm using version 2.0.0-M6 with `californium-core` and `scandium` modules. Apart from this missing class, I haven't run into any other issues yet on API 16.

boaks:
Cool, I prepare a fix.

boaks:
@Thomas-Vos 

If possible, checkout the PR #568 and verify, if it works for you.

Thomas-Vos:
I tested the pull request, after I removed the copied StandardCharsets file from my project, on Android API 16. It now works correctly without any issues. I left one comment on the pull request. Thank you very much for the fast support!

boaks:
Discussing the PR I remembered, that I changed the minSdkVersion from 15 to 19 because `System.lineSeparator()` is used but marked with API 19. So are you sure, that this doesn't cause issues using older android versions?

Thomas-Vos:
Looks like another thing that might need to be changed. This definitely won't work below API 19. A quick search in this library showed that `System.lineSeparator()` is mostly used in tests and `toString()` methods. I am not calling the `toString()` methods of the Californium library anywhere in my project, so that explains why I haven't run into this issue yet. So yes, this might cause issues depending on how someone uses Californium.

sbernard31:
I find a way to check API level compatibility with maven.
(see #570)
I played a bit with it and it seems we are compatible with API 19.
[...]

PR #568, opened 2018-03-05
Fix for issue #567.

boaks:
Replacement of java.nio.charset.StandardCharsets, which requires java
1.7, to support older android versions.

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/RandomAccessBlockTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/RandomAccessBlockTest.java
@@ -29,7 +29,6 @@ import static org.junit.Assert.assertTrue;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 
 import org.eclipse.californium.category.Medium;
@@ -38,6 +37,7 @@ import org.eclipse.californium.core.CoapResource;
 import org.eclipse.californium.core.CoapResponse;
 import org.eclipse.californium.core.CoapServer;
 import org.eclipse.californium.core.coap.BlockOption;
+import org.eclipse.californium.core.coap.CoAP;
 import org.eclipse.californium.core.coap.CoAP.ResponseCode;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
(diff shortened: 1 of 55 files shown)


californium-core/src/test/java/org/eclipse/californium/core/test/RandomAccessBlockTest.java
```diff
@@ -168,7 +168,7 @@ public void testServerReturnsIndividualBlocks() throws Exception {
 		private BlockwiseResource(String name, String responsePayload) {
 			super(name);
 			this.responsePayload = responsePayload;
-			buf = ByteBuffer.wrap(responsePayload.getBytes(StandardCharsets.US_ASCII));
+			buf = ByteBuffer.wrap(responsePayload.getBytes(CoAP.UTF8_CHARSET));
```
Thomas-Vos: Is this supposed to be a different charset or is this a typo?

sbernard31:
No objection, I just ask myself which android API are we focusing ?

I understand that this fix will help us to support API 16.
If I well understand before this fix we was API 19 compliant.

I try to look if there is maven plugin to ensure android API level compatibility but I didn't find anything good... (just this old (unmaintained?) project : https://code.google.com/archive/p/compliancechecker/)

Here is [Android_version_history](https://en.wikipedia.org/wiki/Android_version_history)
And here [some statistics](https://developer.android.com/about/dashboards/index.html#Platform) about uses of android version. (This masked some devices as it only counts devices connected to googleplay store)

I suppose the idea for now about android compliance is just best effort driving by community needs.

boaks:
I changed the minSdkVersion from 15 to 19 with commit 0d3a6f9d4e727d760d564b8512d6086dee0b3c81 3 months ago, 
mainly for the `System.lineSeparator()`.

I don't plan to really support older versions, but, if someone else feels it works with small changes, that's OK for me. Therefore I created this PR.

sbernard31:
> if someone else feels it works with small changes, that's OK for me

That's ok for me too. 

I forgot we have an android demo. Should create a build about it to ensure that we are compatible with a given version of android API ?
I try to build it but we need a very recent version of gradle (on jenkins there is only the v3.1)
Maybe using gradle wraper ? (I'm not an expert about Gradle)

boaks:
Do you think on a static compatibility check?
I would not spend time on that. 
The most issues of the last half year only appeared at runtime.

sbernard31:
I rather continue the discussion in #567

boaks:
Mark the TLS and File NetworkConfig to be excluded from the check.
Android developer may chose to configure by code or use the AssetManager / InputStream variants.

sbernard31:
LGTM !
(It was hard to search "real" changes between all the big  `System.lineSeparator()`/`StringUtil.lineSeparator()`  refactoring ;-) )

Just one open question : 
Using @NotForAndroid to ignore compatibility check is great ! :+1: 
Should we also use it as user documentation ? if yes, I think we should also mark `createWithFile`, `createStandardWithFile` too.

boaks:
> Should we also use it as user documentation ? if yes, I think we should also mark createWithFile, createStandardWithFile too.

Yes, your right.
I also forgot to make the "toString(InetSocketAddress)" check, if the getHostString is available.

boaks:
I tested it successful with API 19 and 23.
So we will see, if it works with older API version or if we get new error issues :-).

PR #569, opened 2018-03-06
Add constructor with attributes map.

boaks:
Support cloning.


sbernard31:
LGTM

PR #571, opened 2018-03-08
Add UdpEndpointContext for mixed matching.

boaks:
Prevent `DtlsEndpointContext` matching with `UdpEndpointContext` using a
`UdpEndpointContextMatcher`. 
Leshan use the same `ObservationStore` for secure and unsecure communication. Without this change, a message received with the unsecure connector would match messages sent with a secure connector, because the `UdpEndpointContextMatcher` will not check for additional context fields, which are unknown by that matcher..

### DIFF
--- a/element-connector/src/test/java/org/eclipse/californium/elements/UdpEndpointContextMatcherTest.java
+++ b/element-connector/src/test/java/org/eclipse/californium/elements/UdpEndpointContextMatcherTest.java
@@ -1,0 +1,67 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial creation
+ ******************************************************************************/
+package org.eclipse.californium.elements;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+import java.net.InetSocketAddress;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class UdpEndpointContextMatcherTest {
+
+	private static final InetSocketAddress ADDRESS = new InetSocketAddress(5683);
+	private static final InetSocketAddress CHANGED_ADDRESS = new InetSocketAddress(5684);
+	
+	private EndpointContext connectorContext;
+	private EndpointContext addressContext;
+	private EndpointContext messageContext;
+	private EndpointContext changedAddressContext;
+	private EndpointContext secureMessageContext;
+	private EndpointContextMatcher matcher;
+
+	@Before
+	public void setup() {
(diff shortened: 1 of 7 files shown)


element-connector/src/test/java/org/eclipse/californium/elements/UdpEndpointContextMatcherTest.java
```diff
@@ -0,0 +1,62 @@
...
+		assertThat(matcher.isToBeSent(messageContext, connectorContext), is(true));
+		assertThat(matcher.isToBeSent(secureMessageContext, connectorContext), is(false));
+	}
+
+	@Test
+	public void testResponse() {
+		assertThat(matcher.isResponseRelatedToRequest(messageContext, messageContext), is(true));
+		assertThat(matcher.isResponseRelatedToRequest(messageContext, secureMessageContext), is(false));
+		assertThat(matcher.isResponseRelatedToRequest(secureMessageContext, messageContext), is(false));
+		assertThat(matcher.isResponseRelatedToRequest(addressContext, messageContext), is(true));
```
sbernard31: If I well understand this should not happened ? the request context will be an UdpEndpointContext or a DtlsEndpointContext ?

element-connector/src/test/java/org/eclipse/californium/elements/UdpEndpointContextMatcherTest.java
```diff
@@ -0,0 +1,62 @@
...
+		assertThat(matcher.isResponseRelatedToRequest(messageContext, addressContext), is(false));
```
sbernard31: This should not happened too ? right ? (I mean reponse context could not be an addressContext ?)

boaks:
> A question about UdpMatcher, when we correlate request and response, should we check peer address/port ?

I would not check the peer's address/port, because the unique token stuff was to decouple from that. Logically it would be the localport/interfaces, but it seems to be not available.

element-connector/src/test/java/org/eclipse/californium/elements/UdpEndpointContextMatcherTest.java
boaks: Yes, it should not happen. I just add the test to be complete.

element-connector/src/test/java/org/eclipse/californium/elements/UdpEndpointContextMatcherTest.java
boaks: right

element-connector/src/test/java/org/eclipse/californium/elements/UdpEndpointContextMatcherTest.java
sbernard31: that's confused me little when I read it :)

sbernard31:
> I would not check the peer's address/port, because the unique token stuff was to decouple from that. 

I thought that we want to ensure that address/port was the same for unsecured connection. For me this is not linked to the unique token stuff. (but this not so important to me)

boaks:
Revert `UdpEndpointContextMatcher` and you see the issue.

boaks:
> I thought that we want to ensure that address/port was the same for unsecured connection. For me this is not linked to the unique token stuff. (but this not so important to me)

Then the UdpMatcher would depend on the security level of the connector.

sbernard31:
Maybe I was not clear, I'm ok with your fix. I think this is necessary.
Reading the PR, I see we don't check address and port for unsecure connection. I just ask if this is really what we want because I thought we want to check it. (so if we add this check this is in addition of what you have done here)

boaks:
AH, I hope I understand now. We can add that check to the UdpEndpointcontextMatcher.

PR #575, opened 2018-03-09
Certpath

boaks:
Replaces #554 


boaks:
Seems that we have to do it sequentially :-)

Issue #572, opened 2018-03-08
Move org.eclipse.californium.scandium.auth to element-connector


boaks:
Branch 2.0.0

As already mentioned in the discussion of an other issue (https://github.com/eclipse/californium/issues/548#issuecomment-365201179) and PR #554 I would like to move the "auth" package to the "element-connector" module.
This will not only prevent #554 to introduce a dependency from "element-connector-tcp" to "scandium", it would also allow "californium-core" to be extended to create `Principals` out of the `URI` user info.
The move would require to move the `Base64` as well.

sbernard31:
Could you details details the `Principals`/`URI` part ?

boaks:
"detail" will be the wrong term for the next weeks/months.

It's a raw idea to have something as 

coaps://user@host/

creates a `PreSharedKeyIdentity`.

But for now I'm only able to focus on PR #554.

PR #578, opened 2018-03-12
Move auth from scandium to element-connector.

boaks:
Issue #572


sbernard31:
:+1:

PR #579, opened 2018-03-13
Extend strict/relaxed response matching modes for plain coap.

boaks:

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/EndpointContextMatcherFactory.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/EndpointContextMatcherFactory.java
@@ -20,6 +20,8 @@
  *                                                    to EndpointContextMatcherFactroy.
  *                                                    Add PRINCIPAL mode.
  *    Achim Kraus (Bosch Software Innovations GmbH) - add TlsEndpointContextMatcher
+ *    Achim Kraus (Bosch Software Innovations GmbH) - extend strict/relaxed modes for
+ *                                                    plain coap. 
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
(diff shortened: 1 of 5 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/EndpointContextMatcherFactory.java
```diff
@@ -38,54 +40,68 @@
...
 		case PRINCIPAL:
-			return new PrincipalEndpointContextMatcher();
+			if (!CoAP.PROTOCOL_UDP.equalsIgnoreCase(protocol)) {
+				return new PrincipalEndpointContextMatcher();
+			}
+		}
+		if (CoAP.PROTOCOL_UDP.equalsIgnoreCase(protocol)) {
+			return new UdpEndpointContextMatcher(true);
+		} else {
+			return new StrictDtlsEndpointContextMatcher();
```
sbernard31: Just a details but I feel this a bit simpler to read like this : 
```java
switch (mode) {
	case RELAXED:
		if (CoAP.PROTOCOL_UDP.equalsIgnoreCase(protocol)) {
			return new UdpEndpointContextMatcher(false);
		} else {
			return new RelaxedDtlsEndpointContextMatcher();
		}
	case PRINCIPAL:
		if (!CoAP.PROTOCOL_UDP.equalsIgnoreCase(protocol)) {
			return new PrincipalEndpointContextMatcher();
		} else {
			return new UdpEndpointContextMatcher(true);
		}
	case STRICT: 
	default:
		if (CoAP.PROTOCOL_UDP.equalsIgnoreCase(protocol)) {
			return new UdpEndpointContextMatcher(true);
		} else {
			return new StrictDtlsEndpointContextMatcher();
		}
	}
}
```

boaks:
You can create the endpoints with other matcher combinations.
Or use different NetworkConfigs. The setup in the Factory is just the default.

boaks:
I changed the switch/case.
FMPOV PRINCIPAL mode is more relaxed and therefore I disabled the address checking.

PR #580, opened 2018-03-14
Change return type from entries Set to Map.

boaks:
With the unmodifiable map it's possible to return that directly.


sbernard31:
LGTM

PR #584, opened 2018-03-16
Workaround for openjdk URI bug.

boaks:
JDK-8199396

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -49,6 +49,7 @@
  *                                                    constructors
  *    Achim Kraus (Bosch Software Innovations GmbH) - replace byte array token by Token
  *    Achim Kraus (Bosch Software Innovations GmbH) - add token generator
+ *    Achim Kraus (Bosch Software Innovations GmbH) - workaround for open jdk URI bug
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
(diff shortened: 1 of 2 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -611,11 +610,22 @@ public InetSocketAddress getAddress() {
 	public URI getUri() {
 		URI uri = null;
 		try {
-			InetSocketAddress address = getAddress();
+			InetSocketAddress socketAddress = getAddress();
 			String scheme = CoAP.getSchemeForProtocol(connector.getProtocol());
-			uri = new URI(scheme, null, address.getHostString(), address.getPort(), null, null, null);
-		} catch (URISyntaxException e) {
-			LOGGER.warn("URI", e);
+			String host = socketAddress.getAddress().getHostAddress();
```
sbernard31: Is it normal that `socketAddress.getAddress().getHostAddress()` are using instead of `address.getHostString()`

boaks: For a local networkinterface the hostname seems for me to be not that important.

sbernard31: It makes sense.

So about #567, the last address.getHostString() which could cause issue is in `TLSClientConnector.createSllEngine`/`TLSServerConnector.createSllEngine`

sbernard31:
LGTM

Issue #587, opened 2018-03-19
 removed observe token on getting unexpected ETag


charybr:
Hi,

In one scenario we noticed that on receiving unexpected ETag in blockwise transfer, observe token was removed but CoapHandler doesnt get notified about observe being deleted. We understand that getting unexpected token is server side issue, but removal of observe was not notificied to CoapHandler.

We are using californium 2.0.0.SNAPSHOT. 
Attached log file
[protocolAdapter.log](https://github.com/eclipse/californium/files/1824208/protocolAdapter.log)

Sequence as explained below:
1. CON notification came at 2018-03-19 00:02:27,711. observe token 7f513394c4cacfdc.
```
2018-03-19 00:02:27,711 DEBUG CoapEndpoint-UDP-0.0.0.0/0.0.0.0:0#1 org.eclipse.californium.core.network.BaseMatcher - re-created exchange from original observe request: CON-GET    MID=33629, Token=[7f513394c4cacfdc], OptionSet={"Uri-Host":"f2ca1c3e-43c0-43ff-837c-83e6e477fa33", "Observe":0, "Uri-Path":["indsens","abc"]}, no payload
2018-03-19 00:02:27,711 DEBUG CoapEndpoint-UDP-0.0.0.0/0.0.0.0:0#1 org.eclipse.californium.core.network.stack.ReliabilityLayer - acknowledging CON response
```

2. Blockwise transfer got started, block transfer token - 8c961e677019d37e
```
2018-03-19 00:02:27,711 DEBUG CoapEndpoint-UDP-0.0.0.0/0.0.0.0:0#1 org.eclipse.californium.core.network.stack.BlockwiseLayer - created tracker for KeyUri[coap://f2ca1c3e-43c0-43ff-837c-83e6e477fa33/indsens/abc, [fd0080000000000000000000001000e7]:5683] inbound block2 transfer [currentNum=0, currentSzx=0, bufferSize=2048, complete=false, random access=false, observe=201], transfers in progress: 7, CON-2.05   MID=29395, Token=[7f513394c4cacfdc], OptionSet={"ETag":0xc4, "Observe":201, "Content-Format":"application/cbor", "Max-Age":1800, "Block2":"(szx=1/32, m=true, num=0)"}, 
a2 64 64 61 74 61 a4 62 69 64 18 a6 64 74 79 70 65 67 63 6f 6e 74 72 61 6b 64 6d 65 74 61 a5 61
```

[...]

PR #589, opened 2018-03-20
Don't remove observe on cancel of notify.

boaks:
Fix issue #587


sbernard31:
LGTM

boaks:
> any reason about commenting instead of deleting ?

Uups, to focused on update PR #551 to #590. I will delete them.

PR #590, opened 2018-03-20
Fix leaks, issue #429

boaks:
These are the rebased & partial squashed commits from PR #551.


boaks:
https://github.com/eclipse/californium/pull/551#issuecomment-370788286

> (I will probably have no more input soon)

@sbernard31 
Your last comment in PR #551 is now 14 days ago.
Any progress in testing in it? If not, I guess it will be better, if we just merge this update PR. Otherwise it will take more time to frequently update it, than to test it :-).

boaks:
(contains PR #589 also)

sbernard31:
@boaks  I missed this comment sry about.
You can merge it, except  I'm still not OK about Exchange.debug set to true and/or not configurable.
sry again.

boaks:
@sbernard31 
Thanks for the feedback! Though I'm currently busy with other jobs, the "delay" doesn't matter :-).
So let me sum-up:
- about the exchange house-keeping, the new design seems to be agreed
- about the threading, the solution using a stripe execution seems to agreed for "now" (mainly because alternatives would take much more time to develop).
- about the Exchange.debug, if that's the last major concern, I will change it.

So, let me finish my other jobs and afterwards I will pick up this PR again and will finish my work on it to get it merged.

boaks:
I spent some time more and squashed some of the commits.
- the observer-test-client is extended with a onComplete timeout to overcome the bugs (fixed with the later commits). So it may be used to compare somehow the behaviour before and after.
- The synchronization changes in the blockwiselayer are isolated, but testing without changing that doesn't show the deadlocks as before, so I kept the old synchronization there (even if I don't think, that this is really working).
 
So, from my side, I think I spent enough time into that and I hope the leaks are gone. I will merge it on wednesday, If no objection.

PR #601, opened 2018-04-04
FETCH/PATCH framework added and tests updated

T0bbe:
OSCORE processing uses FETCH as dummy value during Observe, see [OSCORE draft](https://github.com/core-wg/oscoap). OSCORE also supports PATCH. 

Signed-off-by: Tobias Andersson <tobias.andersson@ri.se>


boaks:
Thanks!

PR #602, opened 2018-04-05
Configure the X.509 certificate chain validation

msangoi:
This PR is a starting point to introduce some flexibility in the validation of the peer's certificate chain.

I introduce a CertificateVerifier interface which is close to the `X509TrustManager` class of java. In this version, the `DTLSSession` is available, adding the possibility to get some additional information (e.g. negociated cipher suite).

Signed-off-by: Manuel Sangoi <msangoi@sierrawireless.com>

### DIFF
--- a/scandium-core/pom.xml
+++ b/scandium-core/pom.xml
@@ -136,7 +136,8 @@
 							org.eclipse.californium.scandium.dtls,
 							org.eclipse.californium.scandium.dtls.cipher,
 							org.eclipse.californium.scandium.dtls.pskstore,
-							org.eclipse.californium.scandium.dtls.rpkstore
+							org.eclipse.californium.scandium.dtls.rpkstore,
+							org.eclipse.californium.scandium.dtls.x509
 						</Export-Package>
 						<Private-Package>
 							org.eclipse.californium.scandium.util
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -43,6 +43,8 @@ import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
 import org.eclipse.californium.scandium.dtls.pskstore.PskStore;
 import org.eclipse.californium.scandium.dtls.rpkstore.TrustAllRpks;
 import org.eclipse.californium.scandium.dtls.rpkstore.TrustedRpkStore;
+import org.eclipse.californium.scandium.dtls.x509.CertificateVerifier;
+import org.eclipse.californium.scandium.dtls.x509.StaticCertificateVerifier;
 
 /**
  * A container for all configuration options of a <code>DTLSConnector</code>.
@@ -76,6 +78,7 @@ public final class DtlsConnectorConfig {
 
 	private InetSocketAddress address;
 	private X509Certificate[] trustStore;
+	private CertificateVerifier certificateVerifier;
 
 	/**
 	 * Experimental feature : Stop retransmission at message receipt
@@ -304,6 +307,16 @@ public final class DtlsConnectorConfig {
 	 */
 	public X509Certificate[] getTrustStore() {
 		return trustStore;
+	}
+
+	/**
+	 * Gets the verifier in charge of validating the peer's certificate chain
+	 * during the DTLS handshake.
+	 *
+	 * @return the certificate chain verifier
+	 */
+	public CertificateVerifier getCertificateVerifier() {
+		return certificateVerifier;
 	}
 
 	/**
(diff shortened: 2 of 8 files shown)


boaks:
> This last point means we should probably go later for an async API with callback

I would prefer, that such a async API should be only an optional possibility and those, we don't need it, can still use the simple sync API. With that, it seems to be not irrelevant, how the sync API is designed.

Just to add: 
The current TLS implementation, AFAIK, is also using sync API to the trust- and key-managers. So FMPOV currently, for the none PSK variants, it seems to be more promising to validate the trust, when the `RawData` is passed to the `RawDataChannel`.

sbernard31:
> FMPOV currently, for the none PSK variants, it seems to be more promising to validate the trust, when the RawData is passed to the RawDataChannel.

You mean using Correlation context ?
I do not understand why this is better ? Could you explain ?

boaks:
> You mean using Correlation context ?
> I do not understand why this is better ? Could you explain ?

Because calling "RawDataChannel.receiveData" doesn't expect that something happens before the call returns. Therefore it's easy to release the "connector thread" there and hook in a async trust API. 
Basically it could be checked, if the session of the rawdata is already trusted to forward the data directly to the CoapEndpoint, or, if not already trusted, the RawDatat may be queued/handled the by a async Trust API.

sbernard31:
I think I understand, but as we already have the problem with PskStore and we would not be able to plug  on "rawdatachannel" for PSK too, we should probably go for a way to handle async store anyway.

In this case, reuse this async mechanism for X509 too makes more sense to me and this should be more consistent.

boaks:
> In this case, reuse this async mechanism for X509 too makes more sense to me and this should be more consistent.

... but will not work for TLS (and therefore, I think @AlexITC should consider to check, if such a implementation wouldn't be a solution for his project).

But let's concentrate on this PR:
FMPOV, keeping the sync API in good shape is preferred, I don't think we should remove the sync API it for a async API. Therefore I would prefer to have the API similar to the `X509TrustManager`. If this is not possible this time, then I agree to merge this PR as it is. Maybe someone can contribute that later.

AlexITC:
@boaks I think the ideal scenario is to let the users choose between a sync or async API, in our case, the async one would be quite useful to avoid our blocking calls.

boaks:
> I think the ideal scenario is to let the users choose between a sync or async API

sure, therefore I commented on @sbernard31's

> This last point means we should probably go later for an async API with callback, so the first remarks about using X509TrustManager is maybe not so relevant. (as this is a sync API) 

that for me, the sync API will stay **RELEVANT**!

But again, Alexis, how do you want to introduce a async API into TLS? The oracle java seems to provide only sync API.

AlexITC:
Honestly, I don't know if that's even possible, the few pieces that I have seen don't give the opportunity to write async code on the handshake related parts.

boaks:
@sbernard31 
so is the sync APi still "not so relevant"?

@msangoi 
Do you think, you will be able to spend more time in it, or should we go for merge it "as it is"?

sbernard31:
> so is the sync APi still "not so relevant"?

I have no strong opinion about that.
 - I feel that it makes sense to try to make  the scandium and element-connector-tcp API as much similar as possible.
 - I also feel that generally reuse java interface is a good point.
 - About async API for TLS/Netty, https://github.com/netty/netty/issues/7862 seems to confirm there is no good way to plug this check at handshake time.

In other hand,
 - I also understand the @msangoi's argument about : _"In this version, the DTLSSession is available, adding the possibility to get some additional information (e.g. negociated cipher suite)."_
 - I also feel that async API would be really better.



So I feel this PR will be the first try and not the last one about that ;)

boaks:
@sbernard31 

I feel, we discussed different things.
I'm not against a async API. But I think, that the sync API will stay relevant.
I understood your very first comment above as

> This last point means we should probably go later for an async API with callback, so the first remarks about using X509TrustManager is maybe not so relevant. (as this is a sync API)

"the sync API is  going to be removed for a async API anyway", 

and that's,what I disagree.

But sure, if @msangoi can/will not spend more time, this PS is better than without if :-).

(In my opinion, using a micro service for credential management should be implemented by a "local caching client", which prefetches the credentials on startup and get later changes potentially delayed, but decoupled from the handshake calls. So the access to the credentials will always be sync and fast. Even with 1K trust certificates, this seems to be better to spend 2MB for that (2K per cert * 1K). Or 50000 psk-credentials with (24 byte id, 16 bytes secret) would also fit in 2MB. Bulk-Transfer on init would take only a couple of seconds. But if someone likes to do it in an other way, this will be OK, until it's not the intention to remove the sync API).

sbernard31:
> "the sync API is going to be removed for a async API anyway"

I thought that only one API was enough, but we could have both if this is really need. I suppose we could see that later, right ?

boaks:
>  I suppose we could see that later, right ?

Sure, but we should see it not too late :-)

boaks:
@msangoi 
Is it possible, that your rebase and solve the conflicts (I guess they are caused by the lineSeparator move)?
If it's currently not possible for you to invest the time I can do that also.

msangoi:
Hi @boaks. Sorry I didn't have much time last week to work on this PR. I'll made those changes tomorrow. 
What was the final decision about the API? Moving to `X509TrustManager`? I though that having access to the `DTLSSession` was valuable but I'm fine with changing it if you all think it is better.

boaks:
Hi, FMPOV, there is no decision (for now). According Simon, a async API potentially remove the sync API. I'm not really convinced about that remove,but we will see.
So my current recommendation would be, rebase, fix the changes, but stay with your API
(for now, if we want to change it, we may do that later).

msangoi:
Following @sbernard31 comment about the `null` certificateVerifier, I've made some changes to ensure the behaviour is still the same. 
The `CertificateVerifier` interface is now very close to the `X509TrustManager` one. We can change it later if needed.

boaks:
FMPOV, I would go to merge it, but change some things afterwards.

As far as I understand it, the `CertificateVerifier` is always created, even, if only PSK or RPK should be supported. `getAcceptedIssuers` is then used to determine, if x509 is supported.
I would prefer to use the `CertificateVerifier` to determine, if x509 should be supported. The accepted issuers are more a hint used from the server to help the client to chose the right certificate. Especially, if the set/list of "accepted issuers" is large (e.g. one per tenant), and the client knows, which certificate it should send, I would recommend not to send that list and therefore return an empty list,  

But that's a detail, which could be changed after merge.

sbernard31:
> I would prefer to use the CertificateVerifier to determine, if x509 should be supported. 

It makes sense to me.

PR #603, opened 2018-04-06
Integration hooks for OSCORE

T0bbe:
Integration hooks for OSCORE. Mostly visibility changes. OSCORE number is set to 9 which it most likely will be set to once the draft process is finished.

Signed-off-by: Tobias Andersson <tobias.andersson@ri.se>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
@@ -75,6 +75,7 @@ public final class OptionSet {
 	private Integer      size1;
 	private Integer      size2;
 	private Integer      observe;
+	private byte[]		 oscore;
 	
 	// Arbitrary options
 	private List<Option> others;
@@ -104,6 +105,7 @@ public final class OptionSet {
 		size1               = null;
 		size2               = null;
 		observe             = null;
+		oscore				= null;
 		
 		others              = null; // new LinkedList<>();
 	}
@@ -132,6 +134,7 @@ public final class OptionSet {
 		block1 = null;
 		block2 = null;
 		observe = null;
+		oscore = null;
 		if (others != null)
 			others.clear();
 	}
@@ -163,6 +166,8 @@ public final class OptionSet {
 			block2          = new BlockOption(origin.block2);
 		
 		observe = origin.observe;
+		if(origin.oscore != null)
+			oscore	= origin.oscore.clone();
 		
 		others              = copyList(origin.others);
 	}
@@ -1270,7 +1275,50 @@ public final class OptionSet {
 	public static boolean isValidObserveOption(final int value) {
 		return value >= 0 && value <= MAX_OBSERVE_NO;
 	}
-
+	
+	/**
+	 * Returns the byte array value of the OSCore option.
+	 * @return the OSCore value or null if the option is not present
+	 */
+	public byte[] getOscore() {
+		return oscore;
+	}
+	
+	/**
+	 * Checks if the OSCore option is present.
+	 * @return true if present
+	 */
+	public boolean hasOscore() {
+		return oscore != null;
+	}
+	
+	/**
+	 * Replaces the Oscore option with oscore.
+	 * Returns the current OptionSet object for a fluent API.
+	 * 
+	 * @param oscore the new Oscore value
+	 * @return this OptionSet
+	 * @throws NullPointerException if oscore is null
+	 */
+	public OptionSet setOscore(byte[] oscore){
+		if(oscore != null){
+			this.oscore = oscore.clone();
+		}else{
+			throw new NullPointerException("Oscore cannot be null.");
+		}
+		return this;
+	}
+	
+	/**
+	 * Removes the OSCore options.
+	 * Returns the current OptionSet object for a fluent API.
+	 * @return this OptionSet
+	 */
+	public OptionSet removeOscore(){
+		oscore = null;
+		return this;
+	}
+	
 	/**
 	 * Checks if an arbitrary option is present.
 	 * @param number the option number
@@ -1352,6 +1400,8 @@ public final class OptionSet {
 			options.add(new Option(OptionNumberRegistry.SIZE1, getSize1()));
 		if (hasSize2())
 			options.add(new Option(OptionNumberRegistry.SIZE2, getSize2()));
+		if(hasOscore())
+			options.add(new Option(OptionNumberRegistry.OSCORE, getOscore()));
 		
 		if (others != null)
 			options.addAll(others);
@@ -1386,6 +1436,7 @@ public final class OptionSet {
 			case OptionNumberRegistry.SIZE1:          setSize1(option.getIntegerValue()); break;
 			case OptionNumberRegistry.SIZE2:          setSize2(option.getIntegerValue()); break;
 			case OptionNumberRegistry.OBSERVE:        setObserve(option.getIntegerValue()); break;
+			case OptionNumberRegistry.OSCORE:		  setOscore(option.getValue()); break;
 			default: getOthersInternal().add(option);
 		}
 		return this;
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -170,7 +170,7 @@ public class CoapEndpoint implements Endpoint {
 	private static final Logger LOGGER = LoggerFactory.getLogger(CoapEndpoint.class.getCanonicalName());
 
 	/** The stack of layers that make up the CoAP protocol */
-	private final CoapStack coapstack;
+	protected CoapStack coapstack;
 
 	/** The connector over which the endpoint connects to the network */
 	private final Connector connector;
@@ -178,7 +178,7 @@ public class CoapEndpoint implements Endpoint {
 	private final String scheme;
 
 	/** The configuration of this endpoint */
-	private final NetworkConfig config;
+	protected final NetworkConfig config;
 
 	/** The matcher which matches incoming responses, akcs and rsts an exchange */
 	private final Matcher matcher;
@@ -392,7 +392,7 @@ public class CoapEndpoint implements Endpoint {
 				throw new IllegalArgumentException("Connector must be a UDPConnector to use apply configuration!");
 			}
 		}
-
+		
 		Executor exchangeExecutionHandler = new Executor() {
 
 			@Override
@@ -685,7 +685,7 @@ public class CoapEndpoint implements Endpoint {
 	 * will then give them to the matcher, the interceptors, and finally send
 	 * them over the connector.
 	 */
-	private class OutboxImpl implements Outbox {
+	public class OutboxImpl implements Outbox {
 
 		@Override
 		public void sendRequest(final Exchange exchange, final Request request) {
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/OptionNumberRegistry.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/OptionNumberRegistry.java
@@ -55,6 +55,9 @@ public final class OptionNumberRegistry {
 	public static final int BLOCK2			= 23;
 	public static final int BLOCK1			= 27;
 	public static final int SIZE2			= 28;
+	
+	//TODO temporary assignment
+	public static final int OSCORE			= 9;
 
 	/**
 	 * Option names.
@@ -83,7 +86,10 @@ public final class OptionNumberRegistry {
 		public static final String Block2			= "Block2";
 		public static final String Block1			= "Block1";
 		public static final String Size2			= "Size2";
+		
+		public static final String Object_Security  = "Object-Security";
 	}
+	
 	
 	/**
 	 * Option default values.
@@ -131,6 +137,7 @@ public final class OptionNumberRegistry {
 			return optionFormats.STRING;
 		case ETAG:
 		case IF_MATCH:
+	    case OSCORE:          
 			return optionFormats.OPAQUE;
 		default:
 			return optionFormats.UNKNOWN;
@@ -227,6 +234,7 @@ public final class OptionNumberRegistry {
 		case IF_NONE_MATCH:
 		case OBSERVE:
 		case ACCEPT:
+		case OSCORE:
 		default:
 			return true;
 		case ETAG:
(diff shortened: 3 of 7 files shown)


boaks:
Is it possible, that you provide the repo/commit, where I can see, how these changes are used?

I will comment my concerns in your changes, but it may be easier to provide alternatives, if I see, how it is used.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -94,9 +94,9 @@
 	private final AtomicReference<EndpointContext> destinationContext = new AtomicReference<EndpointContext>();
 
 	/** The type used for requests (CON is default) */
-	private Type type = Type.CON;
+	protected Type type = Type.CON;
```
boaks: Why should this be more exposed than with useNONs/useCONs?

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
```diff
@@ -107,7 +107,7 @@
 	private MessageDeliverer deliverer;
 
 	/** The list of endpoints the server connects to the network. */
-	private final List<Endpoint> endpoints;
+	protected List<Endpoint> endpoints;
```
boaks: Why should this be more exposed than with available methods?

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
```diff
@@ -1270,7 +1283,44 @@ public OptionSet removeObserve() {
...
+	/**
+	 * Removes the Code option.
+	 * Returns the current OptionSet object for a fluent API.
+	 * @return this OptionSet.
+	 */
+	public OptionSet removeCode(){
+		this.code = null;
+		return this;
+	}
+
```
boaks: I miss a extension of  `public OptionSet addOption(Option option)`
Is this forgotten or intended not to be provided?

T0bbe:
Repo (branch 2.0.x, folder oscore-cf): https://bitbucket.org/lseitz/oscore/commits/branch/2.0.x

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
T0bbe: Erasing endpoints made by super. To make sure that we use OSCoreEndpoints.

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
T0bbe: This is just forgotten.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
T0bbe: This might be able to stay private. The tests runs fine with a small change on my end.

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
T0bbe: We could change it to CoAP Code, or CoAP Request Code.

boaks: May be I misunderstand the intention of "code" in OptionSet. FMPOV the OptionSet is for options, so I assumed the "code" is the new option with the value specified in OSCORE 8.1, the "Object-Security Value", and the pending IANA request in "13.1. CoAP Option Numbers Registry". But now it seems to be even no option and may be just put here for convenience.

So:
- how Is the new option (object-security) handled, if not included in OptionSet?
- why is a none option added to the OptionSet?

californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
```diff
@@ -862,4 +862,16 @@ public static Request newPut() {
...
+	 * Seta a new value for the request code.
+	 * 
+	 * Needed for obfuscation in OSCORE.
+	 * 
+	 * @param value
+	 */
+    public void setCode(Code value) {
+       this.code = value;
+        
+    }
```
boaks: FMPOV, the current Request/Reponse/Message is pretty unclear about what could be set/changed in which stage of processing the message. With adding this setter, it get's even more unclear. 

`public static byte[] decrypt(Request request) `

seems to change so much of the request, that I would try to go for creating a new request (forwarding/copying the MessageObserver) instead of exchanging that much of the request.

boaks:
Thanks for the repo link!

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
boaks:
```
    public OSCoreServer(NetworkConfig config, int... ports) {
        super(config,ports);
        this.unprotected = new ArrayList<>();
        
        //Erase the endpoints created by super
        this.endpoints = new ArrayList<>();
        this.unprotected = new ArrayList<>();
        
        // create endpoint for each port
        for (int port:ports) {
            addEndpoint(new OSCoreEndpoint(port, config, this.unprotected));
        }
    }
```

do you try to call super with a empty list of ports (just `super(config);`)? 
Then there would be no requirement to remove the useless created endpoints.

If you insist on doing it in your way, can you please ensure, that the already created endpoint are proper released (calling destroy())? (But I would strongly recommend to go for the empty port list).

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -1070,7 +1092,7 @@ public void cancelObservation(Token token) {
 	 *
 	 * @param task the task
 	 */
-	private void runInProtocolStage(final Runnable task) {
+	protected void runInProtocolStage(final Runnable task) {
```
boaks: The most scope changes seems to be not used in `OSCoreEndpoint`.

Maybe, if we offer a better way to create a special `CoapStack`, subclassing may be obsolete at all.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
boaks: OSCoreClient

```
    @Override
    protected Request send(Request request, Endpoint outEndpoint) {
        OSCoreEndpoint ep = (OSCoreEndpoint) outEndpoint;

        request.getOptions().addOption(new Option(OptionNumberRegistry.OSCORE));
        // use the specified message type
        request.setType(super.type);

        if (this.blockwise!=0) {
            request.getOptions().setBlock2(new BlockOption(BlockOption.size2Szx(this.blockwise), false, 0));
        }

        ep.sendRequest(request);
        return request;
    }
```

Any reason, why OSCORE requires to add the (default) type here? The original coap client just apply that type in some cases, but offer also advanced methods to send requests with an other type.

californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
T0bbe: - But then why don't make an argument to remove more of the class' setters? (I followed the class' already established setters/getters).
- The current way is obviously more convenient, but I could just do as you say if it feels/looks better for the code.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -74,7 +74,7 @@
 /**
  * A Matcher for CoAP messages transmitted over UDP.
  */
-public final class UdpMatcher extends BaseMatcher {
+public class UdpMatcher extends BaseMatcher {
```
boaks: I can't see, that OSCORE is "extending" the `UdpMatcher`. Can you point me to the extension.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BaseCoapStack.java
```diff
@@ -47,7 +47,7 @@
 
 	private List<Layer> layers;
 	private final Outbox outbox;
-	private final StackTopAdapter top;
+	protected StackTopAdapter top;
```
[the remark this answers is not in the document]
T0bbe: You are correct. It is not used anymore.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
T0bbe: You are correct. It is not used anymore.

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
T0bbe: Exactly, just for convenience. Suggestions: add object-sec as a new option and handle CoaP Code as an Other-option.

None option?

californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
boaks: The current Message API design is not mine :-). 

But I got aware, that it's hard to know, what is when possible and what will fail, when called at the wrong time. Therefore I have my doubt, if it is a good idea to add that setter.

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
T0bbe: I like your way. I will change it.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
T0bbe: Not anymore, i think. I will do some testing, but I think that I can remove it safely.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
boaks: > A lot of variables needs to be accessible for the OSCoreEndpoint since the CoapEndpoint's constructor has become very strict. 

I can't see, where you access these variables, can you point me to that?
You access `this.coapstack` but the others seems to be not used by `OSCoreEndpoint`.

The number of CoapEndpoint constructor has grown over the time by adding optional parameters. Therefore I changed it to a "Builder2, which unfortunately, is not sub-class friendly. But your constructors can simply call "this()" and so only one constructor must then call "super()" and use the "null"-defaults. 

> If you brought back one of these early constructors which don't need a crazy amount of input the exposure for CoapEndpoint would not need to be changed as much.

I don't see, that this is related. But maybe, you can point me to the usage of these variables in OSCORE.

boaks:
I hope, the changes in Token processing from the begin of the year are reflected in your Token processing but from my first impression, this is currently not the case. To handle address change for long term observations, californium is using a "global unique" token, not longer a "destination endpoint unique" token (which will be too hard to handle address changes).

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
boaks: I prefer to have OptionSet for options and not for a "easy OptionJuggle". May be you consider to change the processing in a way, where options kept separated from other artefacts. A approach, which creates a different Message/Request/Response instance could help to sort the functions in an other way.

boaks:
I'm really wondering, what the intention about the `Token` in the OSCORE implementation is.
On the one side, using 2 bytes seems to indicate, that tokens are intended to be unique per destination endpoint. But the `ByteId` will then fail. Or is the intention to have only a few request in process? Or would it be possible to switch to use the `TOKEN_SIZE_LIMIT` from the `NetworkConfig`?

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
T0bbe: I like your suggestion and I will make the changes.

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
```diff
@@ -163,6 +166,7 @@ public OptionSet(OptionSet origin) {
 			block2          = new BlockOption(origin.block2);
 		
 		observe = origin.observe;
+		oscore	= origin.oscore.clone();
```
boaks: check for "oscore != null" before call clone()?

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
boaks: I miss the remove of this change.

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
```diff
@@ -1270,7 +1275,46 @@ public OptionSet removeObserve() {
...
+	
+	/**
+	 * Replaces the Oscore option with oscore.
+	 * Returns the current OptionSet object for a fluent API.
+	 * 
+	 * @param oscore the new Oscore value
+	 * @return this OptionSet
+	 */
+	public OptionSet setOscore(byte[] oscore){
+		if(oscore != null)
```
boaks: The other setter are using exceptions to indicate argument errors. I would go for a NullPointerException. Please add that also to the javadoc.

boaks:
Just to mention: if you use the eclipse ide, we have a code formatter.
It's not that easy, because e.g. OptionSet seems to no comply with that formatter :-) (historical reasons), but maybe you can comply to use tabs and ws as the other lines nearby.

boaks:
To the other committers:
The "hooks" have now been reduced a lot. Thanks T0bbe!
So FMPOV, I would go for merge this hooks soon. To have oscore been working with 2.0.x  (or again), will take some time and I guess, there are some more changes waiting :-). But I think it's a good start.
Any other opinions?

T0bbe:
I will be certain to use the formatter.

Do we have a time line for the merge of the hooks?

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
boaks: Still pending to remove or do you require this change?

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
boaks: Any information on that?

boaks:
> Do we have a time line for the merge of the hooks?

FMPOV two issues are open. Maybe these two changes are required, but for now, I think they may be improved. When done with that, I would mark the PR with "merged soon - if no objection" for a couple of days, and, if no concerns are written, then I will merge it.

T0bbe:
Sounds great. Thanks for the quick reply :)

PR #605, opened 2018-04-09
Unknown PSK Identity and Bad PSK should be both handled in a same way

sbernard31:
This PR aims to fix this issue : https://bugs.eclipse.org/bugs/show_bug.cgi?id=533258.

An "unknown PSK identity" will be ignored as a "bad PSK".

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
@@ -689,7 +689,7 @@ public class ServerHandshaker extends Handshaker {
 		if (psk == null) {
 			throw new HandshakeException(
 					String.format("Cannot authenticate client, identity [%s] is unknown", identity),
-					new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, session.getPeer()));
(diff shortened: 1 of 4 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -689,7 +689,7 @@ private void createCertificateRequest(final ClientHello clientHello, final DTLSF
 		if (psk == null) {
 			throw new HandshakeException(
```
boaks: It may be academic, but wasn't the discussion, that in this case, a random secret is assumed to "fake" the calculation times?

sbernard31: As we don't answer at all, there is no calculation time to detect.

But if we had chosen to send "decrypt_error" alert, we should have to do something like this. (I mean fake calculation)

This is one of the reason which make this solution simple to implement.

boaks: :-). If we do a parallel handshake, we may see the difference there :-).

boaks:
LGTM

PR #607, opened 2018-04-11
Extend demos server and client.

boaks:
Add argument -i for id and pw to client and make endpoints for server
selectable.

### DIFF
--- a/demo-apps/cf-extplugtest-server/src/main/java/org/eclipse/californium/extplugtests/ExtendedTestServer.java
+++ b/demo-apps/cf-extplugtest-server/src/main/java/org/eclipse/californium/extplugtests/ExtendedTestServer.java
@@ -21,6 +21,7 @@ package org.eclipse.californium.extplugtests;
 import java.io.File;
 import java.net.SocketException;
 import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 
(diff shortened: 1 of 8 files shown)


demo-apps/cf-extplugtest-server/src/main/java/org/eclipse/californium/extplugtests/ExtendedTestServer.java
```diff
@@ -60,7 +61,8 @@ public void applyDefaults(NetworkConfig config) {
 			config.setInt(Keys.MAX_RESOURCE_BODY_SIZE, DEFAULT_MAX_RESOURCE_SIZE);
 			config.setInt(Keys.MAX_MESSAGE_SIZE, DEFAULT_BLOCK_SIZE);
 			config.setInt(Keys.PREFERRED_BLOCK_SIZE, DEFAULT_BLOCK_SIZE);
-			config.setInt(Keys.EXCHANGE_LIFETIME, 24700); // 24.7s instead of 247s
+			config.setInt(Keys.EXCHANGE_LIFETIME, 24700); // 24.7s instead of
+															// 247s
```
sbernard31: (a bit out of topic :/)
Do we have an issue with our formatter ? 
(It seems we have a 120 width line limit for code and 80 width line for comments and so this make some strange behavior like this for inline comments)

boaks:
Sometimes the formatter creates funny stuff ...

PR #611, opened 2018-04-13
[backport for 1.0-x] Unknown PSK Identity and Bad PSK should be both handled in a same way

sbernard31:
This is backport fix of #605 for the 1.0-x branch.

531fe52 is not directly related but as this test failed on my machine, I backport the `Request.setURI` implementation of 2.0.x branch.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
@@ -640,7 +640,7 @@ public class ServerHandshaker extends Handshaker {
 		if (psk == null) {
 			throw new HandshakeException(
 					String.format("Cannot authenticate client, identity [%s] is unknown", identity),
-					new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, session.getPeer()));
+					new AlertMessage(AlertLevel.FATAL, AlertDescription.UNKNOWN_PSK_IDENTITY, session.getPeer()));
 		}
 		
 		session.setPeerIdentity(new PreSharedKeyIdentity(identity));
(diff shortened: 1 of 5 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -640,7 +640,7 @@ private DTLSFlight receivedClientHello(ClientHello message) throws HandshakeExce
 		if (psk == null) {
 			throw new HandshakeException(
 					String.format("Cannot authenticate client, identity [%s] is unknown", identity),
-					new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, session.getPeer()));
+					new AlertMessage(AlertLevel.FATAL, AlertDescription.UNKNOWN_PSK_IDENTITY, session.getPeer()));
```
sophokles73: Wouldn't it make more sense to use `AlertLevel.WARNING` and leave `DTLSConnector.handleExceptionDuringHandshake` untouched?

boaks: I'm not sure about the "sense". 
It seems, that until now, WARNING is only used for close and therefore the implementation seems not to send other WARNING alert records to the other peers. But I would expect them to be send.
So to use WARNING instead of FATAL seems FMPOV only working in the above case, because the implementation is not as I would expect it.
Therefore I would prefer to stay with FATAL, not that later the sending of WARNING needs to b fixed and the breaks this issue again.

boaks: @sophokles73 
Do you still want to use a WARNING or could we merge it as it is?

boaks: LGTM!

Six days waiting for @sophokles73 to explain his comment seems to be enough and we can simply add such changes later.

Issue #609, opened 2018-04-13
Cipher init in every message decryption


joechen2010:
DTLS performance penalty is the decrypt the message in CCMBlockCipher. the cipher init every message and it 's a synchronized action, should be move to DTLSession when it creating.
`Cipher cipher = Cipher.getInstance(CIPHER_NAME); `
`cipher.init(Cipher.ENCRYPT_MODE, new SecretKeySpec(key, KEY_TYPE));`
found branch: 2.0.0-M9 

thread dump :

"DTLS-Connection-Handler-47" #171 daemon prio=5 os_prio=0 tid=0x0000000000868000 nid=0x66e86 runnable [0x00007f1d8b7f6000]
   java.lang.Thread.State: RUNNABLE
                at java.util.Arrays.copyOfRange(Arrays.java:3664)
                at java.lang.String.<init>(String.java:207)
                at java.lang.String.toUpperCase(String.java:2810)
                at java.security.Provider$ServiceKey.<init>(Provider.java:872)
                at java.security.Provider$ServiceKey.<init>(Provider.java:865)
                at java.security.Provider.getService(Provider.java:1039)
                - locked <0x00000003c0b1b8f8> (a com.sun.crypto.provider.SunJCE)
                at sun.security.jca.ProviderList$ServiceList.tryGet(ProviderList.java:444)
                at sun.security.jca.ProviderList$ServiceList.access$200(ProviderList.java:376)
                at sun.security.jca.ProviderList$ServiceList$1.hasNext(ProviderList.java:486)
                at javax.crypto.Cipher.getInstance(Cipher.java:513)
                at org.eclipse.californium.scandium.dtls.cipher.CCMBlockCipher.decrypt(CCMBlockCipher.java:310)
                at org.eclipse.californium.scandium.dtls.Record.decryptAEAD(Record.java:619)
                at org.eclipse.californium.scandium.dtls.Record.decryptFragment(Record.java:369)
                at org.eclipse.californium.scandium.dtls.Record.decryptApplicationMessage(Record.java:880)
                at org.eclipse.californium.scandium.dtls.Record.getFragment(Record.java:848)
                at org.eclipse.californium.scandium.dtls.Record.getFragment(Record.java:817)
                at org.eclipse.californium.scandium.DTLSConnector.processApplicationDataRecord(DTLSConnector.java:700)
                - locked <0x000000053cb4b450> (a org.eclipse.californium.scandium.dtls.DTLSSession)
[...]

PR #612, opened 2018-04-13
Issue #609, reuse cipher.

boaks:
This variant reuses the Cipher instance.
With Record.CIPHER_INIT_ON_MESSAGE it's possible to test with or without init for each message.
Doesn't differ much, if init is repeated:

// init once
5000000 requests in 385612ms, 12966 reqs/s
5000000 requests in 381313ms, 13112 reqs/s
5000000 requests in 387289ms, 12910 reqs/s

// init on message
5000000 requests in 392561ms, 12736 reqs/s
5000000 requests in 388110ms, 12882 reqs/s
5000000 requests in 384122ms, 13016 reqs/s

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSConnectionState.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSConnectionState.java
@@ -18,14 +18,20 @@
  *    Kai Hudalla (Bosch Software Innovations GmbH) - add toString()
  *    Kai Hudalla (Bosch Software Innovations GmbH) - improve JavaDocs, add method for retrieving
  *                                                    maximum ciphertext expansion of cipher suite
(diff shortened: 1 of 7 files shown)


boaks:
Though the dtls-connection-state holds the cipher, I don't see a issue with a multiple uses dtls session.

sbernard31:
I'm not sure Cipher is stateless or threadsafe. (https://stackoverflow.com/questions/6957406/is-cipher-thread-safe)

Before this modification, Cipher was encapsulate in decrypt/encrypt so we are ok.

Now we move it in DTLSConnectionState, so potentially the Cipher could be used by several thread at the same time. I mean nothing prevent this in CCMBlockCipher.

My understanding is thanks to the StripedExecutor we are OK.

But as I'm no so confident, I share my thought with you :-)

boaks:
Withdrawn for the more general PR #614.

PR #614, opened 2018-04-21
Make cipher threadlocal.

boaks:
Some numbers of a thread-local implementation:

5000000 requests in 407513ms, 12269 reqs/s
5000000 requests in 407701ms, 12263 reqs/s
5000000 requests in 406746ms, 12292 reqs/s

Seem to be a little slower than #612 

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CCMBlockCipher.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CCMBlockCipher.java
@@ -76,6 +76,7 @@ public class CCMBlockCipher {
 		 * </pre>
 		 * 
 		 * Return remaining bytes in number.
+		 * 
 		 * <pre>
 		 * blockSize = 16;
 		 * number = 0x20103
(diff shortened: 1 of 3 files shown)


boaks:
In the meantime I prefer this PR, even if it's a little slower, it has the smallest side effect and could be applied to all branches.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CCMBlockCipher.java
```diff
@@ -307,7 +324,7 @@ protected int xorInt(int offset, int end, int number) {
 		 */
 
 		// instantiate the underlying block cipher
-		Cipher cipher = Cipher.getInstance(CIPHER_NAME);
+		Cipher cipher = threadLocalCipher.get();
```
sbernard31: if `Cipher.getInstance(CIPHER_NAME)` failed next line will raise an NPE. Should we handle this in a better way ?

boaks: I'm not sure, if coaps makes sense without AES. 
So this would only report a nicer error in those cases, where dtls could not be used at all.

sbernard31: I agree, but I think at least a log would be valuable

boaks: OK, I will  add that.

boaks: Done.

boaks:
5000000 requests in 412077ms, 12133 reqs/s
5000000 requests in 408966ms, 12225 reqs/s
5000000 requests in 409569ms, 12207 reqs/s

For the last variant, which supports also other ciphers.

PR #616, opened 2018-04-23
Use availability of a certificate verifier also to enable x509 support.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
@@ -336,13 +336,12 @@ public abstract class Handshaker {
 			int epoch = candidate.getEpoch();
 			if (epoch < session.getReadEpoch()) {
 				// discard old message
-				LOGGER.debug(
-						"Discarding message from peer [{}] from past epoch [{}] < current epoch [{}]",
-						new Object[]{getPeerAddress(), epoch, session.getReadEpoch()});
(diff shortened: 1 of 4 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -975,7 +970,14 @@ protected final void expectChangeCipherSpecMessage() {
 	 */
 	public void verifyCertificate(CertificateMessage message) throws HandshakeException {
 		if (message.getCertificateChain() != null) {
-			certificateVerifier.verifyCertificate(message, session);
+			if (certificateVerifier != null) {
+				certificateVerifier.verifyCertificate(message, session);
+			} else {
```
sbernard31: A certificate Verifier == null means that we don't support X.509 right ?

If we don't support X.509 the certificate type/cipher suite negotiation should have failed before.
So if we still receive a CertificateMessage with a CertificateChain, I don't know if we should log and raise bad certificate.

I feel this is more an error from the foreign peer (or a bug at negotiation time).
I feel this is more an HANDSHAKE_FAILURE and the log is more about invalid "Certifcate message ..."

boaks: Or unexpected_message?
Because sending such a certificate in that case is caused by a invalid implementation,

sbernard31: unexpected_message sounds better :+1:

Issue #619, opened 2018-04-25
Session Ticket implemented ? 


sbernard31:
While I'm trying to implement #617 (Session lifetime), I faced some code about Ticket Session.
But I'm not sure to understand if [rfc5077](https://tools.ietf.org/html/rfc5077) is implemented ?

It seems to me this work is not really finished ? Am I right ?

2nd question : is SessionCache really used ? and what is the purpose of this ?

boaks:
For me this look like a idea to persist or transfer the dtls session information.
I guess, this was done for the "cluster" or DTLS session overspaning a server restart.
I don't think it's intended to be used for RFC5077.

sbernard31:
> For me this look like a idea to persist or transfer the dtls session information.
> I guess, this was done for the "cluster" or DTLS session overspaning a server restart.

To allowing session resumption in such use cases . I think you're right.

But, this part of the code directly refer the RFC5077 : https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SessionTicket.java#L75

@sophokles73, do you have more information about that ? Do you know if this is actually used ?

sbernard31:
> At least it was my intention to support RFC 5077 when I wrote the code.

Ok, but this implementation is not finished ? I mean you just lay the foundation, right ?

sophokles73:
IIRC the idea was to have the ticket and codec in place so that it can already be used to store session state in a SessionCache. But I did not implement the HELLO extension nor the server's NewSessionTicket message, i.e. a client currently cannot resume a session based on a Session Ticket.

sbernard31:
Yep

Issue #617, opened 2018-04-24
Define a lifetime for DTLS session


sbernard31:
The TLS 1.2 spec  says: _"An upper limit of 24 hours is suggested for session ID lifetimes, since an attacker who obtains a master_secret may be able to impersonate the compromised party until the corresponding session ID is retired."_

We should implement this lifetime, making it configurable and setting it by default to 24h.
A short term solution could be to make `DTLSConnector(final DtlsConnectorConfig configuration, final ResumptionSupportingConnectionStore connectionStore)` public. This way anyone could implement its own store and so add session/connection lifetime.

(This seems to be a [community needs](https://dev.eclipse.org/mhonarc/lists/leshan-dev/msg00976.html))

boaks:
https://tools.ietf.org/html/rfc5246#appendix-F.1.4

> Sessions cannot be resumed unless both the client and server agree.
> If either party suspects that the session may have been compromised,
> or that certificates may have expired or been revoked, it should
> force a full handshake. An upper limit of 24 hours is suggested for
> session ID lifetimes, since an attacker who obtains a master_secret
> may be able to impersonate the compromised party until the
> corresponding session ID is retired. Applications that may be run in
> relatively insecure environments should not write session IDs to
> stable storage.

So very first we should define, if this applies to "Resuming Sessions" or also to "connections". Though

```
public synchronized Connection get(final InetSocketAddress peerAddress) {
	return checkValidity(connections.get(peerAddress));
}
```

seems to remove not only sessions, when they are tried to be resumed, it removes also "connections". Therefore I would guess, that this may have some "unwanted side effects". One I see, is that in cases of "expired connections", terminating the connection is not longer indicated to the other peer.

boaks:
Yes, I would go for two values (0 for disabled and **NO** "<" defined relation between the two values).

sophokles73:
I am not sure if I can follow. Looking at it from a server's perspective, I would argue that it is irrelevant whether a session is established and in use or whether it is resumed by a client at a later time. My understanding of the spec is that the lifetime of the session ID of the established session must/should not exceed 24h. Based on that I would think that the server terminates an established session once it has exceeded the limit and also reject attempts from clients to resume the session. The latter could be implemented by simply evicting all of the connection's  state from the ConnectionStore. Or am I mistaken?

sbernard31:
[...]

PR #620, opened 2018-04-26
Add a lifetime for DTLS session

sbernard31:
This PR aims to add a session lifetime (see #617)

I made `DTLSConnector(final DtlsConnectorConfig configuration, final ResumptionSupportingConnectionStore connectionStore)` `public` to allow other implementation of `InMemoryConnectionStore` and so other implementation of Session lifetime.

This is a simple implementation without more thread, sessions are cleaned on "access" method of `InMemoryConnectionStore`.

I could add a DTLSConfig parameter if we agree on those first commits.

The sessionCache and sessionTicket is not so clear for me, we could discuss about this in #619.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
@@ -64,10 +64,12 @@ import org.eclipse.californium.elements.util.LeastRecentlyUsedCache.Predicate;
 public final class InMemoryConnectionStore implements ResumptionSupportingConnectionStore, SessionListener {
 
 	private static final Logger LOG = LoggerFactory.getLogger(InMemoryConnectionStore.class.getName());
+	private static final int DEFAULT_SESSION_LIFETIME = 24 * 60 * 60; // 24h
 	private static final int DEFAULT_CACHE_SIZE = 150000;
 	private static final long DEFAULT_EXPIRATION_THRESHOLD = 36 * 60 * 60; // 36h
 	private final LeastRecentlyUsedCache<InetSocketAddress, Connection> connections;
 	private final SessionCache sessionCache;
+	private final long sessionLifeTime; // in seconds
 
 	/**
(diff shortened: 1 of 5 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
```diff
@@ -272,4 +294,31 @@ public void handshakeCompleted(final InetSocketAddress peer) {
...
+	 */
+	private Connection checkValidity(Connection connection) {
+		if (connection == null) {
+			return null;
+		}
+		if (!connection.hasEstablishedSession()) {
+			if (!connection.hasSessionTicket()) {
+				return connection;
+			} else if (!connection.getSessionTicket().isExpired(sessionLifeTime)) {
+				return connection;
```
boaks: I think, this check is better done in `deepfind` and return null, when expired.

Connections with tickets live only very short in `DTLSConnector.resumeExistingSession`.
They are not stored in the connection store and therefore need not to be removed from that.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
```diff
@@ -272,4 +294,31 @@ public void handshakeCompleted(final InetSocketAddress peer) {
...
+			}
+		} else if (!connection.getEstablishedSession().isExpired(sessionLifeTime)) {
+			return connection;
+		}
+
+		// connection expired
+		connections.remove(connection.getPeerAddress(), connection);
+		removeSessionFromCache(connection);
```
boaks: `removeSessionFromCache` will not remove session from cache, if the connection is created for a ticket.
See comment above; I would prefer to process the expired tickets in `deepfind`.

sbernard31:
Thx @boaks  for the review .
I will waiting we agree about what we should do #617 before to take this in account.

sbernard31:
This PR is no longer needed.
#1007 obsoleted it.

PR #623, opened 2018-05-02
Fix 511

### LINKED ISSUE #511: PSK identity requires extension for SNI
Branch 2.0.x

The `PreSharedKeyIdentity` doesn't obey a optional indicated server name. Therefore the `Principal` equals check may report accidently the same `Prinicipal`, if the same identiy name is used for different servers.

Requires extension of `PreSharedKeyIdentity` and fixes in `ServerHandshaker` and `ClientHandshaker`.

sophokles73:
This PR should fix #511.

The (server side) DTLSConnetor now scopes the `PreSharedKeyIdentity` created during a PSK based handshake to the *virtual host* that has been conveyed in the client's SNI HELLO extension during a PSK based handshake.

The virtual host that is sent in the SNI extension is passed into the (client side) DTLSConnector as part of the `RawData` message passed in to the *send(RawData)* method. The `DataSerializer` sets the virtual host to the value of a CoAP request's *Uri-Host* value. The connector will refuse to send a message over an established connection if the virtual host that the session is scoped to doesn't match the message's Uri-Host (if set).


element-connector/src/main/java/org/eclipse/californium/elements/auth/PreSharedKeyIdentity.java
```diff
@@ -90,6 +145,13 @@ public boolean equals(Object obj) {
 		} else if (!identity.equals(other.identity)) {
 			return false;
 		}
+		if (virtualHost == null) {
```
boaks: Compare `name` instead of `identity` may do it.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -892,13 +858,15 @@ public Builder setConnectionThreadCount(int threadCount) {
...
+		 * Sets the timeout of automatic session resumption in milliseconds.
 		 * <p>
 		 * The default value is {@code null}, no automatic session resumption.
 		 * 
+		 * @param timeoutInMillis the number of milliseconds or {@code null} to disable automatic
+		 *                        session resumption
 		 * @return this builder for command chaining.
 		 */
-		public Builder setAutoResumptionTimeoutMillis(long timeoutInMillis) {
+		public Builder setAutoResumptionTimeoutMillis(Long timeoutInMillis) {
```
boaks: I'm not sure, if this brings a pay-off. null would only be required, if either the default is changed or a already set value should be overwritten.

element-connector/src/main/java/org/eclipse/californium/elements/auth/PreSharedKeyIdentity.java
```diff
@@ -21,32 +21,82 @@
...
+			result = prime * result + ((identity == null) ? 0 : identity.hashCode());
+			result = prime * result + ((virtualHost == null) ? 0 : virtualHost.hashCode());
+			this.hash = result;
+
+			StringBuilder b = new StringBuilder();
+			if (virtualHost != null) {
+				b.append(virtualHost).append(":");
+			}
+			b.append(identity);
+			this.name = b.toString();
```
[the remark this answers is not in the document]
sophokles73: ???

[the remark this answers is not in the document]
sophokles73: ???

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
sophokles73: you mean there should be no need to explicitly set it to `null`?

element-connector/src/main/java/org/eclipse/californium/elements/auth/PreSharedKeyIdentity.java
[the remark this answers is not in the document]
boaks: > final int prime = 31;
> int result = 1;
> result = prime * result + ((identity == null) ? 0 : identity.hashCode());
> result = prime * result + ((virtualHost == null) ? 0 : virtualHost.hashCode());
> this.hash = result;

calculates a hash from `identity` and `virtualHost`, which functionality seems to be the same as `name.hashCode()` (the functionality not the value!). So I simply would rely on that `String.hashCode()` implementation, I can't see a need to precalculate a hash in that case.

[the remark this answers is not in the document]
sophokles73: Oh, I see, you want to use the hash of the *name* property instead of calculating the hash from the identity and virtual host properties. That makes sense. :+1:

boaks:
FMPOV this PR contains also a lot of "cleanup" (e.g. reduced visibility of keys/ids see issue #624), but I'm not convinced, that this are all improvements.

Generally the PR assumes, that only requests triggers new DTLS handshakes. According https://tools.ietf.org/html/rfc7252#section-9.1.2 that's right. But one of the intentions of the introduction of `EndpointContext` was to overcome this limitation and make it possible, to match based on the principal (if they are unique). So, if currently a "notify" triggers a dtls handshake, SNI would not work ;-(.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -676,26 +683,32 @@ private void createCertificateRequest(final ClientHello clientHello, final DTLSF
...
-
-		LOGGER.debug("Client [{}] uses PSK identity [{}]",
-				new Object[]{getPeerAddress(), identity});
+		byte[] psk = pskStore.getKey(getIndicatedServerNames(), identity);
+		String virtualHost = null;
 
 		if (getIndicatedServerNames() == null) {
-			psk = pskStore.getKey(identity);
+			LOGGER.debug("Client [{}] uses PSK identity [{}]",
+					new Object[]{getPeerAddress(), identity});
```
boaks: Recently, we don't use the explicit `new Object[]{ ... }`

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -676,26 +683,32 @@ private void createCertificateRequest(final ClientHello clientHello, final DTLSF
...
 		} else {
-			psk = pskStore.getKey(getIndicatedServerNames(), identity);
+			ServerName serverName = getIndicatedServerNames().getServerName(NameType.HOST_NAME);
+			if (serverName == null) {
+				LOGGER.debug("client provided invalid SNI extension which doesn't include a hostname");
+			} else {
+				virtualHost = new String(serverName.getName(), ServerName.CHARSET);
+				LOGGER.debug("Client [{}] uses PSK identity [{}] for server [{}]",
+						new Object[]{getPeerAddress(), identity, virtualHost});
```
boaks: Same here, recently, we don't use the explicit `new Object[]{ ... }`

sophokles73:
> So, if currently a "notify" triggers a dtls handshake, SNI would not work ;-(.

Let me try to understand the scenarion
1. A client has started to observe a resource on the server by means of a get request which included a Uri-Host option containing the *virtual host*
2. The connection times out, dies, whatever
3. The server's resource changes and the server thus sends a notification to the client, which triggers a new DTLS handshake (now with roles reversed)

Can you explain what you mean by *SNI would not work* in this scenario?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
sophokles73: are we following any *good practice* as to which places we correct it and which not? When you guys work on a class, do you replace all occurrences with the log4j based syntax or just in those places that you are working on anyway?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/InMemoryPskStore.java
```diff
@@ -21,56 +22,46 @@
...
-
-	private static byte[] getKeyFromMap(final String identity, final Map<String, byte[]> keyMap) {
-
-		byte[] result = null;
-		if (keyMap != null) {
-			byte[] key = keyMap.get(identity);
-			if (key != null) {
-				// defensive copy
-				result = Arrays.copyOf(key, key.length);
+			synchronized (scopedKeys) {
```
boaks: I'm not sure, if `addKnownPeer` should be called dynamically during usage/handshakes. I would guess, that, if this `PskStore` is intended to be used dynamically during execution time, a `removeKnownPeer` is also required. If a dynamic usage should be supported, I would prefer to synchronize on `this`.

boaks:
> Can you explain what you mean by SNI would not work in this scenario?

In that scenario it works :-)! 

Now complicate it a little more, e.g. for the LWM2M use-case:

1. The LWM2M device starts as coap- and dtls-client and sends it's registration request along with the SNI extension. 
1. The LWM2M server then start a observe (dtls established, now acts as coap-client).
1. Time passes, dtls "connection" and session lost.
1. The LWM2M device send a notify the LWM2M server. That would trigger the LWM2M device trigger again a dtls handshake as dtls-client, but the SNI is not available.

According RFC7252 this is not possible, but therefore we introduced the `EndpointContext`.

boaks:
https://tools.ietf.org/html/rfc6066#section-3

> Currently, the only server names supported are DNS hostnames;

But in fact, our intention is to use it as tenant, not related to the DNS hostname!
I guess we should document that at a prominent location.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/InMemoryPskStore.java
sophokles73: by *dynamic*, I suppose, you mean adding and removing peers during runtime, i.e. after the connector has been started, right? I didn't really think about that, since the *addKnownPeer* method had been there before already. However, adding a *removeKnownPeer* method seem reasonable to me. But why do you think it would be better to synchronize on the class instance itself?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks: AFAIK we don't apply a "good practice". 
But for me it seems to be a new logging statements so I would go for the style without `new Object[]{ ... }`.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/InMemoryPskStore.java
boaks: Because the synchronization belongs to both collections. Otherwise we may add a comment, that accessing the other collection should also be protected by synchronization the other. Therefore it seems for me to be more straight to synchronize on `this`.

sophokles73:
> But in fact, our intention is to use it as tenant, not related to the DNS hostname!

Why shouldn't this be related to a hostname? What about `my-tenant.coap.eclipse.org`?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/InMemoryPskStore.java
sophokles73: I would rather add the comment and synchronize on an internal object instead of the class instance itself. Otherwise other code can interfere with the synchronization simply by also synchronizing on the class instance as well .

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
sophokles73: ok, then I will replace it in the code that I have *touched* and leave the rest *as-is*

boaks:
> Why shouldn't this be related to a hostname? What about my-tenant.coap.eclipse.org?

Ah, OK, then maybe a mapping for that hostname into a tenant name would be required.
With this PR all request, which use a URL with hostname and COAPS, will use SNI, right?
I would prefer to have that configurable. And the issue with https://github.com/eclipse/californium/pull/623#issuecomment-386321239, the LWM2M information reporting interface, should alos be solved or at least documented, that it's not intended to work with that.

element-connector/src/main/java/org/eclipse/californium/elements/auth/PreSharedKeyIdentity.java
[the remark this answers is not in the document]
sophokles73: taking another look, I now feel that comparing the two properties individually in equals() and computing a hash based on the two properties is more appropriate. Otherwise, the following two PreSharedKeyIdenties would be considered equal:
```
PreSharedKeyIdentity idOne = new PreSharedKeyIdentity("iot.eclipse", "org:device-1");
PreSharedKeyIdentity idTwo = new PreSharedKeyIdentity("iot.eclipse:org", "device-1");
```

sophokles73:
> With this PR all request, which use a URL with hostname and COAPS, will use SNI, right?

Yes. Do you see a problem with that? vertx' HTTP client also by default adds the SNI extension using the hostname from the URI in https requests. However, we might want to change the PskStore's lookup strategy for keys to fall back to *getKey(String)* if it cannot find a key using *getKey(String, String)*.

sophokles73:
> And the issue with #623 (comment), the LWM2M information reporting interface, should alos be solved or at least documented, that it's not intended to work with that.

FMPOV it is not so much a matter of *it is not intended to work for the LWM2M case* but more of *How do we make this work?*

I had been playing around with including the virtual host name in the EndpointContext but so far, haven't been successful getting it to work. However, I am motivated to give it another shot :-)

element-connector/src/main/java/org/eclipse/californium/elements/auth/PreSharedKeyIdentity.java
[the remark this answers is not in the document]
boaks: Are you sure, that a DNS hostname is allowed to contain ":org"?
FMPOV either special characters are reserved as separator and using them within the ids is a violation and reported as error, or, a "character escape rule" must be defined (e.g. if the ":"  should be part of the hostname, it must be encoded using %3A and it should not be de-escaped when using for the PSK name..

boaks:
> Yes. Do you see a problem with that?

It may be a (small) "waste of bytes". 
And it changes the behaviour of californium. Therefore I would prefer, to have a configuration switch for that.

element-connector/src/main/java/org/eclipse/californium/elements/auth/PreSharedKeyIdentity.java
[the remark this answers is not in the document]
sophokles73: > Are you sure, that a DNS hostname is allowed to contain ":org"?

That's a very good point :-) It isn't. I will add a check in both ServerName as well as PreSharedKeyIdentity to reject illegal DNS hostnames.

boaks:
If SNI will be the default, and if a hostname is provided in the url, then the (updated?) `InMemoryPskStore` must be "filled" with care, because all peer and keys must then be added for all possible hostnames. FMPOV, introducing a "wildcard" for SNIs in that `InMemoryPskStore` would be a more usable solution. Currently, the cf-secure demo-apps example doesn't work any more. It requires additonal PSK setup (e.g. `pskStore.setKey(PSK_IDENTITY, PSK_SECRET, "localhost");`, if used with "localhost").

sophokles73:
> introducing a "wildcard" for SNIs in that InMemoryPskStore would be a more usable solution

That is what I meant with
> However, we might want to change the PskStore's lookup strategy for keys to fall back to getKey(String) if it cannot find a key using getKey(String, String).

sophokles73:
thanks for the feedback so far. I am closing this PR and will replace it with a set of new ones which have a smaller scope, hoping to make reviewing them easier ...

PR #625, opened 2018-05-03
Fix length serialization

boaks:
Fix encoded length according javadoc specification.


sbernard31:
LGTM

PR #628, opened 2018-05-07
Add ASN.1 DER decoder.

boaks:
Read length and key algorithms from ASN.1 DER encoded data.
For future support of additional key algorithms.

### DIFF
--- a/element-connector/src/test/java/org/eclipse/californium/elements/util/Asn1DerDecoderTest.java
+++ b/element-connector/src/test/java/org/eclipse/californium/elements/util/Asn1DerDecoderTest.java
@@ -1,0 +1,221 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial creation
+ ******************************************************************************/
+package org.eclipse.californium.elements.util;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+import java.io.IOException;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+/**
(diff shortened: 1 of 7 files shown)


element-connector/src/test/java/org/eclipse/californium/elements/util/Asn1DerDecoderTest.java
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * Verifies behavior of {@link Asn1DerDecoder}.
+ */
+public class Asn1DerDecoderTest {
+
+	/**
+	 * DH subject public key, ASN.1 DER / Base64 encoded.
+	 */
+	private static final String DH_BASE64 = "MIIBpjCCARsGCSqGSIb3DQEDATCCAQwCgYEA/X9TgR11EilS30qcLuzk5/YRt1I870QAwx4/gLZRJmlFXUAiUftZPY1Y+r/F9bow9subVWzXgTuAHTRv8mZgt2uZUKWkn5/oBHsQIsJPu6nX/rfGG/g7V+fGqKYVDwT7g/bTxR7DAjVUE1oWkTL2dfOuK2HXKu/yIgMZndFIAccCgYEA9+GghdabPd7LvKtcNrhXuXmUr7v6OuqC+VdMCz0HgmdRWVeOutRZT+ZxBxCBgLRJFnEj6EwoFhO3zwkyjMim4TwWeotUfI0o4KOuHiuzpnWRbqN/C/ohNWLx+2J6ASQ7zKTxvqhRkImog9/hWuWfBpKLZl6Ae1UlZAFMO/7PSSoCAgIAA4GEAAKBgH5J+o19W2ct7iGFz0/dLMaYLjCuw7TdaU2QtzZb5FmGj1TyglARYb9V3nKoqifSKlgnwFU8RBu61Sw5/gZYhAeap8kvPH7dwIrBNc4wbt5CMdicCZlSluOPrX6mYn9HzvuIaS0V8G11soSHikCCIp9gFeMLfI0AtbPOYDYD0jHA";
```
sophokles73: what about generating KeyPairs for the arbitrary algorithms using a KeyPairGenerator instead of specifying the raw bytes? I guess this is also ok but generating the keys would improve transparency FMPOV.

boaks: I had that KeyPairGenerator in my first implementation. But I remember, that slow systems will take a time to calculate them. Therefore I generated them in advance to speed up the tests

sophokles73: Slow systems? You mean, just an i5 with 3.5 GHz? IMHO we can safely assume that these test cases are not run on a CoAP client ;-) But I will leave it up to you ...

boaks: So stay with the "fast base64" :-). 
Not to slow down the other tests running on the build system.
Any other comments?

sophokles73: No, all is well.

PR #630, opened 2018-05-08
[#511] Remove hash and equals from EndpointContext implementations.

### LINKED ISSUE #511: PSK identity requires extension for SNI

sophokles73:
The hash and equals methods are not used for matching responses with
requests or for sending a message.

This is the first (small) step in a sequence of PRs that in total should then fix #511.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/UdpEndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/UdpEndpointContext.java
@@ -22,10 +22,10 @@ import java.net.InetSocketAddress;
  */
 public class UdpEndpointContext extends MapBasedEndpointContext {
 
-	static final String KEY_PLAIN = "PLAIN";
+	public static final String KEY_PLAIN = "PLAIN";
 
 	/**
-	 * Creates a new context for an IP address and port.
+	 * Creates a new context for a socket address.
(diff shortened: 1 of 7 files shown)


element-connector/src/main/java/org/eclipse/californium/elements/UdpEndpointContext.java
```diff
@@ -25,7 +25,7 @@
 	static final String KEY_PLAIN = "PLAIN";
```
boaks: Still package scope?

boaks:
Except for the package scope, LGTM.
Maybe the scope will be change in an other PR, then I would have no objection.

element-connector/src/main/java/org/eclipse/californium/elements/UdpEndpointContext.java
sophokles73: I will gladly make it public again :-)

PR #633, opened 2018-05-09
[#511] Use virtual host from EndpointContext in SNI extension.

sophokles73:
The ClientHandshaker by default will include an SNI extension in its
CLIENT_HELLO if the message that triggered the handshake is targeted at
a host name instead of a literal IP. However, SNI can be disabled using
the sniEnabled connector configuration property.


scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -145,6 +144,8 @@
 	 */
 	private Long autoResumptionTimeoutMillis;
 
+	private boolean sniEnabled = true;
```
boaks: With PR #336 we changed the design for this configuration to use references. 
This allows a upper layer to provide it's own defaults by checking, if the value was set, or a default should be used (that's the null case). I would prefer, if this `sniEnable` gets changed to `Boolean`.

sophokles73: So, the idea would be to also add a getter for this property in the builder and then set the property to its default (true) in the *build()* method if it hasn't been set explicitly yet, right?

boaks: yes.

boaks:
Would it be possible to add

			builder.setSniEnabled(false);

to demo-apps/cf-secure/.../SecureClient.java?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
```diff
@@ -243,6 +262,37 @@ void setSessionIdentifier(SessionId sessionIdentifier) {
...
+	public boolean isSniSupported() {
+		return peerSupportsSni;
+	}
+
+	void setSniSupported(boolean flag) {
+		this.peerSupportsSni = flag;
+	}
+
 	public DtlsEndpointContext getConnectionWriteContext() {
 		return new DtlsEndpointContext(peer, peerIdentity, sessionIdentifier.toString(),
```
boaks: Please add the SNI/virtualHost to the context.

sophokles73: That will be in the next step ...

boaks: OK, then please also add this to the read-context.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -205,6 +206,19 @@ public Boolean isAddressReuseEnabled() {
...
+	/**
+	 * Checks whether the connector should support the use of the TLS
+	 * <a href="https://tools.ietf.org/html/rfc6066#section-3">
+	 * Server Name Indication extension</a> in the DTLS handshake.
+	 * <p>
+	 * The default value of this property is {@code true}.
+	 * 
+	 * @return {@code true} if SNI should be used.
+	 */
+	public boolean isSniEnabled() {
```
boaks: The getter is also intended to return a "Boolean" reference, so someone can test for null to check, if the default is intended to be used. You may check leshan builders to see, how it is intended to work.

sophokles73: sure, oversaw this ...

boaks:
LGTM

Issue #508, opened 2017-12-20
DTLS session renegotiation


katkol:
I have a scenario where the dtls client hibernates for a while and during that time the IP address of the device changes. I understand from the wiki page (https://github.com/eclipse/californium/wiki/DTLS-Connection-Renegotiation-vs.-Resumption) that the scandium DTLS verifies for any change in IPaddress when 'DtlsConnector.send' is invoked. I tried to the same scenario, but what I observe in the logs and wireshark is that the client just takes the established session from connection pool and trying to send the encrypted information to the dtls server which is silently discarding the message as it cannot identify the client. In this scenario the client application is running on a different pc from the server.

The log of client and the server are attached here.  Only the dtls session steps after the hibernation are provided in the logs for easy reading. 
dtls client - [java0.log](https://github.com/eclipse/californium/files/1574634/java0.log)
dtls  server - [java1.log](https://github.com/eclipse/californium/files/1574639/java1.log)

Scadium version used in this project - 2.0.0-M5

boaks:
Go for `2.0.0-SNAPSHOT` and try to setup
 
`DtlsConnectorConfig.Builder.setAutoResumptionTimeoutMillis(long);`

In too many cases, the peer is not aware of address/port changes, because these are artefacts of NATs and other intermediate systems.

katkol:
Hi,

can you please provide the maven repository for 2.0.0-SNAPSHOT. I tried with 
 <dependency>
    <groupId>org.eclipse.californium</groupId>
    <artifactId>scandium</artifactId>
    <version>2.0.0-SNAPSHOT</version>
</dependency>

but, it failed to build.

boaks:
Try to add 

> <repository>
>       <id>repo.eclipse.org</id>
>       <name>Californium Repository</name>
>       <url>https://repo.eclipse.org/content/repositories/californium/</url>
> </repository>
> 

Otherwise, try to clone californium and build it with maven locally.

PR #634, opened 2018-05-11
added cf-Android Dtls example

vikram919:
This pull request adds a example usage of dtls server and client setup for android app.
Signed-off-by: Vikram <gopuvikram818@gmail.com>

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Request.java
@@ -357,12 +357,12 @@ public class Request extends Message {
 			}
 		}
 
-		String scheme = uri.getScheme().toLowerCase();
+		String uriScheme = uri.getScheme().toLowerCase();
 		// The Uri-Port is only for special cases where it differs from
 		// the UDP port, usually when Proxy-Scheme is used.
 		int port = uri.getPort();
 		if (port <= 0) {
-			port = CoAP.getDefaultPort(scheme);
+			port = CoAP.getDefaultPort(uriScheme);
 		}
 
 		EndpointContext destinationContext = getDestinationContext();
@@ -377,9 +377,9 @@ public class Request extends Message {
 		}
 
 		if (destinationContext == null) {
-			setDestinationContext(new AddressEndpointContext(new InetSocketAddress(destination, port), null));
+			setDestinationContext(new AddressEndpointContext(new InetSocketAddress(destination, port), getOptions().getUriHost(), null));
 		}
-		this.scheme = scheme;
+		this.scheme = uriScheme;
 		// do not set the Uri-Port option unless it is used for proxying
 		// (setting Uri-Scheme option)
 
@@ -537,8 +537,7 @@ public class Request extends Message {
 	 * 
 	 * @throws IllegalStateException if no destination endpoint context is
 	 *             available and the destination is missing
-	 * @deprecated
-	 * Note: intended to be removed with {@link #setDestination(InetAddress)} and 
+	 * @deprecated intended to be removed with {@link #setDestination(InetAddress)} and 
 	 * {@link #setDestinationPort(int)}
 	 */
 	public void prepareDestinationContext() {
@@ -547,7 +546,10 @@ public class Request extends Message {
 			if (destination == null) {
 				throw new IllegalStateException("missing destination!");
 			}
-			context = new AddressEndpointContext(destination, destinationPort);
+			context = new AddressEndpointContext(
+					new InetSocketAddress(destination, destinationPort),
+					getOptions().getUriHost(),
+					null);
 			super.setDestinationContext(context);
 		}
 	}
--- a/californium-core/src/test/java/org/eclipse/californium/core/coap/RequestTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/coap/RequestTest.java
@@ -197,6 +197,29 @@ public class RequestTest {
 		assertThat(req.getOptions().getUriHost(), is("localhost"));
 	}
 
+	/**
+	 * Verifies that the destination context contains the Uri-Host
+	 * option value.
+	 */
+	@Test
+	public void testSetURISetsVirtualHostOnDestinationContext() {
+
+		assumeTrue(dnsIsWorking());
+		Request req = Request.newGet().setURI("coap://localhost");
+		assertThat(req.getDestinationContext().getVirtualHost(), is("localhost"));
+	}
+
+	/**
+	 * Verifies that the destination context does not contain a virtual
+	 * host if a literal IP address is used as the target.
+	 */
+	@Test
+	public void testSetURIDoesNotSetVirtualHostOnDestinationContextForLiteralIP() {
+
+		Request req = Request.newGet().setURI("coap://127.0.0.1");
+		assertThat(req.getDestinationContext().getVirtualHost(), is(nullValue()));
+	}
+
 	@Test
 	public void testSetURISetsDestinationPortBasedOnUriScheme() {
 		Request req = Request.newGet().setURI("coap://127.0.0.1");
--- a/element-connector/src/main/java/org/eclipse/californium/elements/AddressEndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/AddressEndpointContext.java
@@ -35,6 +35,8 @@ public class AddressEndpointContext implements EndpointContext {
 
 	private final Principal peerIdentity;
 
+	private final String virtualHost;
+
 	/**
 	 * Creates a context for an IP address and port.
 	 * 
@@ -48,6 +50,7 @@ public class AddressEndpointContext implements EndpointContext {
 		}
 		this.peerAddress = new InetSocketAddress(address, port);
 		this.peerIdentity = null;
+		this.virtualHost = null;
 	}
 
 	/**
@@ -57,7 +60,7 @@ public class AddressEndpointContext implements EndpointContext {
 	 * @throws NullPointerException if provided peer address is {@code null}.
 	 */
 	public AddressEndpointContext(InetSocketAddress peerAddress) {
-		this(peerAddress, null);
+		this(peerAddress, null, null);
 	}
 
 	/**
@@ -68,10 +71,23 @@ public class AddressEndpointContext implements EndpointContext {
 	 * @throws NullPointerException if provided peer address is {@code null}.
 	 */
 	public AddressEndpointContext(InetSocketAddress peerAddress, Principal peerIdentity) {
+		this(peerAddress, null, peerIdentity);
+	}
+
+	/**
+	 * Create endpoint context with principal.
+	 * 
+	 * @param peerAddress socket address of peer's service
+	 * @param virtualHost the name of the virtual host at the peer
+	 * @param peerIdentity peer's principal
+	 * @throws NullPointerException if provided peer address is {@code null}.
+	 */
+	public AddressEndpointContext(InetSocketAddress peerAddress, String virtualHost, Principal peerIdentity) {
 		if (peerAddress == null) {
 			throw new NullPointerException("missing peer socket address, must not be null!");
 		}
 		this.peerAddress = peerAddress;
+		this.virtualHost = virtualHost == null ? null : virtualHost.toLowerCase();
 		this.peerIdentity = peerIdentity;
 	}
 
@@ -111,11 +127,16 @@ public class AddressEndpointContext implements EndpointContext {
 	}
 
 	@Override
+	public final String getVirtualHost() {
+		return virtualHost;
+	}
+
+	@Override
 	public String toString() {
 		return String.format("IP(%s)", getPeerAddressAsString());
 	}
 
-	protected String getPeerAddressAsString() {
+	protected final String getPeerAddressAsString() {
 		return StringUtil.toString(peerAddress);
 	}
 }
--- a/element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 Bosch Software Innovations GmbH and others.
+ * Copyright (c) 2016, 2018 Bosch Software Innovations GmbH and others.
  * 
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -29,8 +29,19 @@ import org.eclipse.californium.elements.util.StringUtil;
  */
 public class DtlsEndpointContext extends MapBasedEndpointContext {
 
+	/**
+	 * The name of the attribute that contains the DTLS session ID.
+	 */
 	public static final String KEY_SESSION_ID = "DTLS_SESSION_ID";
+	/**
+	 * The name of the attribute that contains the <em>epoch</em> of the
+	 * DTLS session.
+	 */
 	public static final String KEY_EPOCH = "DTLS_EPOCH";
+	/**
+	 * The name of the attribute that contains the cipher suite used with
+	 * the DTLS session.
+	 */
 	public static final String KEY_CIPHER = "DTLS_CIPHER";
 
 	/**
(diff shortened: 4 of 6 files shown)


demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ConfigureDtls.java
```diff
@@ -0,0 +1,86 @@
...
+    public static final String PSK_IDENTITY = "password";
+    public static final byte[] PSK_SECRET = "sesame".getBytes();
+    private static final String TRUST_NAME = "root";
+    private static final char[] TRUST_STORE_PASSWORD = "rootPass".toCharArray();
+    private static final char[] KEY_STORE_PASSWORD = "endPass".toCharArray();
+
+    public static void loadCredentials(Context context, DtlsConnectorConfig.Builder dtlsConfig, String alias){
+        try {
+
+            // load key store
```
boaks: Any reason not to use the `SslContextUtil`?

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ConfigureDtls.java
```diff
@@ -0,0 +1,86 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Institute for Pervasive Computing, ETH Zurich and others.
```
boaks: The Copyright (c) refelct the year and the authors  organization. For private persons it's just their name.
So I guess, in your case its':

* Copyright (c) 2018 Bala Vikram Reddy Gopu and others

(hope I put your names in the right order :-).

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
```diff
@@ -89,6 +107,15 @@ protected void onPreExecute() {
...
         protected CoapResponse doInBackground(String... args) {
             CoapClient client = new CoapClient(args[0]);
+            if (isSecure) {
+                CoapEndpoint.CoapEndpointBuilder endpointBuilder = new CoapEndpoint.CoapEndpointBuilder();
+                DtlsConnectorConfig.Builder dtlsConfig = new DtlsConnectorConfig.Builder();
+                dtlsConfig.setClientOnly();
+                ConfigureDtls.loadCredentials(getApplicationContext(), dtlsConfig, CLIENT_NAME);
+                DTLSConnector dtlsConnector = new DTLSConnector(dtlsConfig.build());
+                endpointBuilder.setConnector(dtlsConnector);
+                client.setEndpoint(endpointBuilder.build());
```
boaks: With that, every request creates it's own dtls connector, always trigger a new handshake and leave to old one running on the devices RAM. Is that intended? If not consider to use

`        EndpointManager.getEndpointManager().setDefaultEndpoint( ... );`

Please ensure, that you test your App also for several request, and also with a request after a longer pause.
If you intend to use it with an other server then "localhost", or to use you App also "on the way", please also check, if it works, when changing your location (sometime you may get a new ip address, if you move with your device). If it seems to stuck and the request are reported to timeout,  then you may require a "DTLS reset session" Menu or Button and/or configure "AutoResumptionTimeoutMillis" (see also issue #508).

If it's just intended to run for localhost, I would appreciate, if that is mentioned.

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
```diff
@@ -89,6 +107,15 @@ protected void onPreExecute() {
 
         protected CoapResponse doInBackground(String... args) {
             CoapClient client = new CoapClient(args[0]);
+            if (isSecure) {
```
boaks: Means: only if the localhost dtls service is running, dtls could be used. Do yo see something else, which may determine, if you want to use dtls/coaps or plain/coap?

boaks:
Is there any reason, why the DTLS endpoint is put into a own service and server? I would just start the localhost coap server and add the dtls endpoind as well.

Did you read through the android tutorials about the usage of "resources"? My feeling is, that such credentials, if not stored in a "save store", are more assets. Or, if you just want to use the demo-certs, why don't load them from the classpath using the SslContextUtil (e.g. as in demo-apps/cf-plugtest-client
ClientInitializer)?

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ConfigureDtls.java
vikram919: I thought it is obligatory for every class, now I will change it.

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
vikram919: creating new dtls connector on every request is a mistake, I will correct it now. For now I would add just local dtls server, in future commits I will add the feature of using with other dtls server as well.
I will add session auto resumption after few millis, and as well a button to stop the session.

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ConfigureDtls.java
vikram919: Initially I tired to use android URI to locate the path of the resource folder (truststore.bks and keystore.bks locations) and call .toString() method on it to get the classpath provided to SslContextUtil but it did not work.

vikram919:
I had a similar thought of adding dtls endpoint to server service, now I would do that. 
If I use ".p12" type stores, loading certificates is causing a problem so I switched to "BKS" type. It would be nice if demo certs could also add .bks stores as well. I have no further knowledge why is it the case, but for me "BKS" type worked without any problems.

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ConfigureDtls.java
boaks: Try

```
    private static final char[] KEY_STORE_PASSWORD = "endPass".toCharArray();
    private static final String KEY_STORE_LOCATION = "certs/keyStore.p12";
    private static final char[] TRUST_STORE_PASSWORD = "rootPass".toCharArray();
    private static final String TRUST_STORE_LOCATION = "certs/trustStore.p12";
    private static final String CLIENT_NAME = "client";
    private static final String CLIENT_TRUST = null; // all certificates in trust store

    .... load certificates method ...

        SslContextUtil.Credentials clientCredentials = null;
        Certificate[] trustedCertificates = null;
        try {
            clientCredentials = SslContextUtil.loadCredentials(
                    SslContextUtil.CLASSPATH_SCHEME + KEY_STORE_LOCATION, CLIENT_NAME, KEY_STORE_PASSWORD,
                    KEY_STORE_PASSWORD);
            trustedCertificates = SslContextUtil.loadTrustedCertificates(
                    SslContextUtil.CLASSPATH_SCHEME + TRUST_STORE_LOCATION, CLIENT_TRUST, TRUST_STORE_PASSWORD);
        } catch (GeneralSecurityException e) {
            e.printStackTrace();
        } catch (IOException e) {
            e.printStackTrace();
        }
```

boaks:
> If I use ".p12" type stores, loading certificates is causing a problem

I use it for android 4.4 and 6.0. So which android version you are using? 
Maybe you can check my snippet above, if loading the certs from p12 works on your device.
If not, can you provide us the error messages?
(Maybe you need to a android slf4j implementation. I use a very simple "bare" implementation I implemented on my own.)

vikram919:
I used type "BKS" on android 8.0 Oreo and it works fine.  For  type ".p12" here is the exception:
````java
Caused by: java.lang.IllegalArgumentException: no trusted x509 certificates found in 'classpath://certs/trustStore.p12' for 'server'!
        at org.eclipse.californium.elements.util.SslContextUtil.loadTrustedCertificates(SslContextUtil.java:291)
        at org.eclipse.californium.examples.ConfigureDtls.loadCredentials(ConfigureDtls.java:45)
        at org.eclipse.californium.examples.ServerService.onStartCommand(ServerService.java:59)
        at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:3677)
        at android.app.ActivityThread.-wrap21(Unknown Source:0) 
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1801) 
        at android.os.Handler.dispatchMessage(Handler.java:105) 
        at android.os.Looper.loop(Looper.java:164) 
        at android.app.ActivityThread.main(ActivityThread.java:6938) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:327) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1374)

**Comment by vikram919** [comment]
I would summarize the changes as discussed:
1. Added dtls client option in menu, it now no where related to dtls local host.
2. Replaced "Run Dtls Server" option in menu with "Run Dtls Local Server" option, this indicate dtls server running only on localhost.
3.  AutoResumptionTime for client dtls connection, choosen 200ms as it takes an average of 165ms to start a new connection.
4. Loading .bks key and trust stores with SslContextUtil.

Thank you for your time.

**Comment by boaks** [comment]
>  I used type "BKS" on android 8.0 Oreo and it works fine. For type ".p12" here is the exception:

` no trusted x509 certificates found in 'classpath://certs/trustStore.p12' for 'server'`

Sure the californium demo-cersts trustStore doesn't contain a certificate for alias "server", but for "ca" and "root". Maybe you first use some keystore tools to list the contained certificates. Or you apply my snippet in https://github.com/eclipse/californium/pull/634#discussion_r187686443 ! This contains

`private static final String CLIENT_TRUST = null; // all certificates in trust store`

So using "null" will just load all certificates from the truststore.

**Comment by boaks** [comment]
> Added dtls client option in menu, it is now no where related to dtls local host.

Lets imagine, the coap-client is something like a web-browser. Do you know, where the web-browser has it's secure checkbox or menu? Or how do the most http-client decide to use htttps? Wouldn't it be great, if the coap-client does the same (and it does)? If you find the answer, you will see, how easy it is :-).

> Replaced "Run Dtls Server" option in menu with "Run Dtls Local Server" option, this indicate dtls server running only on localhost.

I still prefer to have only one "run local server"! It's so easy to setup a coap-server with both, a coap and a coaps endpoint. It's much easier to just add the second coap-endpoint instead of your switching between secure an insecure setup.

> AutoResumptionTime for client dtls connection, choosen 200ms as it takes an average of 165ms to start a new connection.

According https://github.com/eclipse/californium/pull/634#discussion_r187658022 the idea is to test the communication with different use-case, including "longer" pauses between requests. Does this show any defects? Only, if you got affected by request timeouts using coaps, then you may check, if a "DTLS reset session" Menu or Button and/or configure "AutoResumptionTimeoutMillis" helps to solve the timeout. So did you check that? Or did you just add the setting ahead of your tests? If it takes 165ms for a handshake with the "localhost", I can't see, that this handshake time is related to the AutoResumptionTime in any way! With that setup, any request, which is sent 200ms after the last message is exchanged, will trigger again a handshake. This will only slow down and is in danger to be missinterpreted as intended example.

So my sumup is, that, if this should help as "good" example, then I think, the above remarks should be applied to it. 
Currently it's possible to use californium with android, but using dtls comes with some challenges. So my current strategy is to use my own "advanced app" to ensure, that DTLS and the TCP/TLS flavour is also working. But though the challenges are not alll solved, I didn't publish it.

**Comment by vikram919** [comment]
> So using "null" will just load all certificates from the truststore

Yeah correct now it works like charm. Really! this solves an issue in my master thesis as well.

> Lets imagine, the coap-client is something like a web-browser. Do you know, where the web-browser has it's secure checkbox or menu to enable https? Or how do the most http-client decide to use htttps? Wouldn't it be great, if the coap-client does the same (and it does)? If you find the answer, you will see, how easy it is :-)

For this before each GET request, uri need to be checked  for ex: CoAP.COAP_URI_SCHEME for unsecure coap and CoAP.COAP_SECURE_URI_SCHEME for dtls secured coap. Is this what you mean?

> I still prefer to have only one "run local server"! It's so easy to setup a coap-server with both, a coap and a coaps endpoint. It's much easier to just add the second coap-endpoint instead of your switching between secure an insecure setup.

yeah sure!, now the server is configured with both unsecure and secure ports.

File: demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java [hunk 5]
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015 Institute for Pervasive Computing, ETH Zurich and others.
+ * Copyright (c) 2018 Institute for Pervasive Computing, ETH Zurich and others.
```
boaks: You may add 2018, but not replace the 2015.

vikram919: yeah I will correct it

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
```diff
@@ -100,7 +153,23 @@ protected void onPostExecute(CoapResponse response) {
                 ((TextView)findViewById(R.id.textContent)).setText(response.getResponseText());
             } else {
                 ((TextView)findViewById(R.id.textCodeName)).setText("No response");
+                client.shutdown();
```
boaks: Why do you want to call shutdown()? Does your intention match the javadoc definition of the function?

boaks:
> For this before each GET request, uri need to be checked for ex: CoAP.COAP_URI_SCHEME for unsecure coap and CoAP.COAP_SECURE_URI_SCHEME for dtls secured coap. Is this what you mean?

Not fully. Yes, using the schema should do it. So `coap://localhost` will use plain coap over udp, and "coaps://localhost" will use secure coap over dtls/udp. But the good thing is, you don't need to handle 
this "before" :-), the coap-client does it for you! Just don't set the endpoint in the coap-client explicit, setup the plain and secure endpoint using the `EndpointManager` and you will detect, that the client chose the right endpoint magically (see `CoapCient.getEffectiveEndpoint(Request)`). So in both cases, client and server, using the available function instead of implementing a special solution will make the code much easier and then could be used as "good" example and not only as "example". 

> yeah sure!, now the server is configured with both unsecure and secure ports.

No! Now the server is configured to use "plain only" or "plain and dtls". And I guess, it's just not tested against misuse. local started, then start secure => I guess, it crashs. 
I don't see, whey it should be important, to distinguish between a "plain only" and a "plain and dtls" local server. Therefore I would just reduce it one, the one with "plan and dtls". This reduces the menu and will make the testing easier.

boaks:
> Yeah correct now it works like charm.

OK, so why do you stick to use the .bks, introducing the next copy of the demo certs in a different format? It will be easier to maintain only the ".jks" and ".p12" and so I would prefer, if these p12 format could be used.

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
```diff
@@ -103,4 +123,19 @@ protected void onPostExecute(CoapResponse response) {
             }
         }
     }
+
+    private void initCoapEndpoint(String uri){
```
boaks: Hm, this seems to be called for every request. Is this your intention?

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
```diff
@@ -103,4 +123,19 @@ protected void onPostExecute(CoapResponse response) {
...
+            dtlsConfig.setClientOnly();
+            ConfigureDtls.loadCredentials(getApplicationContext(), dtlsConfig, CLIENT_NAME);
+            DTLSConnector dtlsConnector = new DTLSConnector(dtlsConfig.build());
+            endpointBuilder.setConnector(dtlsConnector);
+        }else if(uri.startsWith(CoAP.COAP_URI_SCHEME+"://")){
+            UDPConnector connector = new UDPConnector();
+            endpointBuilder.setConnector(connector);
+        }
+        EndpointManager.getEndpointManager().setDefaultEndpoint(endpointBuilder.build());
+    }
```
boaks: Please, read my comment https://github.com/eclipse/californium/pull/634#issuecomment-388541186! 

Your idea may seem to work, but it's too complicated and not as intended. The most examples of californium usually just create a endpoint use it and drop it on termination. So these other examples are not too useful in this case. Therefore I recommend to setup the EndpointManager with both plain and secure endpoints, and then just provide the url to the coap client, and let CoapClient.getEffectiveEndpoint(Request) do the job. When you change your code to that pattern, you will see, how easy it is :-).

boaks:
And one hint: you work on your clone of the 2.0.x branch. That will crash, if something else gets merged earlier or someone pushs to the 2.0.x branch (commiter can do that :-) ). 

So read https://github.com/eclipse/californium/blob/master/CONTRIBUTING.md#making-your-changes you missed topic 3, "Create a new branch for your changes based on the 2.0.x branch".

To correct this just rename your current working branch "2.0.x" into something as "android_dtls" and push it with that name to your repo. Then fetch from the eclipse repo of californium, checkout 2.0.x and then push that original "2.0.x" to your repository using the option "force".

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ConfigureDtls.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.eclipse.californium.elements.util.SslContextUtil;
+import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
+import org.eclipse.californium.scandium.dtls.pskstore.StaticPskStore;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.security.cert.Certificate;
+
+public class ConfigureDtls {
```
boaks: With some comments you may share your knowledge others ...

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ServerService.java
```diff
@@ -12,34 +12,46 @@
...
 
     CoapServer server;
+    private static final int DTLS_PORT = 5684;
+    private static final int CoAP_PORT = 5683;
+    public static final String SERVER_NAME = "server";
 
     @Override
     public void onCreate() {
-        this.server = new CoapServer();
+        this.server = new CoapServer(CoAP_PORT);
```
boaks: Any reason, why the plain endpoint is create in "onCreate" and the secure in "onStartCommand"? Did you read the android docu for the "on???" ?

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ConfigureDtls.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final boolean RPK_MODE = false;
+    public static final String PSK_IDENTITY = "password";
+    public static final byte[] PSK_SECRET = "sesame".getBytes();
+    private static final String TRUST_NAME = null; // loads all the certificates
+    private static final String KEY_STORE_LOCATION = "certs/keyStore.p12";
+    private static final String TRUST_STORE_LOCATION = "certs/trustStore.p12";
+    private static final char[] TRUST_STORE_PASSWORD = "rootPass".toCharArray();
+    private static final char[] KEY_STORE_PASSWORD = "endPass".toCharArray();
+
+    public static void loadCredentials(Context context, DtlsConnectorConfig.Builder dtlsConfig,String alias){
```
boaks: I can't see, that `context` is longer required. So remove it?

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ServerService.java
vikram919: onCreate is called only once so we can initialize our server and its's endpoint. For onStartCommand used to run code whenever the service is called.

demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
boaks: Javadoc for service, I can't see the relation to the UI thread or asyn task. Did you mix something up? And the rest of the comment is also a mystery for me.
In which thread can you call your initCoapEndpoint and in which thread not? 
And just check the exceptions! "p12" was also "not working", but in fact it works and the exception was caused by a different mistake. In my app an setup all endpoints in the EndpointManager, plain and secure, in a call based in onCreate(). So simply pass the uri to the coapclient and leave the logic, which endpoint is selected, to that. As I wrote now several times.

PR #636, opened 2018-05-12
Added cf-Android dtls example

vikram919:
This pull request adds dtls example to existing cf-android app. It also address the activity life cycle of MainActivity "onDestroy" method which is missing in existing version. 
In current cf-Android example, when the app is destroyed without unchecking Local Server checkbox from menu options, the local server remain running in background.

### DIFF
--- a/demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ConfigureDtls.java
+++ b/demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/ConfigureDtls.java
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Vikram and others.
+ * Contributors:
+ *    Vikram - Initial creation
+ *    Achim Kraus (Bosch Software Innovations GmbH) - introduce configurable
+ *                                                    key store type and
+ *                                                    InputStreamFactory.
+ ******************************************************************************/
+package org.eclipse.californium.examples;
+
+import android.content.Context;
+
+import org.eclipse.californium.elements.util.SslContextUtil;
+import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
+import org.eclipse.californium.scandium.dtls.pskstore.StaticPskStore;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.security.cert.Certificate;
+
+/**
+ * class ConfigureDTLS.<br>
+ * This class is used to configure dtls connector for both client and server side connections.
+ * It allows to configure dtls connector in three different modes.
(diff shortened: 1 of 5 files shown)


boaks:
Looks much cleaner now than the previous PR!

In one comment of the previous PR, I recommended to use a android slf4j to see the error messages from californium. When I tried to find one, I wasn't really succesful. So, did you find one?
With my own "print all" variant, I get the following error message when I start the app:

>     writing properties to file /Californium.properties {NetworkConfig.java:422 - NetworkConfig.store}
> 05-13 16:11:08.121 5020-5020/org.eclipse.californium.examples.android W/bridge: cannot write properties to file /Californium.properties: /Californium.properties: open failed: EROFS (Read-only file system) {NetworkConfig.java:426 - NetworkConfig.store}
> 
 
Maybe you can check the NetworkConfig javadoc for the intentional use with android.

vikram919:
> writing properties to file /Californium.properties {NetworkConfig.java:422 - NetworkConfig.store}

did you add write permissions in your AndroidManifest.xml?
`<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />`

> In one comment of the previous PR, I recommended to use a android slf4j to see the error messages from californium.

yes, but I used the parent project calfornium slf4j to print the error message to logcat.

> When I tried to find one, I wasn't really succesful. So, did you find one?

https://stackoverflow.com/questions/10372394/android-best-logger-for-logging-into-file
from this link I found that slf4j-android/slf4j can only print log data to logcat but not to a file.
for the intended purpose this link propose using logback-android.

> Maybe you can check the NetworkConfig javadoc for the intentional use with android.
I will look into it and get to you if it works :)

boaks:
Sorry, I was unprecise:
It's your commit (this PR), which prints the error, if it uses a working slf4j android implementation (as my "print all" variant of the slf4j). 
So, if something is required to prevent from that error, you may add it to your PR.
But, please, first check, if you see the error in your logs. Then try your fix. Or, read the javadoc, as mentioned above.

boaks:
@vikram919 

I checked the logback implementation for android and it reports the error.
Thank's a lot! Though it's a new library, I create a CQ, https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16330

I'm not sure, how long that takes. And though you changes are not the root cause of the NetworkConfig error, I think we should merge it first without the logging. So, please remove the dependency from the gradle, when the CQ get's approved, we can add it again.

vikram919:
@boaks 

Thank you very much for your time.

boaks:
The CQ for android logback was approved.
I added your stuff again and committed it into 2.0.x.
Thanks again!

PR #639, opened 2018-05-14
[#511] Include virtual host in DtlsEndpointContext matching.

sophokles73:
Make sure that messages destined for host X are not sent over a DTLS
session established with host Y.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/KeySetEndpointContextMatcher.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/KeySetEndpointContextMatcher.java
@@ -89,4 +89,28 @@ public class KeySetEndpointContextMatcher implements EndpointContextMatcher {
 		return Collections.unmodifiableSet(new CopyOnWriteArraySet<String>(Arrays.asList(keys)));
 	}
 
+
+	/**
+	 * Checks if two endpoint contexts have the same virtual host property value.
+	 * 
+	 * @param firstContext The first context.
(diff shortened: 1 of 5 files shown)


boaks:
Is there any difference in `RelaxedDtlsEndpointContextMatcher` and `StrictDtlsEndpointContextMatcher` according the usage of the introduced `isSameVirtualHost`?

Or was the main intention of that, the using TLS may fail?

boaks:
I would prefer to add a parameter to the `KeySetEndpointContextMatcher` constructor, which defines, if the `isSameVirtualHost` is tested or not.

sophokles73:
> Is there any difference in RelaxedDtlsEndpointContextMatcher and StrictDtlsEndpointContextMatcher according the usage of the introduced isSameVirtualHost?
Or was the main intention of that, when using TLS, it may fail?

Yes, both matchers should consider the virtual host, shouldn't they?

sophokles73:
> I would prefer to add a parameter to the KeySetEndpointContextMatcher constructor, which defines, if the isSameVirtualHost is tested or not.

You mean, introducing this flag instead of overriding the *isResponseRelatedToRequest* and *isToBeSent* methods?

boaks:
> You mean, introducing this flag instead of overriding the isResponseRelatedToRequest and isToBeSent methods?

Yes! 
It would make it also a little easier, to enable that for coap over tls, when that implementation is updated to also fully support SNI.

PR #641, opened 2018-05-14
Improve plugtest checker.

boaks:
Fix NullPointerException on missing payload.
Print errors related to the code in readable format. 


boaks:
Changes only the plugtest checker, so I would like to merge it tomorrow, if no objection.

PR #643, opened 2018-05-15
[#511] Extend PreSharedKeyIdentity with virtual host property.

sophokles73:
PSK identities can now be scoped to a virtual host in order to prevent
identity clashes when SNI is used.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
@@ -35,6 +35,7 @@ import org.eclipse.californium.elements.util.StandardCharsets;
 public final class PrincipalSerializer {
 
 	private static final int PSK_LENGTH_BITS = 16;
+	private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
 
 	private PrincipalSerializer() {
 	}
@@ -91,7 +92,7 @@ public final class PrincipalSerializer {
 
 	private static void serializeIdentity(final PreSharedKeyIdentity principal, final DatagramWriter writer) {
 		writer.writeByte(ClientAuthenticationType.PSK.code);
-		byte[] virtualHost = principal.getVirtualHost() == null ? new byte[0] :
+		byte[] virtualHost = principal.getVirtualHost() == null ? EMPTY_BYTE_ARRAY :
 			principal.getVirtualHost().getBytes(StandardCharsets.UTF_8);
 		writeBytesWithLength(PSK_LENGTH_BITS, virtualHost, writer);
 		writeBytesWithLength(PSK_LENGTH_BITS, principal.getIdentity().getBytes(StandardCharsets.UTF_8), writer);
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 - 2018 Bosch Software Innovations GmbH and others.
+ * Copyright (c) 2016, 2018 Bosch Software Innovations GmbH and others.
  * 
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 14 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
```diff
@@ -91,7 +91,10 @@ public static void serialize(final Principal principal, final DatagramWriter wri
 
 	private static void serializeIdentity(final PreSharedKeyIdentity principal, final DatagramWriter writer) {
 		writer.writeByte(ClientAuthenticationType.PSK.code);
-		writeBytesWithLength(PSK_LENGTH_BITS, principal.getName().getBytes(StandardCharsets.UTF_8), writer);
+		byte[] virtualHost = principal.getVirtualHost() == null ? new byte[0] :
```
boaks: I would prefer the have a `static final byte[] EMPTY = {};` which could be reused.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -547,18 +549,23 @@ private void createServerHello(final ClientHello clientHello, final DTLSFlight f
...
-			indicatedServerNames = serverNameExt.getServerNames();
-			serverHelloExtensions.addExtension(ServerNameExtension.emptyServerNameIndication());
-			session.setSniSupported(true);
-			LOGGER.debug(
-					"Using server name indication received from peer [{}]",
-					clientHello.getPeer());
+			if (sniEnabled) {
+				// store the names indicated by peer for later reference during key exchange
+				indicatedServerNames = serverNameExt.getServerNames();
+				serverHelloExtensions.addExtension(ServerNameExtension.emptyServerNameIndication());
```
boaks: Maybe we can add a comment here, that according RFC6066, section 3, page 7, the server SHALL include also the sni extension and the value SHALL be empty. Otherwise the comment above "store ... later" may be misinterpreted as conflict with the "emptyServerNameIndication".

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -677,26 +684,34 @@ private void createCertificateRequest(final ClientHello clientHello, final DTLSF
...
+		byte[] psk = pskStore.getKey(getIndicatedServerNames(), identity);
+		String virtualHost = null;
 
 		if (getIndicatedServerNames() == null) {
-			psk = pskStore.getKey(identity);
+			LOGGER.debug("client [{}] uses PSK identity [{}]", getPeerAddress(), identity);
 		} else {
-			psk = pskStore.getKey(getIndicatedServerNames(), identity);
+			// SNI is enabled and the client has indicated a virtual host
+			ServerName serverName = getIndicatedServerNames().getServerName(NameType.HOST_NAME);
```
boaks: Multiple `getIndicatedServerNames()` maybe clearer, if get once and stored in a local variable.

boaks:
Maybe stuff for the next PR:

RFC6066, section 4, page 7:

> A server that implements this extension MUST NOT accept the request
> to resume the session if the server_name extension contains a
> different name. Instead, it proceeds with a full handshake to
> establish a new session. When resuming a session, the server MUST
> NOT include a server_name extension in the server hello.
> 

Test added to `DTLSConnectorTest`:

```
	@Test
	public void testConnectorPerformsFullHandshakeWhenResumingWithDifferentSni() throws Exception {
		// Do a first handshake
		RawData raw = RawData.outbound("Hello World".getBytes(), new AddressEndpointContext(serverEndpoint, "server.one", null), null, false);
		givenAnEstablishedSession(raw, true);
		byte[] sessionId = establishedServerSession.getSessionIdentifier().getId();

		// Force a resume session the next time we send data
		client.forceResumeSessionFor(serverEndpoint);
		Connection connection = clientConnectionStore.get(serverEndpoint);
		assertArrayEquals(sessionId, connection.getEstablishedSession().getSessionIdentifier().getId());
		client.start();

		// Prepare message sending
		final String msg = "Hello Again";
		CountDownLatch latch = new CountDownLatch(1);
		clientRawDataChannel.setLatch(latch);

		// send message
		RawData data = RawData.outbound(msg.getBytes(), new AddressEndpointContext(serverEndpoint, "server.two", null), null, false);
		client.send(data);
		assertTrue(latch.await(MAX_TIME_TO_WAIT_SECS, TimeUnit.SECONDS));

		// check session id was not equals
		connection = clientConnectionStore.get(serverEndpoint);
		Assert.assertThat(sessionId, not(equalTo(connection.getEstablishedSession().getSessionIdentifier().getId())));
		assertClientIdentity(RawPublicKeyIdentity.class);
	}
```

fails, so the session is resumed. So create a new issue?

sophokles73:
yes, I have also though about this but for now came to the conclusion that our current session resumption code doesn't consider it as well. Thus, I think a new issue would be the right thing to do. However, thanks for bringing it up and providing the test case :+1:

boaks:
LGTM

PR #657, opened 2018-06-07
Health logger thread must be stopped

sbernard31:
This aims to fix this issue : https://github.com/eclipse/leshan/issues/502

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
@@ -59,6 +59,7 @@ import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.coap.Token;
 import org.eclipse.californium.core.network.Exchange.KeyMID;
 import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.network.config.NetworkConfigDefaults;
 import org.eclipse.californium.core.network.deduplication.Deduplicator;
 import org.eclipse.californium.core.network.deduplication.DeduplicatorFactory;
 import org.eclipse.californium.elements.util.DaemonThreadFactory;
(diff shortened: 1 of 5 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -116,7 +116,7 @@ public InMemoryMessageExchangeStore(final NetworkConfig config, TokenGenerator t
 
 	private void startStatusLogging() {
 
-		final int healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL, 60); // seconds
+		final int healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL, 0); // seconds
```
boaks: I would prefer to add the `0` to `NetworkConfigDefaults` as field and use then that field in all places, where the default is intended.

boaks:
By the way, I'm not sure, if other failures raise, when something is `stopped`. 
Usually, I don't think, this stop/start is a too good idea. In other cases this was just done to overcame some errors, which should be fixed. I also would guess, that this only happens in (load) test scenarios. But anyway, if it cleans up, its better :-).

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
```diff
@@ -48,7 +48,7 @@
 	private volatile boolean enableStatus;
 
 	public InMemoryObservationStore(NetworkConfig config) {
-		int healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL, 60); // seconds
+		int healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL, 0); // seconds
```
boaks: And yes, adding the explicit default should have been done already earlier by me. But maybe, you can add it now :-).

boaks:
FMPOV `stop` is not used normally. And frequently just as "work around" for bugs instead of fixing them.

boaks:
LGTM

sbernard31:
Ok but in this case, how do we stop the scheduler without stop() method ?
(I ask just to know)

boaks:
In which normal case do you stop the scheduler?
If It's on a real shutdown, daemon threads will be cleaned up anyway.
If it's a stop/start, why is this intended to be done?
Or did I miss useful cases?

sbernard31:
I think the intention is about creating several clients which do something then stop. (simulator?, tests?)
If the HEALTH_STATUS_LOGGER is activated a new scheduler is created each time you create a new client and will be destroy only when the jvm will be stopped.

Issue #656, opened 2018-06-02
EC Diffie Hellman key exchange authenticated with PSK?


vikram919:
Currently 
```java
TLS_PSK_WITH_AES_128_CBC_SHA256
TLS_PSK_WITH_AES_128_CCM_8_SHA256 
```
are the two cipher suits that are available for PSK mode.
Can I know why 
```java
TLS_DHE_PSK_WITH_AES_128_CBC_SHA 
TLS_DHE_PSK_WITH_AES_128_CCM8_SHA
``` 
are not supported by californium project, is it a plan for future releases ?

sophokles73:
IIRC we focused on implementing the cipher suites that are/were mandatory for LWM2M 1.0. Since the DHE based PSK cipheres were not mandatory, we did not implement them.

However, there seems to be a need for PSK based ciphers that provide perfect forward secrecy in order to address the problem of plaintext handshake messages in DTLS 1.2. So, implementing the ciphers defined in [1] could be a valuable addition. Personally, I would be more than happy to accept a PR for it. However, it would probably require some refactoring in Scandium first, in order to more cleanly separate the DHE related code from the particular cipher(s).

[1] https://tools.ietf.org/html/rfc4279#section-3

vikram919:
@sophokles73 
> Personally, I would be more than happy to accept a PR for it.

I am glad to work on it.

> However, it would probably require some refactoring in Scandium first, in order to more cleanly separate the DHE related code from the particular cipher(s)

Thanks for the information, I will do that.

vikram919:
https://github.com/eclipse/californium/pull/658

PR #658, opened 2018-06-08
Added support for ECDHE_PSK mode

vikram919:
This pull request as per https://tools.ietf.org/html/rfc5489#section-2, adds an additional cipher suite that use a PSK to authenticate a ECDH exchange. This cipher suite protects from dictionary attacks and provides Perfect Forward Secrecy (PFS) while using pre shared keys.
https://github.com/eclipse/californium/issues/656

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -26,6 +26,7 @@
  *                                                    trustStore := null, disable x.509
  *                                                    trustStore := [], enable x.509, trust all
  *    Bosch Software Innovations GmbH - remove serverNameResolver property
+ *    Vikram (University of Rostock) - added CipherSuite TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256                        
  *******************************************************************************/
 
 package org.eclipse.californium.scandium.config;
@@ -113,10 +114,10 @@ public final class DtlsConnectorConfig {
 	/** store of the PSK */
 	private PskStore pskStore;
 
-	/** the private key for RPK and X509 mode */
+	/** the private key for RPK and X509 mode, right now only EC type is supported */
 	private PrivateKey privateKey;
 
-	/** the public key for both RPK and X.509 mode */
+	/** the public key for RPK and X.509 mode, right now only EC type is supported */
 	private PublicKey publicKey;
 
 	/** the certificate for RPK and X509 mode */
@@ -999,6 +1000,7 @@ public final class DtlsConnectorConfig {
 				switch (suite) {
 				case TLS_PSK_WITH_AES_128_CCM_8:
 				case TLS_PSK_WITH_AES_128_CBC_SHA256:
+				case TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256:
 					verifyPskBasedCipherConfig();
 					break;
 				case TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8:
@@ -1050,6 +1052,7 @@ public final class DtlsConnectorConfig {
 			if (config.pskStore != null) {
 				ciphers.add(CipherSuite.TLS_PSK_WITH_AES_128_CCM_8);
 				ciphers.add(CipherSuite.TLS_PSK_WITH_AES_128_CBC_SHA256);
+				ciphers.add(CipherSuite.TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256);
 			}
 
 			config.supportedCipherSuites = ciphers.toArray(new CipherSuite[0]);
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
@@ -37,6 +37,7 @@
  *                                                    trustStore := [], enable x.509, trust all
  *    Achim Kraus (Bosch Software Innovations GmbH) - fix usage of literal address 
  *                                                    with enabled sni
+ *    Vikram (University of Rostock) - added ECDHE_PSK mode                                                
  ******************************************************************************/
 package org.eclipse.californium.scandium.dtls;
 
@@ -51,7 +52,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
-import org.eclipse.californium.elements.auth.PreSharedKeyIdentity;
 import org.eclipse.californium.elements.auth.RawPublicKeyIdentity;
 import org.eclipse.californium.elements.auth.X509CertPath;
 import org.eclipse.californium.elements.util.StringUtil;
@@ -63,7 +63,7 @@ import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
 import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography;
 import org.eclipse.californium.scandium.dtls.pskstore.PskStore;
 import org.eclipse.californium.scandium.util.ByteArrayUtils;
-import org.eclipse.californium.scandium.util.ServerName;
+import org.eclipse.californium.scandium.util.PskUtil;
 import org.eclipse.californium.scandium.util.ServerNames;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -254,6 +254,10 @@ public class ClientHandshaker extends Handshaker {
 
 				case PSK:
 					serverKeyExchange = (PSKServerKeyExchange) handshakeMsg;
+					break;
+				
+				case ECDHE_PSK:
+					receivedServerKeyExchange((EcdhPskServerKeyExchange) handshakeMsg);
 					break;
 					
 				case NULL:
@@ -450,6 +454,31 @@ public class ClientHandshaker extends Handshaker {
 				new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, getPeerAddress()));
 		}
 	}
+	
+	/**
+	 * This method is called after receiving {@link ServerKeyExchange} message in ECDHE_PSK mode 
+	 * to extract the ServerDHEParams that includes the ephemeral public key.
+	 * 
+	 * @param message
+	 * @throws HandshakeException
+	 */
+	private void receivedServerKeyExchange(EcdhPskServerKeyExchange message) throws HandshakeException {
+		if (serverKeyExchange != null && (serverKeyExchange.getMessageSeq() == message.getMessageSeq())) {
+			// discard duplicate message
+			return;
+		}
+		serverKeyExchange = message;
+		ephemeralServerPublicKey = message.getPublicKey();
+		try {
+			ecdhe = new ECDHECryptography(ephemeralServerPublicKey.getParams());
+		} catch (GeneralSecurityException e) {
+			throw new HandshakeException(
+				String.format(
+					"Cannot create ephemeral keys from domain params provided by server: %s",
+					e.getMessage()),
+				new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, getPeerAddress()));
+		}
+	}
 
 	/**
 	 * The ServerHelloDone message is sent by the server to indicate the end of
@@ -484,62 +513,22 @@ public class ClientHandshaker extends Handshaker {
 			generateKeys(premasterSecret);
 			break;
 		case PSK:
-			String virtualHostName = session.getVirtualHost();
-			String identity = null;
-			byte[] psk = null;
-			PreSharedKeyIdentity pskIdentity = null;
-
-			if (sniEnabled && virtualHostName != null) {
-				ServerNames virtualHost = ServerNames.newInstance().add(ServerName.fromHostName(virtualHostName));
-				if (!session.isSniSupported()) {
-					LOGGER.warn("client is configured to use SNI but server does not support it, PSK authentication is likely to fail");
-				}
-				// look up identity in scope of virtual host
-				identity = pskStore.getIdentity(session.getPeer(), virtualHost);
-				if (identity == null) {
-					AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, session.getPeer());
-					throw new HandshakeException(
-							String.format(
-									"No Identity found for peer [address: %s, virtual host: %s]",
-									session.getPeer(), virtualHostName),
-							alert);
-				} else {
-					psk = pskStore.getKey(virtualHost, identity);
-					if (psk == null) {
-						AlertMessage alert = new AlertMessage(AlertLevel.FATAL,	AlertDescription.HANDSHAKE_FAILURE, session.getPeer());
-						throw new HandshakeException(
-								String.format("No pre-shared key found for [virtual host: %s, identity: %s]",
-										virtualHostName, identity),
-								alert);
-					} else {
-						pskIdentity = new PreSharedKeyIdentity(virtualHostName, identity);
-					}
-				}
-			} else {
-				identity = pskStore.getIdentity(session.getPeer());
-				if (identity == null) {
-					AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, session.getPeer());
-					throw new HandshakeException(
-							String.format("No Identity found for peer [address: %s]", session.getPeer()), alert);
-				} else {
-					psk = pskStore.getKey(identity);
-					if (psk == null) {
-						AlertMessage alert = new AlertMessage(AlertLevel.FATAL,	AlertDescription.HANDSHAKE_FAILURE, session.getPeer());
-						throw new HandshakeException(
-								String.format("No pre-shared key found for [identity: %s]", identity), alert);
-					} else {
-						pskIdentity = new PreSharedKeyIdentity(identity);
-					}
-				}
-			}
-			LOGGER.debug("Using PSK identity: {}", pskIdentity);
-			session.setPeerIdentity(pskIdentity);
-			clientKeyExchange = new PSKClientKeyExchange(identity, session.getPeer());
-			premasterSecret = generatePremasterSecretFromPSK(psk);
+			PskUtil pskUtilPlain = new PskUtil(sniEnabled, session, pskStore);
+			LOGGER.debug("Using PSK identity: {}", pskUtilPlain.getPskIdentity());
+			session.setPeerIdentity(pskUtilPlain.getPskIdentity());
+			clientKeyExchange = new PSKClientKeyExchange(pskUtilPlain.getPskIdentity().getIdentity(), session.getPeer());
+			premasterSecret = generatePremasterSecretFromPSK(pskUtilPlain.getPreSharedKey(), null);
 			generateKeys(premasterSecret);
-
 			break;
-
+		case ECDHE_PSK:
+			PskUtil pskUtil = new PskUtil(sniEnabled, session, pskStore);
+			LOGGER.debug("Using PSK identity: {}", pskUtil.getPskIdentity());
+			session.setPeerIdentity(pskUtil.getPskIdentity());
+			clientKeyExchange = new EcdhPskClientKeyExchange(pskUtil.getPskIdentity().getIdentity(), ecdhe.getPublicKey(), session.getPeer());
+			byte[] otherSecret = ecdhe.getSecret(ephemeralServerPublicKey).getEncoded();
+			premasterSecret = generatePremasterSecretFromPSK(pskUtil.getPreSharedKey(), otherSecret);
+			generateKeys(premasterSecret);
+			break;
 		case NULL:
 			clientKeyExchange = new NULLClientKeyExchange(session.getPeer());
 
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
@@ -1,0 +1,168 @@
+/*******************************************************************************
+ * Copyright 2018 University of Rostock, Institute of Applied Microelectronics and Computer Engineering
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Vikram (University of Rostock)- Initial creation, adapted from ECDHClientKeyExchange
+ ******************************************************************************/
+package org.eclipse.californium.scandium.dtls;
+
+import static org.eclipse.californium.elements.util.StandardCharsets.UTF_8;
+
+import java.net.InetSocketAddress;
+import java.security.PublicKey;
+import java.security.interfaces.ECPublicKey;
+import java.security.spec.ECParameterSpec;
+import java.security.spec.ECPoint;
+import java.util.Arrays;
+
+import org.eclipse.californium.elements.util.DatagramReader;
+import org.eclipse.californium.elements.util.DatagramWriter;
+import org.eclipse.californium.elements.util.StringUtil;
+import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography;
+import org.eclipse.californium.scandium.util.ByteArrayUtils;
+
+/**
+ * Generates client ephemeral ECDH keys for Dtls ECDH_PSK mode.
+ * 
+ * @see <a href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a> for details.
+ */
+public final class EcdhPskClientKeyExchange extends ClientKeyExchange {
+
+
+	protected static final int LENGTH_BITS = 8; // opaque point <1..2^8-1>
+	private static final int IDENTITY_LENGTH_BITS = 16; // opaque <0..2^16-1>;
+
+	/**
+	 *See <a href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a>.
+	 */
+	private final byte[] identityEncoded;
+	/**
+	 *See <a href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a>.
+	 */
+	private final String identity;
+	private final byte[] pointEncoded;
+	
+	/**
+	 * Creates a new key exchange message for an identity hint and a public key.
+	 * 
+	 * @param hint - PSK identity as clear text
+	 * @param clientPublicKey - ephemeral public key of client
+	 * @param peerAddress
+	 * @throws NullPointerException if either hint or clietPublicKey are {@code null}
+	 */
+	public EcdhPskClientKeyExchange(String hint, PublicKey clientPublicKey, InetSocketAddress peerAddress) {
+		super(peerAddress);
+		if (hint == null) {
+			throw new NullPointerException("identity cannot be null");
+		}
+		if (clientPublicKey == null) {
+			throw new NullPointerException("ephemeral public key cannot be null");
+		}
+		this.identity = hint;
+		this.identityEncoded = hint.getBytes(UTF_8);
+		ECPublicKey publicKey = (ECPublicKey) clientPublicKey;
+		ECPoint point = publicKey.getW();
+		ECParameterSpec params = publicKey.getParams();
+			
+		this.pointEncoded = ECDHECryptography.encodePoint(point, params.getCurve());
+	}
+	
+	/**
+	 * Creates a new key exchange message for an identity hint and a public key.
+	 * 
+	 * @param hintEncoded - opaque encoded PSK identity hint for server
+	 * @param pointEncoded - ephemeral public key of client (encoded point)
+	 * @param peerAddress
+	 * @throws NullPointerException if either hintEncoded or pointEncoded are {@code null}
+	 */
+	public EcdhPskClientKeyExchange(byte[] hintEncoded, byte[] pointEncoded, InetSocketAddress peerAddress) {
+		super(peerAddress);
+		if (hintEncoded ==null) {
+			throw new NullPointerException("identity cannot be null");
+		}
+		if (pointEncoded == null) {
+			throw new NullPointerException("epehemeral public key cannot be null");
+		}
+		this.identityEncoded = Arrays.copyOf(hintEncoded, hintEncoded.length);
+		this.identity = new String(this.identityEncoded,UTF_8);
+		this.pointEncoded = Arrays.copyOf(pointEncoded, pointEncoded.length);
+	}
+
+	@Override
+	public byte[] fragmentToByteArray() {
+		DatagramWriter writer = new DatagramWriter();
+		writer.write(identityEncoded.length, IDENTITY_LENGTH_BITS);
+		writer.writeBytes(identityEncoded);
+		writer.write(pointEncoded.length, LENGTH_BITS);
+		writer.writeBytes(pointEncoded);
+		return writer.toByteArray();
+	}
+	
+	/**
+	 * Deserialize byte array to key exchange message.
+	 * 
+	 * @param byteArray
+	 * @param peerAddress
+	 * @return {@code EcdhPskClientKeyExchange}
+	 * @throws NullPointerException if either byteArray or peerAddress is {@code null}
+	 */
+	public static HandshakeMessage fromByteArray(byte[] byteArray, InetSocketAddress peerAddress) {
+		if (byteArray == null) {
+			throw new NullPointerException("byte array cannot be null");
+		}
+		if (peerAddress == null) {
+			throw new NullPointerException("peer address cannot be null");
+		}
+		DatagramReader reader = new DatagramReader(byteArray);
+		int identityLength = reader.read(IDENTITY_LENGTH_BITS);
+		byte[] identityEncoded = reader.readBytes(identityLength);	
+		int length = reader.read(LENGTH_BITS);
+		byte[] pointEncoded = reader.readBytes(length);
+		return new EcdhPskClientKeyExchange(identityEncoded, pointEncoded, peerAddress);
+	}
+		
+
+	@Override
+	public int getMessageLength() {
+		return 3 + identityEncoded.length + pointEncoded.length;
+	}
+
+	/**
+	 * This method returns the ephemeral public key (encoded point) from {@link ClientKeyExchange}.
+	 * 
+	 * @return encoded point in byte array
+	 */
+	public byte[] getEncodedPoint() {
+		return Arrays.copyOf(pointEncoded, pointEncoded.length);
+	}
+
+	@Override
+	public String toString() {
+		StringBuilder sb = new StringBuilder();
+		sb.append(super.toString());
+		sb.append("\t\t Encoded identity value: ");
+		sb.append(ByteArrayUtils.toHex(identityEncoded)).append(StringUtil.lineSeparator());;
+		sb.append("\t\tEC Diffie-Hellman public value: ");		
+		sb.append(ByteArrayUtils.toHexString(pointEncoded));
+		sb.append(StringUtil.lineSeparator());
+		return sb.toString();
+	}		
+	
+	/**
+	 * This method returns the PSK identity as clear text.
+	 * 
+	 * @return psk identity
+	 */
+	public String getIdentity() {
+		return identity;
+	}
+}
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
@@ -1,0 +1,305 @@
+/*******************************************************************************
+ * Copyright 2018 University of Rostock, Institute of Applied Microelectronics and Computer Engineering
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Vikram (University of Rostock)- Initial creation, adapted from ECDHServerKeyExchange
+ ******************************************************************************/
+package org.eclipse.californium.scandium.dtls;
+
+import static org.eclipse.californium.elements.util.StandardCharsets.UTF_8;
+
+import java.net.InetSocketAddress;
+import java.security.GeneralSecurityException;
+import java.security.KeyFactory;
+import java.security.interfaces.ECPublicKey;
+import java.security.spec.ECParameterSpec;
+import java.security.spec.ECPoint;
+import java.security.spec.ECPublicKeySpec;
+import java.util.Arrays;
+
+import org.eclipse.californium.elements.util.DatagramReader;
+import org.eclipse.californium.elements.util.DatagramWriter;
+import org.eclipse.californium.elements.util.StringUtil;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertDescription;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertLevel;
+import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography;
+import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography.SupportedGroup;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Generates server ephemeral ECDH keys for Dtls PSK mode.
+ * <p>
+ * Server must send the {@code ServerKeyExchange} message even if the PSK identity hint is not provided.
+ * 
+ * @see <a href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a>
+ */
+public final class EcdhPskServerKeyExchange extends ServerKeyExchange {
+	
+	private static final Logger LOGGER = LoggerFactory.getLogger(EcdhPskServerKeyExchange.class.getCanonicalName());
+	
+	private static final int IDENTITY_HINT_LENGTH_BITS = 16;
+	private static final String MSG_UNKNOWN_CURVE_TYPE = "Unknown curve type [{}]";
+	private static final int CURVE_TYPE_BITS = 8;
+	private static final int NAMED_CURVE_BITS = 16;
+	private static final int PUBLIC_LENGTH_BITS = 8;
+	
+	/**
+	 * The algorithm name to generate elliptic curve keypairs. See also <a href=
+	 * "http://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#KeyPairGenerator"
+	 * >KeyPairGenerator Algorithms</a>.
+	 */
+	private static final String KEYPAIR_GENERATOR_INSTANCE = "EC";
+
+	/**
+	 * The PSK identity MUST be first converted to a character string, and then
+	 * encoded to octets using UTF-8. See <a
+	 * href="http://tools.ietf.org/html/rfc4279#section-5.1">RFC 4279</a>.
+	 */
+	private byte[] hintEncoded;
+	
+	/** The hint in cleartext. */
+	private String hint;
+	
+	/** The ECCurveType */
+	
+	/** parameters are conveyed verbosely; underlying finite field is a prime 
+	 * field
+	*/ 
+	private static final int EXPLICIT_PRIME = 1;
+	
+	/** parameters are conveyed verbosely; underlying finite field is a 
+	 * characteristic-2 field
+	*/ 
+	private static final int EXPLICIT_CHAR2 = 2;
+	
+	/** a named curve is used */
+	private static final int NAMED_CURVE = 3;
+
+	/** ephemeral public key */
+	private ECPublicKey publicKey = null;
+
+	private ECPoint point = null;
+	private byte[] pointEncoded = null;
+
+	private final int curveId;
+	
+	// TODO right now only named curve is supported
+	private int curveType = NAMED_CURVE;
+	
+	/**
+	 * Creates a new key exchange message with psk hint as clear text and ServerDHParams.
+	 * 
+	 * @see <a href="http://tools.ietf.org/html/rfc4279#section-3">RFC 4279</a>
+	 * @param pskHint - preshared key hint in clear text
+	 * @param ecdhe - {@code ECDHECryptography}
+	 * @param clientRandom - nonce
+	 * @param serverRandom - nonce
+	 * @param namedCurveId - ec curve used 
+	 * @param peerAddress
+	 * @throws NullPointerException if any of the arguments ecdhe, 
+	 * clientRandom and serverRandom are {@code null}
+	 */
+	public EcdhPskServerKeyExchange(String pskHint, ECDHECryptography ecdhe, Random clientRandom, Random serverRandom, 
+			int namedCurveId, InetSocketAddress peerAddress) {	
+		super(peerAddress);	
+		if (ecdhe == null) {
+			throw new NullPointerException("ECDHECryptography class object cannot be null");
+		}
+		if (clientRandom == null || serverRandom == null) {
+			throw new NullPointerException("nonce cannot be null");
+		}
+		if (pskHint != null) {
+			this.hint = pskHint;		
+		} else {
+			this.hint = "";
+		}
+		this.hintEncoded = hint.getBytes(UTF_8);
+		this.curveId = namedCurveId;
+		publicKey = ecdhe.getPublicKey();
+		ECParameterSpec parameters = publicKey.getParams();
+		point = publicKey.getW();
+		pointEncoded = ECDHECryptography.encodePoint(point, parameters.getCurve());
+	}
+	
+	/**
+	 * Creates a new key exchange message with ServerDHParams.
+	 * 
+	 * @see <a href="http://tools.ietf.org/html/rfc4279#section-3">RFC 4279</a>
+	 * @param ecdhe - {@code ECDHECryptography}
+	 * @param clientRandom - nonce
+	 * @param serverRandom - nonce
+	 * @param namedCurveId - ec curve used 
+	 * @param peerAddress
+	 * @throws NullPointerException if any of arguments ecdhe, 
+	 * clientRandom and serverRandom are {@code null}
+	 */
+	public EcdhPskServerKeyExchange(ECDHECryptography ecdhe, Random clientRandom, Random serverRandom, 
+			int namedCurveId, InetSocketAddress peerAddress) {
+		this(null, ecdhe, clientRandom, serverRandom, namedCurveId, peerAddress);
+	}
+	
+	private EcdhPskServerKeyExchange(byte[] hintEncoded, int curveId, byte[] pointEncoded, InetSocketAddress peerAddress) throws HandshakeException {		
+		super(peerAddress);
+		this.curveId = curveId;
+		this.hintEncoded = hintEncoded;
+		this.hint = new String(this.hintEncoded, UTF_8);
+		if (pointEncoded == null) {
+			throw new NullPointerException("ephemeral public key cannot be null");
+		}
+		this.pointEncoded = Arrays.copyOf(pointEncoded, pointEncoded.length);
+		// re-create public key from params
+		SupportedGroup group = SupportedGroup.fromId(curveId);
+		if (group == null || !group.isUsable()) {
+			throw new HandshakeException(
+				String.format("Server used unsupported elliptic curve (%d) for ECDH", curveId),
+				new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, peerAddress));
+		} else {
+			try {
+				point = ECDHECryptography.decodePoint(pointEncoded, group.getEcParams().getCurve());
+				KeyFactory keyFactory = KeyFactory.getInstance(KEYPAIR_GENERATOR_INSTANCE);
+				publicKey = (ECPublicKey) keyFactory.generatePublic(new ECPublicKeySpec(point, group.getEcParams()));
+			} catch (GeneralSecurityException e) {
+				LOGGER.debug("Cannot re-create server's public key from params", e);
+				throw new HandshakeException(
+					String.format("Cannot re-create server's public key from params: %s", e.getMessage()),
+					new AlertMessage(AlertLevel.FATAL, AlertDescription.INTERNAL_ERROR, peerAddress));
+			}
+		}
+	}
+
+	// TODO this is called 4 times for Flight 4
+	@Override
+	public byte[] fragmentToByteArray() {
+		DatagramWriter writer = new DatagramWriter();
+		writer.write(hintEncoded.length, IDENTITY_HINT_LENGTH_BITS);
+		writer.writeBytes(hintEncoded);
+		switch (curveType) {
+		// TODO add support for other curve types
+		case EXPLICIT_PRIME:
+		case EXPLICIT_CHAR2:
+			break;
+		case NAMED_CURVE:
+			writeNamedCurve(writer);
+			break;
+		default:
+			LOGGER.warn(MSG_UNKNOWN_CURVE_TYPE, curveType);
+			break;
+		}
+		return writer.toByteArray();
+	}
+	
+	
+	private void writeNamedCurve(DatagramWriter writer) {
+		writer.write(NAMED_CURVE, CURVE_TYPE_BITS);
+		writer.write(curveId, NAMED_CURVE_BITS);
+		writer.write(pointEncoded.length, PUBLIC_LENGTH_BITS);
+		writer.writeBytes(pointEncoded);
+	}
+	
+	/**
+	 * Deserialize byte array to key exchange message.
+	 * 
+	 * @param byteArray
+	 * @param peerAddress
+	 * @return {@code EcdhPskServerKeyExchange}
+	 * @throws HandshakeException - if the byte array includes unsupported curve
+	 * @throws NullPointerException if either byteArray or peerAddress is {@code null}
+	 */
+	public static HandshakeMessage fromByteArray(byte[] byteArray, InetSocketAddress peerAddress) throws HandshakeException {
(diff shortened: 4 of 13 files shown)


boaks:
Hi @vikram919 

I did a fast check, but got exceptions in `HandshakeMessage` ("Unknown key exchange algorithm"). I also can't see unit tests, which ensures that it is working. So, could you please check your implementation and add some unit tests?

vikram919:
@boaks 
I will add it, thanks for the feedback

vikram919:
@boaks 
Can I know how you had tested it?

boaks:
using a updated cf-secure :-).

boaks:
And don't forget to check the handshake using wireshark. 
Not that it just doesn't use a PSK/DHE variant.

vikram919:
now i have corrected it, can you please try it now? May be i have ignored the changes while i was committing.

boaks:
The exception is fixed. 
But I still think, that unit tests, which validate, that the cipher suite works, are required.

vikram919:
@boaks 
As requested a wire shark sniff to see whether the devices agree on given cipher suite:
ciphersuitessniff

sophokles73:
It looks like you have actually implemented ECDHE_PSK key exchange because it relies on ECC based keys (as generated by `ECDHECryptography`). There exists a draft RFC (https://datatracker.ietf.org/doc/draft-ietf-tls-ecdhe-psk-aead/?include_text=1) which seems to define cipher suites based on ECDHE and PSK, however it doesn't seem to be a recommendation yet.

My understanding of https://tools.ietf.org/html/rfc4279#section-3, however, is that standard Diffie-Hellman is supposed to be used.

sophokles73:
> But, TLS 1.3 define the psk_exchange_modes with (EC) DHE means either of them can be used, so I followed it. I can also implement DHE one as well?

Well, since Scandium implements DTLS 1.2, the ClientHello extension defined in TLS 1.3 has no relevance in this context. We need to focus on how it is done in DTLS 1.2. My point is actually more, that the cipher suites that you define are not matching the key exchange algorithm you have implemented. Note that this would also be problematic in DTLS 1.3. When a Scandium based server selects a DHE_PSK based cipher suite and then actually performs an ECDHE based key exchange, the handshake will fail, because the client (if implemented correctly) will expect the server's key exchange message to contain standard DH params as defined in https://tools.ietf.org/html/rfc2246#appendix-A.4.2 instead of ECDH params as defined in https://tools.ietf.org/html/rfc4492#section-5.4.

vikram919:
@sophokles73 

> Measured by computational cost to the TLS peers, ECDHE appears today
   to offer a much stronger key exchange mechanism than FFDHE.

https://tools.ietf.org/html/rfc7919#section-8.3

This pull request concentrate only adding one cipher suite TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA  as specified in https://tools.ietf.org/html/rfc5489#section-2.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -450,6 +459,30 @@ private void receivedServerKeyExchange(ECDHServerKeyExchange message) throws Han
 				new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, getPeerAddress()));
 		}
 	}
+	
+	/**
+	 * This method is called after receiving {@link ServerKeyExchange} message in DHE_PSK mode
```
sophokles73: ECDHE_PSK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -484,62 +517,22 @@ private void receivedServerHelloDone(ServerHelloDone message) throws HandshakeEx
...
-					}
-				}
-			}
+			configurePskCredentials();
 			LOGGER.debug("Using PSK identity: {}", pskIdentity);
 			session.setPeerIdentity(pskIdentity);
-			clientKeyExchange = new PSKClientKeyExchange(identity, session.getPeer());
-			premasterSecret = generatePremasterSecretFromPSK(psk);
+			clientKeyExchange = new PSKClientKeyExchange(pskIdentity.getIdentity(), session.getPeer());
+			premasterSecret = generatePremasterSecretFromPSK(psk,null);
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+import org.eclipse.californium.scandium.util.ByteArrayUtils;
+
+/**
+ * class EcdhPskClientKeyExchange. <br>
+ * This class EcdhPskClientKeyExchange represents the {@link ClientKeyExchange} in ECDHE_PSK mode. See <a
+ * href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a> for details.
+ * @author Vikram
+ *
+ */
+public final class EcdhPskClientKeyExchange extends ClientKeyExchange{
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.Arrays;
+
+import org.eclipse.californium.elements.util.DatagramReader;
+import org.eclipse.californium.elements.util.DatagramWriter;
+import org.eclipse.californium.elements.util.StringUtil;
+import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography;
+import org.eclipse.californium.scandium.util.ByteArrayUtils;
+
+/**
+ * class EcdhPskClientKeyExchange. <br>
```
sophokles73: remove this line

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+ * This class EcdhPskClientKeyExchange represents the {@link ClientKeyExchange} in ECDHE_PSK mode. See <a
```
sophokles73: you do not need to spell out the class name, JavaDoc already does this for you ...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+ * href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a> for details.
+ * @author Vikram
```
sophokles73: remove this (and all other) *author* JavaDoc tags that you added.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+	
+	/**
+	 * Constructor {@link EcdhPskClientKeyExchange}. The {@link EcdhPskClientKeyExchange} message include 
+	 * an psk identity as clear text for server to identify which psk key is used, 
+	 * and the ephemeral public key of the client.  
+	 * @param hint - PSK identity as clear text
+	 * @param clientPublicKey - ephemeral public key of client
+	 * @param peerAddress
+	 */
+	public EcdhPskClientKeyExchange(String hint, PublicKey clientPublicKey, InetSocketAddress peerAddress) {
```
sophokles73: null checks?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+	 */
+	private final byte[] identityEncoded;
+	/**
+	 *See <a href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a>.
+	 */
+	private final String identity;
+	private final byte[] pointEncoded;
+	
+	/**
+	 * Constructor {@link EcdhPskClientKeyExchange}. The {@link EcdhPskClientKeyExchange} message include
```
sophokles73: no need to repeat the constructor name, no need to link to the class

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+		this.identityEncoded = hint.getBytes(UTF_8);
+		ECPublicKey publicKey = (ECPublicKey) clientPublicKey;
+		ECPoint point = publicKey.getW();
+		ECParameterSpec params = publicKey.getParams();
+			
+		this.pointEncoded = ECDHECryptography.encodePoint(point, params.getCurve());
+	}
+	
+	/**
+	 * Constructor {@link EcdhPskClientKeyExchange}.
```
sophokles73: A simple description like
```
Creates a new key exchange message for an identity hint and a public key.
```

would suffice ...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+		this.pointEncoded = ECDHECryptography.encodePoint(point, params.getCurve());
+	}
+	
+	/**
+	 * Constructor {@link EcdhPskClientKeyExchange}. 
+	 * @param hintEncoded - opaque encoded PSK identity hint for server
+	 * @param pointEncoded - ephemeral public key of client (encoded point) 
+	 * @param peerAddress
+	 */
+	public EcdhPskClientKeyExchange(byte[] hintEncoded, byte[] pointEncoded, InetSocketAddress peerAddress) {
```
sophokles73: null checks?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+		return writer.toByteArray();
+	}
+	
+	/**
+	 * This method returns the {@link EcdhPskClientKeyExchange} message from the byte array.
+	 * @param byteArray
+	 * @param peerAddress
+	 * @return
+	 */
+	public static HandshakeMessage fromByteArray(byte[] byteArray, InetSocketAddress peerAddress) {
```
sophokles73: null checks?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+		DatagramWriter writer = new DatagramWriter();
+		writer.write(identityEncoded.length, IDENTITY_LENGTH_BITS);
+		writer.writeBytes(identityEncoded);
+		writer.write(pointEncoded.length, LENGTH_BITS);
+		writer.writeBytes(pointEncoded);
+		return writer.toByteArray();
+	}
+	
+	/**
+	 * This method returns the {@link EcdhPskClientKeyExchange} message from the byte array.
```
sophokles73: remove link

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+	 * @param byteArray
+	 * @param peerAddress
+	 * @return
```
sophokles73: you could document here that the returned message is a `{@code EcdhPskClientKeyExchange}` ...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,146 @@
...
+		return Arrays.copyOf(pointEncoded, pointEncoded.length);
+	}
+
+	@Override
+	public String toString() {
+		StringBuilder sb = new StringBuilder();
+		sb.append(super.toString());
+		sb.append("\t\t Encoded identity value: ");
+		sb.append(ByteArrayUtils.toHex(identityEncoded));
+		sb.append("\n\t\tDiffie-Hellman public value: ");
```
sophokles73: this should probably be called *EC Diffie-Hellman public key*

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,231 @@
...
+import org.eclipse.californium.elements.util.StringUtil;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertDescription;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertLevel;
+import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography;
+import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography.SupportedGroup;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * class EcdhPskServerKeyExchange.<br>
```
sophokles73: see above,
remove author tag

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,231 @@
...
+	private ECPoint point = null;
+	private byte[] pointEncoded = null;
+
+	private final int curveId;
+	
+	// TODO right now only named curve is supported
+	private int curveType = NAMED_CURVE;
+	
+	/**
+	 * Constructor {@link EcdhPskServerKeyExchange} called on server side. The {@link EcdhPskServerKeyExchange}
```
sophokles73: see above

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,231 @@
...
+	 * message structure is composed of an optional parameter opaque psk identity and ServerDHParams 
+	 * that includes the type of ec curve used and the server public key (point encoded).
+	 * @param ecdhe - {@link ECDHECryptography}
+	 * @param clientRandom - nonce
+	 * @param serverRandom - nonce
+	 * @param namedCurveId - ec curve used 
+	 * @param peerAddress
+	 * @throws GeneralSecurityException
+	 */
+	public EcdhPskServerKeyExchange(ECDHECryptography ecdhe, Random clientRandom, Random serverRandom,
```
sophokles73: null checks?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,231 @@
...
+	private final int curveId;
+	
+	// TODO right now only named curve is supported
+	private int curveType = NAMED_CURVE;
+	
+	/**
+	 * Constructor {@link EcdhPskServerKeyExchange} called on server side. The {@link EcdhPskServerKeyExchange} 
+	 * message structure is composed of an optional parameter opaque psk identity and ServerDHParams 
+	 * that includes the type of ec curve used and the server public key (point encoded).
+	 * @param ecdhe - {@link ECDHECryptography}
```
sophokles73: no need to add a link, JavaDoc automatically generates a link to the parameter's type. Please provide a description of the parameter's role/purpose ...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,231 @@
...
+	 * See <a href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a> */
+	private void writeNamedCurve(DatagramWriter writer) {
+		writer.write(NAMED_CURVE, CURVE_TYPE_BITS);
+		writer.write(curveId, NAMED_CURVE_BITS);
+		writer.write(pointEncoded.length, PUBLIC_LENGTH_BITS);
+		writer.writeBytes(pointEncoded);
+	}
+	
+	/**
+	 * This method returns the {@link EcdhPskServerKeyExchange} message from byte array.
```
sophokles73: see above

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,231 @@
...
+	 * @param byteArray
+	 * @param peerAddress
+	 * @return
+	 * @throws HandshakeException - if the byte array includes unsupported curve
+	 */
+	public static HandshakeMessage fromByteArray(byte[] byteArray, InetSocketAddress peerAddress) throws HandshakeException {
```
sophokles73: null checks?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,231 @@
...
+	 */
+	public int getCurveId() {
+		return curveId;
+	}
+
+	@Override
+	public String toString() {
+		StringBuilder sb = new StringBuilder();
+		sb.append(super.toString());
+		sb.append("\t\tDiffie-Hellman public key: ");
```
sophokles73: EC Diffie-Hellman public

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -563,26 +564,35 @@ protected void calculateKeys(byte[] masterSecret) {
 	 * concatenate a uint16 with the value N, N zero octets, a second uint16
 	 * with the value N, and the PSK itself.
 	 * 
-	 * @param psk
-	 *            the preshared key as byte array.
-	 * @return the premaster secret.
+	 * @param psk - preshared key as byte array
+	 * @param otherSecret - "other secret" either is zeros or comes from DHE_PSK.
```
sophokles73: please update the method's description (JavaDoc) to reflect the role of the *other secret* and provide a pointer/link to the corresponding RFC

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -563,26 +564,35 @@ protected void calculateKeys(byte[] masterSecret) {
...
-		byte[] zero = ByteArrayUtils.padArray(new byte[0], (byte) 0x00, length);
-
-		byte[] premasterSecret = ByteArrayUtils.concatenate(lengthField, ByteArrayUtils.concatenate(zero, ByteArrayUtils.concatenate(lengthField, psk)));
-
-		return premasterSecret;
+		byte[] pskLengthField = new byte[2];
+		pskLengthField[0] = (byte) (pskLength >> 8);
+		pskLengthField[1] = (byte) (pskLength);
+		
+		if(otherSecret == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -627,7 +633,20 @@ private void createServerKeyExchange(final ClientHello clientHello, final DTLSFl
...
 			break;
-
+		
+		case ECDHE_PSK:
+			
+			try {
+				ecdhe = new ECDHECryptography(negotiatedSupportedGroup.getEcParams());
+				serverKeyExchange = new EcdhPskServerKeyExchange(ecdhe, clientRandom, serverRandom,
+				negotiatedSupportedGroup.getId(), session.getPeer());
+			break;
```
sophokles73: indentation

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -970,4 +1000,37 @@ protected boolean isFirstMessageReceived(final HandshakeMessage handshakeMessage
...
+			throw new HandshakeException(
+					String.format("cannot authenticate client, identity [%s] is unknown for server [%s]",
+							identity, virtualHost),
+					new AlertMessage(AlertLevel.FATAL, AlertDescription.UNKNOWN_PSK_IDENTITY, session.getPeer()));
+		} else {
+			session.setPeerIdentity(new PreSharedKeyIdentity(virtualHost, identity));
+			if(otherSecret==null) {
+				return generatePremasterSecretFromPSK(psk, null);
+			}else {
+				return generatePremasterSecretFromPSK(psk, otherSecret);
```
sophokles73: what is the difference if otherSecret is `null` ?

scandium-core/src/test/java/org/eclipse/californium/scandium/config/DtlsConnectorConfigTest.java
```diff
@@ -70,7 +72,7 @@ public void testBuilderSetsPskCipherSuitesWhenPskStoreIsSet() {
 		DtlsConnectorConfig config = builder.setPskStore(new StaticPskStore("ID", "KEY".getBytes())).build();
 		assertTrue(config.getSupportedCipherSuites().length > 0);
 		for (CipherSuite suite : config.getSupportedCipherSuites()) {
-			assertThat(suite.getKeyExchange(), is(KeyExchangeAlgorithm.PSK));
+			assertThat(suite.getKeyExchange(), Matchers.either(Matchers.is(KeyExchangeAlgorithm.PSK)).or(Matchers.is(KeyExchangeAlgorithm.ECDHE_PSK)));
```
sophokles73: you should probably import the Matchers statically ...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
vikram919: Do I have to introduce Assert.assertNotNull(clientPublicKey);  for this?

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -113,10 +114,10 @@
 	/** store of the PSK */
 	private PskStore pskStore;
 
-	/** the private key for RPK and X509 mode */
+	/** the ECDSA private key for RPK and X509 mode */
```
boaks: If the intention would be to limit the scope to ECDSA, I guess it would be expressed by using the EC specific subtypes. Right now I would just mention, that only EC is supported.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -97,6 +98,10 @@
 	private final List<CipherSuite> preferredCipherSuites;
 
 	protected Integer maxFragmentLengthCode;
+	
+	private PreSharedKeyIdentity pskIdentity;
+	
+	private byte[] psk;
```
boaks: these both looks like "two return values" from `configurePskCredentials`.
Maybe you find a other solution.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
boaks: See DataParser.parseMessage

https://github.com/eclipse/californium/blob/2.0.x/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataParser.java

And please don't forget to document the throws NullPointerException.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
vikram919: do you have any ideas or pattern to return two values from one method?

boaks: You may add a "PskUtil" which has a constructor implementing the `configurePskCredentials`.
Then use that PskUtil instance only as local variable.

boaks: PskUtil should then contain the two above fields psk and pskIdentity.

scandium-core/src/main/java/org/eclipse/californium/scandium/util/PskUtil.java
```diff
@@ -0,0 +1,120 @@
...
+	 * @throws NullPointerException if either session or pskStore is {@code null}
+	 */
+	public PskUtil(DTLSSession session, PskStore pskStore) throws HandshakeException {
+		if(session == null) {
+			throw new NullPointerException("Dtls session cannot be null");
+		}
+		if(pskStore == null) {
+			throw new NullPointerException("psk store cannot be null");
+		}
+		configurePskCredentials(session, pskStore);
```
boaks: I would just delete line 60-63, I don't see the benefit of split the implementation.

sophokles73:
@vikram919 
it is very hard to keep track of changes you made to the PR in response to our comments if you force push a single commit on the PR branch. Please simply add more commits to the PR with your changes in the future, so that we can review in a reasonable amount of time. Once we are done reviewing, all commits can be squashed into an appropriate number of individual commits and then merged.

vikram919:
@sophokles73 
Sorry for that, I will keep this in mind

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.Arrays;
+
+import org.eclipse.californium.elements.util.DatagramReader;
+import org.eclipse.californium.elements.util.DatagramWriter;
+import org.eclipse.californium.elements.util.StringUtil;
+import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography;
+import org.eclipse.californium.scandium.util.ByteArrayUtils;
+
+/**
+ * This class EcdhPskClientKeyExchange represents the {@link ClientKeyExchange} in ECDHE_PSK mode. See <a
```
sophokles73: please write a proper JavaDoc comment (see http://www.oracle.com/technetwork/articles/java/index-137868.html for inspiration)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,163 @@
...
+	 * Creates a new key exchange message for an identity hint and a public key.
+	 * 
+	 * @param hint - PSK identity as clear text
+	 * @param clientPublicKey - ephemeral public key of client
+	 * @param peerAddress
+	 * @throws NullPointerException if either hint or clietPublicKey are {@code null}
+	 */
+	public EcdhPskClientKeyExchange(String hint, PublicKey clientPublicKey, InetSocketAddress peerAddress) {
+		super(peerAddress);
+		if(hint == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,163 @@
...
+			throw new NullPointerException("identity cannot be null");
+		}
+		if(clientPublicKey == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,163 @@
...
+	 * Creates a new key exchange message for an identity hint and a public key.
+	 * 
+	 * @param hintEncoded - opaque encoded PSK identity hint for server
+	 * @param pointEncoded - ephemeral public key of client (encoded point)
+	 * @param peerAddress
+	 * @throws NullPointerException if either hintEncoded or pointEncoded are {@code null}
+	 */
+	public EcdhPskClientKeyExchange(byte[] hintEncoded, byte[] pointEncoded, InetSocketAddress peerAddress) {
+		super(peerAddress);
+		if(hintEncoded ==null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,163 @@
...
+			throw new NullPointerException("identity cannot be null");
+		}
+		if(pointEncoded == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,163 @@
...
+	/**
+	 * Deserialize byte array to key exchange message.
+	 * 
+	 * @param byteArray
+	 * @param peerAddress
+	 * @return {@code EcdhPskClientKeyExchange}
+	 * @throws NullPointerException if the byte array is {@code null}
+	 */
+	public static HandshakeMessage fromByteArray(byte[] byteArray, InetSocketAddress peerAddress) {
+		if(byteArray == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,245 @@
...
+import org.eclipse.californium.elements.util.StringUtil;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertDescription;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertLevel;
+import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography;
+import org.eclipse.californium.scandium.dtls.cipher.ECDHECryptography.SupportedGroup;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class EcdhPskServerKeyExchange represents {@link ServerKeyExchange} in DHE_PSK mode. The <a
```
sophokles73: proper comment

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,245 @@
...
+	 * @param namedCurveId - ec curve used 
+	 * @param peerAddress
+	 * @throws GeneralSecurityException
+	 * @throws NullPointerException if any of arguments ecdhe, 
+	 * clientRandom, serverRandom are {@code null}
+	 */
+	public EcdhPskServerKeyExchange(ECDHECryptography ecdhe, Random clientRandom, Random serverRandom, 
+			int namedCurveId, InetSocketAddress peerAddress) throws GeneralSecurityException {	
+		this(namedCurveId, peerAddress);
+		if(ecdhe == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,245 @@
...
+			throw new NullPointerException("ECDHECryptography class object cannot be null");
+		}
+		if(clientRandom == null || serverRandom == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,245 @@
...
+		}
+		publicKey = ecdhe.getPublicKey();
+		ECParameterSpec parameters = publicKey.getParams();
+		point = publicKey.getW();
+		pointEncoded = ECDHECryptography.encodePoint(point, parameters.getCurve());
+	}
+	
+	private EcdhPskServerKeyExchange(int curveId, byte[] pointEncoded, InetSocketAddress peerAddress) throws HandshakeException {		
+		this(curveId, peerAddress);
+		if(pointEncoded == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,245 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class EcdhPskServerKeyExchange represents {@link ServerKeyExchange} in DHE_PSK mode. The <a
+ * href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a> also define that if server must send the {@link ServerKeyExchange}
+ * message even if the PSK identity hint is not provided.
+ *
+ */
+public final class EcdhPskServerKeyExchange extends ServerKeyExchange{
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,245 @@
...
+	 * Deserialize byte array to key exchange message.
+	 * 
+	 * @param byteArray
+	 * @param peerAddress
+	 * @return {@code EcdhPskServerKeyExchange}
+	 * @throws HandshakeException - if the byte array includes unsupported curve
+	 * @throws NullPointerException if byteArray is {@code null}
+	 */
+	public static HandshakeMessage fromByteArray(byte[] byteArray, InetSocketAddress peerAddress) throws HandshakeException {
+		if(byteArray == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -563,26 +564,36 @@ protected void calculateKeys(byte[] masterSecret) {
...
-
-		byte[] premasterSecret = ByteArrayUtils.concatenate(lengthField, ByteArrayUtils.concatenate(zero, ByteArrayUtils.concatenate(lengthField, psk)));
-
-		return premasterSecret;
+		int pskLength = psk.length;
+
+		byte[] pskLengthField = new byte[2];
+		pskLengthField[0] = (byte) (pskLength >> 8);
+		pskLengthField[1] = (byte) (pskLength);		
+		if(otherSecret == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -563,26 +564,36 @@ protected void calculateKeys(byte[] masterSecret) {
...
+			byte[] zero = ByteArrayUtils.padArray(new byte[0], (byte) 0x00, pskLength);
+			byte[] premasterSecret = ByteArrayUtils.concatenate(pskLengthField, ByteArrayUtils.concatenate(zero, ByteArrayUtils.concatenate(pskLengthField, psk)));
+			return premasterSecret;
+		}
+		else {
```
sophokles73: put on same line

scandium-core/src/main/java/org/eclipse/californium/scandium/util/PskUtil.java
```diff
@@ -0,0 +1,115 @@
...
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertDescription;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertLevel;
+import org.eclipse.californium.scandium.dtls.DTLSSession;
+import org.eclipse.californium.scandium.dtls.HandshakeException;
+import org.eclipse.californium.scandium.dtls.pskstore.PskStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Called from ClientHandshaker class to retrieve preshared key identity
```
sophokles73: This is not the purpose of this class (to be called from ...), is it?

scandium-core/src/main/java/org/eclipse/californium/scandium/util/PskUtil.java
```diff
@@ -0,0 +1,115 @@
...
+	
+	/**
+	 * Retrieves preshared key identity and preshared key for the given dtls session and psk store.
+	 * @param session - {@code DTLSSession}
+	 * @param pskStore - {@code PskStore}
+	 * @throws HandshakeException
+	 * @throws NullPointerException if either session or pskStore is {@code null}
+	 */
+	public PskUtil(DTLSSession session, PskStore pskStore) throws HandshakeException {
+		if(session == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/util/PskUtil.java
```diff
@@ -0,0 +1,115 @@
...
+			throw new NullPointerException("Dtls session cannot be null");
+		}
+		if(pskStore == null) {
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/util/PskUtil.java
```diff
@@ -0,0 +1,115 @@
...
+					throw new HandshakeException(
+							String.format("No pre-shared key found for [identity: %s]", identity), alert);
+				} else {
+					this.pskIdentity = new PreSharedKeyIdentity(identity);
+				}
+			}
+		}		
+	}
+	
+	/** {@code PreSharedKeyIdentity} for the dtls session*/
```
sophokles73: please add a proper JavaDoc comment

scandium-core/src/main/java/org/eclipse/californium/scandium/util/PskUtil.java
```diff
@@ -0,0 +1,115 @@
...
+	public PreSharedKeyIdentity getPskIdentity() {
+		return this.pskIdentity;
+	}
+	/** returns the preshared key for the dtls session */
```
sophokles73: please add a proper JavaDoc comment

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
sophokles73: we put a whitespace between `if` and the opening bracket

vikram919: I will correct that

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -86,10 +86,10 @@ public EcdhPskClientKeyExchange(String hint, PublicKey clientPublicKey, InetSock
 	 */
 	public EcdhPskClientKeyExchange(byte[] hintEncoded, byte[] pointEncoded, InetSocketAddress peerAddress) {
 		super(peerAddress);
-		if(hintEncoded ==null) {
+		if( hintEncoded ==null) {
```
sophokles73: whitespace needs to be inserted *before* the opening bracket ...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -86,10 +86,10 @@ public EcdhPskClientKeyExchange(String hint, PublicKey clientPublicKey, InetSock
...
 			throw new NullPointerException("identity cannot be null");
 		}
-		if(pointEncoded == null) {
+		if( pointEncoded == null) {
```
sophokles73: whitespace needs to be inserted *before* the opening bracket ...

vikram919:
In reference to https://tools.ietf.org/html/rfc5489#section-2,
There is an optional parameter opaque identity value sent by the server in ServerKeyExchange, the spec also mentions even though this identity value is absent, server must send the message.

Currently client only parse the EcdhPskServerKeyExchange message assuming that server will not send the identity value. In future while using with other coap servers (not californium sever) might cause inter operability issues if psk identity is mentioned.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,247 @@
...
+
+/**
+ * Generates server ephemeral ECDH keys for Dtls PSK mode.
+ * <p>
+ * Server must send the {@code ServerKeyExchange} message even if the PSK identity hint is not provided.
+ * 
+ * @see <a href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a>
+ *
+ */
+public final class EcdhPskServerKeyExchange extends ServerKeyExchange{
```
sophokles73: whitespace

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,247 @@
...
+	 * @param ecdhe - {@code ECDHECryptography}
+	 * @param clientRandom - nonce
+	 * @param serverRandom - nonce
+	 * @param namedCurveId - ec curve used 
+	 * @param peerAddress
+	 * @throws GeneralSecurityException
+	 * @throws NullPointerException if any of arguments ecdhe, 
+	 * clientRandom, serverRandom are {@code null}
+	 */
+	public EcdhPskServerKeyExchange(ECDHECryptography ecdhe, Random clientRandom, Random serverRandom,
```
sophokles73: should allow specification of an identity hint

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	private EcdhPskServerKeyExchange(int namedCurveId, InetSocketAddress peerAddress) {
+		super(peerAddress);
+		this.curveId = namedCurveId;
+	}
+
+	// TODO this is called 4 times for Flight 4
+	@Override
+	public byte[] fragmentToByteArray() {
```
sophokles73: the byte array produced by this method does not comply with https://tools.ietf.org/html/rfc5489#section-2 FMPOV since it does not contain the (potentially empty) identity hint. Even if there is no identity hint, you need to at least write the length of the hint (i.e. 0 bytes) to the array. Compare with PSKServerKeyExchange

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,247 @@
...
+	 * @return {@code EcdhPskServerKeyExchange}
+	 * @throws HandshakeException - if the byte array includes unsupported curve
+	 * @throws NullPointerException if byteArray is {@code null}
+	 */
+	public static HandshakeMessage fromByteArray(byte[] byteArray, InetSocketAddress peerAddress) throws HandshakeException {
+		if (byteArray == null) {
+			throw new NullPointerException("byte array cannot be null");
+		}
+		DatagramReader reader = new DatagramReader(byteArray);
+		int curveType = reader.read(CURVE_TYPE_BITS);
```
sophokles73: IMHO this will fail because the byte array will contain two bytes indicating the identity hint length at this position ...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,247 @@
...
+		case NAMED_CURVE:
+			length = 4 + pointEncoded.length;
+			break;
+		default:
+			LOGGER.warn(MSG_UNKNOWN_CURVE_TYPE, curveType);
+			break;
+		}		
+		return length;
+	}
+	/**
```
sophokles73: insert empty line

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
boaks: Before the `{`, :-)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
vikram919: added identity hint in the constructor

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
vikram919: implemented

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -97,23 +110,50 @@
 	 * @throws NullPointerException if any of arguments ecdhe, 
 	 * clientRandom, serverRandom are {@code null}
 	 */
-	public EcdhPskServerKeyExchange(ECDHECryptography ecdhe, Random clientRandom, Random serverRandom, 
+	public EcdhPskServerKeyExchange(String hint, ECDHECryptography ecdhe, Random clientRandom, Random serverRandom, 
 			int namedCurveId, InetSocketAddress peerAddress) throws GeneralSecurityException {	
-		this(namedCurveId, peerAddress);
+		super(peerAddress);	
+		if (hint == null) {
```
sophokles73: For convenience  we should also accept `null` here, don't you think? BTW you have not included the *hint* param in the list of params that would cause a NullPointerException anyway ;-)

vikram919: yes you are correct !, for convenience null is better, thank you for the suggestion

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -110,19 +111,20 @@
 	 * @throws NullPointerException if any of arguments ecdhe, 
 	 * clientRandom, serverRandom are {@code null}
 	 */
-	public EcdhPskServerKeyExchange(String hint, ECDHECryptography ecdhe, Random clientRandom, Random serverRandom, 
+	public EcdhPskServerKeyExchange(String pskHint, ECDHECryptography ecdhe, Random clientRandom, Random serverRandom,
```
sophokles73: when you rename the parameter, don't forget to adapt the JavaDoc accordingly ...
In this context: most of the parameters lack a proper description i the JavaDoc comment and it is unclear, when to expect a `GeneralSecurityException` to be thrown. For a user, all of this is vital information in order to be able to use this class properly, don't you think?

vikram919: done

sophokles73:
@vikram919 

can you rebase and squash?

vikram919:
done

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,307 @@
...
+	 * @param byteArray
+	 * @param peerAddress
+	 * @return {@code EcdhPskServerKeyExchange}
+	 * @throws HandshakeException - if the byte array includes unsupported curve
+	 * @throws NullPointerException if byteArray is {@code null}
+	 */
+	public static HandshakeMessage fromByteArray(byte[] byteArray, InetSocketAddress peerAddress) throws HandshakeException {
+		if (byteArray == null) {
+			throw new NullPointerException("byte array cannot be null");
+		}
```
boaks: I'm not sure, why only one of two parameter is checked for null.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskClientKeyExchange.java
```diff
@@ -0,0 +1,163 @@
...
+		return Arrays.copyOf(pointEncoded, pointEncoded.length);
+	}
+
+	@Override
+	public String toString() {
+		StringBuilder sb = new StringBuilder();
+		sb.append(super.toString());
+		sb.append("\t\t Encoded identity value: ");
+		sb.append(ByteArrayUtils.toHex(identityEncoded));
+		sb.append("\n\t\tEC Diffie-Hellman public value: ");
```
boaks: Please replace the "\n" by `StringUtil.lineSeparator()`.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
vikram919: PR updated

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -563,26 +564,35 @@ protected void calculateKeys(byte[] masterSecret) {
...
-		byte[] premasterSecret = ByteArrayUtils.concatenate(lengthField, ByteArrayUtils.concatenate(zero, ByteArrayUtils.concatenate(lengthField, psk)));
-
-		return premasterSecret;
+		int pskLength = psk.length;
+
+		byte[] pskLengthField = new byte[2];
+		pskLengthField[0] = (byte) (pskLength >> 8);
+		pskLengthField[1] = (byte) (pskLength);		
+		if (otherSecret == null) {
+			byte[] zero = ByteArrayUtils.padArray(new byte[0], (byte) 0x00, pskLength);
```
boaks: the DatagramWriter (setup up with the proper length) will do a better job.

boaks:
@vikram919 

Please update your local 2.0.x branch before your rebase on it. 
There may be a conflict caused by the commit c279676025683ec4c25547251d6b5f13e7ff2419 but the comment "Replace getLocalHost() by getLoopbackAddress()." should guide you to solve it.

boaks:
@vikram919 

Please update your local 2.0.x branch before your rebase on it. 
There may be a conflict caused by the commit c279676025683ec4c25547251d6b5f13e7ff2419 but the comment "Replace getLocalHost() by getLoopbackAddress()." should guide you to solve it.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
vikram919: I added DatagramWriter instead of ByteArrayUtils.concatenate(), is this what you mean?

boaks: I mean, that the implementation creates a lot of "short living array objects". Though the handshake is not too frequently executed, it doesn't really harm. But sometime, such "examples" are then copied into a much more frequently use scenario and then it harms.

```byte[] zero = ByteArrayUtils.padArray(new byte[0], (byte) 0x00, pskLength);```

That code first create a empty byte[] array (length 0), 
Then it creates a byte[] array of length pskLength, initialized with 0s (that's how a new array is defined in java).
Then, it fills in the 0s into the already initialized array. 
That's exactly the "bad examples", I don't like. But yes it works and doesn't harm here, because it's rarely executed ...

**Comment by boaks** [comment]
It's hard for me to spend enough time into this PR. 

I have still two doubts:
1. `sniEnabled` smells to be broken. Please analyse the **complete** usage before this PR, and then prove, that your changes don't break that original function. Maybe @sophokles73 can help you, he introduced `sniEnabled` 1 commit ago.

2. I hope the hint and identity is processed proper. But currently I will neither be able to spend my time into getting familiar enough with the spec. nor do I have the time to check, if it works with other implementations. For that, I would just merge it and see, if others do that compatibility check :-).

File: scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java (hunk 60 above)
Comment (vikram919) [reply scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java]: yes you are right,
```java
byte[] zero = new byte[pskLength];
Arrays.fill(zero, (byte)0x00);
```
This is better than this. Am I correct?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -558,31 +561,52 @@ protected void calculateKeys(byte[] masterSecret) {
...
-		byte[] premasterSecret = ByteArrayUtils.concatenate(lengthField, ByteArrayUtils.concatenate(zero, ByteArrayUtils.concatenate(lengthField, psk)));
-
-		return premasterSecret;
+		int pskLength = psk.length;
+
+		byte[] pskLengthField = new byte[2];
+		pskLengthField[0] = (byte) (pskLength >> 8);
+		pskLengthField[1] = (byte) (pskLength);		
+		DatagramWriter writer = new DatagramWriter();
+		if (otherSecret == null) {
```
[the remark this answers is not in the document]
vikram919: I changed it!

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -970,4 +1000,33 @@ protected boolean isFirstMessageReceived(final HandshakeMessage handshakeMessage
 //		}
 //		return result;
 //	}
+	
+	private byte[] configurePskCredentials(String identity, byte[] psk, byte[] otherSecret) throws HandshakeException {
```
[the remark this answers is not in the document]
vikram919: Excuse me for not understanding this point, can you explain how can we use this in standard PSK key exchange?

scandium-core/src/main/java/org/eclipse/californium/scandium/util/PskUtil.java
```diff
@@ -0,0 +1,124 @@
...
+	private static final Logger LOGGER = LoggerFactory.getLogger(PskUtil.class.getName());
+	
+	private byte[] psk;
+	
+	private PreSharedKeyIdentity pskIdentity;
+	
+	/**
+	 * Support Server Name Indication TLS extension.
+	 */
+	protected boolean sniEnabled = true;
```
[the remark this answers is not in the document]
vikram919: I can introduce a boolean in the constructor to disable it?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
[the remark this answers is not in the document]
sophokles73: well, *receivedClientKeyExchange(final PSKClientKeyExchange message)* basically contains all of the code in this method as well. FMPOV *receivedClientKeyExchange* could therefore simply invoke *configurePskCredentials* with the identity, the PSK and `null` for the otherSecret param, couldn't it? Very much like *receivedClientKeyExchange(final EcdhPskClientKeyExchange message)* already does ...

scandium-core/src/main/java/org/eclipse/californium/scandium/util/PskUtil.java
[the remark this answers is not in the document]
sophokles73: In the original code this is set from the DtlsConnectorConfig instance. IMHO it should still be possible to configure it using this mechanism.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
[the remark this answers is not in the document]
vikram919: got it!

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/EcdhPskServerKeyExchange.java
```diff
@@ -0,0 +1,310 @@
...
+	}
+	
+	/** 
+	 * ServerKeyExchange message is always sent even if psk identity hint is not present.
+	 * 
+	 * @see <a href="https://tools.ietf.org/html/rfc5489#section-2">RFC 5489</a> 
+	 * */
+	private void writeNamedCurve(DatagramWriter writer) {
+		/*when psk hint is not present then hintEncoded length is 0 */
+		writer.write(hintEncoded.length, IDENTITY_HINT_LENGTH_BITS);
```
sophokles73: writing the identity hint in this method is very confusing (and potentially wrong) because it should be done regardless of the type of curve, shouldn't it? IMHO it should be moved up into *fragmentToByteArray* or into its own method ...

scandium-core/src/main/java/org/eclipse/californium/scandium/util/PskUtil.java
```diff
@@ -36,29 +36,27 @@
...
 	
-	/**
-	 * Support Server Name Indication TLS extension.
-	 */
-	protected boolean sniEnabled = true;
-	
 	/**
 	 * Retrieves preshared key identity and preshared key for the given dtls session and psk store.
 	 * 
+	 * @param isSniEnabled - true if current {@code DTLSSession} supports
```
sophokles73: this is not true. this flag is explicitly set by the user and indicates whether the DtlsConnector supports SNI *at all*.

sophokles73:
LGTM
@boaks WDYT?

boaks:
@vikram919 
Please, pull the current 2.0.x, squash and rebase your PR with that current 2.0.x.

We will then merge it and afterwards we will see, 
if we get some comments about the interoperability with other implementations :-).

boaks:
In the meantime I check the interoperability with openssl. 
It works, client and server. Great! Thanks a lot!

PR #662, opened 2018-06-13
Introduce coap-stack-factory.

boaks:
Mainly intended to support oscore.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -52,6 +52,7 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - workaround for open jdk URI bug
  *    Achim Kraus (Bosch Software Innovations GmbH) - add striped execution
  *                                                    based on exchange
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add coap-stack-factory
  ******************************************************************************/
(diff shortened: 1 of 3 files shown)


sbernard31:
Introduce the new interface `CoapStackFactory` sounds good. :+1: 

But I'm not sure to understand the benefits of the static default stuff . Why not just create a default implementation and call its constructor in the coapEndpoint.build() method.

If this is about saving few byte by avoiding to instantiate 1 factory by endpoint, I would just go for a static final attribute in coapEndpointBuilder statically initialized. (I mean without the set/get lazy loading)

Anyway, those remarks are just details so LGTM

boaks:
The main intention is, that the `CoapStackFactory` could also be replaced generally. 
And not to run into "race-conditions", I want to "lock" that decision. So, if the general factory is setup, the factory can only changed by providing an other factory to the builder each time creating a new coap endpoint with a different factory. So this provides flexibility but makes it also easy for OSCORE, to replace the factory only once.

sbernard31:
> And not to run into "race-conditions", I want to "lock" that decision. 

if you choose to keep it, this sounds reasonable.

> So this provides flexibility but makes it also easy for OSCORE, to replace the factory only once.

I didn't look at OSCORE, so I don't know if it helps. From my experience generally when you create a custom factory its easy to keep it to hand to reuse it, that's why I thought that all the default stuff was not so useful.

And maybe I'm a bit septic because of all the troubles I had with NetworkConfig.getStandard :grin:

PR #674, opened 2018-06-21
Replace getLocalHost() by getLoopbackAddress().

boaks:
In rare case the test may fail cause by DNS issues, which are not part of the test.


boaks:
I guess, it's simply forgotten to be replaced.

PR #675, opened 2018-06-22
Fix leak in InMemoryMessageExchangeStore

sbernard31:
see https://github.com/eclipse/californium/issues/429#issuecomment-399443684 for more details


californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -193,6 +193,10 @@ public void runStriped() {
 				}
 			}
 
+			@Override
+			public void onSendError(Throwable error) {
+				exchange.executeComplete();
+			}
```
boaks: I would prefer to move it into the `ExchangeCleanupLayer`. (I changed/edited my comment therefore in the issue):

californium-core/src/test/java/org/eclipse/californium/TestTools.java
```diff
@@ -128,4 +130,31 @@ public static boolean waitForCondition(long timeout, long interval, TimeUnit uni
 		}
 		return check.isFulFilled();
 	}
+	
+	/**
+	 * Find a free IP port
+	 */
+	public static int findFreePort() {
```
boaks: Can you please explain, why that kind of "findFreePort" should be introduced again?
In my experience, this was one of the causes for "rare false reds".

californium-integration-tests/src/test/java/org/eclipse/californium/integration/test/SecureTest.java
```diff
@@ -0,0 +1,118 @@
...
+	 * Ensure there is no leak when we try to send a request to an absent peer
+	 */
+	@Test
+	public void testSecureGetHandshakeTimeout() throws Exception {
+		// Create an endpoint
+		createEndpoint();
+
+		// Send a request to an absent peer
+		CoapClient client = new CoapClient("coaps", InetAddress.getLoopbackAddress().getHostAddress(),
+				TestTools.findFreePort());
```
boaks: Instead of introducing "findFreePort" (and additionally using "setReuseAddress(true)" there), I would prefer tot open the DatagramSocket(0) in the test start, use it's assigned port for the test, and shutdown the socket in the test shutdown. And please, no "setReuseAddress(true)".

californium-integration-tests/src/test/java/org/eclipse/californium/integration/test/SecureTest.java
```diff
@@ -0,0 +1,118 @@
...
+		CountingHandler handler = new CountingHandler();
+		client.get(handler);
+
+		// wait until error or response
+		handler.waitForLoadCalls(1, 1000, TimeUnit.MILLISECONDS);
```
boaks: Maybe you spend/add a `waitForErrorCalls` ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
boaks: Cleanup on the "ReliabilityLayer" will only work for CON, but NON request may still leak. 
Even if leshan usually doesn't use such NON requests, I would prefer to add the cleanup to the "ExchangeCleanupLayer" extending/add that with a cleanup for "CON responses". 
And I feel, that checking to use the "failed()" callback (instead of "onSendError") )is also a good idea.

californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
```diff
@@ -186,16 +188,23 @@ public boolean isEmpty() {
...
-					exchangeStore, new UdpEndpointContextMatcher(checkAddress), COAP_STACK_TEST_FACTORY);
+				EndpointContextMatcher matcher) {
+			super(connector, applyConfiguration, config, new RandomTokenGenerator(config), observationStore,
+					exchangeStore, matcher, COAP_STACK_TEST_FACTORY);
 			this.exchangeStore = exchangeStore;
 			this.observationStore = observationStore;
 			this.requestChecker = new RequestEventChecker();
 		}
 
+		private CoapTestEndpoint(InetSocketAddress bind, NetworkConfig config,
```
boaks: Maybe you add the stuff to the constructor, which calls this.

californium-core/src/test/java/org/eclipse/californium/TestTools.java
sbernard31: I just want a free port to be "sure" my hanshake will timeout.

sbernard31: I will to that. This is probably even better as this ensure that nothing will answer on this port for the duration of the test.

californium-integration-tests/src/test/java/org/eclipse/californium/integration/test/SecureTest.java
sbernard31: What is the problem with setReuseAddress(true) ? 
It seems you spend bad time because of it :)

californium-integration-tests/src/test/java/org/eclipse/californium/integration/test/SecureTest.java
sbernard31: I can do that.

californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
sbernard31: Sry I didn't get the point.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
sbernard31: The real question : "Where to add the exchange.executeComplete?"

About NON, CON and ReliabilityLayer, I agree.

About ExchanCleanupLayer, I suspect there is a potential problem because request send by blockwiseLayer will no go through this layer. :(

A good practice :  The class which add the element in the store should be responsible to remove it.
So I would add a CleanerMessageObserver before to add in the store. This way we be sure that all element added has a cleaner attached. WDYT ?

boaks: AFAIK, the `MessageObserver` are copied from the outer requests to the single blocks requests. So I think, it should work with `ExchangeCleanupLayer`.
And Exchanges for notifies are not stored in the `InMemoryMessageExchangeStore`. 
But in the end, I guess without testing the blockwise & notifies, we will not be sure, if the leak is fixed, regardless which implementation is chosen.

californium-integration-tests/src/test/java/org/eclipse/californium/integration/test/SecureTest.java
boaks: In rare case on unix, the client and server have been operating at the same port, and then failed with extremely irritating logs. And in other OSs it may even not be effective at all. So it adds a lot of undetermined behaviour. Yes, I guess it will work 100 times without an error, but as more as you repeat the test, you will see the reds :-).

californium-core/src/test/java/org/eclipse/californium/core/test/MessageExchangeStoreTool.java
boaks: This is an additional private constructor, which is only called by one public constructor. So if the
`new UDPConnector(bind), true, ... new UdpEndpointContextMatcher(checkAddress)` is moved into the public, the private must not be added.

sbernard31: Ok I can do that.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
sbernard31: > But in the end, I guess without testing the blockwise & notifies, we will not be sure, if the leak is fixed, regardless which implementation is chosen.

Do you propose to add tests for blocks & notifiies, I don't know if this will be easy to test. This probably means lockstep tests with DTLS.

You didn't really give arguments again just adding cleaner observer in matcher just before to add in the `ExchangeStore`. Why does it sound to you a bad idea ?

It's seems to me clearer to put closer the add/clean codes and we are sure that all requests will have its cleanerObserver.

boaks: > Do you propose to add tests for blocks & notifiies, 

Yes.

> I don't know if this will be easy to test. This probably means lockstep tests with DTLS.

Though the issue is based on the `onError()` or `onSendError()` I would just call them for additional tests. With the secure test you already provided you proofed already your analysis.

Alternatively it may be possible to use the extplugtest BenchmarkClient with the reverse-observe together with DTLS and the NAT to see, if the implementation is robust enough for that scenario.
 
>You didn't really give arguments again just adding cleaner observer in matcher just before to add in the ExchangeStore. Why does it sound to you a bad idea ?

Yes, I didn't provide arguments, because FMPOV your premise with blockwise isn't true. So why change something, if I didn't expect the issue you see? And if you add the cleanup close to the exchange, but the trigger (cancel) is for/from a outer request, then you must forward that trigger. So, this would change some more. 
And so I think, the most important thing is to provide some tests then we see, if it works with the ExchangeCleanupLayer or if a redesign is required, as you propose.

sbernard31: > So why change something, if I didn't expect the issue you see?

Because you could find that the code will be better structured and less error prone like this for the future.

>And so I think, the most important thing is to provide some tests then we see, if it works with the ExchangeCleanupLayer or if a redesign is required, as you propose.

I will try to add more tests.

boaks:
Just to mention, if you prefer to merge this PR and check, if it works "better" for you, I have no objection. Just don't close the issue :-). I guess with the stuff discussed above, it otherwise would take too much time for now.

sbernard31:
I added some new tests.

cede7b1 is because error on block request was not catched. (If this modification make sense I could add log in `catch`)

I tried to move the `exchange.executeComplete` in `ExchangeCleanupLayer` but `testObserveFailureBeforeToSendDuringBlockNotification` failed.

boaks:
A lot of different thing to comment.
You replaced the explicit "lower()" and "upper()" by "super". I'm not really sure, which variant fit's more. FMPOV, especially in the blockwise layer, it may be easier to understand, if lower() and upper() is explicitly used.

boaks:
Generally catch `RuntimeException` and report it as "onSendError" is a interesting idea.
But without investing time it's hard for me to understand the consequences. And in my experience californium seems not to be affected too much by such `RuntimeException`, or do you see them in your logs or does someone else see it in the logs?

boaks:
And yes, your right, currently processing a blockwise notification will use a new request exchange, which is not handled by the `ExchangeCleanupLayer`. I think, there is also an other issue, which caused by that layering: if a new none blockwise notify was received, a deprecated blockwise notify will be downloaded before it the gets ignored. At least I'm not sure, if the filter in blockwise sees the previous newer none blockwise notify. So, also for that case, it seems, that some changes will improve the functionality.
About couple that cleanup closer to the "add", I'm not sure about the consequences, but sure, one consequence will be, that it is decoupled from the stack layering.

sbernard31:
> You replaced the explicit "lower()" and "upper()" by "super". I'm not really sure, which variant fit's more. FMPOV, especially in the blockwise layer, it may be easier to understand, if lower() and upper() is explicitly used.

I totally agree that using lower() and upper() is clearer.
I do that to "factorize" the try/catch in AbstractLayer if you have a better idea, I take ;).
I introduce all of this to be able to catch and report error in the stack to the end users. (I detect this with the tests I write)

> Generally catch RuntimeException and report it as "onSendError" is a interesting idea.

This is needed for my test. I could try to change my test but I also find this could be a valuable feature.

>And in my experience californium seems not to be affected too much by such RuntimeException

Without my change, the Exception will probably be caught in `StripedExchangeJob` but without exposing it to end user (just an error log)

> I think, there is also an other issue, which caused by that layering: if a new none blockwise notify was received, a deprecated blockwise notify will be downloaded before it the gets ignored. At least I'm not sure, if the filter in blockwise sees the previous newer none blockwise notify. So, also for that case, it seems, that some changes will improve the functionality.

I 'm not sure to understand. Do you think that moving cleanup closer to "add" will have others benefits ?

> About couple that cleanup closer to the "add", I'm not sure about the consequences, but sure, one consequence will be, that it is decoupled from the stack layering.

Should I propose a solution like this ?

boaks:
> I 'm not sure to understand. Do you think that moving cleanup closer to "add" will have others benefits ?
No. But the order of the layers has more pitfalls than only the cleanup.

boaks:
>     Generally catch RuntimeException and report it as "onSendError" is a interesting idea.
> 
> This is needed for my test. I could try to change my test but I also find this could be a valuable feature.
> 
>     And in my experience californium seems not to be affected too much by such RuntimeException
> 
> Without my change, the Exception will probably be caught in StripedExchangeJob but without exposing it to end user (just an error log)

My point is, that if these Exceptions do not occur (too often) and the catches are only introduced for testing, then I'm not sure, if this is a good idea. You may call the `Message.setSendError()` just before you throw the exception in your `ErrorInjector` instead.

boaks:
> Should I propose a solution like this ?

Only, if you can spend lot of time in testing :-).   
I pushed an additional commit, which adds the cleanup to the blockwise notifies. 
(branch "exchange_leak_bw")
FMPOV, it may not fix all, but should fix the aware issue.

boaks:
I pushed a modified `ErrorInjector` to the exchange_leak_bw. It calls `Message.setSendError()` direct and don't need the catchs.

sbernard31:
Ok here is a second try, this time trying to move cleaning stuff in the `Matcher`, closer to the "adding exchange" code.

I reused your `ErrorInjector` to limit this PR to fixing onSendError leak, but I continue to think that catching/reporting errors in the stack could be a good improvement and we should keep in mind tht this kind of error/exception raised in the stack could currently bring to memory leaks too.

californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
```diff
@@ -290,6 +290,34 @@ public void cancelObserve(Token token) {
...
+
+			@Override
+			public void failed() {
+				complete("failed");
+			}
+
+			private void complete(final String action) {
+				if (exchange.executeComplete()) {
+					if (exchange.isOfLocalOrigin()) {
+						LOG.debug("{}, {} request [MID={}, {}]", exchange, exchange.getRequest().getMID(),
```
boaks: Uups, I forgot to add the `action` as first parameter to the log message in my commit 28e20c025f2328be491b25c2d777192106c97533.

sbernard31: Uups, I didn't read this part of the code, too many trust on you ;-)

Did I change for something like : 

```java
LOG.debug("{}, {} request [MID={}, Token={}]", exchange, action, exchange.getRequest().getMID(),
								exchange.getRequest().getToken());
```

Why should we check isOfLocalOrigin ? Why sometime we are using response MID/token and sometime request ones ?

boaks: The local origin is for requests, the remote origin for CON responses (at least in my PR).

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -151,6 +153,7 @@ public void sendResponse(final Exchange exchange) {
 				// by a peer that wants to cancel the observation
 				// these NON notifications will later be removed from the
 				// exchange store when Exchange.setComplete() is called
+				addCleaner(exchange, response);
```
boaks: If this NON notification fails to be sent, the server-side observe will be removed.

boaks:
After analysing some more, I think, that for "notifications" it seems to be easier to not complete them and leave the cleanup to the `Exchange.removeNotifications()`. With that, I'm not sure, if moving the cleanup to the add really pays-off. because it requires also "special cases" there.

boaks:
By the way, I recognized, that cancel a observation by rejecting a previous notify is broken :-).
PR #686 may fix it.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
sbernard31: I will add a test for this.

sbernard31: I added the test without changing anything in my code and it seems to pass.
I will try to better understand this tomorrow.

sbernard31: I rethink about that :
- the comment explain that the exchange is added only to be able to handle a RST to this notification. So if we don't succeed to send the notification there is no need to store an exchange to be able to handle a RST message.
- If my test  is correct,  it seems cleaning the exchange will not remove the observe relation.

So I think there is no issue here.

boaks: :-).
Analysing it, only the exchange is completed (removing all old NON notifies), but the observe relation is currently not canceled on exchange completion.

Though some "resend responses scenarios" required a `Exchange.retransmitResponse()`, the new change is send and could then be rejected.

I changed to reject a previous notify, and then it fails.
(I also cleaned-up the test  little, see my PR #683, commit https://github.com/eclipse/californium/pull/683/commits/c29f1d0226061fc3b3ac4e8139fc072cbb945fb5)

boaks: Just let me add:
Though the time  scope of exchanges and the effect of complete seems to have changed a lot over the time, I'm not sure, if the fact, that complete doesn't cancel the observe, is an bug or intended. AFAIK, the coap-server-side seems to keep the exchange incomplete by intention (see UdpMatcher.sendResponse()) and so I'm not sure, if it is a good idea to complete it on failure.

sbernard31: > Though the time scope of exchanges and the effect of complete seems to have changed a lot over the time

This is not so clear to me but I believe before  there was the `complete()` and `completeCurrentRequest()` which meant 2 different things : complete the global/parent exchange or the child/inner one. In the current case I think we want to complete only the current request but I think we can do that anymore ? (completeCurrentRequest was removed right?)

Finally we always face the same issue about the exchange concept which hold to many concept ... This is an exchange at Message Coap sense, but also something which hold observe relation, but also transparent blockwise stuff. :'(
I really hope we can change this one day together.

> I changed to reject a previous notify, and then it fails.

That's right. So I can remove the `addCleaner` for notification (even for CON) but I suspect that there is a lot of exchanges which will not be clean as soon as it can be... but I suppose we could investigate this in another PR ...

californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
```diff
@@ -31,9 +32,25 @@
...
 		super(bufferSize, contentFormat);
+		if (request != null) {
+			this.request = request;
+			// canceling parent request must cancel block request.
+			request.addMessageObserver(new MessageObserverAdapter() {
+
+				@Override
+				public void onCancel() {
+					synchronized (Block1BlockwiseStatus.this) {
+						if (currentblockRequest != null && !currentblockRequest.isCanceled()) {
```
boaks: The check "!currentblockRequest.isCanceled()" seems to be required to prevent from stack overflow. 
The "synchronized" scope of the "currentblockRequest.setCanceled(true)", which then potentially calls back different observers, smells for future trouble.
Accessing the "currentblockRequest" seems to be also a source "race condition"; what happens if the application thread calls "cancel", while the stack prepares a new block request?
I'm not sure, if the implementation before will not also have such issues, but this snippet really smells for problems.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
boaks: As mentioned in https://github.com/eclipse/californium/pull/675#issuecomment-402064904 with the "corner cases", I'm not sure, if moving the "clean-up" to the "add" pays-off.

It's not only the current implementation of the Exchange and complete.
That may be changed to reflect this awareness (e.g. call the `removeHandler` for the failed message). But make it a little more complex, what should happen, if sending a retransmission fails?
If you "complete" but afterwards receive a matching response?
E.g. the cloud server (usually the DTLS server) sends a request, by retransmission it reaches the "auto resumption" timeout and start a handshake, which fails. But the device received the request, reached also the "auto resumption" and then successfully answers the request, the server will not match that response. OK there may be many, many pitfalls, so I'm not sure, if this is realistic.
So FMPOV, just fixing the "aware leaks", as PR #683 ,has also it's pros.

sbernard31: I think the issue you describe with retransmission will happen with both PR (this one and yours).

I don't even know what could be the behavior expected ? :tired_face: 
For example, we could also imagine that if a request failed to be send we don't want to stop the retransmission ... in that case "failedToSend" should not complete the exchange but be handle like a "sent" and so start timer for a retranmission ... but if there is an handhake with DTLS retransmission timeout, CoAP retransmission timing would not be not exactly what it is expected, They would be delayed by the DTLS retransmission timeout.

boaks: > I think the issue you describe with retransmission will happen with both PR (this one and yours).

Sure, both don't handle such advanced and theoretical cases. 
What I want to express is, that the "move" doesn't really  pay off, because the rules for handling failures are not that simple.

sbernard31: I get it but the californium code base is already very complex, and I still have the impression that we constantly add new corner case without succeed to make the code simpler. :worried: 

Anyway, if you think #683  is better, go for it.

boaks:
#683 is squashed, a new PR (#687) from bsinno is created, and I already merged it.

sbernard31:
Thx @boaks

PR #650, opened 2018-05-30
Basic OSCORE

T0bbe:
This code provides the basic OSCORE functionality, which means that the support for Observe and Blockwise are not included in this PR.

See draft OSCORE [1] for additional details.

[1] https://github.com/core-wg/oscoap

PR #679, opened 2018-06-26
Basic OSCORE, post review

T0bbe:
#650 Moving the pull request to a new one because of the amount of comments.

This code provides the basic OSCORE functionality, which means that the support for Observe and Blockwise are not included in this PR.

See draft OSCORE [1] for additional details.

[1] https://github.com/core-wg/oscoap

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
@@ -169,7 +169,7 @@ public final class OptionSet {
 		
 		observe = origin.observe;
 		if(origin.oscore != null)
-			oscore	= origin.oscore.clone();
+				oscore	= origin.oscore.clone();
 		
 		others              = copyList(origin.others);
 	}
@@ -1508,5 +1508,4 @@ public final class OptionSet {
 			builder.setLength(builder.length() - 1);
 		}
 	}
-
 }
--- a/oscore-cf/src/main/java/org/eclipse/californium/cose/EncryptCommon.java
+++ b/oscore-cf/src/main/java/org/eclipse/californium/cose/EncryptCommon.java
@@ -1,0 +1,199 @@
+/*******************************************************************************
+
+ * Copyright (c) 2016, Jim Schaad
+ * Copyright (c) 2018, Tobias Andersson, RISE SICS
+ * All rights reserved.
+
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+
+ * Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+
+ * Neither the name of COSE-JAVA nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+     
+ ******************************************************************************/
+package org.eclipse.californium.cose;
+
+import com.upokecenter.cbor.CBORObject;
+import com.upokecenter.cbor.CBORType;
+
+import COSE.AlgorithmID;
+import COSE.Attribute;
+import COSE.CoseException;
+import COSE.HeaderKeys;
+import COSE.Message;
+
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+import org.eclipse.californium.scandium.dtls.cipher.CCMBlockCipher;
+
+/**
+ * 
+ * This class is copied from the COSE Java repository. Changes made: Directly
+ * changed the used cipher to Scandiums CCMBlockCipher code. Removing support
+ * for a wider array of AES algorithms.
+ *
+ */
+public abstract class EncryptCommon extends Message {
+
+	private final int AES_CCM_16_IV_LENGTH = 13;
+	protected String context;
+	protected byte[] rgbEncrypt;
+
+	protected byte[] decryptWithKey(byte[] rgbKey) throws CoseException {
+		CBORObject algX = findAttribute(HeaderKeys.Algorithm);
+		AlgorithmID alg = AlgorithmID.FromCBOR(algX);
+
+		if (rgbEncrypt == null)
+			throw new CoseException("No Encrypted Content Specified");
+
+		switch (alg) {
+		case AES_CCM_16_64_128:
+			AES_CCM_Decrypt(alg, rgbKey);
+			break;
+
+		default:
+			throw new CoseException("Unsupported Algorithm Specified");
+		}
+
+		return rgbContent;
+	}
+
+	void encryptWithKey(byte[] rgbKey) throws CoseException, IllegalStateException {
+		CBORObject algX = findAttribute(HeaderKeys.Algorithm);
+		AlgorithmID alg = AlgorithmID.FromCBOR(algX);
+
+		if (rgbContent == null)
+			throw new CoseException("No Content Specified");
+
+		switch (alg) {
+		case AES_CCM_16_64_128:
+			AES_CCM_Encrypt(alg, rgbKey);
+			break;
+
+		default:
+			throw new CoseException("Unsupported Algorithm Specified");
+		}
+	}
+
+	private int getAES_CCM_IVSize(AlgorithmID alg) throws CoseException {
+		switch (alg) {
+		case AES_CCM_16_64_128:
+			return AES_CCM_16_IV_LENGTH;
+		default:
+			throw new CoseException("Unsupported Algorithm Specified");
+		}
+	}
+
+	private void AES_CCM_Decrypt(AlgorithmID alg, byte[] rgbKey) throws CoseException, IllegalStateException {
+		// validate key
+		if (rgbKey.length != alg.getKeySize() / 8) {
+			throw new CoseException("Key Size is incorrect");
+		}
+
+		// obtain and validate IV
+		final int ivLen = getAES_CCM_IVSize(alg);
+		CBORObject iv = findAttribute(HeaderKeys.IV);
+		if (iv == null) {
+			throw new CoseException("Missing IV during decryption");
+		}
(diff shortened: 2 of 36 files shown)


oscore-cf/src/main/java/org/eclipse/californium/cose/EncryptCommon.java
```diff
@@ -0,0 +1,199 @@
...
+		if (iv.GetByteString().length != ivLen) {
+			throw new CoseException("IV size is incorrect");
+		}
+
+		try {
+			rgbContent = CCMBlockCipher.decrypt(rgbKey, iv.GetByteString(), getExternal(), getEncryptedContent(), 0);
+		} catch (NoSuchAlgorithmException ex) {
+			throw new CoseException("Algorithm not supported", ex);
+		} catch (InvalidKeyException ex) {
+			if (ex.getMessage() == "Illegal key size") {
```
boaks: `==` please change it to `equals()`.

oscore-cf/src/main/java/org/eclipse/californium/oscore/CoapOSExceptionHandler.java
```diff
@@ -0,0 +1,117 @@
...
+	 * @throws OSException
+	 */
+	public static Response manageError(CoapOSException e, Request request) throws OSException {
+		ResponseCode responseCode = e.getResponseCode();
+		Response error = null;
+		boolean includeErrMess = true;
+
+		if (request != null) {
+			if (request.getType() != null) {
+				if (ResponseCode.isClientError(responseCode)) {
```
boaks: What should be the intended behaviour for "isServerError"?

oscore-cf/src/main/java/org/eclipse/californium/oscore/CoapOSExceptionHandler.java
```diff
@@ -0,0 +1,117 @@
...
+		if (errMess == null) {
+			LOGGER.error(Error.ERROR_MESS_NULL);
+			throw new NullPointerException(Error.ERROR_MESS_NULL);
+		}
+
+		if (!response.isConfirmable()) {
+			LOGGER.error("An Empty Message will not be created");
+			return null;
+		}
+		return EmptyMessage.newACK(response);
```
boaks: Did OSCORE specify to use ACK (and not RST) even if the message processing caused an error?

oscore-cf/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
```diff
@@ -0,0 +1,243 @@
...
+import org.slf4j.LoggerFactory;
+
+import org.eclipse.californium.core.coap.Token;
+
+/**
+ * 
+ * Implements the OSCoreCtxDB interface with HashMaps.
+ *
+ */
+public class HashMapCtxDB implements OSCoreCtxDB {
```
boaks: I still miss a "thread safe" implementation.

oscore-cf/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
```diff
@@ -0,0 +1,243 @@
...
+		this.ridMap = new HashMap<>();
+		this.uriMap = new HashMap<>();
+		this.seqMap = new HashMap<>();
+		this.allTokens = new ArrayList<Token>();
+	}
+
+	/**
+	 * @return the singleton instance of this context database
+	 */
+	public static HashMapCtxDB getInstance() {
```
boaks: Should all `CoapEndpoint`s share the same CtxDB? I'm not sure, if the "unique" grants (e.g. Token) will also match for different receiving/sending `CoapEndpoint`s.

oscore-cf/src/main/java/org/eclipse/californium/oscore/CoapOSExceptionHandler.java
```diff
@@ -0,0 +1,117 @@
...
+	 * 
+	 * Process the exception e which was provoked by the received request.
+	 * 
+	 * @param e the exception which was thrown because of the received request.
+	 * @param request the received request.
+	 * @return the appropriate error message in a response.
+	 * @throws OSException
+	 */
+	public static Response manageError(CoapOSException e, Request request) throws OSException {
+		ResponseCode responseCode = e.getResponseCode();
```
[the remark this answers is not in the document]
boaks: Please check the arguments (e.g. for null).

oscore-cf/src/main/java/org/eclipse/californium/oscore/CoapOSExceptionHandler.java
```diff
@@ -0,0 +1,117 @@
...
+		Response error = null;
+		boolean includeErrMess = true;
```
boaks: `includeErrMess` seems to be obsolete

oscore-cf/src/main/java/org/eclipse/californium/oscore/Decryptor.java
```diff
@@ -0,0 +1,272 @@
...
+		}
+	}
+
+	/**
+	 * Replaces the message's options with a new OptionSet which doesn't contain
+	 * any of the non-special E options as outer options
+	 * 
+	 * @param message
+	 */
+	protected static void discardEOptions(Message message) {
```
boaks: Should E options really be silently removed for received messages?

oscore-cf/src/main/java/org/eclipse/californium/oscore/Error.java
```diff
@@ -0,0 +1,64 @@
...
+ *    
+ ******************************************************************************/
+package org.eclipse.californium.oscore;
+
+/**
+ * 
+ * Central location for all potential error messages
+ *
+ */
+public final class Error {
```
boaks: The term/name `Error` is already used by the Java API (https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html ) . Therefore I would prefer to rename to `ErrorDescriptions`.

oscore-cf/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
```diff
@@ -0,0 +1,277 @@
...
+					public void onReadyToSend() {
+						Token token = preparedRequest.getToken();
+						db.addContext(token, ctx);
+						db.addSeqByToken(token, seqByToken);
+					}
+				});
+
+				req = preparedRequest;
+
+				if (req.getToken() != null) {
```
boaks: This will be now added in the `onReadyToSend` callback.
Please remove lines 142-145.

oscore-cf/src/main/java/org/eclipse/californium/oscore/CoapOSExceptionHandler.java
T0bbe: Stop processing.

oscore-cf/src/main/java/org/eclipse/californium/cose/EncryptCommon.java
T0bbe: Of course.

oscore-cf/src/main/java/org/eclipse/californium/oscore/CoapOSExceptionHandler.java
T0bbe: I will have to double check this.

oscore-cf/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
T0bbe: We have not talked much about the multiple endpoint situation. The reason why the db was made into a singleton was to avoid passing it along with the exchange object. Your point is good and I think this needs some discussion.

oscore-cf/src/main/java/org/eclipse/californium/oscore/CoapOSExceptionHandler.java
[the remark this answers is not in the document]
T0bbe: Of course.

oscore-cf/src/main/java/org/eclipse/californium/oscore/CoapOSExceptionHandler.java
T0bbe: Yes.

oscore-cf/src/main/java/org/eclipse/californium/oscore/Decryptor.java
T0bbe: Yes. E options which are put as outer options and which are not supposed to be outer options are removed early on when receiving response/request.

oscore-cf/src/main/java/org/eclipse/californium/oscore/Error.java
T0bbe: I will make up another name, something that is also short.

oscore-cf/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
T0bbe: My mistake.

oscore-cf/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
T0bbe: Will fix.

oscore-cf/src/main/java/org/eclipse/californium/oscore/CoapOSExceptionHandler.java
boaks: Without ACK or RST? Then the coap client will usually repeat its request (at least, if it's a CON).
Just to mention, coap uses two number ranges, one indicating, that the client has to change, the other indicating, that the server is in an error state and the client message maybe processed at a other time. So I don't see, why only client errors should be forwarded. Maybe, currently only client errors are used, but if this is extended in the future, I think, server errors should be sent as well.

T0bbe: In the case of isServerError a different function should be used all together, which in some cases replies with an EmptyMessage. See line 98 in CoapOSExceptionHandler

oscore-cf/src/main/java/org/eclipse/californium/oscore/Error.java
T0bbe: Of course.

oscore-cf/src/main/java/org/eclipse/californium/oscore/Decryptor.java
T0bbe: Adding logging for the action.

T0bbe:
Also includes the re-design, to use CoapStackFactory.

boaks:
@T0bbe 
looks not too bad :-).
Please, squash your two commits and revert the "changes, only containing source formatting" in `OptionSet` and `CoapEndpoint`, and then force push that single commit. 

FMPOV, this should be then the base for the CQ. Smaller changes and improvements could/should then be added later in "smaller portions".

T0bbe:
There is another change/addition that we would like to add. Instead of the client manually have to add/flag the OSCORE option we would like a method in the CoapClient in the lines of "enableOSCORE". This change/addition should make the code easier to understand how to "activate" OSCORE processing.

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
```diff
@@ -135,6 +135,7 @@ public void clear() {
 		block2 = null;
 		observe = null;
 		oscore = null;
+
```
boaks: I still prefer, if this file would not be changed only to format code, 
Please replace it with the previous one.

oscore-cf/src/main/java/org/eclipse/californium/oscore/Decryptor.java
boaks: I still miss the logging.

T0bbe: I had placed the logging in the decryptor. Will fix.

boaks:
Maybe "disableOSCORE" in Request may also be easy. Or just remove the `shouldProtectRequest` from the `ObjectSecureLayer`, then all none OSCORE request should then use a  different CoapClient with a different CoapEndpoint. 

My point here is more, that if this is the first version, then we could start to trigger the CQs.
Smaller changes/PR can then be added later without such a CQ.

@sophokles73 
Any opinion?

californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java
T0bbe: Sorry.

boaks:
I've created the CQs for the three dependency use libraries, cose, cbor, and numbers. 
If you're done with this PR, I would also create a CQ for it. 
(This doesn't mean, that the oscore could not be improved or changed. 
It means, that this is the initial contribution to be checked for IP stuff.)

boaks:
CQ is created, please don't change this PR!

boaks:
> You may now proceed to check the project code attached to this CQ into the
> Californium repository.  A full review will be forthcoming.

PR #683, opened 2018-07-02
Exchange leak bw

### LINKED ISSUE #429: Memory leak in MessageExchangeStore

boaks:
Alternative PR for #675 to fix #429.
Add cleanup to blockwise notifies and injects errors without catching the exceptions.


sbernard31:
I'm currently working on moving the cleanup close to the add.
Even If I think reporting error in coapstack would be an enhancement, I let this aside for now and go only about the leak. (using your ErrorInjector)

boaks:
Close this and create PR via bsinno.

PR #685, opened 2018-07-03
Add deliver duplicate to exchange.

boaks:
If coap requests are forwarded into a "back pressure" mechanism, this extension enables the coap stack to explicit wait for a retransmission of a request to overcome a short time overload of the system. Overloads with longer duration may still be answered by 5.03 to indicate such a longer overload to the client.
This extension is intended to be used for the hono coap-adapter.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
@@ -43,6 +43,7 @@
  *                                                    Issue #487
  *    Achim Kraus (Bosch Software Innovations GmbH) - add checkMID to support
  *                                                    rejection of previous notifications
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add deliver duplicate
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -69,6 +70,7 @@ import org.eclipse.californium.core.coap.Token;
 import org.eclipse.californium.core.network.stack.BlockwiseLayer;
 import org.eclipse.californium.core.network.stack.CoapStack;
 import org.eclipse.californium.core.observe.ObserveRelation;
+import org.eclipse.californium.core.server.MessageDeliverer;
 import org.eclipse.californium.core.server.resources.CoapExchange;
 import org.eclipse.californium.elements.EndpointContext;
 import org.slf4j.Logger;
@@ -91,7 +93,8 @@ import org.slf4j.LoggerFactory;
  * {@link #execute(StripedExchangeJob)}. For convenience the
  * {@link #executeComplete()} is provided to execute {@link #setComplete()}
  * accordingly. Methods, which are documented to throw a
- * {@link ConcurrentModificationException}MUST comply to this execution pattern!
+ * {@link ConcurrentModificationException} MUST comply to this execution
+ * pattern!
  * <p>
  * If the exchange represents a "blockwise" transfer and if the transparent mode
  * is used, the exchange keeps also the (original) request and use the current
@@ -252,6 +255,21 @@ public class Exchange {
 	// The relation that the target resource has established with the source
 	private volatile ObserveRelation relation;
 
+	/**
+	 * Indicates, that the current request is already answered. Access is
+	 * synchronized in scope of {@link #deliverDuplicatesCounter}.
+	 */
+	private boolean currentRequestAnswered;
+	/**
+	 * Indicates, that a duplicate request without answer should be passed to
+	 * the {@link MessageDeliverer}.
+	 */
+	private boolean deliverDuplicate;
+	/**
+	 * Counter for delivered duplicate requests.
+	 */
+	private final AtomicInteger deliverDuplicatesCounter = new AtomicInteger();
+
 	private final AtomicReference<EndpointContext> endpointContext = new AtomicReference<EndpointContext>();
 
 	//If object security option is used, the Cryptographic context identifier is stored here
@@ -341,9 +359,20 @@ public class Exchange {
 	 * Accept this exchange and therefore the request. Only if the request's
 	 * type was a <code>CON</code> and the request has not been acknowledged
 	 * yet, it sends an ACK to the client.
+	 * 
+	 * @throws IllegalStateException if {@link #setupDeliverDuplicate(int)} was
+	 *             called before, but {@link #checkCurrentResponse(Response)}
+	 *             wasn't called to indicate the intended request
+	 *             retransmission.
 	 */
 	public void sendAccept() {
 		assert (origin == Origin.REMOTE);
+		synchronized (deliverDuplicatesCounter) {
+			if (deliverDuplicate) {
+				throw new IllegalStateException("Accept not allowed while waiting for request retransmission!");
+			}
+			currentRequestAnswered = true;
+		}
 		Request current = currentRequest;
 		if (current.getType() == Type.CON && !current.isAcknowledged()) {
 			current.setAcknowledged(true);
@@ -355,9 +384,20 @@ public class Exchange {
 	/**
 	 * Reject this exchange and therefore the request. Sends an RST back to the
 	 * client.
+	 * 
+	 * @throws IllegalStateException if {@link #setupDeliverDuplicate(int)} was
+	 *             called before, but {@link #checkCurrentResponse(Response)}
+	 *             wasn't called to indicate the intended request
+	 *             retransmission.
 	 */
 	public void sendReject() {
 		assert (origin == Origin.REMOTE);
+		synchronized (deliverDuplicatesCounter) {
+			if (deliverDuplicate) {
+				throw new IllegalStateException("Reject not allowed while waiting for request retransmission!");
+			}
+			currentRequestAnswered = true;
+		}
 		Request current = currentRequest;
 		current.setRejected(true);
 		EmptyMessage rst = EmptyMessage.newRST(current);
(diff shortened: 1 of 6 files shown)


sbernard31:
If I well understand the idea is to be able on a coap resource to not handle a request right now and delay it to the nth retransmission ?

boaks:
Yes. FMPOV the "short term" congestion control in CoAP just ignores a request and process it on retransmission. "long term" would be 5.03.
I also tried to "externalize" such a mechanism, but I didn't have a good idea to do so ...

sbernard31:
> I also tried to "externalize" such a mechanism, but I didn't have a good idea to do so ...

Are you talking about 'short term' or 'long term' congestion control ? (I mean you tried to externalize the mechanism added in this PR or a new one with 5.03 response ?)

Did you try this PR, does it really help ?  Did you have a sample of code where you are using it ? (I mean more real use case than the Junit test in this PR).

Stupid question : There is congestioncontrol layer in californium (never played with it). Should it solve this kind of issue ?

As attributes are store in `Exchange`, I ask myself how this behave with transparent block-transfer.

boaks:
> Are you talking about 'short term' or 'long term' congestion control ? 

I talk about "short term" and "long term" overloading by congestion (too many requests for the north-bound system, which therefore reports some back pressure). If it's only for a short term (overloading for a period of 1-2s), then in my opinion, ignore and deliver the duplicate will do it best. 
If the "overload" sticks longer, then I think return 5.03 will be better. That's, what this PR intend to solve. 

The "externalize this mechanism" (implement without changing the Exchange) was something, I thought of, but withdrawn it. It has nothing to do with this PR. If externalize of that mechanism would be an option, it would go into an other PR.

> Did you try this PR, does it really help ? Did you have a sample of code where you are using it ? (I mean more real use case than the Junit test in this PR).

Sure, the hono coap-adapter. But currently that hono PR would depend on californium 2.0.0-SNAPSHOT (with this PR in the californium 2.0.0-SNAPSHOT). Therefore I didn't push it.

> Stupid question : There is congestioncontrol layer in californium (never played with it). Should it solve this kind of issue ?   

AFAIK, that layer targets the coap client, but this PR targets the coap-server.
And when I checked it the last time (3 years ago), it was full of threading issues.
So "congestion-control" has two sides: as long as the clients stick to NSTART-1, a server can decide to response with a delay, which then throttle the effective number of request. Without "north bound system" this would cause excessive processing on the server side, which automatically delays the responses. But with a "decoupled/asynchronous northbound", the coap-server may send very fast it's 5.03, which then generates a lot of useless traffic. 

> I ask myself how this behave with transparent block-transfer.

It's not intended for blockwise. Though transparent blockwise is mostly not "short term", there seems to be no need to use it there.

sbernard31:
Ok.

I didn't see any way to externalize this too, but maybe we can simplify the modification.
Instead of adding the threshold/counter stuff, we can just add the possibility to force to deliver duplicates through the stack. So a user can activate it when needed, check getFailedRetransmission() to delay the handling of the request, then deactivate it.

This means just adding 1 attribute (boolean deliverDuplicate), 1 setter (forceDeliverDuplicate), 1 getter (mustDeliverDuplicate).

boaks:
The failed retransmission counter is for sending a message, I didn't want to reuse it for receiving a request.

```
// server overloaded, request could currently not be processed 
if (!exchange.advanced().requestDeliverDuplicate(2)) {
    exchange.respond(ResponseCode.SERVICE_UNAVAILABLE, t.getMessage());
}
```

would be my snippet to use it. For the first 3 request, just ignore the overload. If it takes longer than these two retransmissions, then response with 5.03.
I would prefer to keep the straight forward usage with the two additional attributes. Reusing a different attribute will not make it easier to understand it.

sbernard31:
You're right about getFailedRetransmission(), its for sending request... 

> would be my snippet to use it. For the first 3 request, just ignore the overload. If it takes longer than these two retransmissions, then response with 5.03.

Yep I get it.

When I review it, I can ensure that was not so easy to understand what this method means exchange.advanced().requestDeliverDuplicate(2).

That sounds really easier to understand to me to separate "counting the duplicate" and "start/stop to force to deliver duplicate".

boaks:
OK, I guess "request" is used term in coap. So I renamed it to "prepare".

sbernard31:
Not really clearer to me.

boaks:
I wouldn't say, that split the logic into several simpler primitive functions and then delegate the implementation of the wanted logic to the caller, makes it easier. I chose to group the operations into that two functions to ensure, that there are proper synchronized. I didn't want to schedule a additional striped execution for that very simple prepare. But if the logic is exposed and delegeted to the caller, the caller context must be a striped execution.

sbernard31:
> I wouldn't say, that split the logic into several simpler primitive functions and then delegate the implementation of the wanted logic to the caller, makes it easier.

I think it does and it also allow to use other criteria than giving a number of "duplicate request" to start or stop this "special" mode.

> But if the logic is exposed and delegeted to the caller, the caller context must be a striped execution.

I also think the CoapResource.handleRequest will be executed in the striped execution.

boaks:
> I also think the CoapResource.handleRequest will be executed in the striped execution.

**No**! That may be the default, but is not required, see `ConcurrentCoapResource`.
If a `ExecutorService` is set, `ServerMessageDeliverer.deliverRequest` executes `CoapResource.handleRequest` using that service (and is therefore **not** stripped). 
So it should be also possible to change the Thread even during `CoapResource.handleRequest` and with that, the execution of the operations in `Exchange.setupDeliverDuplicate` must be handover to the stripped executor or must be synchronized within the Exchange. Though the `Exchange.sendAccept`, `Exchange.sendReject`, and `Exchange.sendResponse` are also accessing the fields protected within the `Exchange.setupDeliverDuplicate`, a exposed solution would require to execute them also stripped. 
(And yes, to delegate the processing into vert.x, I pass over the execution to an other thread ... but that's just my use-case.)

So, a lot of things to obey. Therefore I still belief, that this is the right approach. If other criteria will be required, it will be easier to integrate them into the Exchange, than to expose the primitives. Sure, this means to change the "Exchange" again, but to proper synchronize on this seems for me to be too difficult outside.

sophokles73:
@boaks I do not really see the advantage of doing this in comparison to simply returning a 5.03 which includes a Max-Age option indicating to the client how many seconds it should wait before re-trying.
The client may or may not respect the server's wish or it may decide to not re-try at all, e.g. because the data it tried to send will be outdated before the wait period is over anyway. In any case, the server needs to be prepared to process requests from the client at a pace that is dictated by the client. However, I would rather not introduce a new *special case* into the stack if it doesn't provide a clear advantage that cannot be achieved otherwise.

boaks:
> I do not really see the advantage of doing this in comparison to simply returning a 5.03 which includes a Max-Age option indicating to the client how many seconds it should wait before re-trying.

Without this PR no one may be able to see the advantage :-) . 
The IETF core group has also an other solution idea 
https://tools.ietf.org/html/draft-ietf-core-too-many-reqs-02 .

In my opinion using the "congestion control retransmission approach" for "short term" overloading will result in best performance, because it doesn't require to send a response. It will also be supported "out of the box" by the most clients, which have implemented the CON message transmission. 
5.03 (or the potential new 4.29) will required an additional message and a client must implementation it. 

So, with this PR, everyone can check much easier, what fits best.

boaks:
>  However, I would rather not introduce a new special case into the stack if it doesn't provide a clear advantage that cannot be achieved otherwise.

"congestion control retransmission approach" for "short term" overloading is already implemented! This PR just changes the possibility for asynchronous implementation, where the "overload" is not caused be the "californium process", but by an overload of a northbound system.

sbernard31:
> **No!** That may be the default, but is not required, see ConcurrentCoapResource.

bold + exclamation :fearful:

> If a ExecutorService is set, ServerMessageDeliverer.deliverRequest executes CoapResource.handleRequest using that service (and is therefore not stripped).

I was understanding that the thread model you initiated (:+1:) was to isolate exchange handling in their own thread.
The possibility of setting a custom ExecutorService (and so executing deliverRequest in another thread) existed before this "big" change. And maybe this makes less sense now.

> So it should be also possible to change the Thread even during CoapResource.handleRequest

Of course, and this could make sense for long time processing.
But currently this is not so clear (maybe only to me) what could be done out of the striped execution and what is not.

Typically, you are talking about `Exchange.sendAccept`, `Exchange.sendReject`, and `Exchange.sendResponse`. It makes senses to let this method usable outside the striped execution, but actually I don't know if this is really safe. (For example, I suspect changing currentRequest could be a problem. I don't know if this could happened, maybe with transparent blockwise)

You start to add  "check/assert owner" in lot of public method maybe we should go deeper in this way. And do this check/assert for all methods except the one which could be safely used outside the striped execution.

To return to this use case especially, I understand that synchronization is only used to count correctly duplicates.
So if you get the count value before to go outside of the stripe execution, it should be ok.
Making the start/stop deliver mode thread-safe should not be an issue.
About counting duplicates, I don't even know how much this is critical as duplicate requests could be lost anyway with UDP and this could be consider as a best-effort mechanism.

**To make it shorter :**
As @sophokles73, I'm not so sure of the benefits of this and that's why I'm a bit hesitant to add more complexity to the stack for this.
If you really want to include, I would prefer a less specific solution.
Maybe providing demo/tests could help to ensure of the benefits and/or tests severals implementations.
Is it something you need rapidly ?

sbernard31:
Uups, I forgot to answer to this thread.

I would prefer we provide demo/tests which could help to ensure the benefits and/or tests severals implementations.
But if because of time constraint, you would like to integrate this as an experimental feature, I could understand. (even if experimental feature was rarely removed :-p)

Maybe, as this is mainly an hono drived feature, you should get the last @sophokles73's  point of view ?

boaks:
I have no problem, if we just keep the PR open (unmerged) and see, if someone will try it out.
Though it's not that easy to setup "load tests for hono", my current load test with 600 client shows only 1% "overload". That may get larger, but on the otherside, if nobody has more load, then 5.03 will do it also.

PR #686, opened 2018-07-03
Reject prev notify

boaks:
Testing to cancel a observe by rejecting a previous notification showed a defect.
This PR fix that.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -53,6 +53,8 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - add striped execution
  *                                                    based on exchange
  *    Achim Kraus (Bosch Software Innovations GmbH) - add coap-stack-factory
+ *    Achim Kraus (Bosch Software Innovations GmbH) - use checkMID to support
+ *                                                    rejection of previous notifications
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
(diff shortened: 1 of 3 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -976,13 +978,7 @@ private void receiveEmptyMessage(final EmptyMessage message) {
...
 							@Override
 							public void runStriped() {
-								int mid;
-								if (exchange.getOrigin() == Origin.LOCAL) {
-									mid = exchange.getCurrentRequest().getMID();
-								} else {
-									mid = exchange.getCurrentResponse().getMID();
-								}
-								if (message.getMID() == mid) {
+								if (exchange.checkMID(message.getMID())) {
```
sbernard31: Is it not to the matcher to return exchange only if it match this message ?
(I must to confess, I do not really understand why we need to do this check ?)

boaks: :-). That matching may change over the time, therefore I revalidate it within the stripped execution.

sbernard31: Ok we do the same "kind" of check when receiving response too.

I think this maybe deserve some comment in the code to explain that because it's important to not change this and we don't have tests which could ensure that this is not modified by mistake.

boaks: OK, I added two comments.

sbernard31:
LGTM

Issue #688, opened 2018-07-05
Ignore GET response with unexpected observe option


sbernard31:
There is a behavior change from californium 2.0.0-M9 to 2.0.0-M10.

If we receive a Separated response with an observe option from a GET (without observe option) we will ignore it. (see [UdpMatcher](https://github.com/eclipse/californium/blob/2.0.x/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java#L290))
Just to be clear, The Token match but there is just an unexpected observe option.

I think that was not the case before. The changes seems to be done in 25548e29afb011e4bf1257b0bb856a985b8a9179.

I totally agree that the device should not answer with the observe option but, 
1) should we be so strict ? I mean if there is other option than observe that we don't expect we will not ignore the response ?
2) Maybe there is a technical issue which explain we change the behavior like this ?

I tried to search in the spec and didn't find anything about that.

To be honest, we currently have some "buggy" devices which exactly behave like this and so which does not work anymore with californium 2.0.0-M10. We currently revert to 2.0.0-M9 waiting to find a solution (or workaround) for this.

boaks:
> If we receive a Separated response with an observe option from a GET (without observe option) we will ignore it. 

>I think that was not the case before. The changes seems to be done in 25548e2.

```
if (response.isNotification() && response.getType() != Type.ACK && !exchange.getRequest().isObserve()) {
	// notification for none observe request
	// including observation cancel request
	LOGGER.debug("ignoring notify {}!", response);
	return null;
}
```

Yes, analysing test with "proactive" cancel observes (GET Observe 1) shows, that a notify should not complete that exchange, otherwise it will not be ensured, that the cancel is really executed.

>I totally agree that the device should not answer with the observe option but,
  
I hope, this is just a notification and the real response is on the way. Or did the device just add the option even to a GET with a new token? If it just adds the option accidentally, does it send an ACK before? Maybe we change it again, that it only ignores it, if the exchange didn't receive ACK before.

boaks:
Or we check, if the token is contained in the `ObservationStore`. Only if it's in there, we ignore it in that case.
So the questions are:
is  there an observation with the same token?
is a ACK send (usually) ahead?

sbernard31:
Unfortunately, the device is pretty stupid..
You send it a GET request without observe option.
It send you a ACK with the right MID.
Then a seperated CON response with the right token but with an unexpected observe option ...
(We never send it GET with observe option before...)

[...]

PR #689, opened 2018-07-05
Ignore notifies only for cancel observation requests.

boaks:


boaks:
Fix issue #688

sbernard31:
LGTM (except the commit name :-p)
Thx @boaks

Issue #690, opened 2018-07-06
Is the Executor must be ScheduledExecutor ?


joechen2010:
I don't know why always ScheduledExecutor everywhere, especially coapendpoint, seems there is no schedule feature in it. in our test thread model (async http request with jetty and resteasy -> convert to coap request -> client coap resource)  seems there is performance issues with the jdk ScheduledThreadPoolExecutor, and i saw the BenchmarkServer also implements a customized ScheduledExecutorService, it means you have already found this defect, right?

boaks:
The executor is passed in the CoapStack and used to schedule retransmissions there in the ReliabilityLayer.

```
public void startTimer() {
	if (!exchange.isComplete()) {
		int timeout = exchange.getCurrentTimeout();
		ScheduledFuture<?> f = executor.schedule(this, timeout, TimeUnit.MILLISECONDS);
		exchange.setRetransmissionHandle(f);
	}
}
```

> and i saw the BenchmarkServer also implements a customized ScheduledExecutorService, it means you have already found this defect, right?

According git, that at least 3 years old and in the end, it uses also a ScheduledThreadPool.
 
```
public WorkQueueExecutor(int threads) {
	this(threads, Executors.newScheduledThreadPool(1));
}
```

But maybe @mkovatsc knows the reason.

If you do some performance test, can you post your results?
I'm wondering, if in your chain coap is really a bottleneck.
It may be possible, that especially for outgoing CON the ScheduledExecutor implementation spends too much time into it.

Do you use the CoapClient? Do you use "synchronous requests" (e.g. `CoapClient.get()`) ?
If so, try to use the "asynchronous request" (e.g. `CoapClient.get(CoapHandler handler)`). 

A second tip would be, to disable the deduplication,

```
// Network configuration optimal for performance benchmarks
NetworkConfig.createStandardWithoutFile()
	// Disable deduplication OR strongly reduce lifetime
	.setString(NetworkConfig.Keys.DEDUPLICATOR, NetworkConfig.Keys.NO_DEDUPLICATOR)
```

but FMPOV, in production this may have more drawbacks than benefits.

joechen2010:
Thanks Boaks, the result on our test with ScheduledThreadPoolExecutoris around 18k and the following executor is 35k. all outgoing type is CON, DEDUPLICATOR is disable, and the lifechange time was descred to 1500, also use asynchronous request like request.send(endpoint). 
[...]

PR #692, opened 2018-07-09
Add some tests about proactive cancel

sbernard31:
I added some tests relative to #688.

I think "a test about pro active cancel (ed40c69)" makes sense to be added.
The second one, "test that response with observe option is for simple GET (8fcf2cb)", maybe a little less, but I used it to check #689.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
@@ -203,9 +203,8 @@ public class CoapObserveRelation {
 	 * GET with Observe=1.
 	 */
 	public void proactiveCancel() {
-		sendCancelObserve();
-		// cancel observe relation
 		cancel();
+		sendCancelObserve();
 	}
 
 	/**
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
@@ -1256,4 +1256,54 @@ public class ObserveClientSideTest {
 
 		// @after check there is no leak
 	}
+
+	/**
+	 * This is a bit out of specification but to improve compatibility with
+	 * "buggy" server we would like to ensure that if a server answer to a
+	 * simple GET (without observe) by adding an unexpected observe option we
+	 * will handle it anyway.
+	 */
+	@Test
+	public void testSimpleGetAcceptResponseWithObserveOption() throws Exception {
+		System.out.println("Response with observe option is accepted as response for a GET");
+		respPayload = generateRandomPayload(10);
+		String path = "test";
+
+		// Send simple get request
+		Request request = createRequest(GET, path, server);
+		client.sendRequest(request);
+		server.expectRequest(CON, GET, path).storeBoth("A").go();
+
+		// Send a piggyback response with unexpected observe option
+		server.sendResponse(ACK, CONTENT).loadBoth("A").observe(3).payload(respPayload).go();
+
+		// Ensure we get the response
+		Response response = request.waitForResponse(1000);
+		assertResponseContainsExpectedPayload(response, respPayload);
+
+		// Send another simple get request
+		request = createRequest(GET, path, server);
+		client.sendRequest(request);
+		server.expectRequest(CON, GET, path).storeBoth("B").go();
+		// We don't send ACK to simulate it is lost
+		// Send a separated CON response with unexpected observe option
+		server.sendResponse(CON, CONTENT).loadToken("B").mid(mid).observe(4).payload(respPayload).go();
+		server.expectEmpty(ACK, mid).go();
+
+		// Ensure we get the response
+		response = request.waitForResponse(1000);
+		assertResponseContainsExpectedPayload(response, respPayload);
+
+		// Send another simple get request
+		request = createRequest(GET, path, server);
+		client.sendRequest(request);
+		server.expectRequest(CON, GET, path).storeBoth("C").go();
+
+		// Send NON response with unexpected observe option
+		server.sendResponse(NON, CONTENT).loadBoth("C").observe(5).payload(respPayload).go();
+
+		// Ensure we get the response
+		response = request.waitForResponse(1000);
+		assertResponseContainsExpectedPayload(response, respPayload);
+	}
 }
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
@@ -1306,4 +1306,144 @@ public class ObserveClientSideTest {
 		response = request.waitForResponse(1000);
 		assertResponseContainsExpectedPayload(response, respPayload);
 	}
+
+	/**
+	 * Test proactive cancel
+	 * 
+	 * <pre>
+	 * (actual used MIDs and Tokens my vary!)
+	 * (Established observe relation)
+	 * CON [MID=27914, T=[d5c120197ca1c3ce]], GET, /test, observe(0)    ----->
+	 * <-----   ACK [MID=27914, T=[d5c120197ca1c3ce]], 2.05, observe(1)
+	 * (1st Notification)
+	 * <-----   CON [MID=8001, T=[d5c120197ca1c3ce]], 2.05, observe(3)
+	 * ACK [MID=8001]   ----->
+	 * (Send proactive cancel request)
+	 * CON [MID=27915, T=[d5c120197ca1c3ce]], GET, /test, observe(1)    ----->
+	 * <-----   ACK [MID=27915, T=[d5c120197ca1c3ce]], 2.05
+	 * (Next notifications are rejected) 
+	 * <-----   CON [MID=8002, T=[d5c120197ca1c3ce]], 2.05, observe(4)
+	 * RST [MID=8002]   ----->
+	 * <-----   NON [MID=8003, T=[d5c120197ca1c3ce]], 2.05, observe(5)
+	 * RST [MID=8003]   ----->
+	 * </pre>
+	 */
+	@Test
+	public void testProactiveCancel() throws Exception {
+		System.out.println("Proactive cancel");
+		respPayload = generateRandomPayload(10);
+		String path = "test";
+
+		// Establish observe relation
+		Request request = createRequest(GET, path, server);
+		request.setObserve();
+		respPayload = generateRandomPayload(16);
+		client.sendRequest(request);
+		server.expectRequest(CON, GET, path).storeBoth("OBS").go();
+		server.sendResponse(ACK, CONTENT).loadBoth("OBS").observe(1).payload(respPayload).go();
+		Response response = request.waitForResponse(2000);
+		assertResponseContainsExpectedPayload(response, respPayload);
+
+		// New CON notification
+		String notifyPayload = generateRandomPayload(32);
+		server.sendResponse(CON, CONTENT).loadToken("OBS").observe(3).mid(++mid).payload(notifyPayload).go();
+		server.expectEmpty(ACK, mid).go();
+
+		// Send pro active cancel
+		client.cancelObservation(request.getToken());
+		Request proactiveCancel = createRequest(GET, path, server);
+		proactiveCancel.setToken(request.getToken());
+		proactiveCancel.setObserveCancel();
+		client.sendRequest(proactiveCancel);
+		server.expectRequest(CON, GET, path).storeBoth("OBSCANCEL").go();
+
+		// Send response to proactive cancel
+		String cancelPayload = generateRandomPayload(32);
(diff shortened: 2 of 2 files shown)


californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
```diff
@@ -1327,4 +1327,52 @@ public void testNotificationIsNotHandledAsProactiveCancelResponse() throws Excep
 		assertEquals(mid, cancelResponse.getMID());
 		assertResponseContainsExpectedPayload(cancelResponse, cancelPayload);
 	}
+
+	/**
+	 * Ensure that response with unexpected observe option is not ignored
```
boaks: Can you add the hint, that this is only intended to ensure compatibility with a broken/none RFC-7252 compliant client?

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
```diff
@@ -1256,4 +1256,75 @@ public void testObserveFailureBeforeToSendDuringBlockNotification() throws Excep
...
+	 * <-----   NON [MID=8003, T=[15d4ebce227d141d]], 2.05, observe(5)
+	 * RST [MID=8003]   ----->
+	 * CON [MID=55454, T=[15d4ebce227d141d]], GET, /test, observe(1)    ----->
+	 * CON [MID=55454, T=[15d4ebce227d141d]], GET, /test, observe(1)    ----->
+	 * <-----   CON [MID=8004, T=[15d4ebce227d141d]], 2.05
+	 * ACK [MID=8004]   ----->
+	 * </pre>
+	 */
+	@Test
+	public void testNotificationIsNotHandledAsProactiveCancelResponse() throws Exception {
```
boaks: I would prefer to have different tests for:
- receiving notifications during the cancel request.
- receiving notifications after successful cancel request 
- receiving notifications after failed cancel request

sbernard31: This test is only about : 
 - receiving notifications during the cancel request.

(I changed it a little bit after our discussion in #688, before to push this PR)

Would you like I add tests for : 
- receiving notifications after successful cancel request
- receiving notifications after failed cancel request

Currently the behavior for failed and successful cancel will be the same.

boaks: > Currently the behavior for failed and successful cancel will be the same.

Do you mean, currently 
- receiving notifications after successful cancel request
- receiving notifications after failed cancel request

behaves equal? 
I would guess (hope :-) ), the client will RST notifications, if the cancel is successful.

sbernard31: > I would guess (hope :-) ), the client will RST notifications, if the cancel is successful.

After cancelling, notifications are not RST anymore. (success or not)
That's what I try to explain in https://github.com/eclipse/californium/issues/688#issuecomment-403501405.

There is 2 API, coapclient and coapendpoint.
Coapclient is more high level api than coapendpoint.

With coapclient, you can use CoapObserveRelation.proactiveCancel() and it will behave like you expect.

But coapendpoint, is low-level api, there is one API to send proactive cancel request and one to remove observation at client side (passive cancellation).

FMPOV in a perfect world, with coapendpoint API notification should be always handle as long as we don't remove observation from obervation store.
But there is a side effect and during an ongoing proactive cancel request, notifications are RST. 
As you explain fixing this means adding corner case in the stack and maybe this is not really worthy. In generally, the proactive use case is to do both at the same time. I mean removing observation from the store, then sending a cancel request.

With the low level API, the success of cancel request does not determine if the observation is removed or not. This should be done manually before or after using endpoint.cancelObservation()

In this test, I just send the cancel request.

boaks: Sorry Simon,
your tests seems for me to be smoke tests, where messages are fired in a chaotic, none-compliant way. I thought, your intention is to ensure, that there are no leaks introduced. If I have over seen the variant, where the cancel is responded successful and further notifications are accepted instead of rejected, then this is my fault. But even in the comment you refer, the cancel request is NOT responded with success, instead it timesout, which I would count as "failed"

That's also my comment:
Please provide not only the "smoke/chaos" test, provide also the three compliant test-cases.

boaks: After removing "the smoke", I recognized, that this test doesn't use the `CoapClient`. Therefore the "proactive cancel" doesn't work. I prepared a fix for proactive cancel without `CoapClient`, but using "observe" without a `CoapClient` will not filter the notifications based on their order and requires a `NotificationListener`. 
So at least, this seems for me to be not recommended.
I got also aware, that `Request.waitForResponse()` is still in between. The comments about the notifications are not cleaned up and the implementation sticks also on the old notification processing.

sbernard31: > your intention is to ensure, that there are no leaks introduced.

My intention was to test the use case you described at https://github.com/eclipse/californium/issues/688#issuecomment-402757040

> But even in the comment you refer, the cancel request is NOT responded with success,

I would like to fix this comment but I don't know which one you are talking about. Could you precise which one ?

> Please provide not only the "smoke/chaos" test, provide also the three compliant test-cases.

I think this is not so chaotic, just 2 notifications which comes before we receive the proactive cancel and a lost ACK for the cancel separated response. (This is compliant case)

but I could add more tests, but currently I'm not sure what is the behavior expected.

> After removing "the smoke"

I still don't understand what you consider as smoke in this test. :-/

> but using "observe" without a CoapClient will not filter the notifications based on their order and requires a NotificationListener.
So at least, this seems for me to be not recommended.

I think this is the interest of having this low-level API. Bringing more flexibility.

> I got also aware, that Request.waitForResponse() is still in between. The comments about the notifications are not cleaned up and the implementation sticks also on the old notification processing.

I do not understand the issue here with the use of waitForResponse, I do not want to get notification, I want to ensure that notification received is not consider as response for the proactive cancel request.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
sbernard31: I agree this test is a bit strange and a bit out of spec, I mainly write it to ensure that the issue reported in #688 was fixed.

I can understand, if you think this should not be integrated in the californium test base.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
boaks: "chaotic" means, it's not that easy/obvious, what it tests. At least for me ...

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
boaks: From my side there is no objection to have this test included to ensure, your already analysed issue is still fixed. A added hint would ensure, that no one takes this as compliant example.

sbernard31: Ok I will add this !

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
sbernard31: > "chaotic" means, it's not that easy/obvious, what it tests. At least for me ...

I do not take your remarks badly, I just don't know how to do this better. :-)
How could I test that a notification is not handled as a proactive cancel response in a more easy/obvious to understand way ?

(Maybe just remove the lost ACK ?)

boaks: > I would like to fix this comment but I don't know which one you are talking about. Could you precise which one ?

```
X------ ACK M3              ------ // proactive ACK lost
<------ CON M6, T1          ------ // response of proactive cancel
```

That's what is "chaotic" for me. First "ACK lost", and then response. So the test-case doesn't simply piggybacked response to the request. That makes it hard to see, if it fails, because the test is broken or because the implementation is broken. To understand such stuff, much more time is required.

sbernard31: OK I changed this.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
boaks: > Ok I will add this !

Any due day? I would prefer to have such a comment added before merging

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/ObserveClientSideTest.java
boaks: > OK I changed this.

Any due day? For that issue, we may also merge it (after https://github.com/eclipse/californium/pull/692#discussion_r201326176) and add this stuff later.

sbernard31: I would like to finished some tasks, I postponed too many time for Leshan. I hope I could rework on this before the end of the day.

Did you ask this because you would like to release the M11 ? If yes, I think the tests in this PR is not mandatory for M11.

boaks: Mentally, I shifted M11 already for a week :-).

boaks:
After changing the a little the threading (see issue #690), I struggling with some race-conditions.
Would it be possible, to add `@RunWith(RepeatingTestRunner.class)` locally and check the tests for robustness? I don't say, there are issues with that, but I would prefer, if it get's checked.

sbernard31:
About @RunWith(RepeatingTestRunner.class), default value make the tests very long. Do you advice to keep this default value or to reduce it ?

sbernard31:
I added comment on f605ca0  to explain this is a pretty buggy use case.

About proactive cancel tests: 
- I added a "standard" test about proactive cancel.
- for all tests, I `cancelObservation` before as it seems to be the right/standard way to do. 
- I also tried to make tests less smoky.

For the last commit, I think we should cancelObservation before to send proactive request in coapclient. If you think, I'm wrong I will just not integrate it. (not a big deal)

I ran the 3 tests with  @RunWith(RepeatingTestRunner.class) in eclipse. I don't know if this really what you expect. It sounds good.

boaks:
> make the tests very long

Sure, it repeats the tests 100 times until it fails :-). 
The idea is, that at the end, before merging, you add locally the `@RunWith(RepeatingTestRunner.class)` and do a final check with the Eclipse Test Runner only running that test class. If that passes, don't commit this last change (otherwise the eclipse jenkins will be in deep trouble :-) ).
    
Just to mention.
My race-conditions where caused by 2 protocol threads, I fixed that ensuring the usage of 1 protocol thread again.

californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
```diff
@@ -203,9 +203,8 @@ private void cancel() {
 	 * GET with Observe=1.
 	 */
 	public void proactiveCancel() {
-		sendCancelObserve();
-		// cancel observe relation
 		cancel();
+		sendCancelObserve();
```
boaks: The other order was chosen, to ensure, that the token is not used in the "meantime" (between line 206 and 207) by an other request. But it's academically and not really analysed, if this is really a risk.

boaks:
> I ran the 3 tests with @RunWith(RepeatingTestRunner.class) in eclipse. I don't know if this really what you expect. It sounds good.

Cool, does it mean you run the 3 tests each 100 times successful local on your PC?
Yes, that's what I think is a good practice, at least for new stuff.
 
LGTM

californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
sbernard31: I feel you're right that we may have a potential token collision here, but I think that's was already the case before because `sendCancelObserve` is not synchronous.

If we want to avoid this, the observationStore should contains the observation (and so lock the token) until we complete the proactive cancel exchange.

If we want to do that we should go with something more like you do in #694.

So, I will not integrate this last commit and merge only the tests.

sbernard31:
> Cool, does it mean you run the 3 tests each 100 times successful local on your PC?

Yep.

sbernard31:
f605ca043d1bcafe0f8809a9e7baf8619054da79 and 06727905d8b6cb02d6e98e51533f42cedc238459 are integrated in **2.0.x**.

So I close this PR.

PR #693, opened 2018-07-10
Respond overload

boaks:
Add convenient method for overload response to indicate a "long term"
overload to the client along with an estimated time the retry to send
request.


sbernard31:
LGTM

Issue #465, opened 2017-11-03
Observe Relation: Response 4.06, going to onLoad of CoapHandler


charybr:
Hi,

We noticed that when 4.06 response comes from server for a get-observe request, control goes to onLoad. As a result, CoapHandler doesnt get notified about observe being deleted.

Log attached.

[californium.0.0.log.zip](https://github.com/eclipse/californium/files/1441450/californium.0.0.log.zip)

From logs:
Registering obs 
Nov 03, 2017 6:54:58 PM org.eclipse.californium.core.network.BaseMatcher registerObserve
FINER: registering observe request CON-GET    MID=16232, Token=cfb4e4573c35ff28, OptionSet={"Uri-Host":"8eb76988-7264-4a2d-beca-8169df875083", "Observe":0, "Uri-Path":"weighsens"}, no payload

Got 4.06 response from server.
Nov 03, 2017 6:55:00 PM org.eclipse.californium.core.network.UdpMatcher receiveResponse
FINER: received response ACK-4.06   MID=16232, Token=cfb4e4573c35ff28, OptionSet={"Content-Format":"application/cbor"}, no payload

One more 4.06 response from server.
Nov 03, 2017 6:55:02 PM org.eclipse.californium.core.network.UdpMatcher receiveResponse
FINER: received response ACK-4.06   MID=16232, Token=cfb4e4573c35ff28, OptionSet={"Content-Format":"application/cbor"}, no payload

Removed the observation
Nov 03, 2017 6:55:02 PM org.eclipse.californium.core.network.BaseMatcher$2 onResponse
FINE: Response to observe request with token KeyToken[cfb4e4573c35ff28, fd0080000000000000000000000200af:5683] does not contain observe option, removing request from observation store
Nov 03, 2017 6:55:02 PM org.eclipse.californium.core.observe.InMemoryObservationStore remove
FINER: removed observation for token cfb4e4573c35ff28

This response went to onLoad() of CoapHandler with CoapResponse.advanced().getCode() as 4.06. My query is, should it go to onError() instead of onLoad() ?
Or in onLoad(), should we check for response code 4.06 and take some action?

Best Regards,
Chary

boaks:
Please check PR #467

charybr:
Hi Achim,

Thanks!
I tested the fix. I couldnt see any difference in behavior when response is 4.06 i.e it went to onLoad().

Please clarify, if i missed something.

boaks:
[...]

PR #694, opened 2018-07-10
Fix proactive cancel observation, when not using coapclient.

boaks:

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java
@@ -113,9 +113,8 @@ public abstract class BaseMatcher implements Matcher {
 	 *            jobs with a striped executor.
 	 * @throws NullPointerException if one of the parameters is {@code null}.
 	 */
-	public BaseMatcher( NetworkConfig config,  NotificationListener notificationListener,
-			 TokenGenerator tokenGenerator,  ObservationStore observationStore,
-			 MessageExchangeStore exchangeStore, Executor executor) {
+	public BaseMatcher(NetworkConfig config, NotificationListener notificationListener, TokenGenerator tokenGenerator,
+			ObservationStore observationStore, MessageExchangeStore exchangeStore, Executor executor) {
 		if (config == null) {
 			throw new NullPointerException("Config must not be null");
 		} else if (notificationListener == null) {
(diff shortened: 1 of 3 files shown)


sbernard31:
Could you explain the expected behavior with this fix ?

My tests show that :
1. Notification during ongoing proactive request is still RST.
2. Now the response of the cancel response is also visible via notification listener, I don't think this is a good behavior.
3. Observe relation will be canceled (removed()) at client side for any response which does not contains the observe option. So even if the response of the cancel request is something like METHOD_NOT_ALLOWED or NOT_IMPLEMENTED. Depending on how you interpret the specification this could be acceptable.

boaks:
1. sure, I still see no argument to change that.

2. why? The behaviour **before** this PR would be, to forward all responses (except the first), even those responses which indicates, that the observation is removed by the coap-server. See commit 
d9a993905d1ff4c24dc1bef8ad26bcf581cc202f and issue #465.

3.  "METHOD_NOT_ALLOWED or NOT_IMPLEMENTED" smells for me like "chaos". I'm not sure, why a valid "cancel observation request" should be responded with that response code. And sure, if the "token" is used for chaos, then unexpected things will happen.

boaks:
Just to mention:
Feel free to clean-up the "in between" stuff and provide a clean (and so recommendable) low-level API for the "new" observe stuff (the 2 years ago stuff). If you do so, and you want to provide a alternative implementation for a"proactive cancel" which fulfils more your ideas around the functionality, feel also free to do so. 

As I wrote in 
https://github.com/eclipse/californium/pull/692#discussion_r201319616

> So at least, this seems for me to be not recommended.     
 
I would not recommend to use observe without the `CoapClient`, and therefore, I do not insist in this PR. I provided it more or less, because currently the "low level API" adds the observe to the observation store and so I thought, it should also remove it.

boaks:
@sbernard31 
Do you work on a better solution addressing your 3 topics?

Otherwise, though in my opinion the 3 topics are "as intended by me", do you prefer not to merge this PR even without an alternative solution?

sbernard31:
I need to think a bit more about this, so I would prefer to not merge this for now. (I think this is not so important for the M11 too as coapclient works as expected)

boaks:
OK.

sbernard31:
I sum up my thought here.

My first intuition was CoAP endpoint client was a low level API more complex but flexible to use. (meaning you can do some strange things with it)
In this spirit, I saw : 
 - 1 API to remove the observation at server side was `endpoint.cancelObservation(token)`
 - 1 API to sending a cancel request which only impact relation at server side.
That's why I expected naively.

If this behavior was straight forward to implement I will go for it.
Now we see that implement this would require some corner case in the matcher.
The question is : "Does it worth it ?"

I think most of the time you want to first remove Observation from store (`endpoint.cancelObservation`) and then eventually send a proactive cancel request. So in that case all this details does not really matter and so I would not make the matcher more complex for that.

So,
- If you think we should make this corner case behavior more consistent, I could take time to try to change it in the way I describe above.
- If you think this a corner case we don't really need to support, I would prefer to not integrate this PR as I'm not sure this bring a really more consistent behavior.
- If you really prefer to integrate this PR, go for it, as I said before I do not see a use case where you don't cancel the observation before to send the proactive cancel request anyway.

boaks:
No, as I wrote, I'm not insist in this PR. But then, please cleanup PR #692 to not contain tests, which just show, that this is not fixed :-).

sbernard31:
> But then, please cleanup PR #692 to not contain tests, which just show, that this is not fixed :-).

I will do it ^^

PR #697, opened 2018-07-12
Fix test race condition in async client test

boaks:
Also cleanup RepeatingTestRunner (used to analyse the sporadic failing test).

### DIFF
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/ClientAsynchronousTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/ClientAsynchronousTest.java
@@ -20,6 +20,8 @@
  *                                      wait cycles
  *    Achim Kraus (Bosch Software Innovations GmbH) - use CoapNetworkRule for
  *                                                    setup of test-network
+ *    Achim Kraus (Bosch Software Innovations GmbH) - fix race condition with
+ *                                                    reordered notifications
  ******************************************************************************/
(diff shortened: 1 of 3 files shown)


californium-core/src/test/java/org/eclipse/californium/core/test/ClientAsynchronousTest.java
```diff
@@ -135,40 +140,51 @@ public void testAsyncPostUpdatesResource() throws Exception {
...
 					}
 				}
 			}
 		});
-		assertTrue(latch.await(1500, TimeUnit.MILLISECONDS));
+
+		assertAwait("response missing", barrier, 2000, TimeUnit.MILLISECONDS);
+
 		System.err.println("changed 1");
+		resource.setContent(CONTENT_1 + " - 1");
```
sbernard31: That's was already like this before but why using System.err instead of System.out ?

boaks: I don't know :-). 
If it's intended to be corrected, I would go for the logger ...
But in the very first place, I want to get rid of the race condition.

boaks:
> What was exactly the issue ? some notifications was discarded because of order ?

Yes, and counting the notifications will fail, if a notification is dropped caused by reordering.

sbernard31:
So you wait between each notifications, thx to CyclicBarrier to avoid this.

LGTM.

PR #698, opened 2018-07-12
Add IntendedTestException to omit stacktrace.

boaks:
Reduce information overflow in logs during tests.


sbernard31:
OK why not.

Issue #702, opened 2018-07-18
Message cleared silently when shutdown endpoint.


rkimsb2:
on destroy method of UDPConnector, it just clear outgoing message queue.

> outgoing.clear();

Could we trigger onError method of all RawData on this time?

I think the queue does not contains so many messages.

sbernard31:
I suppose this could make sense :thinking:

sophokles73:
The only potential problem I see is that Scandium has no clue about what is happening in the *onError* handler and thus, how long this will take. In the worst case, Scandium needs to wait forever. So I would suggest to also set a maximum time that Scandium allows for the *onError* handler to run before going on to the next pending outbound message. WDYT?

boaks:
Currently, californium doesn't use such the time limit mechanism for callbacks at all. 
We only document some constraints as in `MessageCallback.onContextEstablished`.
If someone "blocks" in callbacks, he will block californium.
I don't think,changing that makes too much sense.

sbernard31:
I agree with @boaks

rkimsb2:
> If someone "blocks" in callbacks, he will block californium.

This means the user who uses the callback has to know what they are doing.

rkimsb2:
And there are a few ways to do this.

> 1. give a graceful options to destroy(shutdown)
> 2. give a property options to triggering onError method.
> 3. just triggers onError method.

and maybe another way is exists.

IMPOV, 1 is so heavy. it should be changed all other types of connectors and even the destroy method of endpoint interface should do so.

And the others, I don't have insight what should we choose.

boaks:
> This means the user who uses the callback has to know what they are doing. Right?

Yes. 
As far as I can see, it would be possible to execute the `onSent()` and `onError(Throwable error)` in the protocol-stack-threading-context, but the `onContextEstablished(EndpointContext context)` MUST be executed in the connector sender thread context (it must be processed completely before the message is actually sent). So delegate the onError callback into the protocol-stack-threading-context would prevent the connector from blocking on shutdown. But I'm not sure, if this really helps.

> And if it right, it seems we could trigger onError with exception of something appropriate type.

Sure, just make a proposal. By the way, do you have a [ECA](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Contributor_Agreement) ? 
if so, you may prepare a PR for the `onError` calls along with your preferred exception :-).
If you want to do that, please obey 
[...]

PR #706, opened 2018-07-25
Trigger onError method of message callback when the connector is shut…

rkimsb2:
…down.

Aimed issue #702

Signed-off-by: Rokwoon Kim <k862390@gmail.com>

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java
@@ -33,6 +33,7 @@
 package org.eclipse.californium.elements;
 
 import java.io.IOException;
+import java.io.InterruptedIOException;
 import java.net.DatagramPacket;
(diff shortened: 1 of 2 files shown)


boaks:
FMPOV, the `void send(RawData msg)` should also call the `onError`, when running is `false`.

rkimsb2:
To @boaks 

I changed this PR like this. Could you check this?

And.. I consider the variable running should be AtomicBoolean or not. And I think this doesn't have to be changed. Because of it's calling timing.

rkimsb2:
Yup. I changed the message. :)

boaks:
Thanks! 
If no objection, I would like to merge this on Monday.

rkimsb2:
> Should we do try to have the same behavior for DTLSConnector ?

I changed void send method of DtlsConnector like this.

And about void sendMessage(RawData, DTLSSession) method, it seems already doing handles of failure.

PR #708, opened 2018-07-30
Fix double incrementing of pending outgoing message downcounter.

boaks:
The runnable incremented the downcounter at the start and finally, and
so broke the pending outgoing message limitation.


sbernard31:
LGTM

PR #710, opened 2018-08-03
Fix stop

boaks:
Some failure in continues running tests showed the NullPointerException.
Therefore the tests and fix are added.


boaks:
My test systems shows some more issues, so I improved the fix.

Issue #712, opened 2018-08-06
2. DTLS session resumption fails, if session cache is used.


boaks:
2.0.x:
If a DTLS session is resumed twice, that fails with

> Client [...] tries to resume non-existing session [ID=...], performing full handshake instead ... 

That's caused by a successful resumption calling `terminateConnection`, which ends up in `InMemoryConnectionStore.remove`. There, the old connection is removed, but also the session id from the `sessionCache`. 
Therefore, the next resumption will fail to validate the session id in the session cache.

boaks:
Fixed with PR #713

PR #713, opened 2018-08-07
Test stop dtls

boaks:
Add tests for start and stop the dtls connector.
Fix issue #702 and #712 

### DIFF
--- a/element-connector/src/main/java/eu/javaspecialists/tjsn/concurrency/stripedexecutor/StripedExecutorService.java
+++ b/element-connector/src/main/java/eu/javaspecialists/tjsn/concurrency/stripedexecutor/StripedExecutorService.java
@@ -290,10 +290,21 @@ public class StripedExecutorService extends AbstractExecutorService {
         try {
             shutdown();
             List<Runnable> result = new ArrayList<>();
-            for (SerialExecutor ser_ex : executors.values()) {
-                ser_ex.tasks.drainTo(result);
+            for (Runnable job : executor.shutdownNow()) {
+                if (job instanceof SerialJob) {
+                    // unwrap serial jobs 
+                    SerialJob serialJob = (SerialJob)job; 
+                    result.add(serialJob.job);
+                    serialJob.executor.drainTo(result);
+                    removeEmptySerialExecutor(serialJob.executor.stripe, serialJob.executor);
+                }
+                else {
+                    result.add(job);
+                }
             }
-            result.addAll(executor.shutdownNow());
+            for (SerialExecutor ser_ex : executors.values()) {
+                ser_ex.drainTo(result);
+            }
             return result;
         } finally {
             lock.unlock();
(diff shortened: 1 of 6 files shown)


boaks:
I work on the renew of this PR ;-(.

Double synchronization in `stop`, and I understand the striped `shutdown` more now, and so I work on a better fix. 
I also changed the ConnectionStore remove from "using the parameter type to remove the session from the session cache or not". to using an additional method with an explicit parameter to remove it from the session cache. 

> Double synchronization in stop

That's also somehow the result of changing the same stuff over too long time again and again ... it makes blind. (`stop` was also adapted in PR #699).

@sbernard31
Should we make a next milestone release without PR #699?  
Or would it be OK to include PR #699 soon, before the next milestone?

boaks:
New version:
- cleaner StripedExecutorService fix.
- cleaner ResumptionSupportingConnectionStore fix/API
- more rare exceptions fixed.

sbernard31:
> cleaner StripedExecutorService fix.

I'm not sure to understand the fix,  but If you think this is an improvement you could consider to contribute it to https://github.com/kabutz/javaspecialists/tree/master/src/main/java/eu/javaspecialists/tjsn/concurrency/stripedexecutor
There is also some tests in this repo, maybe you could check your modification with it : https://github.com/kabutz/javaspecialists/tree/master/src/test/java/eu/javaspecialists/tjsn/concurrency/stripedexecutor

boaks:
My plan is to remove the `StripedExecutorService`, this fix should only help temporary to overcome the error occurred when executing the jobs returned by `shutdownNow`.

I run that test

https://github.com/kabutz/javaspecialists/tree/master/src/test/java/eu/javaspecialists/tjsn/concurrency/stripedexecutor

successfully. These test don't execute the returned jobs, therefore they don't show the defect.
The wrapper jobs created in `SerialExecutor.execute()` removes that executor, when its serial job queue `tasks` is empty. In `shutdownNow` the `tasks` are taken, therefore each wrapper job now tries to remove the `SerialExecutor`, but only the first could do this successful.

boaks:
Left notice in source project.
https://github.com/kabutz/javaspecialists/issues/2

sbernard31:
> My plan is to remove the StripedExecutorService

what do you plan to use instead ?

boaks:
I've tested a prototype, where the SerialExecutor is the Exchange and the serial tasks queue is implemented in the Exchange. This obsoletes the stipes map and the more frequently temporary serial tasks queues. If I can spent more time in it, I would generalize it and use it for scandium also. But I guess, is quite some time in the future :-). 

Therefore the fix, but I also don't want to spend too much time in it.

boaks:
I will merge this latest on Monday. 
I don't see that benefit in waiting for the feedback for
https://github.com/kabutz/javaspecialists/issues/2
especially, because this PR overlaps #699.

If we got other feedback from the `StripedExecutorService` origin, we can update it again :-)

PR #714, opened 2018-08-09
Lru no update on read

boaks:
Update LRU last-access time conditionally. 
Use that to update DTLS session timestamp only on valid MAC.

Additional changes and fixes:
Change values() to return a Collection instead of an Iterator. Enables
use in for-loops. Use nanoseconds for threshold to speedup tests.
Fixes ConcurrentModificationException, when eviction occurs during
iteration over entries.


sbernard31:
LGTM

Issue #716, opened 2018-08-11
Make it possible to register SessionListener instances in a custom subclass of DTLSConnector


lhotari:
Currently it's not possible to make use of the `org.eclipse.californium.scandium.dtls.SessionListener` interface for custom purposes. 

My usecase is about having custom logging (to a custom logging solution) when a DTLS handshake fails. 
The `SessionListener` has `handshakeFailed` callback which suits my usecase. The problem is that DTLSConnector doesn't have a way to register custom `SessionListener` instances to new Handshaker instances. 

I'd be happy with a solution that requires subclassing DTLSConnector. For example adding this kind of method would be helpful:
```
   protected void handshakeInitialized(final Handshaker handshaker) {

   }
```
 This method would be called where there is currently calls to `addSessionCacheSynchronization(handshaker)`. I guess the code could be refactored to take care of both, adding a new method `initializeHandshake` which calls `addSessionCacheSynchronization` and `handshakeInitialized` and replacing the calls to `addSessionCacheSynchronization` by calls to `initializeHandshake`. Maybe this proposal isn't useful and there are better ways. I'm just trying to lower the barrier for getting the change done so that it would be possible to register custom SessionListener instances...

boaks:
Please, check if the merged PR fits your requirements and close this issue, if so.

lhotari:
Thanks @boaks , good work. I'm happy with the changes.

boaks:
@lhotari 

That's long time ago. I'm not sure, if you still use it.
If so, could you check, if PR #1869 works for you as well? 
Then I would deprecate to old way (`onInitializeHandshaker`) and remove it in  4.0 :-).

lhotari:
> @lhotari 
> 
> That's long time ago. I'm not sure, if you still use it.
> If so, could you check, if PR #1869 works for you as well? 
> Then I would deprecate to old way (`onInitializeHandshaker`) and remove it in  4.0 :-).

#1869 looks great. Thank you @boaks .

Issue #717, opened 2018-08-11
Make it possible to provide custom `ResumptionSupportingConnectionStore` in DTLSConnector


lhotari:
Currently DTLSConnector can only use the `InMemoryConnectionStore` although an interface `ResumptionSupportingConnectionStore` exists. 

The constructor `DTLSConnector(final DtlsConnectorConfig configuration, final ResumptionSupportingConnectionStore connectionStore)` is using default access so a custom subclass cannot call it (it would have to be protected or public in this case). Besides this issue there is an assumption in that construct that `ResumptionSupportingConnectionStore` also implements `SessionListener`. Maybe `ResumptionSupportingConnectionStore` should extend `SessionListener` to make that explicit. 

I'd be happy with a protected constructor that would require subclassing DTLSConnector in order to use a custom `ResumptionSupportingConnectionStore` implementation.

boaks:
Change 

```
DTLSConnector(final DtlsConnectorConfig configuration, final ResumptionSupportingConnectionStore connectionStore)
```

from `package` scope to `protected`?
You aware of the API change of PR #713 (added removes variants)?
We will see, if someone had a intention to use that scope, but principally it would be OK for me to change the scope.

lhotari:
yes `protected` would be ok so that DTLSConnector subclasses could call it.

Besides that it would also be good to resolve the current assumption that connectionStore implements `SessionListener` ([in DTLSConnector constructor](https://github.com/eclipse/californium/blob/23ceec42/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java#L273)). Do you have a solution for that?

API changes in PR #713 look fine to me.

boaks:
Please, check if the merged PR fits your requirements and close this issue, if so.

lhotari:
thanks @boaks , great work. This fits my requirements.

PR #720, opened 2018-08-16
Fix issues #716 and #717.

boaks:
Change scopes to protected to support subclass specific implementations.


sbernard31:
LGTM, but @lhotari should have a look.

boaks:
FMPOV, a small redesign, which uses a SessionCache explicitly as parameter and the add sessionlistner only if a session cache is used, should make it easier to use ResumptionSupportingConnectionStore implementations, which don't support the session cache feature.

boaks:
Please, check it again.

boaks:
@lhotari

Does the redesign of the connection store and session cache work for you?

lhotari:
Thank you @boaks . The change was also good for me. I'm sorry I missed your request for feedback.

PR #721, opened 2018-08-16
Fix NullPointerException, if ccs is processed before the SERVER_HELLO.

boaks:
Move expectChangeCipherSpecMessage after receiving SERVER_HELLO.


sbernard31:
NPE because serverRandom is not set, if CCS is processed before SERVER_HELLO ?

If I'm right : LGTM

boaks:
Yes.

PR #726, opened 2018-08-21
Add more logging to unit tests and connection store.

boaks:
Adjust start/stop test to do more logging.


sbernard31:
Hard to me to not talk again about dump for testing purpose.  
lips sealed

No objection.

boaks:
`InMemoryConnectionStore` is `final`. If preferred, I can remove the final an extend `InMemoryConnectionStore`.

sbernard31:
In a general way, I think final keyword is useful mainly for attribute (and immutability)
For a class, I think this is almost alway a bad idea.

About dump, don't wast your time on this detail, if you think this doesn't worth it. ;)

You can merge this PR. :)

PR #732, opened 2018-08-27
Report expired certificates in unit tests.

boaks:
Help to overcome unit test failures from expired certificate.


sbernard31:
no objection

PR #734, opened 2018-08-27
Fix tcp observes

boaks:
Fix several tcp issues.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
@@ -35,6 +35,8 @@
  *                                                    Add exchange dump to status.
  *    Achim Kraus (Bosch Software Innovations GmbH) - check for modified current requests
  *                                                    or responses.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - move retransmitResponse to
+ *                                                    CoapEndpoint to support tcp.
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
@@ -362,7 +364,6 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 			throw new IllegalArgumentException("exchange does not contain a response");
 		} else {
 			Response currentResponse = exchange.getCurrentResponse();
-			exchange.retransmitResponse();
 			if (registerWithMessageId(exchange, currentResponse) > Message.NONE) {
 				if (exchange.getCurrentResponse() != currentResponse) {
 					throw new ConcurrentModificationException("Current response modified!");
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -55,6 +55,8 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - add coap-stack-factory
  *    Achim Kraus (Bosch Software Innovations GmbH) - use checkMID to support
  *                                                    rejection of previous notifications
+ *    Achim Kraus (Bosch Software Innovations GmbH) - reject messages with MID only
+ *                                                    (therefore tcp messages are not rejected)
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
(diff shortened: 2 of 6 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -362,7 +364,6 @@ public boolean registerOutboundResponse(final Exchange exchange) {
 			throw new IllegalArgumentException("exchange does not contain a response");
 		} else {
 			Response currentResponse = exchange.getCurrentResponse();
-			exchange.retransmitResponse();
```
boaks: That's moved to `CoapEndpoint.sendResponse`. 
Potentially changes the UDP processing of notifies on the coap-server (LWM2M device) also. 
All test are passing, so I think this change doesn't break the UDP stuff.

sbernard31: I'm not sure to understand how works the exchange.restransmitResponse() function.

But it seems that before we do that for all response and now only for request with observe option, is it intended ?

boaks:
In this cases, it's a "received MID". I f no MID is received, it should not be rejected, because rejection is based on the MID. So it's either malformed or tcp, but both should not be rejected.

boaks:
So, is it OK to merge this PR?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: For all responses, which are registered as outbound response, so all CON and notifies.
This PR decouples the retransmitResponse for the registerOutboundResponse and use the observe nature of the request to trigger the retransmitResponse.

sbernard31: I'm sorry but I didn't get the point at all :'(

boaks: > I'm not sure to understand how works the exchange.restransmitResponse() function.

That hasn't changed. The idea was to keep control about the state changes in the exchange. Before I introduced it, a lot of exchange stuff was called several times from several places, which doesn't made it easy to fix the "leaks".

> But it seems that before we do that for all response

For all responses, which are registered as outbound response, so all CON and notifies. See the callers of that `registerOutboundResponse`.
(And for sending responses to duplicate request, but his isn't changed).
 
> and now only for request with observe option, is it intended ?
  
Yes. Usual exchanges only send one response by request. Only observes send more.

boaks: So, do you ask, for the function of the `exchange.restransmitResponse()` or for the changes her in this PR according the use of it?

sbernard31: That was a generic question about what is the purpose of this function (no specific of this PR)

boaks: OK, so does the short hint above help?

sbernard31: Nope :'(
I read `exchange.restransmitResponse()` and this do almost nothing.

boaks:
```
	/**
	 * Prepare exchange for retransmit a response.
	 * 
	 * @throws IllegalStateException if exchange is not a REMOTE exchange.
	 * @throws ConcurrentModificationException, if not executed within the
	 *             {@link StripedExchangeJob}.
	 */
	public void retransmitResponse() {
		assertOwner();
		if (origin == Origin.REMOTE) {
			caller = null;
			complete.set(false);
		} else {
			throw new IllegalStateException(this + " retransmit on local exchange not allowed!");
		}
	}
```

So usually, for remote exchanges, only one response is assumed. If more responses are intended to be sent, call this. If not called, but more than one reponse is sent, then

```
	private int registerWithMessageId(final Exchange exchange, final Message message) {
		enableStatus = true;
		exchange.assertIncomplete(message);
```
will suppress this unintended re-response. The stricter status monitoring helps to make the house keeping proper.

boaks: It seems, we discuss more about the PR, which introduces this exchange-state-monitoring, then about this PR. 
The changes in this PR are quite simple, as explained above. Only deduplication and observes requires to send multiple responses.

sbernard31: > It seems, we discuss more about the PR, which introduces this exchange-state-monitoring, then about this PR.

I agree, bu I was not able to understand without that.

And to be honest, I sill don't understand this changes. Anyway go to merge it.

PR #739, opened 2018-08-29
Add client session cache.

boaks:
The client side requires to store the session by the peer's address also
in order to access the right session when starting the communication. 


sbernard31:
LGTM

PR #740, opened 2018-08-29
Preserve session creation time.

boaks:
Adjust time on setting master-secret.


sbernard31:
Could you explain the purpose of this ? :)

boaks:
Enable custom implementation, which use the session create time to "expire" session tickets or sessions.
And I want to display the session creation time, when I use the client session cache extension on my clients.

sbernard31:
Ok, Why creationTime should be adjusted on setting master-secret  ?

boaks:
Because, if a client tries the resume a session, but ends up in a full handshake, I want to update the time. FMPOV the security level is based on timespan of the master-secret.

sbernard31:
OK, it makes sense !
It can be merged.

PR #745, opened 2018-09-07
Add on connect.

boaks:
Intended to adjust a request timeout, when a connection is triggered.


boaks:
If we could agree on PR #746, I would like to add the "retransmitted flight" with their number to the onConnect callback (onConnect(int flight)).

boaks:
"retransmitted flight" added.

sbernard31:
Could you explain a bit the idea ?

I see that you add a new onConnect event.
The javadoc says: 
```
right before a connector establish a connection. 
Not called, if the connection is already established or the connector doesn't require to establish a connection.
```
What is the difference with onContextEstablished ? (I still see this event as a kind of connect event)

About DTLSconnector, I believe onConnect is called each time we retransmist a flight, I'm didn't see the idea ?

boaks:
OnContextEstablished is "always called", so it's no indicator for relaxed timing.
onConnect is only called,if the connector trigger a "connect" which takes some extra time. Therefore I added also the flight retransmissions for dtls, because the indicate also, that some extra time is required.
Currently I improve additionally the android app. There it's used to collect information about occurred retransmissions (flight or CON messages).

boaks:
> About DTLSconnector, I believe onConnect is called each time we retransmist a flight, I'm didn't see the idea ?

The intention was, that it is called, when a client handshake starts and when a flight is retransmitted. I'm not sure, why you "believe", but if you provide logs,which shows, that I'm wrong, I will consider them.

boaks:
I basically chose "onConnect", because it seems to be in line with the other "on..." of the `MessageObserver`.

sbernard31:
> maybe we could start by rename `onConnect` by `onConnecting`

this is in line with the other "on..." ?

boaks:
The others don't use the "...ing". I guess, it's better to add a second callback for the flight retransmission.
`onDtlsFlightRetransmission(int)`.

sbernard31:
> I guess, it's better to add a second callback for the flight retransmission.
onDtlsFlightRetransmission(int).

I feel it's better too.

> The others don't use the "...ing"

Some are using past participle (onSent, onConnectionEstablished, ...)
Some are using base form (onComplete, onSendError, ..)
Some are using noun (onRetransmission, onAcknowledgement, ...)
I don't see any homogeneous form, so I would not be disturb to add some presentParticiple. (...ing)

What I'm seeing  is  "past participle" or "base form" to express an event already done and "presentParticiple" to express something starting or ongoing.

So, even with onDtlsFlightRetransmission, I will go with onConnecting (if this means handshake started)

--------------------------

In a more general way, I'm a bit disturb that we reference DTLS concept (retransmission and flight) at MessageObserver level but I can not see a better way...

boaks:
OK, I will rename and add it.

sbernard31:
LGTM

PR #747, opened 2018-09-09
Reset master secret, when session resumption is refused.

boaks:
When session resumption is refused, a new session id and master secret
is generated. 
The `DTLSSession.setMasterSecret` refuses to overwrite a master secret set for resumption by the master secret generated by the full handshake.  


sbernard31:
LGTM

PR #749, opened 2018-09-12
Add preSharedKeyIdentity to the ServerHandshaker.

boaks:
Support creating statistics.


boaks:
Alternatively, the already available `clientKeyExchange` could be exposed public.
Then check for specific subclass `PSKClientKeyExchange` and read the identifier there.

@sbernard31 
If you can give us your feedback of your preferred variant, 
FMPOV, we could finalize the work for the 2.0.0-M12 and release that next week.

sbernard31:
I have no strong opinion about that.
Maybe the `clientKeyExchange` way is more flexible has we can access to more data than just the PSKIdentity ? But this is not a strong argument so I'm ok with the current implementation too.

Issue #748, opened 2018-09-10
X509v3 keyusage mandatory ?


sbernard31:
Currently in Scandium, `X509v3 keyusage` is mandatory for some type of certificate.
I mean if the certificate does not include this extension with the `digitalSignature` value there is no way to do an handshake.

(see [CertificateRequest#isSupportedKeyType](https://github.com/eclipse/californium/blob/2.0.x/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java#L437))

I don't know if this is the right behavior ? 
I can not find any clear answer but I would say that if the X509v3 usage is present we should check that digitalSignature is present but if the extension is not present we could accept the certificate.

The [Internet X.509 Public Key Infrastructure Certificate and Certificate Revocation List (CRL) Profile](https://tools.ietf.org/html/rfc5280#section-4.2.1.3) says :
_"The key usage extension defines the purpose (e.g., encipherment,
   signature, certificate signing) of the key contained in the
   certificate.The usage restriction might be employed when a key that
   could be used for more than one operation is to be restricted..."_

sbernard31:
To be totally honest, I face this "issue" creating a certificate with simple openssl command.
I also see that certificates (which was not created by me) in leshan-integration-test don't have any `keyusage` extension.

So this makes me search if this is normal that scandium does not accept this kid of certificate.

boaks:
These extensions are available for openssl

https://access.redhat.com/solutions/28965

But, if you prefer to investigate, if they are really proper used, don't hesitate to do so.
May be the "extended key usage" should be also considered, and/or a method is added to the `CertificateVerifier` which checks the "meetsSigningRequirements".

sbernard31:
Thx @boaks, I already found a way to add this kind of extension using openssl. :)
I just explained how I came to this question.

Searching I found this : https://stackoverflow.com/questions/15353595/is-it-mandatory-to-have-keyusage-in-certificate-if-using-jdk7/18071857#18071857
Which seems to go in a way that this should be present but not mandatory.

I thing checking only if the extension is present could make sense. Especially in case where you are using [`domain-issued certificate`](https://tools.ietf.org/html/rfc6698#section-2.1.1) which is [the only supported case in LWM2M 1.0.](https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/187)

But clearly  this question is not a critical or urgent one.

sbernard31:
[...]

PR #753, opened 2018-09-14
make X509v3 keyusage optional

sbernard31:
See #748.


boaks:
Should "C=CA,L=Ottawa,O=Eclipse IoT,OU=Californium,CN=cf-nosigning" really be stored in the "trusts"? 
I don't feel too comfortable to trust it. I would therefore prefer to have it in the keystore.

boaks:
Please mark it, if you want to have it in 2.0.0-M12

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
```diff
@@ -444,8 +444,8 @@ boolean isSupportedKeyType(X509Certificate cert) {
 
 		for (ClientCertificateType type : certificateTypes) {
 			boolean isCompatibleType = type.isCompatibleWithKeyAlgorithm(cert.getPublicKey().getAlgorithm());
-			boolean meetsSigningRequirements = !type.requiresSigningCapability() ||
-					(type.requiresSigningCapability() && cert.getKeyUsage() != null && cert.getKeyUsage()[0]);
+			boolean meetsSigningRequirements = !type.requiresSigningCapability()
+					|| (type.requiresSigningCapability() && (cert.getKeyUsage() == null || cert.getKeyUsage()[0]));
```
boaks: !a || (a && b) := !a || b

Could this be reduced to the logic equivalent?

boaks: And please add a comment, that it's appreciated, if someone can provide details about the usage of the "key usage" and "extended key usage" related to a "certificate request".

sbernard31:
> I don't feel too comfortable to trust it. I would therefore prefer to have it in the keystore.

Ok I will do that.

> please add a comment, that it's appreciated, if someone can provide details about the usage of the "key usage" and "extended key usage" related to a "certificate request".

Good idea, I will do that too.

> Could this be reduced to the logic equivalent?

Exact.

sbernard31:
Better now ?

boaks:
LGTM

PR #754, opened 2018-09-14
Add clone DtlsConfig and a way to create Builder from existing config

sbernard31:
The use case is for Leshan client.
User can give 1 DTLSconfig and I need to create a new config from this one each time I create a new endpoint.(This is because we have 1 endpoint by server)

[here](https://github.com/eclipse/leshan/blob/master/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/impl/CaliforniumEndpointsManager.java#L66) is the way I'm currently doing/using it.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -421,6 +421,38 @@ public final class DtlsConnectorConfig {
 	}
 
 	/**
+	 * @return a copy of this configuration
+	 */
+	@Override
+	protected Object clone() {
+		DtlsConnectorConfig cloned = new DtlsConnectorConfig();
+		cloned.address = address;
+		cloned.trustStore = trustStore;
+		cloned.certificateVerifier = certificateVerifier;
+		cloned.earlyStopRetransmission = earlyStopRetransmission;
+		cloned.enableReuseAddress = enableReuseAddress;
+		cloned.maxFragmentLengthCode = maxFragmentLengthCode;
+		cloned.retransmissionTimeout = retransmissionTimeout;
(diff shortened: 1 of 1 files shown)


boaks:
Please mark it, if you want to have it in 2.0.0-M12

sbernard31:
Ok I will do that. Each time I think a feature is "urgent" I will use the github milestones feature.

Currently all the issue/PR I opened recently about X509/DTLS are not critical or urgent. So I we integrate it for the 2.0.0-M12 it's cool, if we don't this is not an issue.

boaks:
2.0.0-M12 is for you :-), afterwards I would like to introduce the risky threading stuff ;-).

boaks:
I'm somehow "in between". OK, "clone()" is a standard, but on the other side, we clone the configuration mainly to adapt it afterwards. So, do you plan to use the "clone" also somewhere else than in the `Builder` constructor? for me it doesn't make too much sense, because without the `Builder` the `DtlsConnectorConfig` is "final". So, if it's not planed to be used outside of the Builder, I would appreciate to move the clone-code straight forward into the `Builder` constructor, which then doesn't need a clone with exception in the `DtlsConnectorConfig`.
But, if you feel, the "clone" is a good idea, I have no objection against it.

As written above, I have not preference, if this PR is in or not the 2.0.0-M12. As you want it. let me know, then we either release or merge and release then.

sbernard31:
> for me it doesn't make too much sense, because without the Builder the DtlsConnectorConfig is "final"

I think you're right.
So I changed the PR to not expose the `clone` method anymore.
(If you prefer I move the code directly in the `Builder` constructor, but I feel this make a little more sense like this)

Tell me if it's better ?

If you ok I merge it in 2.0.x to make it available for 2.0.0-M12.

boaks:
LGTM

sbernard31:
Integrated in 2.0.x

PR #756, opened 2018-09-26
Add multiple receiver threads.

boaks:
Move release socket to stop.


sbernard31:
Did you see any improvement about passing from 1  to several "receiver" threads ?

boaks:
Hm, I can't see a difference in the current 2.0x (any more?) !

PR #757, opened 2018-09-27
Add TimeAssume to relax timing failures.

boaks:


boaks:
Currently only applied to limited unit test. Merge to see. if it is effective on the eclipse ci-jenkins.

Issue #668, opened 2018-06-18
Possible MID reuse before lifetime end for token [?], expected MID 65535 but received ?


sbernard31:
We observe some warning in production like this : 
```
 o.e.c.core.network.UdpMatcher [] possible MID reuse before lifetime end for token [XXX], expected MID 65535 but received XXX
```
This catches our attention because the expected MID is **alway** 65535.

I investigated this a bit.
This warning is raised in `UdpMatcher.recieveResponse()`.
This happened when exchange come  from `ObservationStore`.

Why? because we add the observation in the store before we attribute an MID to the request. So we try to serialize a -1 java integer MID and this become 65536 on 16bits.

For now, I'm not sure if this bring to real issue...
But this is a bit confusing and maybe this could lead to unexpected behavior.

The other confusing point is with the `InMemoryObservation` implementation. The request is not "cloned" when we add it, so this bring to a different behavior than a persistent store. (In this case, I think this "issue" is not visible with an `InMemoryObservationStore`.

If we move aside the "always 65535 MID", I'm not sure to understand why we get this warning, I tried to change the code to not reuse the same token with block transfer but with my simulation, I still have this kind of error sporadically.

sbernard31:
> Version/commit?
2.0.0-M10

> I think, it doesn't harm more then producing irritating log messages.
Should we fix this anyway as this could be error prone ?

>One scenario, when this may happen, is the reregistration of an observe.
I think we don't do that. I will try to investigate a bit more.

boaks:
2.0.0-M10, so do you have also a update for issue #429 ?

Are the leaks fixed or still appear?

> Should we fix this anyway as this could be error prone ?

For now I would just not close the issue until I can spent time into check, if it could be changed without to many changes. Or until you check it :-).

boaks:
The log message may also be caused by a ACK after a notify.

sbernard31:
> will NOT work for clustered-observes, though the requestMid will always be NONE.

I just tested on my side. And It works for me with Cf 2.0.0-M12. I mean I still have the issue about MID=65536 instead of the right value. But the notification is correctly received and handled.

As explained above, the -1 MID is serialize then deserialize and become a 65536 MID. So on my side `requestMid` is always 65536 (not NONE).

Did you really face the issue or it is just hypothetic ?

Anyway, I still find this would be better to serialize the good MID.

> should we go to provide the effective request MID with the established context?

Where exactly ?

boaks:
[...]

PR #761, opened 2018-09-30
Fix logging of MID matching.

boaks:
Recreated request from observation stores may not have a valid MID.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -48,6 +48,7 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - change ExchangeObserver
  *                                                    to RemoveHandler
  *                                                    remove "is last", not longer meaningful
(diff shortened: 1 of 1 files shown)


boaks:
Intended for issue #668

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -112,6 +112,7 @@ public void sendRequest(final Exchange exchange) {
 		// for observe request.
 		Request request = exchange.getCurrentRequest();
 		if (request.isObserve() && 0 == exchange.getFailedTransmissionCount()) {
+			exchangeStore.assignMessageId(request);
```
sbernard31: Should we handle the case where `assignMessageId` returns `Message.NONE` ?

boaks:
Check for exhausted MID added.

sbernard31:
:+1:

boaks:
So in some time in the future, we hopefully will see, what caused the warning of #668. :-)

PR #764, opened 2018-10-08
Add act as server only.

boaks:
Inhibit starting handshake.


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1521,6 +1521,9 @@ private void sendMessage(final RawData message) throws HandshakeException {
 		DTLSSession session = connection.getEstablishedSession();
 		SessionTicket ticket = connection.getSessionTicket();
 		if (session == null && ticket == null) {
+			if (config.isServerOnly()) {
+				message.onError(new EndpointMismatchException());
```
sbernard31: Using EndpointMismatchException sounds strange to me.
(If we don't find a more appropriate exception, we should at least add a String explanation to the exception)

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -128,6 +130,9 @@
 	/** does the server require the client to authenticate */
 	private Boolean clientAuthenticationRequired;
 
+	/** does not start handshakes */
+	private Boolean serverOnly;
```
sbernard31: Should we rather go for an enum to anticipate clientOnly ? (Client, Server, Both)

(This make me think again to https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/410 and sry to insist but I really like to know your opinion about that :pray:)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: Using EndpointMismatchException sounds strange to me.
(If we don't find a more appropriate exception, we should at least add a String explanation to the exception)

boaks: Ok, I just token the exception from the tcp part.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: I'm simply not up-to-date with LWM2M,  especially 1.1.

sbernard31: > the client only flag is only used by the builder to verify the configuration. the server only limits the connector function.

I didn't notice that we already have a client only flag. I just think that this could make sense to have the same restriction for client side (I mean limiting connector function). and maybe we could anticipate it.

> I'm simply not up-to-date with LWM2M, especially 1.1.

Just talking about LWM2M 1.0, I know that by the past you didn't like this idea of DTLS exchange role. And I would like to know, if you see a way to handle the failover use case for none queue mode ?

boaks: > by the past you didn't like this idea of DTLS exchange role

Yes, but that is mainly based on the fact, that I don't consider to have stable ip-adresses :-).

sbernard31: Ok, so for stable ip-address, DTLS exchanges roles seems not so bad to you ? To be honest I can not see other way for now ... (except maybe DTLS heartbeat which means more bandwidth consumption ...)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: I will add a meaning full exception or message.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: If a DTLS role exchange works, OK. But in my experience, the CLIENT_HELLO from the server usually mixup the people (and devices) more, than it works :-). Therefore I would prefer to just add the possibility to switch it off.

sbernard31: I totally agree that make is switchable is a good idea.

But the initial remarks was : 

1. Does it makes sense to also have a client only mode ?" (I mean exactly like you implement the serverOnly)
2. If yes, maybe we should anticipate it and use an enum instead of a boolean ? or should we reuse the actual ClientOnly to also limit connector behavior ? (not necessarily in this PR)

sbernard31:
I think you missed the remarks above : 

>There is missing some attributes clonning in clone methods of DtlsConnectorConfig

which is totally impossible to understand. :D

So what I meant was "you forget `serverOnly` attributes in clone methods"

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: AFAIK the embedded device are built without the dtls-server-code to save code memory. But for californium, that is not really worth :-).
So my intention was really just to ensure, that the server doesn't start a handshake. I want to enusre this also for the unit test, which may fail, if the server accidently and unexpected start a handshake for which ever reason.

sbernard31: Ok let's this aside for now, but I still feel that this could make sense to set a DTLS connector in a way it refuse (CLIENT_HELLO) and so act as a client only (At least for testing purpose)

sbernard31:
LGTM

> Ok let's this aside for now, but I still feel that this could make sense to set a DTLS connector in a way it refuse (CLIENT_HELLO) and so act as a client only (At least for testing purpose)

PR #765, opened 2018-10-08
Adapt cookie to use only the binary address instead of the string.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/CookieGenerator.java
@@ -17,9 +17,12 @@
  *                                                    which fails to clone hmac.
  *                                                    Use nanoTime instead of 
  *                                                    currentTimeMillis.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - use binary address instead of
+ *                                                    string
  */
 package org.eclipse.californium.scandium;
 
+import java.net.InetSocketAddress;
 import java.security.GeneralSecurityException;
 import java.security.SecureRandom;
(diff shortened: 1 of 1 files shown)


boaks:
FMPOV, it's just a more direct approach to use the byte[] address and port bytes instead of the string.
I'm not sure, if potentially a previous reverse lookup my mix up the string version, therefore the bytes seem to be more stable. I didn't recognize real trouble with the string, but I would prefer the bytes.

sbernard31:
OK why not, but I didn't get the point of `hmac.update((byte) (port >>> 8));`

boaks:
Hm, the `InetSocketAddress.toString()` contains the port. And assuming devices behind NATs, then the devices may end up in the same ip-address but different ports. If the NAT is somewhere in between, I think the port should be part of the cookie. 
Or what did you have in mind, when you wrote  

> OK why not, but I didn't get the point of hmac.update((byte) (port >>> 8));

sbernard31:
```java
	hmac.update(peer.getAddress().getAddress());
	int port = peer.getPort();
	hmac.update((byte) (port >>> 8));
	hmac.update((byte) port);
```

I would like to know why `hmac.update((byte) (port >>> 8));` AND `hmac.update((byte) port);` ?

boaks:
Mac offers:

```
update(byte input)
update(byte[] input)
update(byte[] input, int offset, int len)
update(ByteBuffer input)
```

Though the port is usually a 16 bit value, I don't see an other way to include a 16 bit port into the hash. Do you have one?

sbernard31:
Sry, I just missed this was just a way to cast the port in 2 bytes ... 
You could merge this one.

PR #767, opened 2018-10-09
Serialize matcher execution.

boaks:
Though the match determines in the receives the exchange, moving the
serial execution into it makes the double checks in the current
implementation obsolete.


californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -225,131 +237,174 @@ public Exchange receiveResponse(final Response response) {
...
-					// too early. Therefore don't return a exchange when the MID
-					// doesn't match.
-					// See issue #275
-					return null;
+			@Override
+			public void run() {
+				boolean checkResponseToken = !exchange.isNotification() || exchange.getRequest() != exchange.getCurrentRequest();
+				if (checkResponseToken && exchangeStore.get(idByToken) != exchange) {
+					LOGGER.debug("ignoring response {}, exchange not longer matching!", response);
+					return;
```
sbernard31: not sure to understand this test ?

boaks: the one part of the check is simple:

`exchangeStore.get(idByToken) != exchange`

Is this exchange still be the on returned for this token? 
Or has this changed in the meantime?

The other part of the check is, when this rule should be applied:

`!exchange.isNotification() || exchange.getRequest() != exchange.getCurrentRequest()`

Check it only, if it is no notification exchange (the check would fail, becaus it's in the observation-store), or if it is a blockwise exchange of the left blocks, which then again, is put into the `exchangeStore`.

I had also a version, which checks the notifies in the observation store. But I'm not sure, if that pays off.

I also think, that I need to extend the benchmark client with more statistic, to ensure, that this doesn't lower the quality (e.g. of blockwise notifies).

sbernard31:
(I have no objection about merging this PR)

Issue #343, opened 2017-06-29
Handling of stale handshakes


boaks:
During the discussion of #335 I was pointed to the current handling of stale handshakes
 
https://github.com/eclipse/californium/issues/335#issuecomment-310660094

FMPOV I agree, that there is no reason to remove a DTLS session from the LRU cache (ConnectionStore), until the storage is required for something else.

But for "incomplete handshakes" I have my doubts, if this would not introduce a vulnerability, because some could fill the ConnectionStore with entries without having credentials, and force so other "authorized" clients to handshake again. I would therefore propose to remove the incomplete handshakes actively earlier.

So any opinions on that?

sbernard31:
> some could fill the ConnectionStore with entries without having credentials, and force so other "authorized" clients to handshake again.

We have the `staleConnectionThreshold` to "limit" this issue. (default value 30 minutes)

> I would therefore propose to remove the incomplete handshakes actively earlier.

What did you have in mind exactly ? When/how do we know we can remove an incomplete handshake ?

boaks:
> We have the staleConnectionThreshold to "limit" this issue. (default value 30 minutes)

I principally understood, that using the same `staleConnectionThreshold` and cleanup only when the cache is full, is a pretty simple implementation

My concern is, that if someone attacks using a lot of incomplete handshakes, the other connections based on successful handshakes, may be "garbage collected". With a shorter threshold for connections related to "ongoing handshakes" and switching to the longer threshold when the handshake gets successful would change the order of removing them and so lower the number of removed "connections based on successful handshakes".

> What did you have in mind exactly ? When/how do we know we can remove an incomplete handshake ?

Define a second, shorter threshold and use that, when a new connection with an ongoing handshake was started.
Switch to the `staleConnectionThreshold` when the handshake get successfully completed.

I would also like to introduce a "active remove of connections with stale handshakes" but this would be more related to "feedback" for a triggering request and so not part of this issue. (So i remove the "actively" from my first post :-) )

sbernard31:
> My concern is, that if someone attacks using a lot of incomplete handshakes, the other connections based on successful handshakes, may be "garbage collected".

Only after `staleConnectionThreshold` seconds. While this timeframe, the connection is "immune". For use case where you are behind a NAT this timeframe should already be very low. (At sierra we set it at 10minutes but I think we can reduce it more than that.) 

[...]

PR #768, opened 2018-10-09
Spoof detection for short resumptions without cookie exchange.

boaks:
As first idea of a dynamic usage of a short resumption without HELLO_VERIFY_REQUEST.
If the number of such pending handshakes exceeds a limit, trigger new
resumption handshakes to be verified by a HELLO_VERIFY_REQUEST.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -228,6 +228,9 @@ public class DTLSConnector implements Connector, RecordLayer {
 	private final LeastRecentlyUsedCache<InetSocketAddress, SerialExecutor> connectionExecutors;
 	private final ResumptionSupportingConnectionStore connectionStore;
 
+	private final int thresholdHandshakesWithoutVerifiedPeer;
+	private final AtomicInteger pendingHandshakesWithoutVerifiedPeer = new AtomicInteger();
+
 	/**
 	 * (Down-)counter for pending outbound messages. Initialized with
 	 * {@link DtlsConnectorConfig#getOutboundMessageBufferSize()}.
@@ -323,8 +326,21 @@ public class DTLSConnector implements Connector, RecordLayer {
 				}
 			};
 			int maxConnections = configuration.getMaxConnections();
+			// calculate absolute threshold from relative.
+			long thresholdInPercent = config.getVerifyPeersOnResumptionThreshold();
+			long threshold = (((long) maxConnections * thresholdInPercent) + 50L) / 100L;
+			if (threshold == 0 && thresholdInPercent > 0) {
+				threshold = 1;
+			}
+			this.thresholdHandshakesWithoutVerifiedPeer = (int) threshold;
+			
+			if (maxConnections < 80) {
+				maxConnections = 100;
+			} else {
+				maxConnections += (maxConnections / 4);
+			}
 			// create cache slightly larger than max connections for new CLIENT_HELLOs
-			this.connectionExecutors = new LeastRecentlyUsedCache<>(maxConnections + (maxConnections / 16), 20);
+			this.connectionExecutors = new LeastRecentlyUsedCache<>(maxConnections, 20);
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1321,7 +1323,13 @@ private boolean isClientInControlOfSourceIpAddress(ClientHello clientHello, Reco
 					if (connection == null || !connection.hasEstablishedSession()) {
 						// only new connections or connections without
 						// established sessions
-						return true;
+						int pending = pendingShortResumptionHandshakes.get();
+						int remainingCapacity = connectionStore.remainingCapacity();
+						if (pending < (remainingCapacity / 8)) {
```
boaks: remaining capacity is wrong, but the basic idea should work :-)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
```diff
@@ -143,6 +143,10 @@ public InMemoryConnectionStore(final int capacity, final long threshold, final S
 
 				@Override
 				public void onEviction(Connection staleConnection) {
+					Handshaker handshaker = staleConnection.getOngoingHandshake();
+					if (handshaker != null) {
+						handshaker.handshakeFailed(new Exception("Connection evicted from store"));
+					}
```
sbernard31: Should this change be integrated in 2.0.x ? (I mean even if we don't integrate this PR?)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: You mean you would like that remaining capacity does not count staled connection ?

sbernard31: You mean you would like that remaining capacity does not count staled connection ?

boaks: Yes (or something similar).
I want to show, how to check pending short resumption and use that number to trigger the HELLO_VERIFY_REQUEST, if that number gets to high. But I'm still investigating. So this just a first idea of such a dynamic defence agains spoofing.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
boaks: I want to execute this in a serialzed executor, not to break this at the very end.
And I would prefer to call this as timeout much earliere, as I already mentioned in #343.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: FMPOV, it would wire up the serial execution, if we pass it into the connection store.
With the move of the flight and the introduction of the timeout, handshakes should not be ongoing, when connections get evicted.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
boaks: FMPOV, it would wire up the serial execution, if we pass it into the connection store.
With the move of the flight and the introduction of the timeout, handshakes should not be ongoing, when connections get evicted.

sbernard31:
Maybe I missed something but I believe that this should not work because of the current behavior of LeastRecentlyUsedCache.remainingCapacity()

``` java
        /**
	 * Gets the number of entries that can be added to this cache without the
	 * need for removing stale entries.
	 * 
	 * @return The number of entries.
	 */
         public final int remainingCapacity()
```

boaks:
With PR #782 stale handshakes will timeout and remove themselves from the store and "give remaining capacity". 

Or do you see an other issue?

sbernard31:
stale handshake will be removed but once InMemoryConnectionStore is full of establishedconnection, your optimization will not work ? even if there is some connection from the store which could be evicted.

Or maybe I missed something ?

boaks:
I'm not sure, how often this will happen. 
And yes, then it would require a additional verify request, before a established connection gets removed. But the handshake will still be possible, as it was before introduction of a resumption without verify request.

boaks:
Hmm, or just make it simpler and compare the "threshold" against "pending resumption handshake without verified peer" directly
(just compare threshold to pendingHandshakesWithoutVerifiedPeer).
It's more a 1. experiment :-).

sbernard31:
About that, just in case you missed Eric's answer : http://ietf.10.n7.nabble.com/HELLO-VERIFY-REQUEST-during-abbreviated-handshake-session-resumption-td574523.html#a575398

boaks:
I didn't miss it. But wasn't it you, that has doubts about the memory resources?
With this protection, we ensure, that these resumption handshakes will not overflow the connection store.
At least, I hope that.

sbernard31:
> With this protection, we ensure, that these resumption handshakes will not overflow the connection store.

:+1:  that makes sense.

PR #782, opened 2018-10-30
Move DTLS flight

boaks:
Check for timeouts in unit test.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSFlight.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSFlight.java
@@ -25,6 +25,9 @@
  *                                                    to stop retransmission when already
  *                                                    hand over to other executor
  *    Achim Kraus (Bosch Software Innovations GmbH) - add dtls flight number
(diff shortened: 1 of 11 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSFlight.java
```diff
@@ -193,59 +230,78 @@ public void setRetransmissionNeeded(boolean needsRetransmission) {
...
 	 * 
-	 * Intended to be called within a synchronized block.
+	 * @return {@code true}, if the first handshake message of the response is
+	 *         received, {@code false}, otherwise.
 	 */
-	private final void cancelCurrentRetransmission() {
-		if (this.retransmitTask != null) {
-			if (!this.retransmitTask.isDone()) {
-				this.retransmitTask.cancel(true);
+	public boolean isReponseStarted() {
```
sbernard31: typo : isResponseStarted

boaks: thanks, fixed.

Issue #752, opened 2018-09-14
Limit certificate type


sbernard31:
Currently when you set a X509 credentials, scandium uses `*_certificate_type` extension to express it support `X.509 (0x00)` or `Raw Public Key  (0x02`.

You can eventually change the order by using `preferRawPublicKeys` in `DtlsConnectorConfig.Builder.setIdentity()`.

I think all of this makes sense.

But still for LWM2M, I would like to limit my client to X.509 and so only send the X.509 `client_certificate_type`.

I currently use a workaround by using a `rpktruststore` which trusts nobody.

Maybe a solution could be to replace `preferRawPublicKeys` by an enum.
(`preferX509`, `preferRPK`, `onlyX509`, `onlyRPK`) ?

boaks:
My only concern is, that since weeks now, I would like to "reformat" the logging in scandium, but I was help up by my own changes :-).  So, I need to wait a little more.

boaks:
Should the (preferX509, preferRPK, onlyX509, onlyRPK) then only affect the client's list or
the server's list as well?
The "onlyRPK" seems to be obsolete, for that case `setIdentity(PrivateKey, PublicKey)` without the `setIdentity(PrivateKey, Certificate[], boolean or enum)`is used.

I also stumbled over:

According https://tools.ietf.org/pdf/rfc7250.pdf#7, page 6, Figure 4, the client's certificate type is a list, the server's a single entry.

Later on page 8,

>    The client_certificate_type and server_certificate_type extensions
>    sent in the client hello each carry a list of supported certificate
>    types, sorted by client preference.  When the client supports only
>    one certificate type, it is a list containing a single element.
> 

Scandiums implementation uses a list of also for the server's certificate type.

I didn't check for erratas nor did I dive deep into it.

sbernard31:
> Should the (preferX509, preferRPK, onlyX509, onlyRPK) then only affect the client's list or
the server's list as well?

I was thinking both client and server.

> The "onlyRPK" seems to be obsolete, for that case setIdentity(PrivateKey, PublicKey) without the setIdentity(PrivateKey, Certificate[], boolean or enum)is used.

`onlyRPK` was for people who provide their pubKey using X509 certificate format but I agree it is maybe a bit weird.

boaks:
The current API and implementation seems to "support always RPK". 
If it should be changed to a selective usage, I would prefer to go for:

```
[...]

PR #789, opened 2018-11-09
Enable configuration to select certificate types.

boaks:
Redesign the configuration of certificate types.
Clean up processing of the certificate type extensions. 
Extend cf-secure demo for DHE_PSK and clean up other examples using the new configuration.


boaks:
Fix for issue #752

boaks:
@sbernard31 

If you can spend time into it, I would propose not to focus too much on single changes. Instead:

- is the configuration OK for you? Especially comments on the validation are welcome.
- is changing from `List<CertificateType>` to `CertificateType[]` OK for you?

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1249,65 +1385,110 @@ public DtlsConnectorConfig build() {
...
+			}
+			if (config.trustCertificateTypes == null) {
+				if (config.trustedRPKs != null) {
+					config.trustCertificateTypes = ArrayUtils.add(config.trustCertificateTypes,
+							CertificateType.RAW_PUBLIC_KEY);
+				}
+				if (config.certificateVerifier != null) {
+					config.trustCertificateTypes = ArrayUtils.add(config.trustCertificateTypes, CertificateType.X_509);
+				}
+			} else {
```
sbernard31: For better readability, I think this could be better to clearly separated default value initialization and consistency check.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1249,65 +1385,110 @@ public DtlsConnectorConfig build() {
...
+				if (ArrayUtils.contains(config.trustCertificateTypes, CertificateType.RAW_PUBLIC_KEY)) {
+					if (config.trustedRPKs == null) {
```
sbernard31: why not trusting All RPK by default like before ?

scandium-core/src/main/java/org/eclipse/californium/scandium/util/ArrayUtils.java
```diff
@@ -0,0 +1,102 @@
...
+	 * Initialize ordered values removing contained item duplicates.
+	 * 
+	 * @param values array of values. Maybe {@code null} for empty array.
+	 * @return array with unique items.
+	 */
+	public static <T> T[] init(T[] values) {
+		if (values != null && values.length > 1) {
+			int current = 1;
+			int index = 1;
+			outer: for (; index < values.length; ++index) {
```
sbernard31: If we keep use of Array, could we found a more readable way ?

boaks:
> I do not understand the benefits of this ? Especially if this imply to add an ArrayUtils class with Tests associated ?

AFAIK, java has only one Collection implementation (`LinkedHashSet`), which is `sorted` and a `set` (means no duplicate entries). Therefore only that would fulfill the required function. Using that would cause also all using APIs to change from List to Set (or LinkedHashSet). For me, this seems to be also not that  straight forward.
 
Changing part of the configuration API to `(Type... list)` makes it easier to use it.
Therefore, starting with that "collection", I tried to keep the arrays[] and put the helper functions in a extra class. For me, that's the directest way to do it. Outside the configuration (or the custom setup code), mainly the 'contains' and iterate is used, and that's very simple for small set's.
But sure, we can replace it with `Set` or `LinkedHashSet`, even, if I don't see the "perfect" match there. Or keep the List, do all checks inside the configuration and return a "unmodifiable" variant.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -422,7 +427,11 @@ public PublicKey getPublicKey() {
 	 * @return the root certificates
 	 */
 	public X509Certificate[] getTrustStore() {
-		return trustStore;
+		if (trustStore == null) {
+			return null;
+		} else {
+			return Arrays.copyOf(trustStore, trustStore.length);
```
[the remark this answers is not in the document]
boaks: In general, sometimes it's cloned, sometimes not. With that I wan't to ensure, that the scope of accidentally changes are limited. Alternatively would be a unmodifiable List.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java

boaks:
Try to use other wording:

I used the `array`, because it's introduce by `(X... list)` and I didn't see a benefit converting that to a `List` or `Set`. On the other side, I don't insist in that, and if you feel better with a (unmodifiable) `List`, then I would go for that.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
sbernard31: My first idea was to propose a builder with maximum default value (and so minimal needed configuration) to be able to have a testable/running application quickly.

But In that case, maybe it's better to force user to ask himself some question about what it trust or not. :thinking:

boaks: The behaviour before is also more based on the history than a common way.
Historical, all RPK's are trusted, but X.509 need special trust. Then it ge's changed into, RPK may be trusted and x.509 may be trusted in general. And both with different defaults (RPK: trust all, x.509: trust none). So I would prefer to change the behaviour to a common model.

sbernard31: Agree for consistency.
Agree for forcing users to explicitly say what they want to trust.

boaks:
So the main question: 
switch (back) to List or keep the array?
I will adjust some other stuff (e.g. ciphersuite) to that decision as well, but in an other PR.

sbernard31:
As,
> Outside the configuration (or the custom setup code), mainly the 'contains' and iterate is used

So, I would go with a (unmodifiable) `List`.

Consistency check about uniqueness could be done easily in `DtlsConnectorConfig`.

> Changing part of the configuration API to (Type... list) makes it easier to use it.

We can keep the API with `Type...` and create a list from it using `Arrays.asList(types)`. It will create an arrayList.

boaks:
OK.

>We can keep the API with Type... and create a list from it using Arrays.asList(types). It will create an arrayList.

That's clear, but e.g. cf-secure  / CredentialsUtil requires some function for array, so that X... could be called.

sbernard31:
I look at `CredentialsUil` quickly and so maybe I  missed something but : 
```java
CertificateType[] types = null;
if (x509 >= 0 && rpk >= 0) {
	if (rpk < x509) {
		types = ArrayUtils.add(types, CertificateType.RAW_PUBLIC_KEY);
		types = ArrayUtils.add(types, CertificateType.X_509);
	} else {
		types = ArrayUtils.add(types, CertificateType.X_509);
		types = ArrayUtils.add(types, CertificateType.RAW_PUBLIC_KEY);
	}
} else if (x509 >= 0) {
	types = ArrayUtils.add(types, CertificateType.X_509);
} else if (rpk >= 0) {
	types = ArrayUtils.add(types, CertificateType.RAW_PUBLIC_KEY);
}
config.setIdentity(serverCredentials.getPrivateKey(), serverCredentials.getCertificateChain(), types);
```

Could be replaced by 
```java
CertificateType[] types = null;
if (x509 >= 0 && rpk >= 0) {
	if (rpk < x509) {
		types = new CertificateType[] {CertificateType.RAW_PUBLIC_KEY,CertificateType.X_509};
	} else {
		types = new CertificateType[] {CertificateType.X_509, CertificateType.RAW_PUBLIC_KEY};
	}
} else if (x509 >= 0) {
	types = new CertificateType[] {CertificateType.X_509};
} else if (rpk >= 0) {
	types = new CertificateType[] {CertificateType.RAW_PUBLIC_KEY};
}
config.setIdentity(serverCredentials.getPrivateKey(), serverCredentials.getCertificateChain(), types);
```
----------------------------------------
Another solution could be to propose 2 API, one with `Type...` another with `List<Type>`. Generally this is what I do when I want to provide more convenience. ( see [LwM2mModel](https://github.com/eclipse/leshan/blob/master/leshan-core/src/main/java/org/eclipse/leshan/core/model/LwM2mModel.java#L36))

boaks:
I update the PR to use a `List`.

boaks:
Though I would like to exchange the `CipherSuite` and CertificateChain to use also a `List` instead of an array, I would prefer to merge this PR soon. Unfortunately I detected also #790, which may also change a lot.

@sbernard31 
If you have time and want to spent it into deep review of this PR, I will wait merging it. But we would also be able to add fixes and changes later. So it would be great if you can give your feedback!
(issue #790 is something I would really like to have fixed soon, my android communication tests are depending on that. I could do a local/temporary System.currentTimeMillis() workaround, but I would prefer to fix it.)

boaks:
> I didn't see any call to java.util.Collections.UnmodifiableList<E>, but maybe I missed it.
> 
> Just in case you don't notice that, Arrays.asList() returns a modifiable fixed size list (meaning "no add" but "set allowed")

That is done in `org.eclipse.californium.scandium.util.ListUtils`.

sbernard31:
I missed it, so I just looked at it and do not understand the idea :confused: 

The idea seems to remove duplicates keeping order ? but that's seems confusing to me ..
If in my list I have :
```
A,B,A,C,A
```
I can find many different output which respect the contract.
```
A,B,C
B,A,C
B,C,A
```
I feel this confusing and I would prefer to just validate there is no duplicate and so let the responsibility to not provide list with duplicate

boaks:
> I feel this confusing and I would prefer to just validate there is no duplicate and so let the responsibility to not provide list with duplicate

OK.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -882,64 +951,174 @@ public Builder setPskStore(PskStore pskStore) {
...
+		 *             chain is <code>null</code>
+		 * @throws IllegalArgumentException if the certificate chain does not
+		 *             contain any certificates, or contains a non-X.509
+		 *             certificate
+		 * @see #setTrustStore(Certificate[])
+		 * @see #setCertificateVerifier(CertificateVerifier)
+		 * @see #setRpkTrustAll()
+		 * @see #setRpkTrustStore(TrustedRpkStore)
 		 */
 		public Builder setIdentity(PrivateKey privateKey, Certificate[] certificateChain,
```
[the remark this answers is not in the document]
boaks: Changed.

scandium-core/src/main/java/org/eclipse/californium/scandium/util/ListUtils.java
```diff
@@ -0,0 +1,53 @@
...
+public class ListUtils {
+
+	/**
+	 * Initialize ordered values, testing for contained item duplicates.
+	 * 
+	 * @param values list of values. Maybe {@code null} for empty list.
+	 * @return unmodifiable list with unique items.
+	 * @throws IllegalArgumentException if duplicate items are contained
+	 */
+	public static <T> List<T> init(List<T> values) {
```
sbernard31: I would not keep the `init` method and rather change it by `hasDuplicate()` or `containsDuplicate()`.

But this is clearly a detail, change it only if it makes sense to you.

PR #791, opened 2018-11-15
Use cipher suite list instead of array

boaks:
Adapt to cipher suite list.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -172,7 +172,7 @@ public final class DtlsConnectorConfig {
 	private X509Certificate[] certChain;
 
 	/** the supported cipher suites in order of preference */
-	private CipherSuite[] supportedCipherSuites;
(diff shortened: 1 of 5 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -885,14 +873,14 @@ public Builder setSupportedCipherSuites(String... cipherSuites) {
 			if (cipherSuites.length == 0) {
```
sbernard31: Maybe we could rewrite it like this. (to avoird duplicate code about checking)
```java
		public Builder setSupportedCipherSuites(String... cipherSuites) {
			if (cipherSuites == null) {
				throw new NullPointerException("Connector must support at least one cipher suite");
			}
			List<CipherSuite> suites = new ArrayList<>(cipherSuites.length);
			for (String suiteName : cipherSuites) {
				CipherSuite suite = CipherSuite.getTypeByName(suiteName);
				if (suite == null) {
					throw new IllegalArgumentException(
							String.format("Cipher suite [%s] is not (yet) supported", suiteName));
				}
				suites.add(suite);
			}
			setSupportedCipherSuites(suites);
			return this;
		}
```

PR #793, opened 2018-11-16
Update file server

boaks:
Update file server and hello world client to support file download without adapting the properties .
 

boaks:
Replaced by PR to master.

PR #794, opened 2018-11-16
Use List for certificates chains instead of array.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
@@ -29,7 +29,6 @@ import java.security.PublicKey;
 import java.security.Signature;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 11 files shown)


boaks:
Last on changing a configuration array to a unmodifiable List.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
```diff
@@ -502,10 +502,10 @@ public SignatureAndHashAlgorithm getSignatureAndHashAlgorithm(X509Certificate[]
 	 * @param cert The certificate chain to test.
 	 * @return {@code true} if all certificates have been signed using one of the supported algorithms.
 	 */
-	boolean isSignedWithSupportedAlgorithm(X509Certificate[] chain) {
+	boolean isSignedWithSupportedAlgorithm(List<X509Certificate> chain) {
 
-		for (int index=0; index < chain.length; ++index) {
-			X509Certificate cert = chain[index];
+		for (int index=0; index < chain.size(); ++index) {
```
sbernard31: for each ?

boaks: Done

Issue #790, opened 2018-11-14
Android: System.nanoTime() stop in deep sleep 


boaks:
Branch 2.0.x:
During test I got aware, that Android stops the `nanoTime` during deep sleeps.
To change that, all calls need to be delegated to a replaceable `nanoTime` provider.

PR #795, opened 2018-11-17
Adjust system nanotime processing.

boaks:


boaks:
Fix issue #790

element-connector/src/main/java/org/eclipse/californium/elements/util/ClockUtil.java
```diff
@@ -0,0 +1,74 @@
...
+
+	/**
+	 * Get expired realtime in nanoseconds.
+	 * 
+	 * If no system specific handler was set before, {@link System#nanoTime()}
+	 * is used.
+	 * 
+	 * @return expired realtime in nanoseconds
+	 */
+	public static long nanoRealtime() {
```
sbernard31: I'm not sure if this is better, but intuitively I would have created a default RealTime implementation which returns `System.nanoTime` instead of doing the check `systemHandler !=  nul`l for each call of nanoRealtime.

boaks: OK.

boaks:
Currently I use it a android app for communication tests. Though it's still underdevelopment and changing too frequently, I haven't published it right now in californium. The issue  in #790 only occurs, if you put the smartphone in sleep mode and then reuse the initialized californium/scandium check. I got aware of that problem using auto resume and waking up the app every 5 minutes, if connectivity is available. Though it's 5 minute real time, but the app slept so it's only about 20s uptime, the auto resume failed. This enables a android app, which want to use the californium/scandium during sleeps.

sbernard31:
:+1:

element-connector/src/main/java/org/eclipse/californium/elements/util/ClockUtil.java
boaks: Done.

PR #803, opened 2018-11-20
Add more flexibility to auto resumption.

boaks:
Introduce none critical context attributes.
Rename inhibitNewConnection in hasCriticalEntries.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/AddressEndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/AddressEndpointContext.java
@@ -112,7 +112,7 @@ public class AddressEndpointContext implements EndpointContext {
 	}
 
 	@Override
-	public boolean inhibitNewConnection() {
+	public boolean hasCriticalEntries() {
 		return false;
 	}
 
--- a/element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
@@ -43,6 +43,13 @@ public class DtlsEndpointContext extends MapBasedEndpointContext {
 	 * the DTLS session.
 	 */
 	public static final String KEY_CIPHER = "DTLS_CIPHER";
+	/**
+	 * The name of the attribute that contains a auto session resumption timeout
+	 * in milliseconds. {@code "0"}, force a session resumption, {@code ""},
+	 * disable auto session resumption. None critical attribute, not considered
+	 * for matching.
+	 */
+	public static final String KEY_RESUMPTION_TIMEOUT = KEY_PREFIX_NONE_CRITICAL + "DTLS_RESUMPTION_TIMEOUT";
 
 	/**
 	 * Creates a context for DTLS session parameters.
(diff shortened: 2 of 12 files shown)


boaks:
I use this, if the network has changed and to optimize the communication for different environment as Wifi and Mobile.
Anyway, in the meantime I think, implementing the draft 
https://github.com/tlswg/dtls-conn-id
makes sense. I hope I can spend some time in the next weeks into it.

sbernard31:
I have no remarks about the code itself.

But I'm not sure to understand the auto-resumption modification.
You want to be able to send a message which are able to change the default auto resumption timeout value ? Why not just changing it at DTLSconnector level ?

> Anyway, in the meantime I think, implementing the draft
https://github.com/tlswg/dtls-conn-id makes sense

Very interesting !

I recently think a lot about clustering and failover and I ask myself if those could not make sense to try to share/persist connection state. What is your feeling about that ?

boaks:
> Why not just changing it at DTLSconnector level ?

for me it seems easier to relate that to the message then to add a setter to the dtls connector.

boaks:
> I recently think a lot about clustering and failover and I ask myself if those could not make sense to try to share/persist connection state. What is your feeling about that ?

The pitfall would be the https://tools.ietf.org/html/rfc6347#section-4.1.2.6

That window (read) and the send-seqn (write) must be synced. Depending on the assumptions of the "cluster node mapping" changes, this may get hard.

sbernard31:
> for me it seems easier to relate that to the message then to add a setter to the dtls connector.

I do not understand why because it doesn't seem to be a property of a message ?
(easier at use ? or easier to implement ? or both :grin: )

> The pitfall would be the https://tools.ietf.org/html/rfc6347#section-4.1.2.6

Very good point! I think at this deeper and that probably means synchronization without cache allowed and we also need to read/write connection in store foreach packet received or sent (seq number too). It sounds not so good, right ?

boaks:
> It sounds not so good, right ?

Depends on the speed of the internal communication compared to the outer.
May be a bulk sync could be used, but these are all just very raw ideas.

boaks:
>    for me it seems easier to relate that to the message then to add a setter to the dtls connector.

> I do not understand why because it doesn't seem to be a property of a message ?
(easier at use ? or easier to implement ? or both 😁 )

In some a couple of cases (e.g. using the EndpointManager) it's not easy to access the DTLSConnector. But the EndpointContext attributes are accessible and intended to specify the details of the communication, especially the connector specific detials. Therefore I put it there.

sbernard31:
> In some a couple of cases (e.g. using the EndpointManager) it's not easy to access the DTLSConnector. But the EndpointContext attributes are accessible and intended to specify the details of the communication, especially the connector specific detials.

It still sounds strange to me and there is a `getConnector` on `CoapEndpoint`. But this is not blocker for me (hopping that we will not put so many config thing in endpointContext). So feel free to merge it like this if you found this better than just accessing to the DTLS connector. 

> Depends on the speed of the internal communication compared to the outer.
May be a bulk sync could be used, but these are all just very raw ideas.

Do you think this is worth it to try to experiment that ? I started to think that was not

boaks:
> Do you think this is worth it to try to experiment that ? I started to think that was not

Currently I see 3 topics:
- implement the connection id draft (new work)
- publish my advanced android client :-) (work to complete, but after the connection id)
- improve the deduplication using a SoftReference cache 
  and also improve the connectionExecutors cache prevent connections with pending jobs from being removed. (first prototype for deduplication is working, but has some small negative performance effects.)

So "cluster" is something, I will not be able to work on this year :-).

(And @sophokles73 reminded me about a 2.0 release to include the coap support in hono.)

PR #804, opened 2018-11-20
Cleanup pseudo random function.

boaks:
Reuse arrays.


sbernard31:
:+1:  !!

(did you see any performance gain ?)

boaks:
No. I just don't like the new and new agains.

boaks:
> About avoiding allocation, I have doubts about the need to make an array copy in RawData.getBytes().
Less easy, DatagrammReader an DatagramWriter are good candidate too.

:-) yes, I guess they also obsolete. They difference to this PR is, that this PR just optimize the function internal allocation and doesn't span that to a larger context.

PR #806, opened 2018-11-22
Add handshake timestamp to DtlsEndpointContext.

boaks:
I guess, one comment will be: "now we can make strict matching really strict" 

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
@@ -38,6 +38,9 @@
  *                                                    process reordered handshake messages
  *    Achim Kraus (Bosch Software Innovations GmbH) - reset master secret, when
  *                                                    session resumption is refused.
+ *    Achim Kraus (Bosch Software Innovations GmbH) - remove unused isClient
+ *                                                    add handshake timestamp for
+ *                                                    session and endpoint context.
  ******************************************************************************/
 package org.eclipse.californium.scandium.dtls;
 
@@ -118,12 +121,6 @@ public final class DTLSSession {
 	private byte[] masterSecret = null;
 
 	/**
-	 * Indicates whether this object represents the <em>client</em> or the <em>server</em>
-	 * side of the connection. The <em>client</em> side is the one initiating the handshake.
-	 */
-	private final boolean isClient;
-
-	/**
 	 * The <em>current read state</em> used for processing all inbound records.
 	 */
 	private DTLSConnectionState readState = new DTLSConnectionState();
@@ -172,6 +169,8 @@ public final class DTLSSession {
 	private String virtualHost;
 	private boolean peerSupportsSni;
 
+	private final String handshakeTimeTag;
+
 	// Constructor ////////////////////////////////////////////////////
 
 	/**
@@ -179,11 +178,9 @@ public final class DTLSSession {
 	 *
 	 * @param peerAddress
 	 *            the remote address
-	 * @param isClient
-	 *            whether the entity represents a client or a server.
 	 */
-	public DTLSSession(InetSocketAddress peerAddress, boolean isClient) {
-		this(peerAddress, isClient, 0, System.currentTimeMillis());
+	public DTLSSession(InetSocketAddress peerAddress) {
+		this(peerAddress, 0, System.currentTimeMillis());
 	}
 
 	/**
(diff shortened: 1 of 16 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
```diff
@@ -182,8 +183,8 @@
 	 * @param isClient
```
sbernard31: remove isclient from javadoc too.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
```diff
@@ -231,8 +232,8 @@ public DTLSSession(SessionId id, InetSocketAddress peerAddress, SessionTicket ti
 	 *            (see <a href="http://tools.ietf.org/html/rfc6347#section-4.2.1">
 	 *            section 4.2.1 of RFC 6347 (DTLS 1.2)</a> for details)
 	 */
-	public DTLSSession(InetSocketAddress peerAddress, boolean isClient, long initialSequenceNo) {
-		this(peerAddress, isClient, initialSequenceNo, System.currentTimeMillis());
+	public DTLSSession(InetSocketAddress peerAddress, long initialSequenceNo) {
```
sbernard31: remove isclient from javadoc too.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
```diff
@@ -251,15 +252,15 @@ public DTLSSession(InetSocketAddress peerAddress, boolean isClient, long initial
 	 *            section 4.2.1 of RFC 6347 (DTLS 1.2)</a> for details)
 	 * @param creationTime creation time of session. Maybe from previous session on resumption.
 	 */
-	public DTLSSession(InetSocketAddress peerAddress, boolean isClient, long initialSequenceNo, long creationTime) {
+	public DTLSSession(InetSocketAddress peerAddress, long initialSequenceNo, long creationTime) {
```
sbernard31: same here

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
```diff
@@ -418,11 +415,13 @@ public int getWriteEpoch() {
 		return writeEpoch;
 	}
 
+	// tests only, currently not used
 	void setWriteEpoch(int epoch) {
 		if (epoch < 0) {
 			throw new IllegalArgumentException("Write epoch must not be negative");
 		} else {
 			this.writeEpoch = epoch;
+			this.writeEpochTag = Integer.toString(epoch);
```
sbernard31: You create an attribute to avoid to create a string each time we create a context ? I'm not sure to see the benefit.
And if there is one, should we rather store the context itself ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
boaks: > And if there is one, should we rather store the context itself ?

that was something, I tried to introduce last year :-).

boaks:
> You want to use timestamp as identifier for the connection? this should work

The currently used epoch is just the same, because 1 is the next to 0. The timestamp of the handshake only used by the side, which initialized it. should work. Or which issue do you see or expect?

sbernard31:
> Or which issue do you see or expect?

Currently I don't see any issue. Using a timestamp as identifier are not so intuitive to me but I think it should work.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
sbernard31: Let me guess: I said I didn't see the benefits ? (or you didn't open the PR?)

boaks:
It's more a "determiner" than a "identifier" :-).

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
boaks: I can also got for convert the ints "on demand" always into strings, if you prefer.

sbernard31: meaning removing `writeepochTag` and `readEpochTag` attributes ?

boaks: yes.

sbernard31: I would prefer. And if we think an optimization like this is worth it. We could go later in that way but targeting  `DtlsEndpointContext` instead.

Issue #649, opened 2018-05-24
SNI PresharedKeyIdentity.getName() API break ?


sbernard31:
I'm currently integrating cf 2.0.0-M10 in Leshan.
The main API break is about PSK/SNI changes.

I don't think so much about integrating SNI support in Leshan so I disable it for now, using `setSniEnabled(false)`

But my test doesn't pass anymore mainly because of  **`PreSharedKeyIdentity.getName()` returns the `identity` prefixed with a `:` even if there is no virtualhost provided**. (see PreSharedKeyIdentity constructor)

Is it the intended behavior ? 

I was missing a bit the discussion about it and maybe I do not understand so well the idea behind it.

sophokles73:
Sorry to hear that it broke your test(s). The reason to (always) include the colon (`:`) is that otherwise you cannot distinguish between an identifier `the-identity` qualified by virtual host name `my.host` and an unqualified identifier called `my.host:the-identity`.
If the *getName* method is used to determine if two identifiers are equal, then this might pose a problem. The *equals* method therefore checks for both properties, the virtual host and the identifier.

However, the `Principal` interface explicitly defines the *equals* and *hashCode* methods which to me indicates that checking for equality should be based on these methods (and probably not on *getName*). Based on this it might be acceptable to return the same name for the two instances described above and add a warning to the *getName* method's JavaDoc pointing out this problem. WDYT?

Sorry for the inconvenience, I didn't expect any particular problem arising from this. @boaks reviewed the PR and also didn't seem to suspect a problem here.

sbernard31:
Now I understand why this is done like this.

> Sorry for the inconvenience, I didn't expect any particular problem arising from this. @boaks reviewed the PR and also didn't seem to suspect a problem here.

No problem, It's not a big deal to me, I replaced in my code `getName()` by `getIdentity`.
But I still ask myself, if in a general way we want that disabling SNI make scandium behave like before ? Maybe this is acceptable like this ?

boaks:
@sbernard31 

Any open stuff for this issue? Or can it be closed?

sbernard31:
I just raise that `getName()` behavior break compatibility with previous californium version even if you deactivate SNI support. Which surprising me.

On our side we fix that using `getIdentity` instead of `getName`.

As @sophokles73 say if you really want to compare 2 `Principal`, you should probably use `equals`. So finally I don't know what is the purpose of `getName` and so if it's ok to return identity prefixed with `:` if there is no virtualhost.

[...]

Issue #644, opened 2018-05-15
DTLS Session Resumption doesn't consider SNI 


boaks:
Branch 2.0.x:

RFC6066, section 4, page 7:

> A server that implements this extension MUST NOT accept the request
> to resume the session if the server_name extension contains a
> different name. Instead, it proceeds with a full handshake to
> establish a new session. When resuming a session, the server MUST
> NOT include a server_name extension in the server hello.
> 

Test added to `DTLSConnectorTest`:

```
	@Test
	public void testConnectorPerformsFullHandshakeWhenResumingWithDifferentSni() throws Exception {
		// Do a first handshake
		RawData raw = RawData.outbound("Hello World".getBytes(), new AddressEndpointContext(serverEndpoint, "server.one", null), null, false);
		givenAnEstablishedSession(raw, true);
		byte[] sessionId = establishedServerSession.getSessionIdentifier().getId();

		// Force a resume session the next time we send data
		client.forceResumeSessionFor(serverEndpoint);
		Connection connection = clientConnectionStore.get(serverEndpoint);
		assertArrayEquals(sessionId, connection.getEstablishedSession().getSessionIdentifier().getId());
		client.start();

		// Prepare message sending
		final String msg = "Hello Again";
		CountDownLatch latch = new CountDownLatch(1);
		clientRawDataChannel.setLatch(latch);

		// send message
		RawData data = RawData.outbound(msg.getBytes(), new AddressEndpointContext(serverEndpoint, "server.two", null), null, false);
		client.send(data);
		assertTrue(latch.await(MAX_TIME_TO_WAIT_SECS, TimeUnit.SECONDS));

		// check session id was not equals
		connection = clientConnectionStore.get(serverEndpoint);
		Assert.assertThat(sessionId, not(equalTo(connection.getEstablishedSession().getSessionIdentifier().getId())));
		assertClientIdentity(RawPublicKeyIdentity.class);
	}
```

fails, so the session is resumed without considering the changed SNI.

boaks:
Using SNI (provide a hostname instead of the ip-address as destination), a resumption fails always!
The Resuming DTLSSession has not virtual name (it's only set for new sessions), therefore the matcher fails.

boaks:
Fixed with PR #816

PR #811, opened 2018-11-28
Dtls cleanup

boaks:
Some cleanup and simplification.
The adapted handling of the connection is part of preparing for CID, when the connection must be looked up also with a CID and not only the address:port.  

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -172,9 +172,7 @@ import org.eclipse.californium.scandium.dtls.DTLSSession;
 import org.eclipse.californium.scandium.dtls.DtlsHandshakeException;
 import org.eclipse.californium.scandium.dtls.HandshakeException;
 import org.eclipse.californium.scandium.dtls.HandshakeMessage;
-import org.eclipse.californium.scandium.dtls.HandshakeType;
 import org.eclipse.californium.scandium.dtls.Handshaker;
-import org.eclipse.californium.scandium.dtls.HelloRequest;
 import org.eclipse.californium.scandium.dtls.HelloVerifyRequest;
 import org.eclipse.californium.scandium.dtls.InMemoryConnectionStore;
 import org.eclipse.californium.scandium.dtls.MaxFragmentLengthExtension;
@@ -354,7 +352,7 @@ public class DTLSConnector implements Connector, RecordLayer {
 
 	private final void sessionEstablished(Handshaker handshaker, final DTLSSession establishedSession)
 			throws HandshakeException {
-		Connection connection = connectionStore.get(handshaker.getPeerAddress());
+		final Connection connection = connectionStore.get(handshaker.getPeerAddress());
 		connectionStore.putEstablishedSession(establishedSession, connection);
 		List<Object> list = handshaker.takeDeferredApplicationData();
 		if (!list.isEmpty()) {
@@ -369,7 +367,7 @@ public class DTLSConnector implements Connector, RecordLayer {
 
 							@Override
 							public void run() {
-								sendMessage(rawData, establishedSession);
+								sendMessage(rawData, connection, establishedSession);
 							}
 						});
 					} else if (message instanceof Record) {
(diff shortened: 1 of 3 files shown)


boaks:
I work on #649 and #644 top on this changes. It would make it less work prone, if this could be merged, because that fix crosses "some" changes here.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -779,27 +779,15 @@ private void receiveNextDatagramFromNetwork(DatagramPacket packet) throws IOExce
 		for (final Record record : records) {
 			try {
 
-				switch (record.getType()) {
-				case HANDSHAKE:
-				case APPLICATION_DATA:
-				case ALERT:
-				case CHANGE_CIPHER_SPEC:
-
```
sbernard31: Maybe it was not so bad to remove discard record before to create a new "task" ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1078,183 +1085,134 @@ private void processChangeCipherSpecRecord(Record record) {
...
-	 * @param record
-	 * @param connection
+	 * @param message handshake message
+	 * @param record record of handshake message
+	 * @param connection connection to process handshake message. May only be
+	 *            {@code null}, if a CLIENT_HELLO is received.
 	 * @throws HandshakeException if the handshake message cannot be processed
 	 */
-	private static void processOngoingHandshakeMessage(final HandshakeMessage message, final Record record, final Connection connection) throws HandshakeException {
+	private void processOngoingHandshakeMessage(final HandshakeMessage message, final Record record, final Connection connection) throws HandshakeException {
```
sbernard31: javadoc say connection can be NULL, if this is true => NPE spotted

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: The bad's are already removed in  `List<Record> Record.fromByteArray` by

```
ContentType contentType = ContentType.getTypeByValue(type);
if (contentType == null) {
	LOGGER.debug("Received DTLS record of unsupported type [{}]. Discarding ...", type);
} else {
	records.add(new Record(contentType, version, epoch, sequenceNumber, fragmentBytes, peerAddress));
}
```

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: No, javadoc says:

> May only be {@code null}, if a CLIENT_HELLO is received.

The null connection is catched up in `receiveNextDatagramFromNetwork` with

```
if (connection == null && (records.size() != 1 || !records.get(0).isNewClientHello())) {
	LOGGER.debug("Discarding {} records from [{}] received without existing connection", records.size(), peerAddress);
	return;
}
```

(And I plan, to extract the processing of the "connection less CLIENT_HELLO" from the serialized execution, but that will be a future step ... )

boaks:
> (I have not so much time to look at californium, sry for delay)

That's fully OK, you do a very great job for leshan. And if things need to be adjusted later, I do not feel, that this will get a too big issue.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: > No, javadoc says:
>
>    May only be {@code null}, if a CLIENT_HELLO is received.

I still understand that when I use this method I can use a null connection.

boaks: And so you propose?
And again, only for a short term, then the processing of "connection less CLIENT_HELLO" will be redesigned in one of the next steps. Then the "maybe null" gets removed.

sbernard31: If this will be removed, ignore this comment. Go to merge :D

boaks: > May only be {@code null}

was used in some places, therefore I checked the wrong ones. At this location you are right and I fixed the javadoc. :-)

PR #816, opened 2018-12-03
Fix SNI issues #649 and #644

boaks:
Use simple PSK principal name, if SNI is disabled.

Also support session resumption with SNI.
Fixes issues #649 and #644.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
@@ -14,6 +14,8 @@
  *    Bosch Software Innovations GmbH - initial creation
  *    Achim Kraus (Bosch Software Innovations GmbH) - use ASN.1 DER encoding
  *                                                    directly for serialization
+ *    Achim Kraus (Bosch Software Innovations GmbH) - distinguish plain and scoped
+ *                                                    identity. issue #649
  *******************************************************************************/
 
 package org.eclipse.californium.scandium.auth;
@@ -34,7 +36,8 @@ import org.eclipse.californium.elements.util.StandardCharsets;
  */
 public final class PrincipalSerializer {
 
-	private static final int PSK_LENGTH_BITS = 16;
+	private static final int PSK_HOSTNAME_LENGTH_BITS = 16;
+	private static final int PSK_IDENTITY_LENGTH_BITS = 16;
 	private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
 
 	private PrincipalSerializer() {
@@ -70,6 +73,10 @@ public final class PrincipalSerializer {
 	 * }
 	 * </pre>
 	 * 
+	 * psk_identity may be scoped by server name indication. To distinguish
+	 * scoped and plain psk_identity, the first byte in the opaque psk_identity
+	 * indicates a scoped identity with 1, or a plain identity with 0.
+	 * 
 	 * @param principal The principal to serialize.
 	 * @param writer The writer to serialize to.
 	 * @throws NullPointerException if the writer is {@code null}.
@@ -92,10 +99,16 @@ public final class PrincipalSerializer {
 
 	private static void serializeIdentity(final PreSharedKeyIdentity principal, final DatagramWriter writer) {
 		writer.writeByte(ClientAuthenticationType.PSK.code);
-		byte[] virtualHost = principal.getVirtualHost() == null ? EMPTY_BYTE_ARRAY :
-			principal.getVirtualHost().getBytes(StandardCharsets.UTF_8);
-		writeBytesWithLength(PSK_LENGTH_BITS, virtualHost, writer);
-		writeBytesWithLength(PSK_LENGTH_BITS, principal.getIdentity().getBytes(StandardCharsets.UTF_8), writer);
+		if (principal.isScopedIdentity()) {
+			writer.writeByte((byte) 1); // scoped
+			byte[] virtualHost = principal.getVirtualHost() == null ? EMPTY_BYTE_ARRAY
+					: principal.getVirtualHost().getBytes(StandardCharsets.UTF_8);
+			writeBytesWithLength(PSK_HOSTNAME_LENGTH_BITS, virtualHost, writer);
+			writeBytesWithLength(PSK_IDENTITY_LENGTH_BITS, principal.getIdentity().getBytes(StandardCharsets.UTF_8), writer);
+		} else {
+			writer.writeByte((byte) 0); // plain
+			writeBytesWithLength(PSK_IDENTITY_LENGTH_BITS, principal.getIdentity().getBytes(StandardCharsets.UTF_8), writer);
+		}
 	}
 
 	private static void serializeSubjectInfo(final RawPublicKeyIdentity principal, final DatagramWriter writer) {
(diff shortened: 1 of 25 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
```diff
@@ -300,8 +297,48 @@ public String getVirtualHost() {
...
+	/**
+	 * Set the server names for the server that this session has been
+	 * established for.
+	 * <p>
+	 * 
+	 * @param serverNames the server names (may be {@code null}).
+	 */
+	public void setServerNames(ServerNames serverNames) {
+		this.virtualHost = null;
+		this.serverNames = serverNames;
```
sbernard31: I'm not so aware about SNI extension but there is several server names for a given session ?

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
```diff
@@ -92,10 +98,16 @@ public static void serialize(final Principal principal, final DatagramWriter wri
...
 		writer.writeByte(ClientAuthenticationType.PSK.code);
-		byte[] virtualHost = principal.getVirtualHost() == null ? EMPTY_BYTE_ARRAY :
-			principal.getVirtualHost().getBytes(StandardCharsets.UTF_8);
-		writeBytesWithLength(PSK_LENGTH_BITS, virtualHost, writer);
-		writeBytesWithLength(PSK_LENGTH_BITS, principal.getIdentity().getBytes(StandardCharsets.UTF_8), writer);
+		if (principal.isScopedIdentity()) {
+			writer.writeByte((byte) 1); // scoped
+			byte[] virtualHost = principal.getVirtualHost() == null ? EMPTY_BYTE_ARRAY
+					: principal.getVirtualHost().getBytes(StandardCharsets.UTF_8);
+			writeBytesWithLength(PSK_LENGTH_BITS, virtualHost, writer);
```
sbernard31: it sounds a bit strange to use `PSK_LENGHT_BITS` constant for `virtualHost` length.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1108,8 +1109,11 @@ public Builder setIdentity(PrivateKey privateKey, Certificate[] certificateChain
 		public Builder setTrustStore(Certificate[] trustedCerts) {
```
sbernard31: should we deprecate `setTrustStore(Certificate[] ...)` and create a new `setTrustStore(X509Certificate[] ...)` ?

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
```diff
@@ -146,10 +158,16 @@ private static X509CertPath deserializeCertChain(final DatagramReader reader) {
...
+		byte scoped = reader.readNextByte();
+		if (scoped == 1) {
+			byte[] bytes = readBytesWithLength(PSK_LENGTH_BITS, reader);
+			String virtualHost = bytes.length == 0 ? null : new String(bytes, StandardCharsets.UTF_8);
+			bytes = readBytesWithLength(PSK_LENGTH_BITS, reader);
+			return new PreSharedKeyIdentity(virtualHost, new String(bytes, StandardCharsets.UTF_8));
+		} else {
+			byte[] bytes = readBytesWithLength(PSK_LENGTH_BITS, reader);
+			return new PreSharedKeyIdentity(new String(bytes, StandardCharsets.UTF_8));
+		}
```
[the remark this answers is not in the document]
boaks: Yes. It's a straight forward solution for the introduced "scope flag".
AFAIK this serializer is mainly used for session-tickets and some bugs are just recently fixed (last 6 months), I hope, it's not  a too big issue. Otherwise, If you would be affected, or someone else is reporting troubles, I can write to old format and append the type at the end (with default scoped).

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
```diff
@@ -167,8 +179,9 @@
 	private volatile long receivedRecordsVector = 0;
 	private long creationTime;
 	private String virtualHost;
+	private ServerNames serverNames;
 	private boolean peerSupportsSni;
-
+
```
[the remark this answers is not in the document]
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/DTLSSession.java
boaks: Several of different types. But the hostname is only one type and currently the only supported :-).

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
boaks: That's not changed. Should we change that to LENGTH_BITS ?
Or other proposal?

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: if  the certs are loaded from a "KeyStore", then they are `Certificate`. 
And, if we change it, we should change it also in `setIdentity`.

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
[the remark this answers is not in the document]
sbernard31: Personally, I'm not affected I report the issue just in case.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
sbernard31: Ok it sounds to be a good argument to keep it like this.

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/PrincipalSerializer.java
sbernard31: It could be,  or adding a `HOST_LENGTH_BITS` (or `VIRTUALHOST_LENGTH_BITS`).

boaks: PSK_HOSTNAME_LENGTH_BITS and PSK_IDENTITY_LENGTH_BITS are now used

boaks:
> 	 * If enabled, the client side should send a server name extension, if the
> 	 * server is specified with hostname rather then with a raw ip-address. The
> 	 * server side support currently includes a server name specific PSK secret
> 	 * lookup and a forwarding of the server name to the CoAP stack in the
> 	 * {@link EndpointContext}. The x509 or RPK credentials lookup is currently
> 	 * not server name specific, therefore the server's certificate will be the
> 	 * same, regardless of the indicated server name.
 
I added some details to the javadoc as well. May be the last thing you put a eye on for this PR :-).

sbernard31:
:+1:

Issue #677, opened 2018-06-25
Ping implementation send Empty Message with Uri-Host option


hirokiht:
According to RFC7252, 

> Empty Message
      A message with a Code of 0.00; neither a request nor a response.
      An Empty message only contains the 4-byte header.

The ping implementation sends an empty message with Uri-Host option, makes it incompatible with other standard compliant CoAP libraries such as libcoap.

vikram919:
IMHO

>  request.setURI(uri); 

> makes it incompatible with other standard compliant CoAP libraries such as libcoap.

Can I know how you have tested it?

this means not setting UriHost option, but to mention about destination address of the empty ping request.

hirokiht:
I discovered when using the CoAP Shell (https://github.com/tzolov/coap-shell), which is based on Callifornium. As libcoap fails the packet sanity check, I fired up Wireshark to take a look. I confirmed it by removing `pdu->used_size != 0` in the `coap_pdu_parse_opt` (pdu.c) in libcoap, then recompiled it.

boaks:
This smells very strong for a RFC ambiguity :-) .

https://tools.ietf.org/html/rfc7252#section-5.10.1

A "ping" of 4 bytes could then not be used, if "multiple virtual servers" are used.
"Explicit Uri-Host and Uri-Port Options are typically used when an endpoint hosts multiple virtual servers."

vikram919:
I have tested with PlugtestClient.java from demo-apps/cf-plugtest-client Californium with coap-server.o libcoap it true that libcoap server prints "WARN discard malformed PDU"  message

hirokiht:
IMHO, a "ping" doesn't really need to reach "virtual server", just the hosting server is good enough. After all, ping is just a simplistic tool. If the virtual server somehow is unable to respond to normal requests, the virtual server (or hosting server if configured so) would be responsible to send error respond codes.

boaks:
I agree, that one interpretation maybe, that the ping targets only the real interface (hosting server). 
But I don't see, that this is the only possible interpretation. It's also possible, that the combination was simply not analysed and so not specified. So I would prefer to have an explicit statement from the IETF core to clarify the expected behaviour.

hirokiht:
In Section 4.1 of RFC7252:

> An Empty message has the Code field set to 0.00.  The Token Length
   field MUST be set to 0 and bytes of data MUST NOT be present after
   the Message ID field.  If there are any bytes, they MUST be processed
   as a message format error.

[...]

PR #828, opened 2018-12-17
Fix serialization of empty messages.

boaks:

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataSerializer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataSerializer.java
@@ -30,6 +30,7 @@
 package org.eclipse.californium.core.network.serialization;
 
 import org.eclipse.californium.core.coap.*;
+import org.eclipse.californium.core.coap.CoAP.Type;
 import org.eclipse.californium.elements.MessageCallback;
 import org.eclipse.californium.elements.RawData;
 import org.eclipse.californium.elements.util.DatagramWriter;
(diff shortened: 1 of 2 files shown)


boaks:
Fix for issue #677

californium-core/src/main/java/org/eclipse/californium/core/network/serialization/DataSerializer.java
```diff
@@ -44,28 +45,45 @@
...
+		if (message.getRawCode() == 0) {
+			// simple serialization for empty message.
+			// https://tools.ietf.org/html/rfc7252#section-4.1
+			if (message.getType() == Type.NON) {
+				throw new IllegalArgumentException("NON message must not use code 0 (empty message)!");
+			}
+			if (!message.getToken().isEmpty()) {
+				throw new IllegalArgumentException("Empty messages must not use a token!");
+			}
+			MessageHeader header = new MessageHeader(CoAP.VERSION, message.getType(), message.getToken(), 0,
```
vikram919: Hi @boaks , LOW: it would be nice to separate serializeHeader to a new method :) rather than duplicating it in both cases

boaks: > it would be nice to separate serializeHeader to a new method

Sure, I can reduce the duplicate of the call to `serializeHeader/writeCurrentByte`, but that would require to do more `if`s, at least for the call to `messageWriter.write(optionsAndPayloadWriter)`.
FMPOV, accepting the duplicate is the simpler approach.

PR #835, opened 2018-12-27
Fix udp matcher deduplication with dtls address changes.

boaks:


sbernard31:
The idea is about handling address change of foreign peer on deduplication.
This is the case where an address is re-asign and so a peer with a different identity will reuse address and MID from another peer, right ?

Could we face the opposite situation ? where we should have detect a deduplication but didn't get it as peer address changed.

This bring again on the table the idea about having a kind of endpoint identifier on which we can rely on ?

boaks:
Discussion in #837 ?

sbernard31:
Ok let's move to #837.

PR #843, opened 2019-01-08
Fix TLS unit test when executed with java 11.

boaks:
Java 11 introduces TLSv1.3, which breaks TLSv1.2 dependent unit tests.
Therefore continue to use TLSv1.2 as default.


sbernard31:
I look at quickly at SslContextUtil.java modification.
No objection

Issue #834, opened 2018-12-26
GET and DELETE request payloads sent but truncated when large


rogierc:
When GET or DELETE requests contain payloads, the payloads are transferred to the server. When the payloads are large, there is - so it seems - no blockwise transfer done, causing truncation of the payload. No exception is thrown, which can be expected to cause problems on the user application level. 

Rfc7252 & rfc7959 don't define payloads in GET and DELETE requests. So, it's not clear to me how this is intended. Shurely, a truncated payload is not. Seems to me either:
- not transfering blockwise is not intended
- transfering payload alltogether is not intended and should throw an exception
- Californium user should not set payloads on these requests.

In the latter case (or latter 2 cases) Mule CoAP Connector - as Cf user - has to prohibit payloads in GET and DELETE requests.

What's your opinion on this?

Used Californium version: 2.0.0-M12

boaks:
Californium offers two blockwise modes, transparent (done by californium) and "pass through" (blocks are passed through and the blocklogic is done in user-code on the top of californium). The modes are controlled by the NetworkConfig property "MAX_RESOURCE_BODY_SIZE". Please refer to the javadoc there (https://github.com/eclipse/californium/blob/2.0.x/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java#L171).

The usage blockwise may also be adjust/optimized , according your network constraints, with the PREFERRED_BLOCK_SIZE and MAX_MESSAGE_SIZE.

By the way, at least it's assumed, that: 
a GET request should not contain payload, the response will.
a DELETE request should not contain payload, the response also should not contain payload.
(a PUT request will contain payload, the response should not.
a POST request may contain payload, and the response may also contain payload).

boaks:
> Rfc7252 & rfc7959 don't define payloads in GET and DELETE requests. So, it's not clear to me how this is intended. 

I guess your issue will be solved with setting the MAX_RESOURCE_BODY_SIZE to a proper value. Beside of "observes" (which are just bad idea's for blockwise transfers :-) ), the method is not considered for the blockwise. I also can't see that other code limits the use of payload related to the used method or return code. FMPOV it's at least good practice, to comply to the assumptions.

rogierc:
> I guess your issue will be solved with setting the MAX_RESOURCE_BODY_SIZE

[...]

PR #844, opened 2019-01-09
Introduce strict payload checks.

boaks:
Check, it message is intended to have payload and thorw exceptions, if
not. Overwrite that check by mark a message to have unintended payload.

issue #834

### DIFF
--- a/californium-proxy/src/test/java/org/eclipse/californium/proxy/HttpTranslatorTest.java
+++ b/californium-proxy/src/test/java/org/eclipse/californium/proxy/HttpTranslatorTest.java
@@ -32,8 +32,8 @@ import org.junit.Test;
 public class HttpTranslatorTest {
 
 	@Test
-	public void testGetHttpEntity() throws Exception {
-		Request req = new Request(Code.GET);
+	public void testPutHttpEntity() throws Exception {
+		Request req = new Request(Code.PUT);
 		req.setPayload("payload");
 		req.getOptions().setContentFormat(MediaTypeRegistry.TEXT_PLAIN);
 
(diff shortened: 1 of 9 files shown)


californium-proxy/src/test/java/org/eclipse/californium/proxy/HttpTranslatorTest.java
```diff
@@ -33,7 +33,7 @@
 
 	@Test
 	public void testGetHttpEntity() throws Exception {
-		Request req = new Request(Code.GET);
+		Request req = new Request(Code.PUT);
```
sbernard31: I maybe don't get the point here, but should we rename the test name too?

californium-proxy/src/test/java/org/eclipse/californium/proxy/HttpTranslatorTest.java
```diff
@@ -42,7 +42,7 @@ public void testGetHttpEntity() throws Exception {
 
 	@Test
 	public void testGetHttpEntityWithJSON() throws Exception {
-		Request req = new Request(Code.GET);
+		Request req = new Request(Code.PUT);
```
sbernard31: same as above ?

californium-proxy/src/test/java/org/eclipse/californium/proxy/HttpTranslatorTest.java
boaks: the next line calls `req.setPayload("payload");` so I changed it.
Yes, the test should be then renamed.

PR #845, opened 2019-01-09
Use full-handshake for resumption with empty session id.

boaks:
Fix connection for pending flight to new connection to prevent
retransmissions after finishing the handshake.


boaks:
@sbernard31 

I spent the afternoon in analyze the behavior of my tradfri gateway when I restart it to reproduce https://github.com/eclipse/smarthome/issues/6065 .

I found, the resuming with an "empty session id" is not according 
https://tools.ietf.org/html/rfc5246#section-7.4.1.3

I also detected, the the pending flight is maintained by the old connection and so not proper stopped when the next flight of the new connection is sent. 

This PR fixes both. And I hope it helps to fix smarthome  tradfri support.

So, if possible, please have a fast check. I think this could be merged soon. If the smarthome project verifies that patch to improve the work, I think we should release 1.0.7 
(but only then :-) ).  

(The 2.0.x PR for the empty session id will come soon.)

Issue #824, opened 2018-12-12
WIP: first implementation of draft-ietf-tls-dtls-connection-id-03


boaks:
Just for those who are interested, I prepared a WIP branch for the draft-ietf-tls-dtls-connection-id-03 implementation. Currently the house keeping need a lot more work. Depending on the answer to my issue https://github.com/tlswg/dtls-conn-id/issues/28 it may get hard, if mixed usage is intended. That would require to update the peers ip-address after the record could be decrypted successfully.

boaks:
https://github.com/eclipse/californium/tree/wip_add_dtls_cid

thomas-fossati:
Build fails for me: see [wip_add_dtls_cid-branch-build.log](https://github.com/eclipse/californium/files/2749038/wip_add_dtls_cid-branch-build.log)

boaks:
> RequestTest.testSetURIRejectsUnresolvableHost Expected exception: java.lang.IllegalArgumentException

```
	@Test(expected = IllegalArgumentException.class)
	public void testSetURIRejectsUnresolvableHost() {
		Request.newGet().setURI("coap://non-existing.host");
	}
```

Does nslookup `non-existing.host` work on your side? 
(And, yes, assuming that this **MUST** cause an error is also not "state of the art". But it's unfortunately not the only one, but one, which is not reported from too many others.)

Anyway, this unittest error seems to be irrelevant for testing cid.

Please try 

```
mvn clean install -DskipTests=true
```

And: currently it's still mostly used with java 1.8 .Newer ones may not work :-).

boaks:
First part is prepared as PR #852

boaks:
I prepared a "interop workaround" for a go wip.

https://github.com/thomas-fossati/dtls/blob/master/CID.md 

Starting the go server and using the sc-dtls-example-client "works on my machine" :-).

So, please `git pull --force`, build californium again and the cd to `demo-apps/run`. 

Start the go-server as described in the link above (please adjust the port to listen to 5684) .
Execute the `sc-dtls-example-client-2.0.0-SNAPSHOT.jar` in `demo-apps/run` using 

```
java -jar sc-dtls-example-client-2.0.0-SNAPSHOT.jar
```
The go-server should display the "HELLO WORLD" message and typing something into the go-server console sends something back.

The go-client and scandium server still requires some work.

@thomas-fossati 
Did you find a answer to
`Does nslookup non-existing.host work on your side?`
[...]

PR #846, opened 2019-01-09
Use full-handshake for resumption with empty session id.

boaks:
Add no server session id to configuration.
Use creation time for endpoint context, if session id is empty.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/MapBasedEndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/MapBasedEndpointContext.java
@@ -142,8 +142,9 @@ public class MapBasedEndpointContext extends AddressEndpointContext {
 			String value = attributes[++index];
 			if (null == key) {
 				throw new NullPointerException((index / 2) + ". key is null");
-			}
-			if (null == value) {
+			} else if (key.isEmpty()) {
+				throw new IllegalArgumentException((index / 2) + ". key is empty");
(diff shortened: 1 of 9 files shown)


boaks:
#845 for 2.0.x.

If "more" server don't provide a session id than the tradfri gateway, the endpoint context must use something else for matching endpoint contexts

sbernard31:
LGTM.

So currently, for session without session id, we use the creation time as identifier.
Meaning that 2 sessions which are created as the same time would have the same identifier. A bit strange but probably not an issue as conflict seems really improbable.

Just in case, some idea to replace that :
As resumption and re-negociation are not allowed that means that this session is strongly bound to the connection, so having a kind of connection id would help. (it would also help to implement a real STRICT endpoint context matcher), so we could :
1. a generate a connection id.
2. or maybe we could find a kind of connection identifier in DTLSConnectionState ?

boaks:
> currently

with cid #824 each connection has a id (unique for a period, so appended with the time, unique), which could then be used. Until yesterday, I didn't even know, that some servers may use empty session id's. So I don't think, this temporary fix will harm.

Otherwise, If you have a lot of time, should we include the cid into M13? :-)

sbernard31:
I'm OK to integrate it like this.
My comments was just to feed the reflexion for future changes.

So it's good to merge professor ! :p

boaks:
My hint to #824 was, that I hope it's the very near future.

PR #847, opened 2019-01-10
#834: UDPConnector now drops truncated payload

sbernard31:
This aims to implement https://github.com/eclipse/californium/issues/834#issuecomment-452740812.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java
@@ -330,7 +330,8 @@ public class UDPConnector implements Connector {
 
 		private Receiver(String name) {
 			super(name);
-			this.size = receiverPacketSize;
+			// we add one byte to be able to detect potential truncation.
+			this.size = receiverPacketSize + 1;
 			this.datagram = new DatagramPacket(new byte[size], size);
 		}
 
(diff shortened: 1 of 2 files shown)


element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java
```diff
@@ -339,16 +340,23 @@ protected void work() throws IOException {
...
-				RawData msg = RawData.inbound(bytes,
-						new UdpEndpointContext(new InetSocketAddress(datagram.getAddress(), datagram.getPort())),
-						false);
-				receiver.receiveData(msg);
+				if (datagram.getLength() >= size) {
+					// too large datagram for our buffer! data could have been
+					// truncated, so we discard it.
+					LOGGER.debug(
+							"UDPConnector ({}) received truncated UDP datagram from {}:{}. Maximum size allowed {}. Discarding ...",
+							effectiveAddr, datagram.getAddress(), datagram.getPort(), size - 1);
```
boaks: In some cases a RST or 4.xx response may be preferred to stop retransmissions and inform the sender, but though I don't think it will be triggered too often, it's OK.

boaks:
LGTM

element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java
sbernard31: Wait ! for DTLS :thinking:, if record is truncated we will not be able to decrypt it, right ?

So, what I proposed is maybe not doable in DTLS ? so I don't know if it is worthy to do it just for unsecured CoAP.

boaks: > Wait ! for DTLS thinking, if record is truncated we will not be able to decrypt it, right ?

Yeap.

> And so Upper layer could try to decode&reply to it.

I would only do that, if someone complains ...
I guess, in the most cases it will not be the californium peer, which has trouble to decode too large messages :-).

PR #855, opened 2019-01-24
Add vm statistic to benchmark.

boaks:
Add also reason to message when clients fails.


sbernard31:
I have a very quick look at this as this concern mainly extplugtest-client and server => no objection.

boaks:
Your right. the idea is to provide some metrics to compare different releases in the future.

PR #856, opened 2019-01-24
Remove deprecated from trust store.

boaks:
Add javadoc with link to CertificateVerifier.


sbernard31:
:+1:

PR #859, opened 2019-01-28
Add Bytes.

boaks:
Use it for Token and SessionId.

(split from #857)


boaks:

element-connector/src/main/java/org/eclipse/californium/elements/util/Bytes.java
```diff
@@ -0,0 +1,127 @@
...
+		return bytes;
+	}
+
+	/**
+	 * Get bytes as (hexadecimal) string.
+	 * 
+	 * @return bytes as (hexadecimal) string
+	 */
+	public final String getAsString() {
+		return StringUtil.byteArray2HexString(bytes, StringUtil.NO_SEPARATOR, 0);
```
sbernard31: We have many "byte to hex string" function : 
- some in `element-connector` : StringUtil
- others in `scandium` : ByteArrayUtils

The scandium version seems to be an optimized one.
Maybe one day we should merge all in `element-connector`. (and removing duplicate one)

boaks: issue #168 

Somehow, I'm aware, but somehow, #168 is a clean-up, which affects many lines and with the stuff in #824 open, I'm afraid of a lot extra merging work. Therefore I shifted that #168 clean-up now many times ;-(. 
Hope, after the "connection id", it's silent enough :-).

PR #860, opened 2019-01-28
Set fragment length on setCookie.

boaks:
Remove additional setFragmentLength after calling setCookie.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
@@ -327,9 +327,7 @@ public class ClientHandshaker extends Handshaker {
 	protected void receivedHelloVerifyRequest(HelloVerifyRequest message) throws HandshakeException {
 
 		clientHello.setCookie(message.getCookie());
-		// update the length (cookie added)
-		clientHello.setFragmentLength(clientHello.getMessageLength());
-		
+
 		flightNumber = 3;
 		DTLSFlight flight = new DTLSFlight(getSession(), flightNumber);
 		flight.addMessage(wrapMessage(clientHello));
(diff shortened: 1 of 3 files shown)


boaks:

sbernard31:
fragment length is calculated from `extensions`, `sessionid`, `supportedCipherSuites`, `compressionMethods` and `cookie` length.

It's a bit strange to me that fragment length is updated only when we are using this setter.

If I well understand, the idea is that generally the fragment lenght is calculated when we convert it to `byteArray`?
Maybe the setter could just flag that the previous fragment length is not correct anymore by setting it to -1 (e.g. : by adding an `invalidFragmentLength` method)

Just an idea. :thinking:

boaks:
The idea of this PR is to move the 

```
clientHello.setCookie(message.getCookie());
// update the length (cookie added)
clientHello.setFragmentLength(clientHello.getMessageLength());
```

`setFragmentLength` into the `setCookie`. That's just a small change, which should damage something.

A redesign of the `HandshakeMessage`moving the fragment stuff to the `FragmentedHandshakeMessage` may be something for the future.
For now I focus on wip cid.

sbernard31:
I try to reformulate because I was maybe not clear.

I agree that your modification should not damage anything.
Just in term of design, I prefer the code before this PR than after, because it sounds strange to me that only this setter updates the fragment length.

As I understand,  the issue for you was the duplicated code. I try to propose another solution which will force the length calculation if the set is called and so you could remove the `clientHello.setFragmentLength(clientHello.getMessageLength());` part.

So replacing `setFragmentLength(clientHello.getMessageLength()` in setCookie by `setFragmentLength(-1);` should do the job. (creating a more explicit method `invalidFragmentLength` would be more clear).

I just tested it and tests passed.

> For now I focus on wip cid.

I will try to look at this tomorrow.

boaks:
Generally, Messages are not considered to be altered after sending.
The only exception from that seems to be the ClientHello, which is modified by the setCookie() after the it is sent the first time. With that, I don't think it's worth to adapt the other functions.

My feeling was, that the split of the field and functions of `HandshakeMessage` and `FragmentedHandshakeMessage` is the root-cause. For the most `HandshakeMessage`s there is not reason to deal with `fragmentOffset` and `fragmentLength` other then set it to `0` and `messageLength`. So these fragment fields may simply be moved to the `FragmentedHandshakeMessage` and replaced by methods in the `HandshakeMessage`, which then get overwritten in `FragmentedHandshakeMessage`.  

Shift that to the future.

PR #861, opened 2019-01-28
Remove copy of master secret from handshaker.

boaks:


boaks:

sbernard31:
Session object could not be shared by several handshaker ?

PR #862, opened 2019-01-28
Redesign wrap message.

boaks:
Remove temporary record list and add messages directly to the flight.


boaks:

PR #864, opened 2019-01-29
Extract requestNextBlock into separate method.

boaks:
Picked from tcp_experimental_features branch for easier merging in the
future.


sbernard31:
It's just moving code into `requestNextBlock`, right  ?
If I'm right => LGTM

boaks:
That's right. The other changes of the `tcp_experimental_features` could then be easy merged. for the last merges I compared that snippet with changes in that part. With that PR, this will be easier.

PR #865, opened 2019-01-30
cleanup client authentication wanted and required defaults.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -857,45 +857,56 @@ public final class DtlsConnectorConfig {
 		}
 
 		/**
-		 * Sets whether the connector wants (requests) DTLS clients to authenticate
-		 * during the handshake. The handshake doesn't fail, if the client didn't
-		 * authenticate itself during the handshake. That mostly requires the client
-		 * to use a proprietary mechanism to authenticate itself on the application
-		 * layer (e.g. username/password). It's mainly used, if the implementation
-		 * of the other peer has no PSK cipher suite and client certificate should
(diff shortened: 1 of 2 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1607,7 +1622,7 @@ public DtlsConnectorConfig build() {
 				}
 			} 
 
-			if (!config.clientAuthenticationRequired && !config.clientAuthenticationWanted
+			if (!clientOnly && !config.clientAuthenticationRequired && !config.clientAuthenticationWanted
```
sbernard31: I think the idea is : if I'm a server and didn't ask for  any form of client authentication then I should not have set a trustcertificatetypes ?

!clientOnly does not mean server only and in case of "role exchange" this could be possible to have `trustCertificateTypes` used for client role and does not ask for client authentication in server role, but I concede this would be a pretty strange use case ... so maybe it's ok like this !

boaks: > I think the idea is : if I'm a server and didn't ask for any form of client authentication then I should not have set a trustcertificatetypes ?

Yes.

> !clientOnly does not mean server only and in case of "role exchange" this could be possible to have trustCertificateTypes used for client role and does not ask for client authentication in server role, but I concede this would be a pretty strange use case ... so maybe it's ok like this !

Though it was more a idea, to check the configuration more strict, because it got very complex, it may also fail for use-case I didn't see. So check only for "server only" :-)-

boaks: Just to mention:
if find some cases, where the validation is too strict, just report or fix it.

PR #866, opened 2019-01-31
Cleanup string utils hex converter.

boaks:


boaks:
May be, the clean up "step by step" has a chance :-).

scandium-core/src/main/java/org/eclipse/californium/scandium/util/ByteArrayUtils.java
```diff
@@ -75,14 +65,7 @@ public static String toHexString(byte[] byteArray) {
...
 	public static String toHex(byte[] byteArray) {
-
-		char[] bytesHexadecimal = new char[byteArray.length * 2];
-		for (int src = 0, dest = 0; src < byteArray.length; src++) {
-			int value = byteArray[src] & 0xFF;
-			bytesHexadecimal[dest++] = BIN_TO_HEX_ARRAY[value >>> 4];
-			bytesHexadecimal[dest++] = BIN_TO_HEX_ARRAY[value & 0x0F];
-		}
-		return new String(bytesHexadecimal);
+		return StringUtil.byteArray2HexString(byteArray);
```
sbernard31: A slightly different behavior. 
Before a null parameter (`byteArray`) raises NPE, now it return null.

boaks: I guess, no one requires the NPE :-).

boaks:
I cleaned up the javadoc :-)

boaks:

PR #867, opened 2019-02-01
Redesign handshaker.

boaks:
Keep associate connection in handshaker.
Split deprecated processing of incoming and outgoing records.
Call on onError for outgoing records on failed handshakes.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -339,7 +339,15 @@ public class DTLSConnector implements Connector, RecordLayer {
 
 				@Override
 				public void handshakeFailed(Handshaker handshaker, Throwable error) {
-					connectionStore.remove(handshaker.getPeerAddress(), false);
+					List<RawData> listOut = handshaker.takeDeferredApplicationData();
+					if (!listOut.isEmpty()) {
+						LOGGER.debug("Handshake with [{}] failed, report error to deferred {} messages",
+								handshaker.getPeerAddress(), listOut.size());
+						for (RawData message : listOut) {
+							message.onError(error);
+						}
+					}
+					connectionStore.remove(handshaker.getConnection(), false);
 				}
 			};
 			int maxConnections = configuration.getMaxConnections();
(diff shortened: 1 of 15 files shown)


boaks:
The current diff from `wip_add_dtls_cid` to `2.0.x` is

74 files changed, 2704 insertions(+), 1039 deletions(-)

So I'M afraid, if the 600 lines here are split more, it will take very long for the other 4000 :-).

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1823,12 +1825,12 @@ private void sendMessage(final RawData message, final Connection connection, fin
 	 *         be send.
 	 * @see EndpointContextMatcher#isToBeSent(EndpointContext, EndpointContext)
 	 */
-	private boolean checkOutboundEndpointContext(final RawData message, final EndpointContext connectionContext) {
+	private boolean checkOutboundEndpointContext(final RawData message, final Connection connection,
```
sbernard31: The connection param is added but seems not to be used.

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorAdvancedTest.java
```diff
@@ -1239,21 +1257,23 @@ public void setDrop(int drop) {
...
 		public void resendLastFlight() throws GeneralSecurityException, IOException {
 			DTLSFlight flight = lastFlight;
+			Connection connection = lastConnection;
 			if (null == flight) {
 				throw new IllegalStateException("no last flight!");
 			}
 			flight.incrementTries();
 			flight.setNewSequenceNumbers();
-			sendFlight(flight);
+			sendFlight(flight, connection);
```
sbernard31: could be just : `sendFlight(flight, lastConnection)`

The idea is to ensure we alway send flight through the same connection  ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
```diff
@@ -223,6 +227,33 @@ public InetSocketAddress getPeerAddress() {
...
+	}
+
+	/**
+	 * Check, if the provided address is the peers address.
+	 * 
+	 * @param peerAddress provided peer address
+	 * @return {@code treu}, if the addresses are equal
+	 */
+	public boolean equalsPeerAddress(InetSocketAddress peerAddress) {
+		if (this.peerAddress == peerAddress) {
```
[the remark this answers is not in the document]
boaks: Or some calls getPeerAddress() and the equalsPeerAddress().
I did apply more the common pattern then analyze it in deep.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: I fix it.

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorAdvancedTest.java
boaks: Yes. In one of the next versions the connection is get by address or cid, so I want to reduce the connection look ups.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
```diff
@@ -223,6 +227,33 @@ public InetSocketAddress getPeerAddress() {
...
+		if (establishedSession != null) {
+			establishedSession.setPeer(peerAddress);
+		}
+	}
+
+	/**
+	 * Check, if the provided address is the peers address.
+	 * 
+	 * @param peerAddress provided peer address
+	 * @return {@code treu}, if the addresses are equal
```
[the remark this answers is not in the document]
boaks: OK

Issue #874, opened 2019-02-11
Javadoc generation broken for Java11+


bdamm:
# Javadoc gen under Java11 is broken

Due to Java11 javadoc defaulting to and validating html5 as the output format, the html 4.0.1 elements in the source's javadocs do not pass. The javadoc plugin must receive:

```
<additionalOptions>-html4</additionalOptions>
```

boaks:
If possible, please retest with the PR #875 

Just to mention:
Californium 2.0 promises to run with java 1.7, to use java 1.11 is rather experimental. 
If you collect experience, your feedback will be welcome. Simply create an issues with the feedback, regardless of failures or success stories :-).

Issue #873, opened 2019-02-11
Javadoc generation broken for Java8+


bdamm:
# JavaDoc has many errors under Java8 and up

Due to a more strict interpretation of warnings, the default Javadoc in Java8 and newer will throw many errors when building docs for californium.

A change to the pom file solves it. Within the javadoc plugin's configuration element:

```
<doclint>none</doclint>
```

boaks:
Fixed with #875, please retest, if possible.

PR #875, opened 2019-02-12
Javadoc

boaks:
Fix javadoc in element-connector and scandium.
Add plugin configuration as work around according issue #873 and #874.


bdamm:
Your change did not work for me. I moved the "configuration" element to outside of and after the "executions" element, and then it all worked fine (under Java 11.)

boaks:
I'm wondering about that. On "my machine" it works with the configuration within the execution. According https://maven.apache.org/guides/mini/guide-configuring-plugins.html (last example) it's within the spec.

Anyway, I moved it out of the execution.

Could you please retest again?

Issue #872, opened 2019-02-11
MulticastSocket used by UdpMulticastConnector always uses the default network interface


aml2610:
Hi! 

## Version
I saw that https://github.com/eclipse/californium/pull/742 added multicast support so I started using 2.0.0-M13 version of californium-core.

## Issue
For my use-case I need to bind the multicast socket to a non-default network interface so whilst trying to do so I discovered that `UdpMulticastConnector` does not set the network interface on the `MulticastSocket` that it uses so the socket always ends up bound to the default interface.

## How to reproduce
When creating a CoAP server you can provide an endpoint that uses a `UdpMulticastConnector` e.g:
```
        ....
        // Add multicast endpoint
        final InetSocketAddress serverSocketAddress = new InetSocketAddress( "<INET_ADDR_OF_CUSTOM_IFACE>", <LOCAL_PORT>);

        final Connector connector = new UdpMulticastConnector(serverSocketAddress,CoAP.MULTICAST_IPv4);

        final CoapEndpoint.Builder builder = new CoapEndpoint.Builder();
        builder.setNetworkConfig(config);
        builder.setConnector(connector);

        server.addEndpoint(builder.build());
        ....
```

If you debug up to `UdpMulticastConnector.start()` and inspect the `MulticastSocket` created inside that method you will see the interface is never set, therefore `Multicast.interfaceSet` remains set to `false` so when `MulticastSocket.joinGroup(group)` is called it will always set the network interface to the default one - can check that by adding a breakpoint to the `setNetworkInterface(defaultInterface);` call inside `MulticastSocket.joinGroup()` method.

## Expected behaviour
I would expect that if I provide an address corresponding to a custom network interface, the multicast socket will be bound to that interface instead of the default one.

## How to fix
I think a possible fix would be something along these lines: inside `UdpMulticastConnector.start()`, after it creates a `MulticastSocket`, call `multicastSocket.setInterface(addr)` with addr being the local address provided to the CoAP endpoint.

I am happy to do the fix if you decide this is indeed an issue and if I am not mistaken in any of the above.

Thanks

boaks:
Currently `MulticastSocket socket = new MulticastSocket(localAddr);` is called, and at least the intention was, that this defines the interface by the provided local address. 
Do you have a network interface, which doesn't have a local address assigned?

boaks:
I investigated on this in the web. At least there are hints, that the `MulticastSocket` is somehow mixed up. I will check that, but it may take some days to find the time.

[...]

PR #880, opened 2019-02-18
Expose multicast socket methods in UDP multicast connector

aml2610:
This fixes a potential issue described in https://github.com/eclipse/californium/issues/872.

Tested with the MulticastClient and MulticastServer, it fixes the issue on the (problematic) environment I have been testing on (macOS high sierra, java 8, californium 2.0.M13).

Depending on the environment, this issue might not occur at all so the fix is only optional to implement, notes have been added to javadoc to instruct users on how to use this if they need it.


element-connector/src/main/java/org/eclipse/californium/elements/UdpMulticastConnector.java
```diff
@@ -28,35 +29,67 @@
...
-	 * @param socketAddress local socket address
+	 *
+	 * @param localAddress local socket address
 	 * @param multicastGroups multicast groups to join
+	 * @throws IOException if an I/O exception occurs while creating the multicast socket
 	 */
-	public UdpMulticastConnector(InetSocketAddress localAddress, InetAddress... multicastGroups) {
+	public UdpMulticastConnector(InetSocketAddress localAddress, InetAddress... multicastGroups) throws IOException {
 		super(localAddress);
+		this.socket = new MulticastSocket(localAddress);
```
boaks: I see, the socket is created here to be able applying the `this.socket.setInterface(inf);`
The downside would be, that stop/start will not work as for the other connectors.

Proposal:
I would go for using a additional field "intfAddress", check that to be not set by a any-address.
Use that field on start. If no explicit "intfAddress" is available, and the localAddress is not a "any-address", use the locaAddress instead for the socket.setInterface. 
The get my use the socket, if available or the additional field, if the socket is null.

aml2610: > check that to be not set by a any-address.
do the check in the constructor? what would happen if the check fails?

> Use that field on start
use it for what? building the socket? I thought you didn't want `setInterface` to be called unless you have to. Do you want it to be called from inside `start`?

To be honest, I think adding an additional "address" field would make things even more confusing in a context where I'm actually trying to reduce complexity of use instead of adding more.
Is there any other problem with this kind of connector working differently apart from consistency?

boaks: > do the check in the constructor? what would happen if the check fails?

No, the check should be done in setMulticastInterface() and throw a IllegalArgumentException. You may also check, if it's a local address. 

> Is there any other problem with this kind of connector working differently apart from consistency?

In my experience, the users frequently try to restart (stop/start) a connector. And that would fail.

boaks: > use it for what? building the socket? 

Use it, if set, during the `start()`, after the socket is actually created.
   
> I thought you didn't want `setInterface` to be called unless you have to.

I can't see the difference here.

> Do you want it to be called from inside `start`?

Yes, that's where the other connectors do their "start io" stuff.

boaks: @sbernard31 

Do you have a opinion/preference on the design?
The proposed one would use the socket more directly and therefore you're somehow closer to the java API.
Using an additional field and delay the actual `setInterface` until the `start()` keeps out current design of the other connectors.

aml2610: @boaks I have looked into this some more and I more or less agree with what you propose. It is right I didn't take into account you might want to restart the connector when I wrote my solution.

But I keep my point about using two different address fields - I think we can use just one unless I am missing something.

I think we can use `setInterface` with `localAddress` but only if `localAddress` is not "any-address". This would make sense since even when I had to `setInterface` on the socket separately, I would always set it to the same address that I would pass to the connector constructor as the `localAddress` param.
Do you see any problem with that?

aml2610: Actually I think we don't even need to check that it's not a "any-address" since inside `multicastSocket.joinGroup` if the multicast interface is not already set it will be set to "any-address" anyway so might as well do it here.

boaks: Just to mention:
I must also agree, that your solution is more straight forward (if the restart is ignored). But then it would become more and more a very special connector.


I already tested to call the `MulticastSocket.setInterface` in `start`.

```
InetSocketAddress localAddress = new InetSocketAddress(port);
Connector connector = new UdpMulticastConnector(localAddress, CoAP.MULTICAST_IPV4);
```

results in:

> 10:54:17.053 ERROR [CoapServer]: cannot start server endpoint [0.0.0.0/0.0.0.0:5683]
> java.net.SocketException: bad argument for IP_MULTICAST_IF: address not bound to any interface
> 	at java.net.TwoStacksPlainDatagramSocketImpl.socketNativeSetOption(Native Method)
> 	at java.net.TwoStacksPlainDatagramSocketImpl.socketSetOption(TwoStacksPlainDatagramSocketImpl.java:146)
> 	at java.net.AbstractPlainDatagramSocketImpl.setOption(AbstractPlainDatagramSocketImpl.java:309)
> 	at java.net.MulticastSocket.setInterface(MulticastSocket.java:471)
> 	at org.eclipse.californium.elements.UdpMulticastConnector.start(UdpMulticastConnector.java:53)
> 	at org.eclipse.californium.core.network.CoapEndpoint.start(CoapEndpoint.java:401)
> 	at org.eclipse.californium.core.CoapServer.start(CoapServer.java:252)
> 	at org.eclipse.californium.examples.MulticastTestServer.main(MulticastTestServer.java:44)
>   
> 

Therefore I mentioned the "any check"

```
MulticastSocket socket = new MulticastSocket(localAddr);
if (!localAddr.getAddress().isAnyLocalAddress()) {
	socket.setInterface(localAddr.getAddress());
}
```

works "on my machine". If works also on yours, then I think, that's the way to go.

sbernard31: @boaks says "this PR design" would not work : 
> In my experience, the users frequently try to restart (stop/start) a connector. And that would fail.

[Looking at the stop code](https://github.com/eclipse/californium/blob/2.0.0-M13/element-connector/src/main/java/org/eclipse/californium/elements/UDPConnector.java#L204), I would agree with him and so this [don't respect the API](https://github.com/eclipse/californium/blob/2.0.0-M13/element-connector/src/main/java/org/eclipse/californium/elements/Connector.java#L54).

If we are both right, I prefer the design which works :grin:

I read the #872 and I'm not sure to understand if this is "normal" to call `socket.setInterface(inf);`. It seems all of this is more or less relative to this [JVM issue](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4701650). (mainly documentation issue)

As the JVM API and behavior is not clear, I can understand the desire to keep close to java API.
So maybe we could go with the @boaks proposal but without this tricks :
_"If no explicit "intfAddress" is available, and the localAddress is not a "any-address", use the locaAddress instead for the socket.setInterface."_
And instead just call `socket.setInterface` if `intfAddress` is available. I'm not even sure if we must check  that `inf` is not `any-address` in `setMulticastInterface`. Those would make us even closer than java.

Anyway, I have no experience with multicast socket that's why maybe I rather prefer to limit our check unless you guys are pretty sure it's the right thing to do.

A question from a newb : is using `joinGroup(SocketAddress mcastaddr, NetworkInterface netIf)` another way to fix your issue ?

boaks: > I read the #872 and I'm not sure to understand if this is "normal" to call socket.setInterface(inf);. It seems all of this is more or less relative to this JVM issue. (mainly documentation issue)

At least, it seems to depend on the used jvm. And the javadoc and a web search didn't brought light into it.

> A question from a newb : is using joinGroup(SocketAddress mcastaddr, NetworkInterface netIf) another way to fix your issue ?

I haven't checked that. I mainly tried the snippet above

```
MulticastSocket socket = new MulticastSocket(localAddr);
if (!localAddr.getAddress().isAnyLocalAddress()) {
	socket.setInterface(localAddr.getAddress());
}
```

On ubuntu 16.04 it work with and without the additional  setInterface call. On Mac it seems to be required. My initial idea to fix that jvm issue was to expose the special MulticastSocket functions. But if the above simplified solution also works on Mac, then go for that.

sbernard31: I missed the last 3 comments before [mine](https://github.com/eclipse/californium/pull/880#discussion_r258435741).

If you are both agree now, that's ok for me.

aml2610:
```MulticastSocket socket = new MulticastSocket(localAddr);
if (!localAddr.getAddress().isAnyLocalAddress()) {
	socket.setInterface(localAddr.getAddress());
}```

This fixes my issue and I like it indeed because it means no change in the way the connector is used externally.
But unfortunately it brings a different issue: testing with the fix in, I now do not get any multicast packets on the server. 
The very useful bug report @sbernard31 referenced has some hint of explanation for that:
> MulticastSocket(SocketAddress) ctor can be used to bind to a specific
address. This is problematic because on many operating systems you can
only receive multicast packets when bound to the wildcard address.

so it's not safe to use anything else than the wildcard address when constructing the `MulticastSocket` (in fact I would add this to the connector constructor's javadoc).

With this in mind, building upon everything in this discussion I propose this:

```
	public UdpMulticastConnector(InetAddress intfAddress, InetSocketAddress localAddress, InetAddress... multicastGroups) {
		super(localAddress);
		this.intfAddress = intfAddress;
		this.multicastGroups = multicastGroups;
	}

	public UdpMulticastConnector(InetSocketAddress localAddress, InetAddress... multicastGroups) {
		this(null, localAddress, multicastGroups);
	}
``` for the constructors and

inside start we do:

```
	public synchronized void start() throws IOException {
		if (this.running)
			return;

		// creates a multicast socket with the given port number
		MulticastSocket socket = new MulticastSocket(localAddr);

		if (intfAddress != null && !intfAddress.isAnyLocalAddress()) {
			socket.setInterface(intfAddress);
		}

		// add the multicast socket to the specified multicast group for
		// listening to multicast requests
		for (InetAddress group : multicastGroups) {
			socket.joinGroup(group);
		}
		init(socket);
	}
```

This is basically @boaks solution but as simplified as possible. Note this also does not affect users for which the old code is working.

Comment (boaks) [reply element-connector/src/main/java/org/eclipse/californium/elements/UdpMulticastConnector.java]: The pain is the difference ;-(.
I will check windows and ubunut, if they work with any in the socket, and a special in the interface.
But not today ...

Comment (boaks) [reply element-connector/src/main/java/org/eclipse/californium/elements/UdpMulticastConnector.java]: I read and tried.
With the one implementations, using the wildcard/any address for the socket itself, uni-cast messages are then received by all interfaces. Only the multi-cast messages are then filter by the "interface-address".

```
// creates a multicast socket with the given port number
MulticastSocket socket = new MulticastSocket(localAddr.getPort());
if (!localAddr.getAddress().isAnyLocalAddress()) {
   socket.setInterface(localAddr.getAddress());
}
```

Therefore the snippet above would enable uni-cast message the be received by all interfaces, which would be quite unexpected.

The solution with two addresses local and intf will allow this and would be closer to the direct use of the `MulticastSocket`. I think, it should be mentioned in the javadoc, that, if someone runs into trouble with that approach, a own `UdpMulticastConnector` implementation may be used with a proper initialization of the  `MulticastSocket` in a "overriden" `void start()` for that case.
Also the link to the bug, Simon mentioned, would be welcome in the javadoc. When the Constructor doesn't throw IOExceptions, then some changes could be reverted.

aml2610: > a own `UdpMulticastConnector` implementation may be used with a proper initialization of the `MulticastSocket` in a "overriden" `void start()` for that case.

that's how I fixed it before raising this 🙂 

So just to confirm you are OK with the solution I described above?

boaks: Yes.

aml2610: Closing this as I have opened [a new one](https://github.com/eclipse/californium/pull/895)

phani50:
I want send response code as 4.29 in the case of too many requests. but i did not find this ResponseCode enum. How can i send this code ?

boaks:
May be you open an new issue for that.
The current SNAPSHOT/HEAD of 2.0.x has this response code.

PR #892, opened 2019-02-26
Early nego blockwise request at short interval

mb590721:


mb590721:
I committed with the bad account, I will fix that tonight

mb590721:
I added the unit test and fixed some typos from my previous contribution

californium-core/src/test/java/org/eclipse/californium/core/test/BlockwiseTransferTest.java
```diff
@@ -220,42 +222,56 @@ public void onCancel() {
...
+	 * Test that consecutive requests to a same resource with early blockwise negotiation are both going through the applicative layer
+	 * @throws InterruptedException
+	 */
+	@Test
+	public void testMultipleEarlyNegotiationWithShortInterval() throws InterruptedException {
+		final int nbApplicativeRequestReceived = applicativeLayerGetRequestCount.get();
+		testGetRequestWithEarlyNegotiation(false, PARAM_SHORT_RESP);
+		testGetRequestWithEarlyNegotiation(false, PARAM_SHORT_RESP);
+		
+		assertEquals("Applicative layer received only one request", 2, applicativeLayerGetRequestCount.get() - nbApplicativeRequestReceived);
```
boaks: I would prefer a implementation, which calls `applicativeLayerGetRequestCount.set(0);` in the `@After`instead of the calculation.

boaks: May be the text "Application layer received not two requests" would describe the failure not the expectation (what, if 3 is returned :-) ).

boaks: I'm not sure, why you use "applicative", I would use "application".

mb590721: > I would prefer a implementation, which calls `applicativeLayerGetRequestCount.set(0);` in the `@After`instead of the calculation.

Sorry but I'm not sure to understand what it would looks like, from what I know the `@After`method is called last to clean the test context. I committed a new version which you may prefer.

> May be the text "Application layer received not two requests" would describe the failure not the expectation (what, if 3 is returned :-) ).

Yes you're right ! I edited the message.

> I'm not sure, why you use "applicative", I would use "application".

A bad translation from French to English :-).

boaks: > Sorry but I'm not sure to understand what it would looks like, from what I know the `@After`method is called last to clean the test context.

Yes, the `BlockwiseTransferTest` contains several tests, and `@After` is executed after each to cleanup. If it's there, and someone adds a tests, which uses your counter again, that test get's also a "clean" counter. 

With the updated approach, every tests, which uses the counter, must reset it by it's own.

```
@After
public void destroyClient() throws Exception {
	clientEndpoint.destroy();
	applicativeLayerGetRequestCount.set(0);
}
```

mb590721: Thanks, I should have made more test on my own and I would have seen that its called after every `@Test`. I had a wrong understanding and thought it was only called only once after all tests were run.

I fixed it and added destroy on the second client endpoint too. Should be ok this time :-)

boaks: @BeforeClass - once per class
@Before - once per test
@Test - the tests
@After - once per test
@AfterClass - once per class

boaks:
Thanks!

Issue #891, opened 2019-02-26
Consecutive early negociation blockwise requests


mb590721:
My issue is based on branch 2.0.x but may be present since erlier on.

When making consecutives requests with early block2 blockwise negociation, the server respond to the following ones without forwarding them to the applicative layer.

It seems that the server forward it when making a request to another resource, the blockwise status lifetime and/or exchange lifetime are outdated.

I think it's the way the blockwise layer determine if it's a new blockwise request or part of an existing one for which new blocks are asked for. In the `BlockwiseLayer.receiveRequest` :
`
KeyUri key = getKey(exchange, request);

Block2BlockwiseStatus status = getBlock2Status(key);
`

Maybe it's intended but it's seems a little weird.

mb590721:
See PR 892 with a dedicated test case

boaks:
Also a good catch. :-).

boaks:
Please check, if PR #893 works for you.

mb590721:
@boaks 

Do you think of any workaround which could be implemented while waiting the next release ?
Decrease at the minimum possible the block2 status lifetime maybe ?

boaks:
Hm, then I would prefer to release 2.0.0-M14 without the new cipher suites.

@sbernard31 
What do you think about a M14?

mb590721:
I checked and reducing blockwise status lifetime would not really be a suitable solution for us anymay,

If you can release an M14 it would be great but otherwise we will rollback the migration and wait for its release.

By the way from what I saw it seems to be present since the M3.

boaks:
I scheduled the M14 release for Friday, 8-March-2019 (see issue #900).
Though I want to ensure, the M14 will work with https://cloud.google.com/community/tutorials/cloud-iot-coap-proxy , I would like to postpone the M14 release to Monday, 11-March-2019. But only, if this doesn't effect you negative. So, please give me your feedback, if Monday is also OK for you.

mb590721:
Monday is ok for us too, you can do as it fits better for you :-).

boaks:
Great, thanks!

boaks:
2.0.0-M14 is released. Please check it. If you have any trouble, don't hesitate to report it. If it works for you, I would appreciate, that the issue gets closed.

mb590721:
I will make some tests tomorrow and keep you up-to-date.

boaks:
Take all the time you need :-) !

mb590721:
Hi,

[...]

PR #893, opened 2019-02-26
Fix issue #891.

boaks:
Forward request for block2 num 0.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -368,7 +368,7 @@ public class BlockwiseLayer extends AbstractLayer {
 				// This is a large POST or PUT request
 				handleInboundBlockwiseUpload(exchange, request);
 
-			} else if (block2 != null) {
+			} else if (block2 != null && block2.getNum() > 0) {
 
 				KeyUri key = getKey(exchange, request);
 				Block2BlockwiseStatus status = getBlock2Status(key);


boaks:
@sbernard31 

Please, have also a look on this :-). It changes the blockwise independent from the "strict block2 configuration".

mb590721:
What if the first response block isn't received by the client and that it retransmit its request for the first block, the server would then pass it to the applicative layer instead of transparently retransmitting the response block right ? 

I'm not too sure about the right way to handle this use case, I don't find the RFC on the blockwise very clear about it. I'm not sure neither it's a real problem, I suppose it depends on what is done in the applicative layer.

mb590721:
Or maybe since the client should reuse the same token it will be handled correctly at the server side by another layer ? 

If yes then it seems good to me.

boaks:
> What if the first response block isn't received by the client and that it retransmit its request for the first block, the server would then pass it to the applicative layer instead of transparently retransmitting the response block right ?

No, retransmission is detected by the `MID:address:port` and handled by the deduplication in the `ReliabilityLayer.receiveRequest`. So it will not be passed again to the application layer.

> I'm not too sure about the right way to handle this use case, I don't find the RFC on the blockwise very clear about it.

The deduplication is handled as for any usual request and is defined in https://tools.ietf.org/html/rfc7252#section-4.5 . Blockwise (RFC7959) just uses such request multiple times. The "complex" point will be, that RFC7959 doesn't specify, how to implement that in a "transparent mode". So it's undefined there, when a request is forwarded to the application layer. If a  different processing is required, you may switch off that "californium transparent mode" by configuring "MAX_RESOURCE_BODY_SIZE" to 0.

boaks:
> Or maybe since the client should reuse the same token it will be handled correctly at the server side by another layer ?

The token is a artefact of a single request-response matching. There is no further token relation for blockwise transfers. Californium uses the same token for traceability of the transparent blockwise transfers. But in fact, it should be possible to patch the `BlockwiseLayer`, to use different tokens for the requests.

sbernard31:
This sounds good to me.

Should we add the test in #892 or something similar ?
Too bad we don't check exchange/blocklayer  memory leak in the class `BlockwiseTransferTest` like we do in `BlockwiseServerSideTest`.

boaks:
> Should we add the test in #892 or something similar ?

I would ask @mb590721 to do so, after this PR gets merged.

> Too bad we don't check exchange/blocklayer memory leak in the class BlockwiseTransferTest like we do in BlockwiseServerSideTest.

That's "out of scope" from issue #891.

boaks:
@mb590721 

You made some comments above. do you want to make some more tests about your remarks and my answers? Or should we merge it and you rebase your added test?

mb590721:
No that's ok for me, I was confused about the retransmission but with your explanations I now understand how it works.

You can merge the PR and I will rebase my branch based on the new 2.0.x

boaks:
> You can merge the PR and I will rebase my branch based on the new 2.0.x

Thanks in advance. I appreciate your analysing and testing of the blockwise! If you find the next issue, don't hesitate to report it.

PR #895, opened 2019-02-27
Add option to supply network interface to be set on the multicast socket

aml2610:
fix for https://github.com/eclipse/californium/issues/872


boaks:
Thanks a lot for all your work and patience!

PR #898, opened 2019-02-27
Cleanup Record.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
@@ -22,6 +22,7 @@
  *    Achim Kraus (Bosch Software Innovations GmbH) - use handshake parameter and
  *                                                    generic handshake messages to
  *                                                    process reordered handshake messages
+ *    Achim Kraus (Bosch Software Innovations GmbH) - cleanup
  ******************************************************************************/
 package org.eclipse.californium.scandium.dtls;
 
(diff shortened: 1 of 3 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -839,21 +865,17 @@ public void setSequenceNumber(long sequenceNumber) throws GeneralSecurityExcepti
 	 * @return the fragment length excluding record headers
 	 */
 	public int getFragmentLength() {
-		return fragmentLength;
+		final byte[] bytes = fragmentBytes;
+		return bytes == null ? 0 : bytes.length;
```
sbernard31: When the length is unknown, should we return -1 ? 
Or something which express clearly this could not be a real fragment length ? 

(and add this to this javadoc?)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -839,21 +865,17 @@ public void setSequenceNumber(long sequenceNumber) throws GeneralSecurityExcepti
...
 	 * 
-	 * Take over peer address from session.
-	 * 
 	 * @param session session to be set
 	 */
 	public void setSession(DTLSSession session) {
 		this.session = session;
-		if (session != null && this.peerAddress == null) {
-			this.peerAddress = session.getPeer();
-		}
```
sbernard31: This was something which has been added recently, and finally this does not make sense, right ?

boaks:
> Maybe this cleaning PR could be a good candidate to change getType behavior, which could call getFragment internally when needed to avoid getFragment call

My feeling is, decrypt with setSession is the right approach. But that requires some more changes so I postponed it for now.

sbernard31:
> My feeling is, decrypt with setSession is the right approach. But that requires some more changes so I postponed it for now.

OK, because each time we set the session we need to decrypt anyway ?
(`updateSession`? :grin: )

boaks:
> OK, because each time we set the session we need to decrypt anyway ?

"each time" was the cause, why I haven't changed it in this PR :-).
On receiving I think, it should be determined and set "once", and that naturally triggers the decryption.
On sending, it's alreay passed into the constructor. So I think, it's "applySession". But not for now ...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks:
```
public synchronized void setSession(DTLSSession session) {
	this.session = session;
	if (session != null) {
		this.peerAddress = null;
	}
}

public InetSocketAddress getPeerAddress() {
	if (session != null) {
		return session.getPeer();
	} else if (peerAddress != null) {
		return peerAddress;
	} else {
		throw new IllegalStateException("Record does not have a peer address");
	}
}
```

That was the original implementation. So this cleanup PR is more a "next step" then a "reverse step".

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks: I've updated the constructors to ensure, that the fragmentBytes are available.

PR #899, opened 2019-02-27
Reduce calls to find a dtls session.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -165,6 +165,7 @@ import org.eclipse.californium.scandium.dtls.AlertMessage;
 import org.eclipse.californium.scandium.dtls.AlertMessage.AlertDescription;
 import org.eclipse.californium.scandium.dtls.AlertMessage.AlertLevel;
 import org.eclipse.californium.scandium.dtls.ApplicationMessage;
+import org.eclipse.californium.scandium.dtls.AvailableConnections;
 import org.eclipse.californium.scandium.dtls.ClientHandshaker;
 import org.eclipse.californium.scandium.dtls.ClientHello;
 import org.eclipse.californium.scandium.dtls.Connection;
@@ -1313,24 +1314,22 @@ public class DTLSConnector implements Connector, RecordLayer {
 			// before starting a new handshake or resuming an established
 			// session we need to make sure that the peer is in possession of
 			// the IP address indicated in the client hello message
-			if (isClientInControlOfSourceIpAddress(clientHello, record, null)) {
-				final Connection handshakeConnection;
-				if (connection != null && connection.hasOngoingHandshakeStartedByClientHello(clientHello)) {
-					handshakeConnection = connection;
-				} else {
-					handshakeConnection = new Connection(peerAddress, new SerialExecutor(getExecutorService()));
-					if (!connectionStore.put(handshakeConnection)) {
+			final AvailableConnections connections = new AvailableConnections(connection);
+			if (isClientInControlOfSourceIpAddress(clientHello, record, connections)) {
+				if (connection == null || !connection.hasOngoingHandshakeStartedByClientHello(clientHello)) {
+					connections.setConnectionByAddress(new Connection(peerAddress, new SerialExecutor(getExecutorService())));
+					if (!connectionStore.put(connections.getConnectionByAddress())) {
 						return;
 					}
 				}
 				try {
 
-					handshakeConnection.getExecutor().execute(new Runnable() {
+					connections.getConnectionByAddress().getExecutor().execute(new Runnable() {
 
 						@Override
 						public void run() {
 							if (running.get()) {
-								processClientHello(clientHello, record, handshakeConnection);
+								processClientHello(clientHello, record, connections);
 							}
 						}
 					});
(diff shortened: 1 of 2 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/AvailableConnections.java
```diff
@@ -0,0 +1,112 @@
...
+ * 
+ * Helper to prevent duplicated calls to
+ * {@link ResumptionSupportingConnectionStore#find(SessionId)}.
+ * 
+ * <ul>
+ * <li>connection associated with the peer's address or cid</li>
+ * <li>connection associated with the session id</li>
+ * </ul>
+ */
+public final class AvailableConnections {
```
sbernard31: Even if I can not see why somebody would like to inherit from this class, I prefer to  avoid final keyword for class.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/AvailableConnections.java
```diff
@@ -0,0 +1,112 @@
...
+	/**
+	 * Connection associated with the session id. May be {@code null}, if no
+	 * connection is associated with the session id.
+	 */
+	private Connection bySessionId;
+	/**
+	 * Indicates, that {@link #setConnectionBySessionId(Connection)} has been
+	 * called.
+	 */
+	private boolean setBySessionId;
```
sbernard31: Why do we need an attribute ? `bySessionId == null` is not the same meaning ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/AvailableConnections.java
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	/**
+	 * Check, if {@link #setConnectionBySessionId(Connection)} was called.
+	 * 
+	 * @return {@code true}, if {@link #setConnectionBySessionId(Connection)}
+	 *         was called. Used to recognize already unsuccessful tries to find
+	 *         a connection associated with the session id.
+	 */
+	public boolean isSetConnectionBySessionId() {
```
sbernard31: `isConnectionBySessionAvailable` ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1511,12 +1527,26 @@ private void startNewHandshake(final ClientHello clientHello, final Record recor
...
+		} else if (!connection.equalsPeerAddress(peerAddress)) {
+			throw new IllegalArgumentException("connection must have records address!");
+		}
+
 		SessionTicket ticket = null;
-		final Connection previousConnection = connectionStore.find(clientHello.getSessionId());
+		if (!connections.isSetConnectionBySessionId()) {
+			connections.setConnectionBySessionId(connectionStore.find(clientHello.getSessionId()));
+		}
+		Connection previousConnection = connections.getConnectionBySessionId();
```
sbernard31: I'm not sure to understand why we set `connections` as we never use it after ?
It could be :
``` java
Connection previousConnection;
if (connections.isSetConnectionBySessionId()) {
    previousConnection = connections.getConnectionBySessionId(); 
} else {
    previousConnection = connectionStore.find(clientHello.getSessionId());
}
```

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/AvailableConnections.java
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/AvailableConnections.java
boaks: Let me try to explain:
With the `VerifyPeersOnResumptionThreshold` the "find" will be called potentially from two functions: `isClientInControlOfSourceIpAddress` and `resumeExistingSession`.
With this PR I want to ensure explicit, that `find`is only called once per received client_hello (may be more within the handshake), even if `null`is returned.
There may be "complexer" explanations, why `find` is "anyway" only called once, but for me this seems to be more fragile.

sbernard31: Ok I get it.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/AvailableConnections.java
sbernard31: Now I get the point. the name would be : `isConnectionBySessionIdKnown`?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: Is the question still pending?
The intention is to limit the calls to `find`, even if `find` has returned `null`.

sbernard31: > I guess, it's more or less a different opinion on simpleness :-)

Yes :)

boaks:
@sbernard31 

Do you have an objection against this PR?
Should I postpone it after M14?
Alternatively we can go for something without `isConnectionBySessionIdKnown` and invest in unit-tests to ensure, that `find` is called only once. I personally prefer the "explicit" approach of this PR.

sbernard31:
You can merge it :+1:

Issue #653, opened 2018-06-01
DTLS anti replay mechanism


Philago:
Hello, 

Is it possible to deactivate the DTLS anti replay mechanism (slidding window) ?

RFC6347 §3.3 states that "DTLS optionally supports record replay detection"

Thanks !

Best regards

sbernard31:
AFAIK, this is not deactivable with scandium.

boaks:
May be you can give us more information about your use-case?

https://tools.ietf.org/html/rfc6347#section-3.3

> DTLS optionally supports record replay detection.  The technique used
>    is the same as in IPsec AH/ESP, by maintaining a bitmap window of
>    received records.  Records that are too old to fit in the window and
>    records that have previously been received are silently discarded.
>    The replay detection feature is optional, since packet duplication is
>    not always malicious, but can also occur due to routing errors.
>    Applications may conceivably detect duplicate packets and accordingly
>    modify their data transmission strategy.

So, do you want such a statistic?

Philago:
Hello,

thank you for you answers.

No I don't want any statistics.

In my use case the data will be sent through different paths (packets 1,2,3,4 are sent on one path, packets 5,6,7,8 on another path,..) with different delays for each path. So depending on the size of the slidding window, if the lastest data sent arrive first, the oldest data sent may be discarded and I do not want that to happen. That's why I am asking if the anti replay mechanism can be deactivated.

Have a good day

boaks:
And a "receive window" of 64 is not enough?
Which size would you prefer or do you really want to disable it?

boaks:
Feel free to add a new comment to this issue or create a new issue, if this topic gets interesting again for you.

boaks:
https://github.com/eclipse/californium/issues/824#issuecomment-469740746

> yes I am still interested in disabling the anti replay mechanism or at least to have the possibility to configure the size of the receive window

OK, so disable would be enough? Should we go for

`NO`: no anti replay
`WINDOW`: anti replay only for the current window, older messages are **NOT** filtered out and will be decrypted (if possible and not causing a MAC error).
`FULL`: full anti replay as currently implemented

Philago:
Hello @boaks ,

just to understand the window case:
- Messages older then the window are accepted
- Messages inside the window are accepted / discarded if sent twice
- Message newer than the window are discared

Am I correct ? If it's the case it's OK on my side

boaks:
Message which are newer than the window, are always forwarded to the MAC check. 
[...]

PR #906, opened 2019-03-13
Add disable anti replay filter.

boaks:
Address issue #653

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -253,6 +253,22 @@ public final class DtlsConnectorConfig {
 	private Boolean useNoServerSessionId;
 
 	/**
+	 * Use anti replay filter.
+	 * 
+	 * @see http://tools.ietf.org/html/rfc6347#section-4.1
+	 */
+	private Boolean useAntiReplayFilter;
+
+	/**
+	 * Use filter for record in window only.
+	 * 
+	 * Messages too old for the filter window will pass the filter.
+	 * 
+	 * @see http://tools.ietf.org/html/rfc6347#section-4.1
+	 */
+	private Boolean useWindowFilter;
+
+	/**
 	 * Logging tag.
 	 * 
 	 * Tag logging messages, if multiple connectors share the same logging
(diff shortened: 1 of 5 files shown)


boaks:
I would like to merge this PR at the end of the week.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1641,6 +1681,40 @@ public Builder setNoServerSessionId(boolean flag) {
...
+		/**
+		 * Use window filter.
+		 * 
+		 * Messages too old for the filter window will pass the filter.
+		 * 
+		 * @param enable {@code true} to enable filter. Default {@code false}.
+		 * @return this builder for command chaining.
+		 * @throws IllegalArgumentException if anti replay window filter is active.
+		 * @see http://tools.ietf.org/html/rfc6347#section-4.1
+		 */
```
sbernard31: Is it a normal use case to do that ? I feel this is more a kind of workaround?
If I am right, should we precise this is not really intended to be used ? (or something like this)

Bonus question, not directly linked to this PR, does the spec support windows size > 64 ? (min size seems to be 32) I didn't check the windows sliding algorithm to know if my question makes sense :grin:

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1748,6 +1822,12 @@ public DtlsConnectorConfig build() {
 			if (config.sniEnabled == null) {
 				config.sniEnabled = Boolean.FALSE;
 			}
+			if (config.useAntiReplayFilter == null) {
+				config.useAntiReplayFilter = true;
+			}
+			if (config.useWindowFilter == null) {
+				config.useWindowFilter = false;
+			}
```
sbernard31: A detail, but it seems that setter does not allow a config where  `config.useAntiReplayFilter=true` and `config.useWindowFilter= true`

But if I set only userWindowsFilter to true, then the default value of useAntiReplayFilter will be true too.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: I guess, yes. But I don't see the benefit. Only 1 issue with that. Others seems not to have that issue. 64 is a long and more will make it more complex :-).

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: True, that's irritating.

DTLSConnector

```
this.useWindowFilter = config.useWindowFilter();
this.useFilter = config.useAntiReplayFilter() || useWindowFilter;

```

so it doesn't harm. But I will change the checks. I didn't wont "ambiguity"  in the configuration.

PR #907, opened 2019-03-13
Cleanup byte array utils.

boaks:
Replace it with common string util.


sbernard31:
LGTM

PR #908, opened 2019-03-14
OSCORE: Support for Observe functionality

rikard-sics:
Pull request with code that adds support for using Observe with OSCORE messages. It also includes a JUnit test for testing Observe with OSCORE.

Finally it does some small fixes in the other OSCORE code and comments.


cf-oscore/src/main/java/org/eclipse/californium/oscore/Encryptor.java
```diff
@@ -209,8 +213,15 @@ protected static Encrypt0Message prepareCOSEStructure(byte[] confidential, byte[
 			}
 		} else {
 			bRes.write(firstByte);
+		}
+		
+		//If the OSCORE option is length 1 and 0x00, it can be empty
+		//See https://tools.ietf.org/html/draft-ietf-core-object-security-16#page-33 point 4
+		byte[] optionBytes = bRes.toByteArray();
```
boaks: The javadoc says "Creates a newly allocated byte array.".
So I'm wondering, why `toByteArray` is called twice.
For `new byte[0]` consider to use `org.eclipse.californium.elements.util.Bytes.EMPTY`

rikard-sics: Yes, it's not needed to call it twice. I have changed that and also use Bytes.EMPTY instead now. See commit 003a99433cfa418f3babef9ee7d1af7d2ae0c2ba.

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
```diff
@@ -0,0 +1,348 @@
...
+ * 
+ * Contributors:
+ *    Matthias Kovatsch - creator and main architect
+ *    Martin Lanter - architect and re-implementation
+ *    Dominique Im Obersteg - parsers and initial implementation
+ *    Daniel Pauli - parsers and initial implementation
+ *    Kai Hudalla - logging
+ *    Achim Kraus (Bosch Software Innovations GmbH) - use CoapNetworkRule for
+ *                                                    setup of test-network
+ *    Achim Kraus (Bosch Software Innovations GmbH) - destroy server after test
```
boaks: I would prefer, that 
``` 
 * Copyright (c) 2018 RISE SICS and others.
 * 
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * and Eclipse Distribution License v1.0 which accompany this distribution.
 * 
 * The Eclipse Public License is available at
 *    http://www.eclipse.org/legal/epl-v10.html
 * and the Eclipse Distribution License is available at
 *    http://www.eclipse.org/org/documents/edl-v10.html.
 * 
 * Contributors: 
 *    Rikard Höglund (RISE SICS) - testing OSCORE Observe messages
```

Is used as header including your comment
```
This test class is based on org.eclipse.californium.core.test.SmallServerClientTest
```

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
```diff
@@ -0,0 +1,348 @@
...
+	@Before
+	public void initLogger() {
+		System.out.println(System.lineSeparator() + "Start " + getClass().getSimpleName());
+		EndpointManager.clear();
+	}
+	
+	//Use the OSCORE stack factory
+	@BeforeClass
+	public static void setStackFactory() {
+		OSCoreCoapStackFactory.useAsDefault();
```
boaks: Executing the test-suite with maven fails.

`OSCoreCoapStackFactory.useAsDefault();` calls `CoapEndpoint.setDefaultCoapStackFactory(new OSCoreCoapStackFactory());` which must be called only once!.

I would got for 
```
public class OSCoreCoapStackFactory implements CoapStackFactory {
	private static AtomicBoolean init = new AtomicBoolean();
...
	public static void useAsDefault() {
		if (init.compareAndSet(false, true)) {
			CoapEndpoint.setDefaultCoapStackFactory(new OSCoreCoapStackFactory());
		}
	}
}
```

cf-oscore/src/main/java/org/eclipse/californium/oscore/Encryptor.java
```diff
@@ -209,8 +214,15 @@ protected static Encrypt0Message prepareCOSEStructure(byte[] confidential, byte[
...
 			bRes.write(firstByte);
-			return bRes.toByteArray();
 		}
-		return bRes.toByteArray();
+		
+		//If the OSCORE option is length 1 and 0x00, it can be empty
+		//See https://tools.ietf.org/html/draft-ietf-core-object-security-16#page-33 point 4
+		byte[] optionBytes = bRes.toByteArray();
+		if(optionBytes.length == 1 && optionBytes[0] == 0x00) {
+			return Bytes.EMPTY;
```
boaks: I guess, this change causes now
`OSCoreTest.testEndcodingObjectSecurityValueCompression` to fail when
`assertArrays(objectSecurityResponse, new byte[] { 0x00 });`

It may be 
`assertArrayEquals(objectSecurityResponse, Bytes.EMPTY);`

(I would prefer to replace the own implementation of `assertArrays` by the one of `org.junit.Assert`. Or do you have a reason for your own implementation?)

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
rikard-sics: Sure, I have now update the file accordingly in commit 2e5ab13698dbb2495c8021975cc5a9a8fb70fa35.

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
rikard-sics: Thanks, I have now changed the code according to your suggestion. Done in commit a057caa93983e255fdf48050c820c7902415dbd5.

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
boaks: :-) I was wrong, we have 2019, sorry :-).
could you please adjust it again?

cf-oscore/src/main/java/org/eclipse/californium/oscore/Encryptor.java
rikard-sics: Yes, I changed the comparison to Bytes.EMPTY instead since that should now be the correct value. I also removed the assertArrays method and instead use the one from org.junit.Assert as you recommend. Done in commit 3bf97445eb5c2bb814e0f2e5b10787b3b3e1b31f

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
rikard-sics: > :-) I was wrong, we have 2019, sorry :-).
> could you please adjust it again?

Sure, it's been updated now :) See commit b0ec659f2ed558535aafea0e01da4f3b92652f30

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Performs tests of Observe message exchanges between an OSCORE server and client
+ * Based on interoperability test specification created by Ericsson:
+ * https://ericssonresearch.github.io/OSCOAP/test-spec5.html
+ * 
+ */
+@Category(Large.class)
+public class OSCoreObserveTest {
+	private CoapServer server;
```
boaks: > java.io.IOError: java.net.SocketException: Use org.eclipse.californium.elements.rule.NetworkRule to define DatagramSocket behaviour!

Last fix for the above error to run `mvn test` successfully for the complete californium project: 

```
public class OSCoreObserveTest {
   @ClassRule
   public static CoapNetworkRule network = new CoapNetworkRule(CoapNetworkRule.Mode.DIRECT, CoapNetworkRule.Mode.NATIVE);
```

That's required, if a datagram socket is involved in the tests. It enables the tests to replace the datagram socket implementation by a in memory variant, which ensures, that records are not dropped unintended. The rule enables the test execution to also select tests, which can only be execute with the "native" datagram socket implementation or the "direct". The rule also ensures, that changes in `NetworkConfig`are rolled back on the end of the test.

boaks:
When the `Rule` is added (see above), please "squash your commits and force push to your branch".
I will then check the PR with the eclipse ci-jenkins and, if the tests succeeds, I will merge it.

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
rikard-sics: Alright, thanks for the information. I had missed that it was needed. It's now been added via commit 7c901cea597079c0bb47004bcb8a4805c7b5c2e5

rikard-sics:
> When the `Rule` is added (see above), please "squash your commits, rebase and force push to your branch".
> I will then check the PR with the eclipse ci-jenkins and, if the tests succeeds, I will merge it.

Thanks for all the assistance with things to fix in the pull request. It should now be squashed, rebased and force pushed.

boaks:
Thanks for all your work and patience.

Issue #870, opened 2019-02-04
Sandbox californium.eclipse.org


boaks:
Dear CoAP community,

I'm happy, that coap sandbox, californium.eclispe.org offers now  for more than a year, is used pretty well. We got several hundred (300-600) different test-hosts each month.

Sometimes my feeling is, that either the performance or the robustness/security is tested.
That's pretty OK for me. Please take into account, that sometimes I deploy a development version and so your results will not be very characteristic. If you want better results, I recommend, that you use your own test-host, where you have much more control. You may either use our "demo-apps/cf-plugtest-server" or "demo-apps/cf-extplugtest-server". 

Currently you would have to build it on your own. 
If there is interest, I add the artifacts to our maven deployment. I would also add some details in the documentation to run it on a linux server. So let me know, if your interested.

rogierc:
The demo's  contain the ETSI plugtests, right? Having these as artefacts will ease automatic plugtesting. So, it would be helpfull for the Mule CoAP Connector tests.

boaks:
> The demo's contain the ETSI plugtests, right? 

Yes, the sandbox runs the Cf-ExtendedPlugtestServer, which contains the Cf-PlugtestServer on ports 5683 and 5684. 

> Having these as artefacts will ease automatic plugtesting. So, it would be helpfull for the Mule CoAP Connector tests.

OK, so I will ask the other committers to push these artefacts to maven central.

boaks:
The cf-plugtest-server is now deployed to SNAPSHOTS and will also be part of the next milestone (2.0.0-M14).

boaks:
Unfortunately, the missing javadoc generation caused and error, when the module was deployed on maven central. It's only available at the eclipse repo https://repo.eclipse.org/content/repositories/californium-releases.
I fixed it for M15, so the next milestone will deploy it also on maven central.

boaks:
Hopefully the 2.0.0-M15 will deploy the artifact to maven central :-).

I prepared a short wiki page for the setup as [old style systemd service](https://github.com/eclipse/californium/wiki/Californium-as-old-style-unix-systemd-service) .
Additional a second wiki page shows how to setup a [DTLS 1.2 firewall](https://github.com/eclipse/californium/wiki/DTLS-1.2-based-firewall).

boaks:
The sandbox server is now available 

[...]

PR #911, opened 2019-03-15
Skip generation of source and javadoc for demo-certs and legal.

boaks:


boaks:
On M14 the deployment of the `cf-pgutest-server` to maven central failed, which seems to be caused by the missing javadoc.

https://github.com/eclipse/californium/issues/870#issuecomment-471416721

Therefore I'm not sure, what is required to deploy modules without source and javadoc to maven central.  That issue requires more time to investigate.

PR #912, opened 2019-03-15
Try to define cipher suites order following best practices.

sbernard31:
Following : https://github.com/eclipse/californium/pull/904#discussion_r265738768

FInally this is more about order documentation, than changing order :grin:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
@@ -56,34 +56,53 @@ public enum CipherSuite {
 
 	// Cipher suites //////////////////////////////////////////////////
 
-	// PSK cipher suites, ordered by default preference, see getPskCiperSuites ///////////////////////////////////////////
-	TLS_NULL_WITH_NULL_NULL(0x0000, CertificateKeyAlgorithm.NONE, KeyExchangeAlgorithm.NULL, Cipher.NULL, MACAlgorithm.NULL, true),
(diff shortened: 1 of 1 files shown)


boaks:
LGTM!

Some notes:
GCM seems to use a "random explicit nonce", which makes it hard/impossible to ensure, that the nonce is not reused. With that, I'm not sure, if the security is effectively better than CCM.

CBC vulnerability: one of the countermeasures mentioned in:
http://www.isg.rhul.ac.uk/tls/Lucky13.html
"Add random time delays to CBC-mode decryption processing."

So the server-side, using java on multiple processors, will do a good job :-).
The client side, using c, may be more affected.

sbernard31:
> GCM seems to use a "random explicit nonce", which makes it hard/impossible to ensure, that the nonce is not reused (but, sure its very probable to be unique). With that, I'm not sure, if the security is effectively better than CCM.

From what I read and with my limited knowledge, I understand that both are pretty much as secure.

I put GCM first because I understood that it is faster on "big system" (PC/android) which is pretty much the Californium target, but I could also make sense to put CCM first as is the recommended for embedded. 

My idea was the limited device should probably not propose a GCM cipher and so in this case CCM will be used, in other case we use the "faster" GCM.

boaks:
Currently the order on the server-side is even ignored! 
It's only important, if it supported on the server side at all.
Only the client_hello order is used to find the first supported common cipher-suite.
I was already asked to make the selection also based on the servers order or customizable (pass in the client and server ordered supported cipher suites). But not for now.

As I mentioned in my PR #904, I assume that, if the order is important, someone will chose a own ordered supported list of cipher suites.

So, just merge it.

PR #914, opened 2019-03-18
Add GCM test.

boaks:
Also extend test of JVM support by key length.
Not all JVMs seems to support AES256.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -1034,11 +1034,13 @@ public final class DtlsConnectorConfig {
 		 * order) during the DTLS handshake when negotiating a cipher suite with
 		 * a peer.
 		 * 
-		 * @param cipherSuites the supported cipher suites in the order of preference
+		 * @param cipherSuites the supported cipher suites in the order of
+		 *            preference
 		 * @return this builder for command chaining
 		 * @throws NullPointerException if the given array is <code>null</code>
 		 * @throws IllegalArgumentException if the given array is empty or
-		 *             contains {@link CipherSuite#TLS_NULL_WITH_NULL_NULL}
+		 *             contains {@link CipherSuite#TLS_NULL_WITH_NULL_NULL}, or
+		 *             contains a cipher suite, not supported by the JVM.
 		 */
 		public Builder setSupportedCipherSuites(List<CipherSuite> cipherSuites) {
 			if (cipherSuites == null) {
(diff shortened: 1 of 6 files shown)


boaks:
Fixes missing AES 256 support on ci-eclipse jenkins with java 7.

boaks:
I would like to merge this ASAP :-).

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1774,6 +1781,11 @@ public DtlsConnectorConfig build() {
 				throw new IllegalStateException("Supported cipher suites must be set either " +
 						"explicitly or implicitly by means of setting the identity or PSK store");
 			}
+			for (CipherSuite cipherSuite : config.supportedCipherSuites) {
+				if (!cipherSuite.isSupported()) {
+					throw new IllegalStateException("cipher-suites " + cipherSuite + " is not supported by JVM!");
+				}
+			}
```
sbernard31: I suspect this test is not needed, because we already test it in setter and for default value `determineCipherSuitesFromConfig` should return only supported cipher ?

(but maybe it's clearer to keep it?)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
```diff
@@ -372,6 +372,25 @@ public static int getOverallMaxCiphertextExpansion() {
...
+	public static List<CipherSuite> getSupportedCipherSuites(List<CipherSuite> cipherSuites) {
+		List<CipherSuite> list = new ArrayList<>();
+		for (CipherSuite suite : cipherSuites) {
+			if (suite.isSupported()) {
+				list.add(suite);
+			}
+		}
+		return list;
+	}
+
```
sbernard31: Just in case, it is not intended : Eclipse says me that this method in not currently used.
(But I'm ok to add to the API)

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: My concern was "should return". I just want to make sure, that the JVM support doesn't make more trouble then intended. But yes, though `determineCipherSuitesFromConfig` only return supported ones, I may also sharp the javadoc.

boaks:
> (I don't know if this is intended but there is no test for PSK/GCM)

I will add one.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
boaks: I removed it.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: Though currently the unit tests of 2.0.x fails without this PR, I would go to merge it. If wanted, we can remove the redundant check later.

boaks:
"Green" again :-).

Issue #913, opened 2019-03-18
Redesign ByteId of cf-oscore to extend Bytes


boaks:
Branch 2.0.x:

In the meantime californium has a central implementation for byte array based ids:

https://github.com/eclipse/californium/blob/2.0.x/element-connector/src/main/java/org/eclipse/californium/elements/util/Bytes.java

I would appreciate, if redesign of `ByteId` extending `Bytes` is analyzed and if it works, a PR is created. 

See Token, SessionId, or ConnectionId for examples.

@rikard-sics 
May be you can spend some time in it :-).

rikard-sics:
> May be you can spend some time in it :-).

Sure, I can take a look at this for a future PR.

boaks:
Thanks!

rikard-sics:
I have now created a pull request to solve this issue:
https://github.com/eclipse/californium/pull/921

boaks:
Fixed with PR #921

PR #921, opened 2019-03-30
OSCORE: Changed ByteId to extend Bytes

rikard-sics:
Redesigned the ByteId class from cf-oscore to extend the existing Bytes class. It provides support for byte arrays as keys.

As reported in the following issue:
https://github.com/eclipse/californium/issues/913


boaks:
Thanks!

PR #924, opened 2019-04-04
Add cid generator.

boaks:
Use cid to support unique cids in cluster nodes.
Enables also to generate cids with encoded flexible length.

Uses alternative `attach` approach.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -170,6 +170,7 @@ import org.eclipse.californium.scandium.dtls.ClientHandshaker;
 import org.eclipse.californium.scandium.dtls.ClientHello;
 import org.eclipse.californium.scandium.dtls.Connection;
 import org.eclipse.californium.scandium.dtls.ConnectionId;
+import org.eclipse.californium.scandium.dtls.ConnectionIdGenerator;
 import org.eclipse.californium.scandium.dtls.ContentType;
 import org.eclipse.californium.scandium.dtls.DTLSFlight;
 import org.eclipse.californium.scandium.dtls.DTLSSession;
@@ -266,6 +267,12 @@ public class DTLSConnector implements Connector, RecordLayer {
 
 	private final List<Thread> receiverThreads = new LinkedList<Thread>();
 
+	/**
+	 * Configure connection id generator. May be {@code null}, if connection id
+	 * should not be supported.
+	 */
+	private final ConnectionIdGenerator connectionIdGenerator;
+
 	private InetSocketAddress lastBindAddress;
 	private int maximumTransmissionUnit = DEFAULT_IPV4_MTU;
 	private int inboundDatagramBufferSize = MAX_DATAGRAM_BUFFER_SIZE;
@@ -323,7 +330,6 @@ public class DTLSConnector implements Connector, RecordLayer {
 	public DTLSConnector(final DtlsConnectorConfig configuration, final SessionCache sessionCache) {
 		this(configuration,
 				new InMemoryConnectionStore(
-						configuration.getConnectionIdLength(),
 						configuration.getMaxConnections(),
 						configuration.getStaleConnectionThreshold(),
 						sessionCache).setTag(configuration.getLoggingTag()));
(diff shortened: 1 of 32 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -266,6 +267,8 @@
 
 	private final List<Thread> receiverThreads = new LinkedList<Thread>();
 
+	private final ConnectionIdGenerator connectionIdGenerator;
+
```
sbernard31: formating issue. (extra tab)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
```diff
@@ -275,6 +276,9 @@ public synchronized boolean put(final Connection connection) {
 			}
 			ConnectionId connectionId = connection.getConnectionId();
 			if (connectionId == null) {
+				if (connectionIdGenerator == null) {
+					throw new IllegalStateException("Connection id generator must be attached before!");
+				}
```
sbernard31: should we move this check in newConnectioId() ?

boaks:
> LGTM !

So I can close PR #923 and we continue with this approach?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
boaks: My idea was, to hide the private `newConnectioId` and report the public `put`. For me this makes it easier to document the requirements.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java
sbernard31: I thought that it would be better to get the check closer from where we use `connectionIdGenerator` attribute.
But we can let it like this if you prefer.

sbernard31:
> So I can close PR #923 and we continue with this approach?

Yep ! (I tried quickly my idea but I'm not sure this is really better)

boaks:
I extended the class javadoc in `ConnectionIdGenerator`. If possible, please give me your feedback. I would like to merge it in the late evening. That will make it easier to test the DTLS CID load-balancer setup :-).

I would also like to start the feedback collection for a 2.0.0-M15, are you OK with that?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ConnectionIdGenerator.java
```diff
@@ -0,0 +1,107 @@
...
+ * 
+ * Responsible for generating ID which identifies scandium connections in store.
+ * 
+ * By default, DTLS defined that IP address and port of the peer are used to
+ * identify the DTLS Connection. The DTLS connection ID draft defines a way to
+ * identify connection using Connection ID and so supports environments where IP
+ * address/port changes. See <a href=
+ * "https://tools.ietf.org/html/draft-ietf-tls-dtls-connection-id-03">draft-ietf-tls-dtls-connection-id-03</a>.
+ * 
+ * The draft enables the peers to chose flexible the level of support or usage.
```
sbernard31: The draft enables the peers to chose **~flexible~** the level of support or usage.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ConnectionIdGenerator.java
```diff
@@ -0,0 +1,107 @@
...
+ * The dtls client peer informs the dtls server peer about its preference using
+ * a new HELLO_EXTENSION in it's CLIENT_HELLO.
+ * <p>
+ * The dtls client can chose:
+ * <dl>
+ * <dt>client doesn't support it</dt>
+ * <dd>the new extension is not included in the hello</dd>
```
sbernard31: hello <=> CLIENT_HELLO ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ConnectionIdGenerator.java
```diff
@@ -0,0 +1,107 @@
...
+ * hello</dd>
+ * </dl>
+ * <p>
+ * If the client doesn't support it, the server must reply with a server hello
+ * without the new hello extension, regardless of the configuration.
+ * <p>
+ * If the client supports or uses it, the server can chose.
+ * <dl>
+ * <dt>server doesn't support it</dt>
+ * <dd>the new extension is not included in the hello</dd>
```
sbernard31: hello <=> SERVER_HELLO ?

PR #925, opened 2019-04-08
Add unix demo setup.

boaks:
Issue #870 
I started to fill more information about running californium into the wiki. 
My plan would be to create links in the wikis to the files contained in this PR. 
Alternatively to the wiki we could prepare md files within this project.


sbernard31:
No objections :)

About "Wiki" vs "md files in repo", I would say it depends if documentation lifecycle is linked to the project lifecycle ?

Just some questions.
1. Why adding a maven module for this ?
2. Your using LVS for default loadbalancing and iptables rules for specifics CID rules, that's the idea ? Maybe I should wait you finished your documentation before to ask this ^^
3. Any reason why install.sh :
 - is a bash script and other sh script ?
 - use sudo internally where others ask for :  _"Requires su rights, execute with sudo!"_

boaks:
> About "Wiki" vs "md files in repo", I would say it depends if documentation lifecycle is linked to the project lifecycle ?

FMPOV, the future will show it. If it get's more linked to the project lifecycle, i will migrate it into .md in the repository.

boaks:
> 1.  Why adding a maven module for this ?

The idea was to be able to deploy these scripts, but If you feel that doesn't make sense, then I remove the pom.
 
> 2. Your using LVS for default loadbalancing and iptables rules for specifics CID rules, that's the idea ? Maybe I should wait you finished your documentation before to ask this ^^

The handshake itself doesn't use a cid (I discussed that, but the decision was to go without that change). Therefore the lvs works as standard UDP load-balancer, which should be no issue for the handshake itself. The cluster-node, which negotiates the handshake with the peer uses then a cid with a fixed node id as prefix (0x01|0x02 ....). If in a quiet period the lvs route times out and then a tls12_cid record is received, that prefix (0x01|0x02) is used to establish a (short time) DNAT route to the node again, which negotiated the handshake. So yes, that's intended. It doesn't consider explicitly a "fail over" scenario. That may work, if a peer, which doesn't receive responses/ACKs would resume the session as usual and the lvs will find a working node. But that's out of scope.
   
> 3.  Any reason why install.sh :

It's not intended to be executed, instead it's a source to copy lines.

sbernard31:
2. 3. Thx for explanation ! :)

1. "The idea was to be able to deploy these scripts" => I can not see how but If you think it's useful, that's ok for me.

Issue #931, opened 2019-04-17
Race condition when processing new CLIENT_HELLO


AndreasHuber-CH:
When doing a performance test we observe occasionally the following issue:
- The client repeats the CLIENT_HELLO after 1 second because it didn't get a response from the server.
- The server processes both CLIENT_HELLO simultanously (I still have to determine why this can happen...)

Then as you can see in the log below, two new Connections get allocated by Scandium for the same handshake one replacing the other. And Scandium responds with two SERVER_HELLO, SERVER_HELLO_DONE messages that contain two different values for the ServerRandom.

Later on (not in the logs below) the handshake fails with MAC verification errors because the client and server did apparently not use the ServerRandom value out of the two returned by Scandium.

```
12:19:59.973 org.eclipse.californium.scandium.DTLSConnector FINE: Processing new CLIENT_HELLO from peer [/10.3.12.25:61533]
12:19:59.973 org.eclipse.californium.scandium.dtls.Record FINE: Parsing message without a session
12:19:59.973 org.eclipse.californium.scandium.dtls.Record FINE: Parsing HANDSHAKE message plaintext [null]
12:19:59.973 org.eclipse.californium.scandium.dtls.HelloExtensions FINE: Peer included an unknown extension type code [13] in its Hello message
12:19:59.973 org.eclipse.californium.scandium.dtls.InMemoryConnectionStore FINE: connection: add CID=1127952BF1 (size 400)
12:19:59.973 org.eclipse.californium.scandium.dtls.InMemoryConnectionStore FINE: connection: CID=1127952BF1 - /10.3.12.25:61533 added!
12:19:59.973 org.eclipse.californium.scandium.DTLSConnector FINE: Received Handshake (22) record from peer [/10.3.12.25:61443]
12:19:59.973 org.eclipse.californium.scandium.DTLSConnector FINE: Processing new CLIENT_HELLO from peer [/10.3.12.25:61533]
12:19:59.973 org.eclipse.californium.scandium.dtls.Record FINE: Parsing message without a session
12:19:59.973 org.eclipse.californium.scandium.dtls.Record FINE: Parsing HANDSHAKE message plaintext [null]
12:19:59.973 org.eclipse.californium.scandium.dtls.HelloExtensions FINE: Peer included an unknown extension type code [13] in its Hello message
12:19:59.973 org.eclipse.californium.scandium.dtls.InMemoryConnectionStore FINE: connection: add CID=B1E211AF41 (size 401)
[...]

PR #932, opened 2019-04-17
Fix race condition of CLIENT_HELLO.

boaks:
Issue #931

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
@@ -40,6 +40,7 @@ package org.eclipse.californium.scandium.dtls;
 
 import java.io.IOException;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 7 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
```diff
@@ -491,13 +511,15 @@ public void sessionEstablished(Handshaker handshaker, DTLSSession session) throw
 		@Override
 		public void handshakeCompleted(Handshaker handshaker) {
 			if (ongoingHandshake.compareAndSet(handshaker, null)) {
+				startedByClient.set(null);
```
boaks: Reset the random has the same effect as ending the handshake in the implementation before. 
With 32 bytes random, I'm not sure, if this should be done at all or at least delayed. 
With a CLIENT_HELLO duplicate "on the way" and a reset random, a new handshake may be trigger instead of ignoring the duplicate.

PR #935, opened 2019-04-18
Shutdown attached executor on destroy even if server is stopped.

sbernard31:
The test was useful to check this in eclipse but :
1. I don't know how it behave if junit tests are launched in parallel
2. It fails because like this in maven : 
```
[ERROR] Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.034 s <<< FAILURE! - in org.eclipse.californium.core.server.CoapServerTest
[ERROR] testStartStopDestroy(org.eclipse.californium.core.server.CoapServerTest)  Time elapsed: 0.021 s  <<< ERROR!
java.io.IOError: java.net.SocketException: Use org.eclipse.californium.elements.rule.NetworkRule to define DatagramSocket behaviour!
	at org.eclipse.californium.core.server.CoapServerTest.testStartStopDestroy(CoapServerTest.java:43)
Caused by: java.net.SocketException: Use org.eclipse.californium.elements.rule.NetworkRule to define DatagramSocket behaviour!
	at org.eclipse.californium.core.server.CoapServerTest.testStartStopDestroy(CoapServerTest.java:43)
```

So we can not merge it, if you prefer.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
@@ -287,27 +287,36 @@ public class CoapServer implements ServerInterface {
 		LOGGER.info("Destroying server");
 		// prevent new tasks from being submitted
 		try {
-			if (running && !detachExecutor) {
-				executor.shutdown(); // cannot be started again
-				try {
-					// wait for currently executing tasks to complete
-					if (!executor.awaitTermination(1, TimeUnit.SECONDS)) {
-						// cancel still executing tasks
-						// and ignore all remaining tasks scheduled for later
-						List<Runnable> runningTasks = executor.shutdownNow();
-						if (runningTasks.size() > 0) {
-							// this is e.g. the case if we have performed an incomplete blockwise transfer
-							// and the BlockwiseLayer has scheduled a pending BlockCleanupTask for tidying up
-							LOGGER.debug("ignoring remaining {} scheduled task(s)", runningTasks.size());
+			if (!detachExecutor)
+				if (running) {
+					executor.shutdown(); // cannot be started again
+					try {
+						// wait for currently executing tasks to complete
+						if (!executor.awaitTermination(1, TimeUnit.SECONDS)) {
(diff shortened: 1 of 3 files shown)


boaks:
LGTM

sbernard31:
Any idea about the NetworkRule exception ?

boaks:
I added some small changes to branch "fix_destroy_rule".

> java.net.SocketException: Use org.eclipse.californium.elements.rule.NetworkRule to define DatagramSocket behaviour!

The nasty stuff of the "direct in process socket" using `DatagramSocket.setDatagramSocketImplFactory(factory)` is, it's not revertible. Therefore it's important to mark all test, which use a DatagramSocket, if they run NATIVE, DIRECT or with BOTH. Therefore maven use DIRECT for all none `NativeDatagramSocketImplRequired` and after the NATIVE for the `NativeDatagramSocketImplRequired`.  Currently all test support the DIRECT. There was a period, where the multicast stuff wasn't supported.
Using `CoapNetworkRule` ensures additionally, that the `NetworkConfig` is always reset and so the changes do not span tests o depend on the order of tests.

boaks:
If the "general scheduler" is initialized in `@BeforeClass`, the test counting threads gets much simpler without redesign the "general scheduler".
See the small changes in branch "fix_destroy_rule".

sbernard31:
Your fix sounds good. Thx @boaks.

Just by curiosity, the DIRECT mode was created to limit issue on overloaded Jenkins instance, right ?

boaks:
> Just by curiosity, the DIRECT mode was created to limit issue on overloaded Jenkins instance, right ?

Kai and I suspected the "normal UDP drop" to be one cause for randomly failing unit tests. Therefore I tried to fix that cause by this "DIRECT (in process communication)". Unfortunately, on the eclipse.jenkins,  the results where only little better.

My current assumption is, that californium tests, which assume, that something happens in 2s, but the jenkins pause for 3s, fail. To mitigate that, I now run the test also on my private dedicated jenkins instances and there my results are much better then on the eclipse one.

Issue #936, opened 2019-04-18
some concerns about `ExecutorsUtil`


sbernard31:
Looking at #934, I looked at `ExecutorsUtil` and I have some concerns.

1.  the static block create ScheduledThreadPoolExecutor which can not be shutdown and which is shared by default.
2.  Experimental SplitScheduledThreadPoolExecutor seems to be activated by default, and can not be deactivated.

boaks:
That mainly caused by the past, where stop or restart of components wasn't considered to be that important. My experience with the `DTLSConnector` and `UDPConnector` was,  that fixing (reported for pending messages  by callback) that is enormous time consuming. Therefore I didn't pay too much attention to stop and restarts and spent my time in other issues.

So, do you have only concerns about have that "general scheduler" always running? Or is there a current issue? 

The "experimental" stuff worked well for a couple of months. 
Or do you have issues except of the javadoc?

sbernard31:
> So, do you have only concerns about have that "general scheduler" always running? Or is there a current issue?

The concerns here are not linked to any bug.
But having a "shared", "always running"  `ScheduledThreadPoolExecutor` sounds not so clean.

> The "experimental" stuff worked well for a couple of months.

There is 2 mode in the code but not easy way to switch from one to another... so any easy way to compare it.

sbernard31:
About removing the static  `ScheduledThreadPoolExecutor` from `ExecutorsUtil`.

1) Do we really need a separated thread pool ? Why not just using the `CoapServer` one ?
2) If we really need it, I experiment some code about this. You could have a look at #937.

boaks:
I have two office days left before my  vacation.  And the office days are filled with a lot of "none open source" stuff. So it's just the wrong time for deep discussion.

1.  (of your initial comment)
```
	/**
	 * Get general scheduled executor.
	 * 
	 * Intended to be used for rare executing timers (e.g. cleanup tasks).
	 * 
	 * @return scheduled executor service
	 */
	public static ScheduledThreadPoolExecutor getScheduledExecutor() {
		return scheduler;
	}
```

I chose that to not pass around the scheduled executor of the `CoapEndpoint`.  At least at the time,  when I tested that idea, the PRs have been large enough. And now my feeling is, that there are other issues with more priority.

2. 
```
SplitScheduledThreadPoolExecutor
```
[...]

PR #727, opened 2018-08-22
Replace striped executor with serial executor.

boaks:
Signed-off-by: Achim Kraus <achim.kraus@bosch-si.com>

PR #937, opened 2019-04-19
add way to customize timer Executor for CoapServer

sbernard31:
(Not intended to be merged)

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -316,6 +316,19 @@ public class CoapClient {
 		return this;
 	}
 
+	private ScheduledThreadPoolExecutor getSecondaryExecutor() {
+		if (secondaryExecutor == null) {
+			synchronized (this) {
+				if (secondaryExecutor == null) {
+					secondaryExecutor = new ScheduledThreadPoolExecutor(1,
+							new NamedThreadFactory("CoapClient(secondary)#"));
+				}
+				this.detachExecutor = false;
+			}
+		}
+		return secondaryExecutor;
+	}
+
 	/**
 	 * Gets the endpoint this client uses.
 	 *
@@ -1170,7 +1183,7 @@ public class CoapClient {
 		if (request.getOptions().hasObserve()) {
 			assignClientUriIfEmpty(request);
 			Endpoint outEndpoint = getEffectiveEndpoint(request);
-			CoapObserveRelation relation = new CoapObserveRelation(request, outEndpoint);
+			CoapObserveRelation relation = new CoapObserveRelation(request, outEndpoint, getSecondaryExecutor());
 			// add message observer to get the response.
 			ObserveMessageObserverImpl messageObserver = new ObserveMessageObserverImpl(handler, request.isMulticast(), relation);
 			request.addMessageObserver(messageObserver);
@@ -1205,7 +1218,7 @@ public class CoapClient {
 		if (request.getOptions().hasObserve()) {
 			assignClientUriIfEmpty(request);
 			Endpoint outEndpoint = getEffectiveEndpoint(request);
-			CoapObserveRelation relation = new CoapObserveRelation(request, outEndpoint);
+			CoapObserveRelation relation = new CoapObserveRelation(request, outEndpoint, getSecondaryExecutor());
 			// add message observer to get the response.
 			ObserveMessageObserverImpl messageObserver = new ObserveMessageObserverImpl(handler, request.isMulticast(), relation);
 			request.addMessageObserver(messageObserver);
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
@@ -50,6 +50,7 @@ import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.californium.elements.exception.ConnectorException;
@@ -105,6 +106,9 @@ public class CoapClient {
 
 	/** The client-specific executor service. */
 	private ExecutorService executor;
+
+	/** Scheduled executor intended to be used for rare executing timers (e.g. cleanup tasks). */
+	private ScheduledThreadPoolExecutor secondaryExecutor;
 
 	/**
 	 * Indicate, it the client-specific executor service is detached, or
@@ -252,16 +256,19 @@ public class CoapClient {
 	 */
 	public CoapClient useExecutor() {
 		boolean failed = true;
-		ExecutorService executor = ExecutorsUtil.newFixedThreadPool(1, new NamedThreadFactory("CoapClient#")); //$NON-NLS-1$
+		ExecutorService executor = ExecutorsUtil.newFixedThreadPool(1, new NamedThreadFactory("CoapClient(main)#")); //$NON-NLS-1$
+		ScheduledThreadPoolExecutor secondaryExecutor = new ScheduledThreadPoolExecutor(1, new NamedThreadFactory("CoapClient(secondary)#"));
 		synchronized (this) {
-			if (this.executor == null) {
+			if (this.executor == null && this.secondaryExecutor == null) {
 				this.executor = executor;
+				this.secondaryExecutor = secondaryExecutor;
 				this.detachExecutor = false;
 				failed = false;
 			}
 		}
 		if (failed) {
 			executor.shutdownNow();
+			secondaryExecutor.shutdown();
 			throw new IllegalStateException("Executor already set or used!");
 		}
 
@@ -277,41 +284,28 @@ public class CoapClient {
 	}
 
 	/**
-	 * Sets the executor service for this client.
+	 * Sets the executor services for this client.
 	 * 
-	 * All handlers will be invoked by this executor. The executor will shutdown
-	 * on {@link #shutdown()}.
-	 * 
-	 * @param executor the executor service
-	 * @return the CoAP client
-	 * @throws IllegalStateException if executor is already set or used.
-	 * @throws NullPointerException if provided executor is null
-	 */
-	public CoapClient setExecutor(ExecutorService executor) {
-		return setExecutor(executor, false);
-	}
-
-	/**
-	 * Sets the executor service for this client.
-	 * 
-	 * All handlers will be invoked by this executor. The executor will shutdown
+	 * All handlers will be invoked by the main executor. The executors will shutdown
 	 * on {@link #shutdown()}, if not detached.
 	 * 
-	 * @param executor the executor service
+	 * @param executor the main executor service
+	 * @param secondaryExecutor intended to be used for rare executing timers (e.g. cleanup tasks).
 	 * @param detach {@code true}, if the executor is not shutdown on
 	 *            {@link #shutdown()}, {@code false}, otherwise.
 	 * @return the CoAP client
 	 * @throws IllegalStateException if executor is already set or used.
-	 * @throws NullPointerException if provided executor is null
+	 * @throws NullPointerException if provided executors are null
 	 */
-	public CoapClient setExecutor(ExecutorService executor, boolean detach) {
-		if (executor == null) {
-			throw new NullPointerException("Executor must not be null!");
+	public CoapClient setExecutors(ExecutorService executor, ScheduledThreadPoolExecutor secondaryExecutor, boolean detach) {
+		if (executor == null || secondaryExecutor == null) {
+			throw new NullPointerException("Executors must not be null!");
 		}
 		boolean failed = true;
 		synchronized (this) {
-			if (this.executor == null) {
+			if (this.executor == null && this.secondaryExecutor == null) {
 				this.executor = executor;
+				this.secondaryExecutor = secondaryExecutor;
 				this.detachExecutor = detach;
 				failed = false;
 			}
@@ -1023,20 +1017,27 @@ public class CoapClient {
 	}
 
 	/**
-	 * Shutdown the client-specific executor service, when not detached. Only
-	 * needed if {@link #useExecutor()} or {@link #setExecutor(ExecutorService)}
-	 * are used (i.e., a client-specific executor service was set).
+	 * Shutdown the client-specific executor service, when not detached. Always
+	 * needed unless you used detached executor.
 	 */
 	public void shutdown() {
 		ExecutorService executor;
+		ExecutorService secondaryExecutor;
 		boolean shutdown;
 		synchronized (this) {
 			executor = this.executor;
+			secondaryExecutor = this.secondaryExecutor;
 			shutdown = !this.detachExecutor;
 			this.executor = null;
+			this.secondaryExecutor = null;
 		}
-		if (shutdown && executor != null) {
-			executor.shutdownNow();
+		if (shutdown) {
+			if (executor != null) {
+				executor.shutdownNow();
+			}
+			if (secondaryExecutor != null) {
+				secondaryExecutor.shutdownNow();
+			}
 		}
 	}
 
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
@@ -31,7 +31,6 @@ import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -114,6 +113,9 @@ public class CoapServer implements ServerInterface {
 
 	/** The executor of the server for its endpoints (can be null). */
 	private ScheduledExecutorService executor;
+
+	/** Scheduled executor intended to be used for rare executing timers (e.g. cleanup tasks). */
+	private ScheduledExecutorService secondaryExecutor;
 	/**
 	 * Indicate, it the server-specific executor service is detached, or
 	 * shutdown with this server.
@@ -178,40 +180,31 @@ public class CoapServer implements ServerInterface {
 		}
 	}
 
-	/**
-	 * Sets the executor service to use for running tasks in the protocol stage.
-	 * 
-	 * @param executor The thread pool to use.
-	 * @throws IllegalStateException if this server is running and a new
-	 *             executor is provided.
-	 */
-	public void setExecutor(final ScheduledExecutorService executor) {
-		setExecutor(executor, false);
-	}
+	public synchronized void setExecutors(final ScheduledExecutorService mainExecutor,
+			final ScheduledExecutorService secondaryExecutor, final boolean detach) {
+		if (mainExecutor == null || secondaryExecutor == null) {
+			throw new NullPointerException("executors must not be null");
+		}
+		if (this.executor == mainExecutor && this.secondaryExecutor == secondaryExecutor) {
+			return;
+		}
+		if (running) {
+			throw new IllegalStateException("executor service can not be set on running server");
+		}
 
-	/**
-	 * Sets the executor service to use for running tasks in the protocol stage.
-	 * 
-	 * @param executor The thread pool to use.
-	 * @param detach {@code true}, if the executor is not shutdown on
-	 *            {@link #destroy()}, {@code false}, otherwise.
-	 * @throws IllegalStateException if this server is running and a new
-	 *             executor is provided.
-	 */
-	public synchronized void setExecutor(final ScheduledExecutorService executor, final boolean detach) {
-		if (this.executor != executor) {
-			if (running) {
-				throw new IllegalStateException("executor service can not be set on running server");
-			} else {
-				if (!this.detachExecutor && this.executor != null) {
-					this.executor.shutdownNow();
-				}
-				this.executor = executor;
-				this.detachExecutor = detach;
-				for (Endpoint ep : endpoints) {
-					ep.setExecutors(executor, ExecutorsUtil.getScheduledExecutor());
-				}
+		if (!this.detachExecutor) {
+			if (this.executor != null) {
+				this.executor.shutdownNow();
 			}
+			if (this.secondaryExecutor != null) {
+				this.secondaryExecutor.shutdownNow();
+			}
+		}
+		this.executor = mainExecutor;
+		this.secondaryExecutor = secondaryExecutor;
+		this.detachExecutor = detach;
+		for (Endpoint ep : endpoints) {
+			ep.setExecutors(this.executor, this.secondaryExecutor);
 		}
 	}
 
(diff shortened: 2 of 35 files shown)


boaks:
At the time of introducing the "general scheduled executor" i just didn't want to change so many files for passing the executor around.

boaks:
Seems to target the scope of the coap-server only and leaves the other usage unchanged?

sbernard31:
> Seems to target the scope of the coap-server only and leaves the other usage unchanged

This not not finished. I would keep the PR small to let you know the general idea.

sbernard31:
The idea is just to start to remove the `ScheduledThreadPoolExecutor` by creating a new attribute.

boaks:
For me, this will be, when completed, neither simpler nor does it pay off.
Anyway, if you prefer it, I have no objection, if you change it.

sbernard31:
>For me, this will be, when completed, neither simpler nor does it pay off.

A clearer explanation here : https://github.com/eclipse/californium/issues/936#issuecomment-491751159

sbernard31:
I think the PR is in a re-viewable state. I tried to decompose it in a lot of short consistent commits.

I finally choose to implement the 2 executor way (1 main and 1 secondary).
So, behavior should not change so much and we should be able to compare with only 1 executor (just by using the same for main and secondary)
If later, we see that there is no gain we can modify the code again to use only one executor. 

For now, I didn't run the benchmark, I plan to do that this afternoon.

I introduce a little behavior change about coapClient.shutdown()
Now the call of shutdown is required even if you don't use useExecutors or setExecutors (..., ..., false). 
(to stop secondary executor)

boaks:
First fast impression: 
A lot of stuff :-). And I miss the update for `ConcurrentCoapResource`.

sbernard31:
> A lot of stuff :-)

It's often like this, the more we wait to remove static stuff the more it's hard to remove it.
I agree this is lot of stuff, but I really tried to make as readable as possible.

> I miss the update for ConcurrentCoapResource

To be honest I missed it but now I looked at it,  I'm not sure this make sense to add anything ?
You would like to add a setter for the `secondaryExecutor` ? Before this executor was shared by all the JVM now by default it is shared by all resources of 1 coapServer. 
I think it's ok for now :thinking: (knowing that nothing prevent a user to change this behavior by extending this class)

boaks:
`CoapClient setExecutors` doesn't specify the requirements for the `secondaryExecutor`. And the implementation also provides no hint, if `null` is supported. 

By the way, the `CoapResource.createClient()` will fail without executor, but that's not related with this PR.

boaks:
CoapEndpoint:
`this.secondaryExecutor = secondaryExecutor != null ? secondaryExecutor : mainExecutor;`

CoapServer:
`if (mainExecutor == null || timerExecutor == null) {`

So `null` or no `null`, that seems to be the question.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -252,10 +256,12 @@ public CoapClient setDestinationContext(EndpointContext peerContext) {
 	 */
 	public CoapClient useExecutor() {
 		boolean failed = true;
-		ExecutorService executor = ExecutorsUtil.newFixedThreadPool(1, new NamedThreadFactory("CoapClient#")); //$NON-NLS-1$
+		ExecutorService executor = ExecutorsUtil.newFixedThreadPool(1, new NamedThreadFactory("CoapClient(main)#")); //$NON-NLS-1$
+		ScheduledThreadPoolExecutor secondaryExecutor = new ScheduledThreadPoolExecutor(1, new NamedThreadFactory("CoapClient(secondary)#"));
```
boaks:
```
if (failed) {
	executor.shutdownNow();
	throw new IllegalStateException("Executor already set or used!");
}
```

Should that be extended for the `secondaryExecutor` ?

boaks: any reason not to use the `ExecutorsUtil.newDefaultSecondaryScheduler`?

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -322,6 +315,19 @@ public CoapClient setExecutor(ExecutorService executor, boolean detach) {
 		return this;
 	}
 
+	private ScheduledThreadPoolExecutor getSecondaryExecutor() {
+		if (secondaryExecutor == null) {
+			synchronized (this) {
+				if (secondaryExecutor == null) {
+					secondaryExecutor = new ScheduledThreadPoolExecutor(1,
```
boaks: 1. any reason not to use the `ExecutorsUtil.newDefaultSecondaryScheduler`?
2. the "double check" pays off?

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -322,6 +315,19 @@ public CoapClient setExecutor(ExecutorService executor, boolean detach) {
...
+							new NamedThreadFactory("CoapClient(secondary)#"));
+				}
+				this.detachExecutor = false;
```
boaks: If a executor and a secondaryExecutor == null is used, the this may overwrite the set `detach`.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -217,9 +217,18 @@
 	/** Parser to convert datagrams to messages. */
 	private final DataParser parser;
 
+	/** A store containing data about message exchanges. */
+	private final MessageExchangeStore exchangeStore;
+
+	/** A Store containing data about observation. */
+	private ObservationStore observationStore;
```
boaks: `final exchangeStore` and non-final `observationStore`, is that intended?

sbernard31:
> CoapClient setExecutors doesn't specify the requirements for the secondaryExecutor. And the implementation also provides no hint, if null is supported.

I missed that, I would go for "null not supported" like the other executor.

> By the way, the CoapResource.createClient() will fail without executor, but that's not related with this PR.

I know that. I was writing a test about that but as this is not related this PR I didn't include in it.

> So null or no null, that seems to be the question.

Both was my response ^^, I choose to be more restrictive for CoapServer API than for CoapEndpoint.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sbernard31: yep

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -449,14 +459,18 @@ public synchronized boolean isStarted() {
...
 
 	@Override
-	public synchronized void setExecutor(final ScheduledExecutorService executor) {
-		if (this.executor != executor) {
-			if (started) {
-				throw new IllegalStateException("endpoint already started!");
-			}
-			this.executor = executor;
-			this.coapstack.setExecutor(executor);
+	public void setExecutors(ScheduledExecutorService mainExecutor, ScheduledExecutorService secondaryExecutor) {
```
boaks: Please provide javadoc for the intended behaviour!

Especially document, if mainExecutor / secondaryExecutor could be `null` and what the intended result.

californium-core/src/main/java/org/eclipse/californium/core/network/Endpoint.java
```diff
@@ -82,16 +82,17 @@
...
-	 * Sets the executor for this endpoint and all its components.
+	 * Sets executors for this endpoint and all its components.
 	 *
-	 * The executor is not managed by the endpoint, it must be shutdown
+	 * Executors are not managed by the endpoint, it must be shutdown
 	 * externally, if the resource should be freed.
 	 *
-	 * @param executor the new executor
+	 * @param mainExecutor executors used for main tasks
+	 * @param secondaryExecutor intended to be used for rare executing timers (e.g. cleanup tasks).
```
boaks: Please document the `null` expectations.

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
```diff
@@ -156,8 +162,8 @@ public synchronized void start() {
 		int healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL,
 				NetworkConfigDefaults.DEFAULT_HEALTH_STATUS_INTERVAL); // seconds
 
-		if (healthStatusInterval > 0 && HEALTH_LOGGER.isDebugEnabled()) {
-			statusLogger = ExecutorsUtil.getScheduledExecutor().scheduleAtFixedRate(new Runnable() {
+		if (healthStatusInterval > 0 && HEALTH_LOGGER.isDebugEnabled() && executor != null) {
```
boaks: So, if the secondary gets `null` (how ever), health status gets disabled.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sbernard31: 1. Because for a client I feel that 1 thread was enough by default.
2. I did that to avoid to go in the synchronized block each time we call `observe`. Do you have any idea to avoid it ? maybe you propose to always  go in the synchronized block ?

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sbernard31: With https://github.com/eclipse/californium/pull/937#issuecomment-493077130 this should not be an issue anymore.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
sbernard31: nope, I will fix it by adding final keyword

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
boaks: 2. I guess, running the benchmarks will show no difference, if the double check is removed.

californium-core/src/main/java/org/eclipse/californium/core/observe/InMemoryObservationStore.java
sbernard31: yep

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
sbernard31: I will add this on `Endpoint` interface but keep in mind that maybe we will remove all of this if finally tests show that using same executor does not change anything.

californium-core/src/main/java/org/eclipse/californium/core/network/Endpoint.java
sbernard31: That's the same comment as above, but a bit better as it ends with a `.` instead of a `!` :grin:

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sbernard31: Currently I tried to run benchmarks on 2.0.x branch (so without my modifications) and it blocks on coap+tcp run :/

Do you want we test it before to remove it or do you want I remove it now ? (personally I have no strong opinion about that)

boaks:
> To be honest I missed it but now I looked at it, I'm not sure this make sense to add anything ?
> You would like to add a setter for the secondaryExecutor ? Before this executor was shared by all the JVM now by default it is shared by all resources of 1 coapServer.
> I think it's ok for now 🤔 (knowing that nothing prevent a user to change this behavior by extending this class)

That executor stuff is in `CoapResource` somehow implemented in a irritating way. The `CoapResource` will always return `null` unless a parent implementation overwrites it. With that, the createClient() will not work, if 
> I missed that, I would go for "null not supported" like the other executor.

gets true.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
boaks: With netty 4.1.34? If so, please update 2.0.x again. If not, just comment out the tcp stuff in the benchmark.sh.

sbernard31:
I fix all your comments. Thx for the review.

About `null` or `not null`, I finally choose `not null` everywhere this is more consistent. (javadoc updated)

boaks:
My first result:

```
org.eclipse.californium.core.network.ExchangeCompleteException: Exchange[L38, complete] already complete!
        at org.eclipse.californium.core.network.Exchange.setComplete(Exchange.java:818)
        at org.eclipse.californium.core.network.stack.ExchangeCleanupLayer.receiveResponse(ExchangeCleanupLayer.java:86)
        at org.eclipse.californium.core.network.stack.ObserveLayer.receiveResponse(ObserveLayer.java:158)
        at org.eclipse.californium.core.network.stack.BlockwiseLayer.receiveResponse(BlockwiseLayer.java:702)
        at org.eclipse.californium.core.network.stack.ReliabilityLayer.receiveResponse(ReliabilityLayer.java:287)
        at org.eclipse.californium.core.network.stack.AbstractLayer.receiveResponse(AbstractLayer.java:89)
        at org.eclipse.californium.core.network.stack.BaseCoapStack.receiveResponse(BaseCoapStack.java:139)
        at org.eclipse.californium.core.network.CoapEndpoint$1.receiveResponse(CoapEndpoint.java:256)
        at org.eclipse.californium.core.network.UdpMatcher$4.run(UdpMatcher.java:397)
        at org.eclipse.californium.elements.util.SerialExecutor$1.run(SerialExecutor.java:277)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
```

And the first "plain" coap benchmarks seems to be significant slower (on win10, java8).
I will see, if this result is the same on ubuntu 18.04.

sbernard31:
> And the first "plain" coap benchmarks seems to be significant slower (on win10, java8).

Interesting :thinking: , what did you call significant slower ?

About the exception you think this is related to this PR ?

boaks:
2.0.x:
> 29878 reqs/s

This PR:
> 18717 reqs/s

We will see, what my PC at home gets.

boaks:
> About the exception you think this is related to this PR ?

I'm not sure about that. It will take several runs to see, if this PR changes the probability. 
And currently I'm still busy with other tasks. May be, I can spent the time next week to test that.

sbernard31:
ouch it's really a significant lost ... :scream: 

On my side I was not able to observe significant changes.
(I ran tests on an old machine on debian stable : Intel® Core™ i7-2620M CPU @ 2.70GHz × 4 / 8GB RAM)

Detailed results :  
Before this PR commit 75f44bb9 : [statistics-75f44bb9.log](https://github.com/eclipse/californium/files/3191796/statistics-75f44bb9.log)
With this PR commit ccf4cc11 : [statistics-ccf4cc11-dynamicExecutor.log](https://github.com/eclipse/californium/files/3191797/statistics-ccf4cc11-dynamicExecutor.log)
With this PR commit ccf4cc11 without double check in coap client : [statistics-ccf4cc11-no2c.log](https://github.com/eclipse/californium/files/3191799/statistics-ccf4cc11-dynamicExecutor-withoutdoublechek.log)

boaks:
Sometimes my "office PC" produces results, which could not be explained :-).
I will retest that at home.

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
```diff
@@ -1023,20 +1030,26 @@ public CoapObserveRelation observe(CoapHandler handler, int accept) {
...
 		boolean shutdown;
 		synchronized (this) {
 			executor = this.executor;
+			secondaryExecutor = this.secondaryExecutor;
 			shutdown = !this.detachExecutor;
 			this.executor = null;
 		}
-		if (shutdown && executor != null) {
-			executor.shutdownNow();
+		if (shutdown) {
```
boaks: Is it intended, not to reset the `this.secondaryExecutor = null`?

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
```diff
@@ -178,41 +180,31 @@ public CoapServer(final NetworkConfig config, final int... ports) {
...
-	 * @param executor The thread pool to use.
-	 * @throws IllegalStateException if this server is running and a new
-	 *             executor is provided.
-	 */
-	public void setExecutor(final ScheduledExecutorService executor) {
-		setExecutor(executor, false);
-	}
+public synchronized void setExecutors(final ScheduledExecutorService mainExecutor, final ScheduledExecutorService secondaryExecutor,  final boolean detach) {
+		if (mainExecutor == null || secondaryExecutor == null) {
+			throw new IllegalArgumentException("executors must not be null");
```
boaks: `NullPointerException` ?

boaks: Indent missing (source formatting)?

boaks:
My results from my "home PC and C"I:
- the benchmarks differs more from "run to run" than for the versions.
- the CI failure rate shows also no difference

[2.0.x statistics.log](https://github.com/eclipse/californium/files/3194963/statistics.log)
[This PR statistics.log](https://github.com/eclipse/californium/files/3194964/statistics.log)
[This PR without double check statistics.log](https://github.com/eclipse/californium/files/3194965/statistics.log)

All benchmarks are executed twice with the server kept running.

> I will add this on Endpoint interface but keep in mind that maybe we will remove all of this if finally tests show that using same executor does not change anything.

I'm not really sure, if that would case changes again. if so, could we consider to merge this after M15 (I scheduled it for the 24. May)?

californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java
sbernard31: I will fix this.

californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
sbernard31: About `NullPointerException` vs `IllegalArgumentException`.

I had in mind the Apaches commons lang way: https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/Validate.html (`IllegalArgument` is used for validate.notNull)

But in recent version NPE is now used : https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/Validate.html

And it seems that in californium NPE is preferred too.

So I will change it.

(I will fix formatting)

sbernard31:
About integration, as I changed lot of code. I would be more comfortable if we integrate this PR just after a milestone release. (so after M15 sounds good)

sbernard31:
Should I merge this in 2.0.x now ?

boaks:
If you want. 

My feeling is, that, if you consider to test it with an other "executor configuration", which changes the API again, I would prefer, that you do that test before. Alternatively, you may encapsulate the "executor configuration" in something as "ExecutionEnvironment" pass this around. At least the "only forwarding" parts will then keep their API.

sbernard31:
> if you consider to test it with an other "executor configuration".

I would like to test if secondary executor is really needed, but there is nothing urgent now and having a release with the current version could help to test easily in production (I mean using same executor for both main and secondary one) and go back with 2 executors if this is not so good. 

> Alternatively, you may encapsulate the "executor configuration" in something as "ExecutionEnvironment" pass this around. At least the "only forwarding" parts will then keep their API.

This PR aims to remove static executor with minimal behavior changes.
Removing secondary Executor in another PR would aim to remove unnecessary complexity (if tests seems confirm this is unnecessary). So I don't really like the idea to add another abstraction like `ExecutionEnvirionnment`. 

So I would go to integrate it now like this.

boaks:
I don't get it. 
If you want to test it, I'm pretty sure you will be able to do so without having the changes in the eclipse - 2.0.x.
If others are intended to test it, they should at least know that, or?
So do you want to create an issue and pin it? (we can unpin the sandbox issue).

sbernard31:
So why we integrate all this configurable experimental stuff in Californium until now, if this is so easy to test it  before ?

boaks:
> So why we integrated all this configurable experimental stuff in Californium until now, if this is so easy to test it before ?

That's a good idea, make it configurable :-).

No, in my opinion, it's "over complex", and a mentioned that at the very beginning.
Anyway, if you want it, merge it.
If you really want a feed back from others, create an issue and pin it.

boaks:
> So why we integrated all this configurable experimental stuff in Californium until now, if this is so easy to test it before ?

FMPOV, a lot of that stuff is very isolated, and if configurable, someone can easily configure it back to the previous behavior.

But the executors are essentially and used "all over".
Last year I also worked on these executors based on issue #690 .
My first PR (#699) about that started on 13. July 2018  and the final PRs (#727 #728) have been merge on 24. September 2018. That makes a couple of months to experiment and test without influence others.

sbernard31:
Honestly,  you can not compare your massive change with those PR. This is neither comparable regarding the size nor the changes involved...

Here, I just changed a static references by a dynamic one. As you seemed worry about this modification, I even chose to go with the less impacting one using 2 schedulers (main and secondary) keeping exactly the same behavior.

Tests are OK. Benchmarks does not reveal any performance issue. I didn't say there is no regression at all but risks are very limited and we have 1 milestone release to detect it.

> No, in my opinion, it's "over complex", and a mentioned that at the very beginning.

What is "over complex" exactly ? I just reuse same mechanism that is used all over the code : setter and start/stop. What do you propose ? keeping the static threadpool reference ? 

> someone can easily configure it back to the previous behavior.

That why I propose to go first with this 2 executors solution. So if you want to use 2 executors like before you can (default behavior). If you want to test with only 1 executor you can also using `setExecutors(sameExecutor,sameExecutor)`.

boaks:
Your question:

> Should I integrate this in 2.0.x now ?

My answer:

> If you want.

My point was: this changes APIs. And you intent to change them again.
So my preference would be, to check, what should be the solution in the end.
 
> What is "over complex" exactly ? I just reuse same mechanism that is used all over the code : setter and start/stop. What do you propose ? keeping the static threadpool reference ?

You now have two executors, and more complex rules about detach. You added also a start-state to the layers of the stack. So, yes FMPOV, the static threadpool is the better trade-off. 

But again:

If you want, it merge it.

sbernard31:
> So, yes FMPOV, the static threadpool is the better trade-off.

I believed that you agree that creating static unshutdownable threadpool was a bad practice.
I believed that you did that because you didn't want to do so many changes at once when you did your executors refactoring. (or you haven't time to focus on this)
But it seems you finally think that the static way if the best trade-off and so the best version of the code ... :thinking:  
Luckily most of java libraries does not go for this kind of trade-off.

Anyway, I will to to merge it tomorrow morning.

boaks:
> I believed that you agree that creating static unshutdownable threadpool was a bad practice.

https://github.com/eclipse/californium/issues/936#issuecomment-491842949

That's a trade-off.

boaks:
I sometimes forget to squash, but in generally I think that's the preferred way. 
Or should we change that?

sbernard31:
My feeling that a commit in master should be consistent and ideally as small as possible.
Sometime commits in a PR are not so consistent, so squash it for grace of consistency is better.

Here I tried to make consistent commits, I mean each commit could have been a separated PR. So, I believed that was better to keep it like this.

Maybe you don't agree with this ? :worried:

boaks:
For me, one commit should solve a task, here the task is only finished with the last commit (remove ...).
I'm not sure, if you can leave out one of the previous commits, I would guess, all are required. 
With that, the most commits on their own doesn't make sense (this comes with the last), and all are required. So for me, it's more realistic to have one commit then pretend that it could be done in smaller steps.

sbernard31:
Commit name are clear. Read it again and you will see this is separated tasks.
Of course the final purpose of each of this separated task is to remove the static executor and I was able to do that only when I did the other tasks. If this was not the case I did do the previous task.

But saying that a task depends of another one doesn't mean this is the same one...

E.g. Does it mean that for CID implementation, you should have done only 1 big commit ? because in the end, the CID task is only finished with the last commit...

Anyway, what do you want from me exactly ? you want I fix 2.0.x history bu squashing ? you want I ask next time before to merge ? you want I let you integrate all requests from now ?

boaks:
My question was simple:

> I sometimes forget to squash, but in generally I think that's the preferred way.
Or should we change that?

That's about the intention for future commits.

sbernard31:
And my answer was not so complicated : 
>My feeling that a commit in master should be consistent and ideally as small as possible.
Sometime commits in a PR are not so consistent, so squash it for grace of consistency is better.

I don't think a simple rule like "always squash PR" produce a clearer git history. It depends on PR content. And of course there is a part of subjectivity. If you add this rule you move the subjectivity and discussion at PR level.

I also answered about why I integrated this PR like this, as I suspect you asked this because you don't agree with that way. Your answer seems to confirm that, because you are using it as example and you even suggest I'm not really honest in my argumentation... :disappointed:  
> So for me, it's more realistic to have one commit then pretend that it could be done in smaller steps.

PR #938, opened 2019-04-22
Add alternative slf4j-android library.

boaks:
Optional support for android app development.
Use complied in configuration and prints all log messages by default.


boaks:
Without the new android app, this PR doesn't make too much sense.

PR #939, opened 2019-04-22
Fix javadoc of ExecutorsUtil.

boaks:
Replace experimental by workaround.


sbernard31:
LGTM

PR #949, opened 2019-05-15
Add documentation to benchmark.sh

boaks:
Explain how to adjust the script and start the required server. 

### DIFF
--- a/demo-apps/cf-extplugtest-client/benchmark.sh
+++ b/demo-apps/cf-extplugtest-client/benchmark.sh
@@ -1,10 +1,48 @@
 # !/bin/sh
 # create benchmark statistic
 
+echo "Californium CoAP Benchmark"
+echo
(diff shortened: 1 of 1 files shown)


sbernard31:
LGTM.

Just to let you know the -d64 is removed since jdk 10 (see https://bugs.openjdk.java.net/browse/JDK-8180954)

sbernard31:
What means the "-Dcalifornium.statistic=M13"  in benchmark.sh ?

boaks:
> What means the "-Dcalifornium.statistic=M13" in benchmark.sh ?

The idea was to run that statistics for every milestone to compare the results. 
With that it marks the result as M13. And yes, it should be updated to M14 (what I have done, when I run them on my machine.)
Just one additional hint: 
with a fast R7/2800 it takes about 2000s. May be I introduce to lower the numbers for slower machines, but its not done.
So feel free to reduce the numbers ...

PR #953, opened 2019-05-19
Fix reported cause on handshake failures.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -382,7 +382,18 @@ public class DTLSConnector implements Connector, RecordLayer {
 							message.onError(error);
 						}
 					}
-					connectionStore.remove(handshaker.getConnection(), false);
+					Connection connection = handshaker.getConnection();
+					if (!connection.hasEstablishedSession()) {
+						connectionStore.remove(connection, false);
+					} else if (connection.getEstablishedSession() == handshaker.getSession()) {
+						// failure after established (last FINISH),
+						// but before completed (first data)
+						LOGGER.warn("Handshake with [{}] failed after session was established!",
+								handshaker.getPeerAddress());
+					} else {
+						LOGGER.warn("Handshake with [{}] failed, but has an established session!",
+								handshaker.getPeerAddress());
+					}
 				}
 			};
 			int maxConnections = configuration.getMaxConnections();
@@ -976,19 +987,26 @@ public class DTLSConnector implements Connector, RecordLayer {
 	 */
 	private void terminateOngoingHandshake(final Connection connection, final Throwable cause, final AlertDescription description) {
 
-		if (connection.hasOngoingHandshake()) {
+		Handshaker handshaker = connection.getOngoingHandshake();
+		if (handshaker != null) {
 			if (LOGGER.isTraceEnabled()) {
 				LOGGER.trace("Aborting handshake with peer [{}]:", connection.getPeerAddress(), cause);
 			} else if (LOGGER.isInfoEnabled()) {
 				LOGGER.info("Aborting handshake with peer [{}]: {}", connection.getPeerAddress(), cause.getMessage());
 			}
-			Handshaker handshaker = connection.getOngoingHandshake();
+			handshaker.setFailureCause(cause);
 			DTLSSession session = handshaker.getSession();
 			AlertMessage alert = new AlertMessage(AlertLevel.FATAL, description, connection.getPeerAddress());
 			if (!connection.hasEstablishedSession()) {
 				terminateConnection(connection, alert, session);
 			} else {
 				// keep established session intact and only terminate ongoing handshake
+				if (connection.getEstablishedSession() == handshaker.getSession()) {
+					// failure after established (last FINISH), but before completed (first data)
+					LOGGER.warn("Handshake with [{}] failed after session was established!", handshaker.getPeerAddress());
+				} else {
+					LOGGER.warn("Handshake with [{}] failed, but has an established session!", handshaker.getPeerAddress());
+				}
 				send(alert, session);
 			}
 			handshaker.handshakeFailed(cause);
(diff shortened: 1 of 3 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -382,7 +382,10 @@ public void handshakeFailed(Handshaker handshaker, Throwable error) {
 							message.onError(error);
 						}
 					}
-					connectionStore.remove(handshaker.getConnection(), false);
+					Connection connection = handshaker.getConnection();
+					if (!connection.hasEstablishedSession()) {
+						connectionStore.remove(connection, false);
+					}
```
sbernard31: In which case, handshake failed but we have an established session ? :thinking:

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1004,26 +1005,49 @@ public final void handshakeCompleted() {
...
+	 * @return cause of failure, or {@code null}, if handshake has failed.
+	 */
+	public Throwable getFailureCause() {
+		return cause;
+	}
+
+	/**
+	 * Set the failure cause.
+	 * 
+	 * Prevent to report secondary causes when executing the cleanup.
```
sbernard31: What are you calling "executing the cleaning" ? (on flight retransmission timeout ?)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: `terminateOngoingHandshake(final Connection connection, final Throwable cause, final AlertDescription description)`

The condition is copied from this function.
The PR is a work-around, I recognized that currently all failing handshakes reports "address change".  A real cleanup of the "alert and termination" requires much more work.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: The main cause of this PR was the `Connection.updatePeerAddress`. Always, if `terminate`/`remove` was called before `handshakeFailed`, the "secondary cause, address changed" was reported.
On pain point is a unit test, which used "handshakeFailed" to check, if the connextion store has removed the connection. With that, "remove" must be called before the "handshakeFailed". Therefore I decided to introduce the "failure cause".

sbernard31: I'm lost :sweat_smile: 

I think I get the point about `updatePeerAddress` which always implies that handshake reports "address changed". 

But I don't understand the link with unit tests ? 

I think `Prevent to report secondary causes when executing the cleanup` javadoc message is a bit confusing. Even if this is the reason why you introduce this. At `Handshake` level this setter is just a new feature. (The real workaround is in `Connection.updatePeerAddress` and if you care about explain it, you should explain it in `updatePeerAddress`)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: It seems you maybe to answer to my question ? 
Do you think or did you face a case where an handshake fails and has an established session ?

It seems impossible to me. :thinking:

boaks: I guess, the original idea was something as:
- peer 1, currently mapped to endpoint 1 establish successful a session.
- peer 2, then gets mapped to endpoint 1 can't establish successful a session.
- peer 1, then gets mapped to endpoint 1 again.

I don't know, if that works or not. I didn't investigate the "termination".
I just toke the snippet and comment from the other location.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: `DTLSConnectorAdvancedTest.LatchSessionListener`

```
Throwable error = clientSessionListener.waitForSessionFailed(timeout, TimeUnit.MILLISECONDS);
assertNotNull("client handshake not failed", error);
assertThat(clientConnectionStore.remainingCapacity(), is(remain));
```

That assumes, that the connection is removed before the `handshakeFailed` is called.

boaks: > I think Prevent to report secondary causes when executing the cleanup javadoc message is a bit confusing. Even if this is the reason why you introduce this. At Handshake level this setter is just a new feature. (The real workaround is in Connection.updatePeerAddress and if you care about explain it, you should explain it in updatePeerAddress)

No the real, implementation is in `Handshaker.handshakeFailed`.
Even if the `updatePeerAddress` would not check the cause, it will not trigger the listeners to be called, if the cause is set before. My intention in add that check to the `updatePeerAddress` was to document that behaviour. Now I think, a comment would be better ...

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: > It seems you didn't answer to my question ?

The answer would have required a investigation. Though I didn't investigate, I provided the information, why the code was inserted.

sbernard31: As we do not allow `renegociation`, it seems to me that if a handshake failed this implies that there was no session established because handshake is only done if session is not established.

@boaks, If you think I'm right maybe we could keep the code like this but just adding a warning (here and in terminateOngoingHandshake)  in case where handshake failed AND session is established. This way we could see if this happens in real life. WDYT ?

boaks: OK, go for the warning.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sbernard31: It's clearer to me now.
But `Prevent to report secondary causes when executing the cleanup.` javadoc sounds still confusing.
Anyway it's not a big deal.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: You can merge it after that.

PR #954, opened 2019-05-20
Update link format RD attributes

pokgak:
This PR updates the attributes used by LinkFormat in californium-core to follow the latest version of the Resource Directory draft ([version 20](https://tools.ietf.org/html/draft-ietf-core-resource-directory-20)).

The changes should not affect other part of the californium repository. From what I could find, only examples californium.tools are affected. I also plan to update the resource directory example there to follow the latest draft.

I split this PR into two commits to fix the mix usage of tabs and spaces first. I can squash it into one commit if necessary.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
@@ -39,13 +39,13 @@ public class LinkFormat {
 	public static final String LINK                  = "href";
 
 	// for Resource Directory
-	public static final String HOST		     		 = "h";
-	public static final String LIFE_TIME     		 = "lt";
-	public static final String INSTANCE		   		 = "ins";
-	public static final String DOMAIN	     		 = "d";
-	public static final String CONTEXT		   		 = "con";
-	public static final String END_POINT     		 = "ep";
-	public static final String END_POINT_TYPE		 = "et";
+	public static final String LIFE_TIME             = "lt";
+	public static final String SECTOR                = "d";
+	public static final String CONTEXT               = "anchor";
+	public static final String BASE                  = "base";
+	public static final String RELATION              = "rel";
+	public static final String END_POINT             = "ep";
+	public static final String END_POINT_TYPE        = "et";
 
 	// for parsing
 	public static final Pattern DELIMITER      = Pattern.compile("\\s*,+\\s*");


boaks:
Thanks for your contribution, and in advance for updating cf-rd in californium.tools.

@sbernard31 
Though I lost the track of the LWM2M 1.0 TS, does this change affect it?
If not, I would go to merge it.

californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
```diff
@@ -39,13 +39,12 @@
 	public static final String LINK                  = "href";
 
 	// for Resource Directory
-	public static final String HOST		     		 = "h";
```
sbernard31: I don't know if this is an issue but the comment is called `Fix tab and space mixup` and the constant `HOST` was removed.

sbernard31:
> Though I lost the track of the LWM2M 1.0 TS, does this change affect it?

In Leshan we don't use this class and to be honest I'm not so aware about this Resource Directory draft. Those new attributes (the old and new ones) doesn't talk to me. I didn't remember to see any of them in LWM2M 1.0 TS.

californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
pokgak: Yeah I missed that. `HOST` is supposed to be removed in the other commit. I pushed a fix already. If its okay, I'll squash the commits to 2 commits again.

pokgak:
As far as I know the attributes is used only in Resource Directory, so it should not affect LWM2M.

boaks:
> In Leshan we don't use this class 

OK, then go for it.

@pokgak 

Please squash your commits to a single one and ensure, that this commit is signed as your first two commits has been. Forced push that commit to your branch. Thanks in advance.

pokgak:
Squashed.

boaks:
Thanks!

And I would really enjoy it, if you could spend your time also in the californium.tools cf-rd!

pokgak:
Thanks to you too.

> And I would really enjoy it, if you could spend your time also in the californium.tools cf-rd!

That is already on my todo list. I plan to use cf-rd for my next project :)

boaks:
@pokgak 

I renamed the references to `LinkFormat.DOMAIN` into `SECTOR` in cf-rd.
Please update "californium.tools" before you apply additional changes.

pokgak:
@boaks noted.

Issue #956, opened 2019-05-22
Adding CoAP Publish Subscribe Model to Californium


wajd:
My team and I have used your library to build a project for a university course that allows client-side use of Californium using the [IETS's RFC for Publish Subscribe](https://datatracker.ietf.org/doc/draft-ietf-core-coap-pubsub/). 
You can see the project [here](https://github.com/wajd/TeamMulticast-CoAP-PubSub). It has 2 simple classes and 1 alternative to WebLink, Topic.
We would like to contribute our features to this library as it uses a lot of Californium classes and we could not successfully mask them. 
Unfortunately we built it on Java 12 and used Californium 2.0.0-M14. We also did not write any test cases, so a lot of our 
Would you want this contribution? What should we do in order to make our contribution compliant with your guidelines? (besides forking californium 2.0.x, using Java 1.7, and adding test cases)

boaks:
I did a very short check of the classes, currently I'm unfortunately can't spent more time on it.
I could compile them with java 1.7, so that seems to be not issue.
I'm not sure about the scope, so correct me:
- that are some classes, you would like to contribute to californium, that implements the client-side
- an example is missing?
- the server-side is missing?
- unit tests are missing?

FMPOV, some guide to show how to use it, would be very welcome. May be I just overseen it.

To contribute that, obey [contribution guide](https://github.com/eclipse/californium/blob/master/CONTRIBUTING.md), mainly the [ECA](http://www.eclipse.org/legal/ECA.php). 
Please add a license header to all java files and name all authors.
If other authors are also involved, we need their permission as well.

I would recommend to move the classes into a package, e.g. "org.eclipse.californium.pubsub.client".

wajd:
The scope of the project was to build a client-side CoAP Publish Subscribe model library that allows developers to have easy access to functions that interact with a Coap PubSub broker. 

The networks lab intends to replace their current use of MQTT in their client side (databases, front-end clients, etc.) with CoAP publish-subscribe, and so we were tasked with building the IETS's RFC for pub-sub so that a front-end client such as a website or app could use it to read data from their sensor networks.

The Networks lab that commissioned the project already had a broker built, this is its [github](https://github.com/PethrusG/CoAP-PubSub-Broker) so we had no need to build the broker/server-side because they already had one.

It would take time to set up a working environment to use any of the functions as it requires the aforementioned broker to be running to communicate with it. 

[...]

PR #959, opened 2019-05-23
Added publish subscribe

wajd:
I added our project that adds publish subscribe functionality as a new module. This is still unfinished and would require further work but most core functionality is available. 
You can test it with this [Coap pubsub broker](https://github.com/PethrusG/CoAP-PubSub-Broker). 
 
refer to this [issue](https://github.com/eclipse/californium/issues/956) for further info or to the readme .

### DIFF
--- a/cf-pubsub/pom.xml
+++ b/cf-pubsub/pom.xml
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>parent</artifactId>
+        <groupId>org.eclipse.californium</groupId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>cf-pubsub</artifactId>
+    <packaging>jar</packaging>
+
+    <name>Cf-PubSub</name>
+    <description>Californium (Cf) PubSub, CoAP Publish-Subscribe implementation</description>
+
+    <properties>
+
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>californium-legal</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.californium</groupId>
+            <artifactId>californium-core</artifactId>
+        </dependency>
+
+        <!-- test dependencies -->
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <configuration>
+                    <descriptors>
+                        <descriptor>enhanced-jar-with-licenses.xml</descriptor>
+                    </descriptors>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Converter.java
+++ b/cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Converter.java
@@ -1,0 +1,265 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Wajd Tohme, Ahmad Hussaein, Petr Kocián, Matias Carlander-Reuterfelt, 
+ *                    Ismail Hilal, Tuna Gersil, and Zainab Alsaadi and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *    Wajd Tohme
+ *    Ahmad Hussaein
+ *    Petr Kocián
+ *    Matias Carlander-Reuterfelt
+ *    Ismail Hilal
+ *    Tuna Gersil
+ *    Zainab Alsaadi
+ *
+ ******************************************************************************/
+
+package org.eclipse.californium.pubsub;
+
+import org.eclipse.californium.core.CoapResponse;
+import org.eclipse.californium.core.WebLink;
+import org.eclipse.californium.core.coap.LinkFormat;
+import org.eclipse.californium.core.coap.MediaTypeRegistry;
+
+import java.util.Set;
+import java.util.concurrent.ConcurrentSkipListSet;
+
+/**
+ * The Converter class is a collection of helper functions that make using Californium {@link WebLink}
+ * and {@link CoapResponse} easier to use. It also holds various uri linting functions
+ * The {@link PubSub} methods return CoapResponse and this helps with using them
+ */
+public class Converter {
+
+    /*helper functions for  CoapResponse */
+
+    /**
+     * creates a Set of WebLinks from the ResponseText of a given CoapResponse
+     *
+     * @param response CoapResponse
+     * @return a Set of WebLink
+     */
+    public static Set<WebLink> getWebLinks(CoapResponse response){
+        return LinkFormat.parse(response.getResponseText());
+    }
+
+    /**
+     * renders a WebLink for each available topic in the ResponseText of a given CoapResponse
+     * This would include the node topics that have child/leaf topics in the returned Set
+     *
+     * @param response CoapResponse
+     * @return a Set of WebLink
+     */
+    public static Set<WebLink> getAllWebLinks(CoapResponse response) {
+        return extractAllWebLinks(getWebLinks(response));
+    }
+
+    /*helper functions for Set<WebLink>*/
+
+    /**
+     * renders an array out of a given Set of WebLinks
+     * useful if you want to access a specific WebLink without casting to an ArrayList
+     *
+     * @param webLinks Set<WebLink>
+     * @return an array of WebLinks
+     */
+    public static WebLink[] getArray(Set<WebLink> webLinks) {
+        return webLinks.toArray(new WebLink[0]);
+    }
+
+
+    /**
+     * renders a WebLink for each available topic in the uris of the WebLinks in the given Set
+     * This would include the node topics that have child/leaf topics in the returned Set
+     *
+     * @param webLinks Set<WebLink>
+     * @return a Set of WebLink
+     */
+    public static Set<WebLink> extractAllWebLinks(Set<WebLink> webLinks) {
+        Set<WebLink> topics = new ConcurrentSkipListSet<WebLink>();
+        WebLink p;
+        for(WebLink w: webLinks) {
+            p = w;
+            while (!cleanUri(p.getURI()).equals("ps")) {
+                topics.add(p);
+                p = getParent(p);
+            }
+        }
+        return topics;
+    }
+
+    /*helper functions for WebLink*/
+
+    /**
+     * renders a WebLink using the given uri and sets its Content type
+     *
+     * @param uri String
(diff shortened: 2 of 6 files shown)


cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Converter.java
```diff
@@ -0,0 +1,264 @@
+/*******************************************************************************
+ * Copyright (c) 2018 RISE SICS and others.
```
boaks: I'm not sure, if your work for the RISE SICS as well. 
If not, just place your university or employer.

boaks:
Do you use the eclipse IDE? If so, would it be possible to import the source formatter "eclipse-formatter-profile.xml" and apply it to your java files?

cf-pubsub/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.californium</groupId>
+            <artifactId>californium-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
```
boaks: hamcrest and junit are currently not used.

slf4j-api also not.

Consider to remove these dependencies, if they are not used.

cf-pubsub/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+        </dependency>
+        <dependency>
+            <groupId>ca.jimr</groupId>
+            <artifactId>gae-mini-profiler</artifactId>
+            <version>2.0.3</version>
```
boaks: At least to compile, this seems to be not used,
Consider to change the scope of the dependency to runtime by adding `<scope>runtime</scope>`

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Converter.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * renders an array out of a given Set of WebLinks
+     * useful if you want to access a specific WebLink without casting to an ArrayList
+     *
+     * @param webLinks Set<WebLink>
+     * @return an array of WebLinks
+     */
+    public static WebLink[] getArray(Set<WebLink> webLinks) {
+        return webLinks.toArray(new WebLink[0]);
```
boaks: Seems to be equivalent to `return new WebLink[0];`

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Converter.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * returns an int of the ContentType of the given WebLink
+     * it parses the ContentType returned to ensure it does not contain extra letters
+     * due to issues faced with the testing broker we used
+     *
+     * @param webLink WebLink
+     * @return an int ContentType
+     */
+    public static int getCT(WebLink webLink) {
```
boaks: Usually we try not to use abbreviations. Please consider to rename it to `getContentType`.

boaks: Also for `setCT` an `getCTString`.

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/PubSub.java
```diff
@@ -0,0 +1,343 @@
...
+            Request req = new Request(CoAP.Code.GET);
+
+            client = new CoapClient(SCHEME, getHost(), getPort(), uri);
+            client.useExecutor();
+            client.setTimeout(timeout);
+
+            req.setURI(client.getURI());
+            req.setObserve();
+
+            config.set(NetworkConfig.Keys.TOKEN_SIZE_LIMIT, 4);
```
boaks: I would recommend to either move this into the constructor, or use a properties file and 
https://github.com/eclipse/californium/blob/2.0.x/demo-apps/cf-extplugtest-server/src/main/java/org/eclipse/californium/extplugtests/ExtendedTestServer.java#L68 with
https://github.com/eclipse/californium/blob/2.0.x/demo-apps/cf-extplugtest-server/src/main/java/org/eclipse/californium/extplugtests/ExtendedTestServer.java#L114 .

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/PubSub.java
```diff
@@ -0,0 +1,343 @@
...
+            RandomTokenGenerator rand = new RandomTokenGenerator(config);
+            Token token = rand.createToken(false);
+            req.setToken(token);
```
boaks: If you set TOKEN_SIZE_LIMIT in the "Standard" NetworkConfig (as done in line 321), it should be effective without generating a token on your own.

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/PubSub.java
```diff
@@ -0,0 +1,343 @@
...
+        }
+
+        /**
+         * Unsubscribes from the uri of the Subscription instance
+         * and shuts down the client
+         */
+        public void unsubscribe() {
+            if (this.relation != null) {
+                relation.proactiveCancel();
+                int mid = relation.getCurrent().advanced().getMID();
```
boaks: That creates a very rare race condition. 
Line 335 may have already changed the current response and no further response will be received.
The while may then last for longer ...

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Topic.java
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * Creates an instance of Topic given a WebLink from which it
+     * extracts the name, URI and Content Type
+     *
+     * @param wl WebLink
+     */
+    public Topic(WebLink wl) {
+        this.uri = wl.getURI().substring(1).split("(/)");
+        this.name = this.uri[this.uri.length - 1];
+        String corchete = wl.toString().substring(wl.toString().indexOf('[') + 1);
```
boaks: I'm not sure, if this is reliable. It seems to read the first attribute value list, regardless of the attributes name. I would prefer a solution, which gets the attribute-value by name.

If that string-search approach is intended, then I would prefer to call `wl.toString()` only once and re-use it. 

Or

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Topic.java
```diff
@@ -0,0 +1,202 @@
...
+        try {
+            this.ct = Integer.parseInt(corchete.substring(0, 1));
+        } catch (NullPointerException | NumberFormatException | ArrayIndexOutOfBoundsException e) {
+            this.ct = Integer.parseInt(corchete.substring(0, 0));
```
boaks: Parse a empty string? 

> Exception in thread "main" java.lang.NumberFormatException: For input string: ""

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Topic.java
```diff
@@ -0,0 +1,202 @@
...
+        return at;
+    }
+
+    /**
+     * Adds to a given ArrayList of Topics all of the topics and subtopics as individuals
+     *
+     * @param topic a Topic which you want to get the subtopics and parent Topics from
+     * @param at    an already existing ArrayList of Topics where you want all the Topics to be
+     */
+    public void getTopics(Topic topic, ArrayList<Topic> at) {
```
boaks: May be consider to make this function `static`?

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Topic.java
```diff
@@ -0,0 +1,202 @@
...
+        String p = sb.toString();
+        return p = p.substring(0, p.length() - 1);
+    }
+
+    /**
+     * @param topic that you want to compare
+     * @return true if given Topic has same URI as instance Topic
+     */
+    public boolean equals(Topic topic) {
+        return (this.getName().equals(topic.getName()) && this.getURIString().equals(topic.getURIString()));
```
boaks: the name is the last part of the URI? So no need to compare it twice?

Try `Arrays.equals(this.uri, topic.uri);`

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Topic.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+}
+
+/**
+ * The TopicComparator class is used for sorting purposes whenever it is needed
+ * to compare and sort Topics. It sorts in terms of alphabetical order and URI length
```
boaks: It's ok, not have unit test, but here I'm sure, they would show a defect.

boaks:
I'm not sure, how much time you can spend, but the

> Copyright (c) 2018 RISE SICS and others.

must use either your university or your employer.

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Converter.java
wajd: We've asked the Network Systems Lab and are awaiting their response on what to put under the copyright. I apologize for forgetting to change it.

wajd:
> Do you use the eclipse IDE? If so, would it be possible to import the source formatter "eclipse-formatter-profile.xml" and apply it to your java files?

We used IntelliJ unfortunately.

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Converter.java
wajd: This function might actually not be needed if this class is to be added to Californium, as someone using Californium should know how to do toArray properly on a Set of WebLink.

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Converter.java
boaks: Any news on that?
If the "Network Systems Lab" doesn't claim the "copyright",  you may also just add the names of the authors. But for a clean legal state, It would be also great, if such a "no-claim" is written in some e-mail :-).

wajd: The networks lab and KTH (same entity) both said they have no copyright. As for us, the authors, we've already signed our names. So I think it should be fine if the RISE copyright is removed and just our names left as authors.

wajd:
> I'm not sure, how much time you can spend to go through the list above. If you can't fix all, that will be also OK. But the
> 
> > Copyright (c) 2018 RISE SICS and others.
> 
> must use either your university or your employer.

I apologize for burdening you further but it seems I have no time to go through the requested changes. If you can make those changes yourself, please go ahead. I've already replied above about the copyright claim.

cf-pubsub/src/main/java/org/eclipse/californium/pubsub/Converter.java
boaks: The eclipse foundation requires a explicit copyright statement.
It's not big thing, just list the same names above instead of the "RISE SICS".

boaks:
> I apologize for burdening you further but it seems I have no time to go through the requested changes. If you can make those changes yourself, please go ahead. I've already replied above about the copyright claim.

Except the copyright, that will be OK.
Please just copy the list of authors to the copyright above. Then squash your 2 commits into 1 and force push that to your PR branch. After that I'm able to create the CQ.

wajd:
I updated the copyright, removed the unneeded dependencies, and renamed the CT methods. I did a squash and I hope I did it right since this is the first time I've done a git squash.

boaks:
Squash is "well done" :-) .

But the copyright should end with "and others."
Would it be possible for you to add that? 
Amend that changes to your last commit and force push it again.

wajd:
done.

boaks:
Thanks!

I will create a CQ. Please don't update this branch any more.
Changes and fixes must be in separate PRs.

Just to mention:
If the contribution is done by one person and below 1000 lines, we can "by pass" the CQ. 
Therefore, if you or one of your team want to contribute some improvements, it's easier, if that is done "alone" :-).

boaks:
Hi all, thanks a lot for your contribution and patience!

PR #963, opened 2019-06-03
Cleanup DTLSConnectorTest.

boaks:
Use ConnectorHelper and remove duplicates.

### DIFF
--- a/scandium-core/src/test/java/org/eclipse/californium/scandium/ConnectorHelper.java
+++ b/scandium-core/src/test/java/org/eclipse/californium/scandium/ConnectorHelper.java
@@ -411,17 +411,17 @@ public class ConnectorHelper {
 	 * A data handler that decrements a latch on successful processing of data.
 	 *
 	 */
-	abstract class LatchDecrementingDataHandler implements DataHandler {
+	static abstract class LatchDecrementingDataHandler implements DataHandler {
 
 		private CountDownLatch latch;
 
-		public LatchDecrementingDataHandler(final CountDownLatch latch){
+		public LatchDecrementingDataHandler(CountDownLatch latch){
 			this.setLatch(latch);
 		}
 
 		@Override
-		public void handleData(InetSocketAddress endpoint, byte[] data) {
-			if (process(data)) {
+		public final void handleData(InetSocketAddress endpoint, byte[] data) {
+			if (process(endpoint, data)) {
 				latch.countDown();
 			}
 		}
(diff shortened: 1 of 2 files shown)


scandium-core/src/test/java/org/eclipse/californium/scandium/ConnectorHelper.java
```diff
@@ -411,17 +411,17 @@ public void clear() {
...
 
-		public LatchDecrementingDataHandler(final CountDownLatch latch){
+		public LatchDecrementingDataHandler(CountDownLatch latch){
 			this.setLatch(latch);
 		}
 
 		@Override
-		public void handleData(InetSocketAddress endpoint, byte[] data) {
-			if (process(data)) {
+		public final void handleData(InetSocketAddress endpoint, byte[] data) {
```
sbernard31: Except for very rare case, I generally don't like to much the idea to not let user override a method.

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
```diff
@@ -808,24 +779,23 @@ public void testConnectorSendsHelloVerifyRequestWithoutCreatingSession() throws
...
 		InetSocketAddress endpoint = new InetSocketAddress(12000);
 
-		DataHandler handler = new DataHandler() {
-
+		LatchDecrementingDataHandler handler = new LatchDecrementingDataHandler(latch) {
 			@Override
-			public void handleData(byte[] data) {
-				receivedRecords.addAll(Record.fromByteArray(data, serverEndpoint, null)); 
-				latch.countDown();
+			public boolean process(InetSocketAddress endpoint, final byte[] data) {
```
sbernard31: any reason to make `data` final ?

scandium-core/src/test/java/org/eclipse/californium/scandium/ConnectorHelper.java
boaks: Isn't this one of that rare cases?
The intention of the class is to count something down. The `final` ensures, that overriding is applied to the new extension point `process`, and so a decision about counting must be provided by the implementation of that new extension point. 
FMPOV, if that `handleData` is intended to be overridden in a way, that `process` doesn't allow, it should be considered to use the `DataHandler` itself instead of this `LatchDecrementingDataHandler`.

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
boaks: "code completion" on the old definition, which uses that final.
I removed the final on all occurrences.
It's completely removed with a later clean up.

PR #964, opened 2019-06-03
Add COUNT and PAGE to LinkFormat for resource directory

pokgak:
Resource directory uses query parameter count and page for lookup pagination. This PR adds this to LinkFormat.


boaks:
Thanks!

PR #967, opened 2019-06-04
Check for client certificate, if client authentication is required.

boaks:
Unit tests will be added after PR #966

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
@@ -112,11 +112,6 @@ public class ServerHandshaker extends Handshaker {
 	private PublicKey clientPublicKey;
 
 	/**
-	 * The client's X.509 certificate chain.
-	 */
-	private CertPath peerCertPath;
-
-	/**
 	 * The cryptographic options this server supports, e.g. for exchanging keys,
 	 * digital signatures etc.
 	 */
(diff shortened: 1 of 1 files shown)


boaks:
Though PR #966 tries to check the handshake failure cause, this PR should be merge first to make it possible, to test that cause as well.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -336,7 +336,13 @@ private void receivedClientCertificate(final CertificateMessage message) throws
...
 		verifyCertificate(clientCertificate);
 		clientPublicKey = clientCertificate.getPublicKey();
-		peerCertPath = message.getCertificateChain();
+		peerCertPath = clientCertificate.getCertificateChain();
+		if (clientAuthenticationRequired && clientPublicKey == null) {
+			LOGGER.debug("Certificate validation failed: x509 could not be trusted!");
+			AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE,
+					session.getPeer());
+			throw new HandshakeException("Client Certificate required!", alert);
+		}
```
sbernard31: Should we do `verifyCertificate(clientCertificate);` after this new check ?

boaks: Done-

PR #970, opened 2019-06-07
Shutdown client in examples. (As now it's mandatory by default)

sbernard31:
Since #937, shutdown client is mandatory in case you don't use a detached executor.

``` java
/**
 * Shutdown the client-specific executor service, when not detached. Always
 * needed unless you used detached executor.
 */
public void shutdown()
```

This is mainly added for didactic purpose.

Issue #736, opened 2018-08-28
CoCoA Congestion Control not working for Server


samkumar:
I am working with the latest version (2.0.0-M11), but I experience similar issues with version 1.0.6.

I am using CoAP in the following scenario: a client repeatedly sends data to a server using POST messages, and the server responds with the return code 2.04 CHANGED (this response is just to conform to REST; we could also just replace it with a CoAP ACK).

This is the base server implementation, and it works fine:

```
import org.eclipse.californium.core.CoapResource;
import org.eclipse.californium.core.CoapServer;
import org.eclipse.californium.core.coap.CoAP;
import org.eclipse.californium.core.server.resources.CoapExchange;

public class Main {
	public static void main(String[] args) {
		CoapServer server = new CoapServer();
		server.add(new CoapResource("anemometer") {
			@Override
			public void handlePOST(CoapExchange exchange) {
				System.out.printf("Got a message of size %d\n", exchange.getRequestPayload().length);
				exchange.respond(CoAP.ResponseCode.CHANGED);
			}
		});
		server.start();
	}
}
```

I have tried to add CoCoA congestion control to this setup. The client (which is using Contiki's Erbium implementation of CoAP) seems to be behaving normally when I use the server above, even when I enable CoCoA on the client. However, I also tried enabling CoCoA on the server, using the following code:
```
import org.eclipse.californium.core.CoapResource;
import org.eclipse.californium.core.CoapServer;
import org.eclipse.californium.core.coap.CoAP;
import org.eclipse.californium.core.network.config.NetworkConfig;
import org.eclipse.californium.core.network.stack.congestioncontrol.Cocoa;
import org.eclipse.californium.core.server.resources.CoapExchange;

public class Main {
	public static void main(String[] args) {
		NetworkConfig config = new NetworkConfig();
		config = config.setBoolean(NetworkConfig.Keys.USE_CONGESTION_CONTROL, true);
		config = config.setString(NetworkConfig.Keys.CONGESTION_CONTROL_ALGORITHM, Cocoa.class.getSimpleName());
		CoapServer server = new CoapServer(config);
		server.add(new CoapResource("anemometer") {
			@Override
			public void handlePOST(CoapExchange exchange) {
				System.out.printf("Got a message of size %d\n", exchange.getRequestPayload().length);
				exchange.respond(CoAP.ResponseCode.CHANGED);
			}
		});
		server.start();
	}
}
```

[...]

PR #975, opened 2019-06-11
Fix congestion control layer

boaks:
Fix the `ConcurrentModificationException` in PR #970
I'm pretty sure, that a lot of concurrency issues in the congestion control are still left.
See also discussion on issue #736


sbernard31:
no objections

PR #971, opened 2019-06-07
Add support for setting ThreadGroup for workers

crea-doo:
I'm using Californium in an OSGi environment for client connections and want to be able to stop and restart these connections. I'm aware of the fact that Californium was not designed with enabling restarts of connections in mind (like written in #936) but wanted to enable this use case:

I got a couple of `java.lang.IllegalThreadStateException` exceptions (similar to the one described in #680) but was able to overcome most of them using the provided setters for custom executors.

Unfortunately the `Worker` defined in the [DTLSConnector](https://github.com/eclipse/californium/blob/2.0.x/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java#L2267) uses the static ThreadGroup `SCANDIUM_THREAD_GROUP` defined in the `NamedThreadFactory`.

This pull requests adds functionality to be able to set a custom ThreadGroup for these workers - the default behaviour doesn't change though.

PR #978, opened 2019-06-12
Explicitly reset daemon of thread groups.

boaks:

### DIFF
--- a/element-connector/src/test/java/org/eclipse/californium/elements/util/NamedThreadFactoryTest.java
+++ b/element-connector/src/test/java/org/eclipse/californium/elements/util/NamedThreadFactoryTest.java
@@ -1,0 +1,178 @@
+
+/*******************************************************************************
+ * Copyright (c) 2019 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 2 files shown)


boaks:
Alternative fix for destroyed thread groups.
Obsoletes workaround of PR #971.

element-connector/src/test/java/org/eclipse/californium/elements/util/NamedThreadFactoryTest.java
```diff
@@ -0,0 +1,87 @@
...
+	public void testThreadGroup() {
+		ThreadGroup group = new ThreadGroup("test-group");
+		group.setDaemon(false);
+		createThread(group);
+		assertFalse("group is destroyed", group.isDestroyed());
+	}
+
+	@Test
+	public void testScandiumThreadGroup() {
+		ThreadGroup group = new ThreadGroup("test-group");
```
sbernard31: :thinking:  I suspect there is a copy paste issue because I don't see any difference between : 
testThreadGroup
testScandiumThreadGroup
testCaliforniumThreadGroup

boaks: Yes, copy & paste. Fixed.

boaks:
May be a little over tested. The test now ensures, that the daemon in the thread groups are reset, even if the calling context is a daemon thread group.

PR #982, opened 2019-06-18
Add content type json and cbor to plugtests.

boaks:


boaks:
That is added to `master`, so I implemented it for `2.0.x`.

PR #983, opened 2019-06-20
OSCORE: Update in functionality for securely re-generating contexts

rikard-sics:
This pull request updates the support for the method described in Appendix B.2. of the OSCORE draft for securely re-generating contexts:
https://tools.ietf.org/html/draft-ietf-core-object-security-16#appendix-B.2

The idea is to perform context re-derivation by essentially using a message exchange together with the Context ID field in the OSCORE option to securely generate a new shared context. This can for instance be used when one device has lost power and information about the mutable parts of a context (e.g. sequence number) but retains information about static parts (e.g. master secret).

This pull request updates support for this functionality to let the client set a flag to indicate that it has lost the mutable parts of an OSCORE context. When that flag is set the first message of the context re-derivation procedure is immediately sent which will be followed by the response from the server. After this the client can proceed with the actual request it intends to send which will act as the third message of the context re-derivation procedure, followed by the response from the server.

In addition the JUnit test for testing this functionality has been updated to be able to run stand-alone without an external server application.

The end goal is that the first message of the re-derivation procedure is not sent at the moment when the client indicates that it has lost the mutable parts of an OSCORE context but rather not sent until the client is actually interested in sending a request to the server. A further pull request will be created for this.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
@@ -1,0 +1,104 @@
+/*******************************************************************************
+ * Copyright (c) 2019 RISE SICS and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Rikard Höglund (RISE SICS)
+ *    
+ ******************************************************************************/
+package org.eclipse.californium.oscore;
+
+import java.io.IOException;
+import java.security.SecureRandom;
+
+import org.eclipse.californium.core.CoapClient;
+import org.eclipse.californium.core.CoapResponse;
+import org.eclipse.californium.core.Utils;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.coap.CoAP.Code;
+import org.eclipse.californium.core.coap.CoAP.ResponseCode;
+import org.eclipse.californium.elements.exception.ConnectorException;
+import org.eclipse.californium.elements.util.Bytes;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Class that implements methods for dynamic re-generation of OSCORE contexts.
+ *
+ * See https://tools.ietf.org/html/draft-ietf-core-object-security-16#appendix-B.2
+ *
+ */
+public class ContextRederivation {
+	
+	/**
+	 * The logger
+	 */
+	private static final Logger LOGGER = LoggerFactory.getLogger(RequestDecryptor.class.getName());
+	
+	/**
+	 * Method for an application to indicate that the mutable parts of an OSCORE context has been lost.
+	 * In such case the context re-derivation procedure is triggered.
+	 * 
+	 * @param uri the URI associated with the context information has been lost for
+	 * @throws CoapOSException if re-generation of the context fails
+	 */
+	public static void setLostContext(String uri) throws CoapOSException
+	{
+		try {
+			rederive(uri);
+		} catch (ConnectorException | IOException | OSException e) {
+			LOGGER.error(ErrorDescriptions.CONTEXT_REGENERATION_FAILED);
+			throw new CoapOSException(ErrorDescriptions.CONTEXT_REGENERATION_FAILED, ResponseCode.BAD_REQUEST);
+		}
+	}
+	
+	/**
+	 * Perform re-derivation of contexts as detailed in Appendix B.2.
+	 * Essentially it uses a message exchange together with the Context ID
+	 * field in the OSCORE option to securely generate a new shared context.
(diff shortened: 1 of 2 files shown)


rikard-sics:
As a next step I would like to make it so that the first message of the re-derivation procedure is not sent at the moment when the client indicates that it has lost the mutable parts of an OSCORE context but rather not sent until the client is actually interested in sending a request to the server.

I tried adding code so the sending of this message is done in the ObjectSecurityLayer class but had issues with that. Perhaps it does not work well to trigger sending of a message when a message is currently in the process of being sent.

Do you have any suggestions for how I can trigger the transmission of a message to happen immediately before, when a client is attempting to send a message?

boaks:
I guess, the intention is, that in `ObjectSecurityLayer.sendRequest` after 
`final OSCoreCtx ctx = db.getContext(uri);` that should trigger a call to `ContextRederivation.setLostContext`. Is that the intention?

So that trigger is in the execution context of the requests `Exchange`, it starts a CoapClient and waits for the response. But maybe the processing of the response requires the  thread, which waits for the response.
I would try it with the async variant of `CoapClient.advanced`, where the response is handled in the `CoapHandler` callback. It's not that easy, because the continue of the processing of the requests Exchange MUST be done in that execution context! Not in that of the `CoapHandler` callback! Therefore the callback must schedules the left work in a `Runnable` to the  execution context of the requests exchange using `Exchange.execute()`.
Try to prepare that. I will help, if it doesn't fully work, the threading in the Exchange is not that easy.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
```diff
@@ -0,0 +1,103 @@
+/*******************************************************************************
+ * Copyright (c) 2018 RISE SICS and others.
```
boaks: Switch to 2019?

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
```diff
@@ -0,0 +1,103 @@
...
+		db.addContext(uri, ctx);
+		
+		//Now send request using the new context
+		String resource = "/rederive"; //Dummy resource to access for context re-derivation
+		String URI = uri + resource;
+		System.out.println(URI);
+		
+		CoapClient c = new CoapClient(URI);
+		Request r = new Request(Code.GET);
+		r.getOptions().setOscore(new byte[0]);
```
boaks: May be replaced by `Bytes.EMPY`

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
```diff
@@ -0,0 +1,103 @@
...
+		System.out.println(Utils.prettyPrint(r));
+		
+		CoapResponse resp = null;
+		resp = c.advanced(r);
```
boaks: Consider to provide also the async approach with

`void advanced(CoapHandler handler, Request request)`

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
rikard-sics: Done in commit 2cb4bc043454e7302e201bc86f6ca4af19a24fe0

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
rikard-sics: Done in commit 2cb4bc043454e7302e201bc86f6ca4af19a24fe0

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
rikard-sics: Is it alright if I wait with this and add it to my next pull request? There I will try to apply your idea from the earlier post.

rikard-sics:
> I guess, the intention is, that in `ObjectSecurityLayer.sendRequest` after
> `final OSCoreCtx ctx = db.getContext(uri);` that should trigger a call to `ContextRederivation.setLostContext`. Is that the intention?
> 
> So that trigger is in the execution context of the requests `Exchange`, it starts a CoapClient and waits for the response. But maybe the processing of the response requires the thread, which waits for the response.
> I would try it with the async variant of `CoapClient.advanced`, where the response is handled in the `CoapHandler` callback. It's not that easy, because the continue of the processing of the requests Exchange MUST be done in that execution context! Not in that of the `CoapHandler` callback! Therefore the callback must schedules the left work in a `Runnable` to the execution context of the requests exchange using `Exchange.execute()`.
> Try to prepare that. I will help, if it doesn't fully work, the threading in the Exchange is not that easy.

Thanks for the feedback. I will try to apply this idea in a coming PR.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
boaks: > Is it alright if I wait with this and add it to my next pull request? 

Sure

rikard-sics: > Sure

Alright sounds good. Then I will come with a new one some time after this one is merged.

boaks:
LGTM.

Please squash.

rikard-sics:
> LGTM.
> 
> Please squash.

Thanks. The commits have now been squashed.

PR #986, opened 2019-06-26
Threadsrule

boaks:
Cleanup unit tests. 
Check for terminating started threads.
Cleanup start/stop synchronization.

Signed-off-by: Achim Kraus <achim.kraus@bosch-si.com>

PR #987, opened 2019-06-27
Coap threadsrule

boaks:
Cleanup californium-core tests.

### DIFF
--- a/californium-core/logback-test.xml
+++ b/californium-core/logback-test.xml
@@ -12,6 +12,10 @@
 		<appender-ref ref="STDOUT" />
 	</logger>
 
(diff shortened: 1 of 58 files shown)


sbernard31:
I'm a bit lost because I have the impression that  I already read some commits in #986. :sweat_smile: 

Anyway, this looks good to me :+1:

boaks:
> I have the impression that I already read some commits in #986

Sure, because it's "always" the introduction of the ThreadsRule and the cleanup of the unit tests. 
That caused some rare race-conditions on `stop`, so I had to reduce the "synchronized section".

PR #989, opened 2019-06-28
Oscore threadsrule

boaks:
Introduce CoapThreadsRule in oscore test.

### DIFF
--- a/cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
+++ b/cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreObserveTest.java
@@ -21,12 +21,11 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 
 import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.net.UnknownHostException;
 import java.util.Random;
 import java.util.Timer;
 import java.util.TimerTask;
 
+import org.eclipse.californium.TestTools;
 import org.eclipse.californium.category.Large;
 import org.eclipse.californium.core.CoapClient;
 import org.eclipse.californium.core.CoapHandler;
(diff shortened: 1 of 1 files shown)


boaks:
@rikard-sics

If possible, please have a look at the changes.
I'm not sure, if you use the code also in other contexts, so that the changes may violate that.

rikard-sics:
> If possible, please have a look at the changes.
> I'm not sure, if you use the code also in other contexts, so that the changes may violate that.

I think the changes you propose should be fine. As far as I understand this will only affect functionality in the OSCoreObserveTest?

But I was thinking about if the test class should be changed further so that clientAddress and serverAddress take into account how you are retrieving the address of the localhost with TestTools.LOCALHOST_EPHEMERAL instead of InetAddress.getLoopbackAddress()?

Also perhaps equivalent changes would be needed in the OSCoreServerClientTest since it is similar and also has a server and client part of the test class.

rikard-sics:
> I'm not sure, if you use the code also in other contexts, so that the changes may violate that.

I do also have another class with JUnit tests that implements a lot of the interoperability tests that Ericsson have defined in a document. This test class I have not commited yet but I am planning to. I can update this code to follow CoapThreadsRule approach.

boaks:
> TestTools.LOCALHOST_EPHEMERAL instead of InetAddress.getLoopbackAddress()?

The difference is that the one is "localhost:0" (a socket-address) and the other "localhost" (only address). Though "localhost:0" was used in many places, I introduced it in `TestTools`. 

My impression is, that though the server and client "unnaturally" share the same context db, the context base on the "hostname" doesn't work! Therefore the test try to solve that by using "127.0.0.1" and "localhost" as name.  Therefore I think, providing different db's to both client and server endpoint will make it easier, because both can then use "localhost"
Generally, I'm not sure, if using the hostname (without the port) is a good idea, but that is not my concern. In the case above it makes it at least fragile (you must take care to use localhost and "127.0.0.1").

rikard-sics:
> The difference is that the one is "localhost:0" (a socket-address) and the other "localhost" (only address). Though "localhost:0" was used in many places, I introduced it in `TestTools`.

Yes, I see your point. I was thinking if there was some way to eliminate the definitions of serverAddress and clientAddress or base them on the ThreadTools.LOCALHOST_EPHEMERAL. (Assuming the client and server get separate context db:s.)

Perhaps for adding the context on the client, instead of:
`db.addContext("coap://" + serverAddress, ctx);`
Using:
`db.addContext("coap://" + TestTools.getUri(serverEndpoint, null), ctx);`
(And similar for adding the context on the server.)

> My impression is, that though the server and client "unnaturally" share the same context db, the context based on the "hostname" doesn't work! Therefore the test try to solve that by using "127.0.0.1" and "localhost" as name. I think, providing different db's to both client and server endpoint will make it easier, because both can then use "localhost"

Yes I agree. The way it is done now it is unnatural that the client and server share the context db. I solved this by using "127.0.0.1" for one and "localhost" for the other but that is a solution just to bypass the issue of them sharing the context db. So it would be better for them to have separate db:s so the hostname can be used for both.

(Strictly speaking the server does not need a correct URI to associate the context with since it will not be sending out any requests and can identity the correct context to use for incoming requests by the Sender ID).

> Generally, I'm not sure, if using the hostname (without the port) is a good idea, but that is not my concern. In the case above it makes it at least fragile (you must take care to use localhost and "127.0.0.1").

Yes this is also true. For instance currently the client cannot handle 2 server applications running on the same IP but on different ports since it just cares about the hostname and not the port. I will try to take care of this in a future PR.

boaks:
> (Strictly speaking I believe the server does not need a correct hostname to associate the context with since it will not be sending out any requests

In LWM2M CoAP is used bidirectional, the device send request to the registration interface, and the data management sends requests to the device.

rikard-sics:
> In LWM2M CoAP is used bidirectional, the device send request to the registration interface, and the data management sends requests to the device.

Yes you are correct. I experienced this in my Leshan work with LWM2M. Such a server does need to set the client hostname correctly since the server will also send requests.

I was thinking particularly about this JUnit test where the server should not need the hostname of the client.

Issue #974, opened 2019-06-11
cf-oscore should set a meaninful PeerIdentity in the EndpointContext


LudwigSeitz:
This concerns cf-oscore in M15.

It would be nice if cf-oscore could set a meaningful value for the PeerIdentity in the EndpointContext, e.g. ID Context || Sender ID from the matching OSCORE context.

LudwigSeitz:
Perhaps some explanation on why this is useful: If two parties use OSCORE, they are implicitly authenticating each other (by possession of the correct OSCORE context), the authenticated identity can be useful e.g. for access control. I believe the authenticated identity is ID Context || Sender ID  (if ID Context is not null, otherwise just Sender ID).

boaks:
I agree. 
Do you see any technical challenge?
Is there anything else hold you up to implement it and contribute it (beside of the time :-) )?
AFAIK, just derive a implementation class from `java.security.Principal` as the other `Identity` implementations in `org.eclipse.californium.elements.auth`.

LudwigSeitz:
Besides time nothing, either Rikard or me could provide this, I just added the issue to document the need and have a discussion if needed.

rikard-sics:
Yes, I have discussed this with Ludwig and can certainly do it in a future pull request.

As far as the best way to do it, would it be to set a Principal, accessible via getPeerIdentity() as with DTLS? Or should I rather extend californium/elements/EndpointContext.java with some other method to get OSCORE-related information?

It could for instance be the case that an application is using both DTLS and OSCORE for a communication channel.

rikard-sics:
> Do you see any technical challenge?
> Is there anything else hold you up to implement it and contribute it (beside of the time :-) )?
> AFAIK, just derive a implementation class from `java.security.Principal` as the other `Identity` implementations in `org.eclipse.californium.elements.auth`.

Just to clarify if this is a potential issue but what if a certain connection is using both DTLS and OSCORE? In such case should information related to both OSCORE and DTLS be stored in the same Principal? Or would it be better to implement a separate method for OSCORE like getOscoreIdentity? Otherwise OSCORE and DTLS may conflict and overwrite each other's peer information.

boaks:
> using both DTLS and OSCORE?

Just my first thoughts: 
Make it configurable. I'm not sure, if someone who uses both DTLS and OSOCRE, requires to get both. I would assume, that being able to select, which one is used, would be a good first approach. If a use-case comes up with both, then it would also be possible to extend it.

[...]

PR #996, opened 2019-07-02
Add principal converter

sophokles73:
The Principal instances created by Scandium when a client has been
authenticated contain only the authentication identifier used by the
client, e.g. the PSK identity or the client cert subject DN.

However, applications will often map the (technical) authentication
identifier used at the DTLS level to a logical application level
identifier which is used to refer to the client.

The PrincipalConverter can be implemented by applications to convert the
technical Principal created by Scandium to a Principal that contains
applications level information which can then be used when processing
CoAP requests.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/auth/AbstractExtensiblePrincipal.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/auth/AbstractExtensiblePrincipal.java
@@ -1,0 +1,61 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial creation
+ *******************************************************************************/
+
+
+package org.eclipse.californium.elements.auth;
+
+import java.security.Principal;
+import java.util.Collections;
+import java.util.Map;
+
+
+/**
+ * A base class for implementing {@link ExtensiblePrincipal}s.
+ *
+ * @param <T> The type of the principal.
+ */
+public abstract class AbstractExtensiblePrincipal<T extends Principal> implements ExtensiblePrincipal<T> {
+
+	private final Map<String, Principal> additionalInfo;
+
+	/**
+	 * Creates a new principal with no additional information.
+	 */
+	protected AbstractExtensiblePrincipal() {
+		this(null);
+	}
+
+	/**
+	 * Creates a new principal with additional information.
+	 * 
+	 * @param additionalInformation The additional information.
+	 */
+	protected AbstractExtensiblePrincipal(Map<String, Principal> additionalInformation) {
+		if (additionalInformation == null) {
+			this.additionalInfo = Collections.emptyMap();
+		} else {
+			this.additionalInfo = Collections.unmodifiableMap(additionalInformation);
+		}
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public final Map<String, Principal> getExtendedInfo() {
+		return additionalInfo;
+	}
+}
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/ApplicationLevelInfoSupplier.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/auth/ApplicationLevelInfoSupplier.java
@@ -1,0 +1,35 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ * 
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial creation
+ *******************************************************************************/
+
+
+package org.eclipse.californium.scandium.auth;
+
+import java.security.Principal;
+import java.util.Map;
+
+/**
+ * A strategy for retrieving additional (application level) information about an authenticated peer.
+ */
+public interface ApplicationLevelInfoSupplier {
+
+    /**
+     * Gets additional information about an authenticated peer.
+     * 
+     * @param peerIdentity The peer identity.
+     * @return The additional information about the peer.
+     */
+	Map<String, Principal> getInfo(Principal peerIdentity);
+}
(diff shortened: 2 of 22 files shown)


boaks:
My feeling is, that this would break the functionality of the `PrincipalSerializer`.
May be adding a `CustomPrincipalSerializer` to the config and  `PrincipalSerializer` fixes it.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -286,6 +287,8 @@
 	 */
 	private ConnectionIdGenerator connectionIdGenerator;
 
+	private PrincipalConverter principalConverter;
```
boaks: New fields must be included in the `protected Object clone()`.

boaks:
The test covers, that the `convert`is called?
A test, which verifies, that the returned Principal is used in the `EndpointContext` would also be interested.

sophokles73:
> A test, which verifies, that the returned Principal is used in the EndpointContext would also be interested.

will do

sophokles73:
> My feeling is, that this would break the functionality of the PrincipalSerializer.

Good point. I'll take a look at the existing serializer and try to come up with some options ...

boaks:
If the `PrincipalConverter` is moved into the "element-connector" module in the package `org.eclipse.californium.elements.auth`, it may be also used with TLS.

boaks:
> My feeling is, that this would break the functionality of the PrincipalSerializer.

> Good point. I'll take a look at the existing serializer and try to come up with some options ...

I think, the `PrincipalSerializer` must be redesigned to remove the `static`s. 
If a custom Prinicpal is used, the serialization may use a ClientAuthenticationType.CUSTOM to differentiate between the standard and custom principals.

boaks:
The new commit seems not to address the serialization issue, or?

element-connector/src/main/java/org/eclipse/californium/elements/auth/AbstractExtensiblePrincipal.java
```diff
@@ -0,0 +1,58 @@
+/**
```
boaks: Please replace that with the californium headers.
We may change that according EPL-2.0 after the 2.0 release.

sophokles73: sorry, settings from my Hono workspace keep getting in the way ...

sophokles73:
> The new commit seems not to address the serialization issue, or?

Well, my hope is that we do not need to include the application specific info in serialization because the information is restored (using the `ApplicationLevelInfoSupplier` every time a session is established. I think I have managed to do it for the standard full and abbreviated handshake. However, I am not sure if it already is done in case of a fail-over between nodes ...

scandium-core/src/main/java/org/eclipse/californium/scandium/auth/ApplicationLevelInfoSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface ApplicationLevelInfoSupplier {
+
+    /**
+     * Gets additional information about an authenticated client.
+     * 
+     * @param clientIdentity The client identity.
+     * @return The additional information for the client.
+     */
+    Set<Principal> getInfo(Principal clientIdentity);
```
boaks: The `Set`-solution may also fit for issue #974 .

Is there any intention to access a specific `Principal` in `Set`?
Any thoughts about using a `Map<String, Principal>` ?

boaks:
Except of the license header of the new files, LGTM.
If possible, please let us know, how you intend to operate on the Set or if a Map would do it better.

sophokles73:
A Map would also work for me. We could probably also use a Map<String, Object> instead.

boaks:
OK, so you didn't use the `Set` with a strong idea behind. I would prefer a `Map<String, Principal>`, as FMPOV this matches best for me.

sophokles73:
ok, I'll make the corresponding changes...

sophokles73:
@boaks I have pushed the changes ...

boaks:
LGTM

sbernard31:
The purpose of this PR is to avoid to "retrieve" each time application level identity from connection level identity, right ?

Why limited this feature to server side ?
Maybe the call of `amendClientPrincipal()` could be done on `sessionEstablished` ?

sophokles73:
> The purpose of this PR is to avoid to "retrieve" each time application level identity from connection level identity, right ?

right

> Why limited this feature to server side ?

Just because we only need it on server side in Eclipse Hono. We only implement a CoAP server which never acts as a client initiating a connection.
However, this could be added in another PR later on (if somebody needs it) ...

sbernard31:
To extends this features to both side server and client, I think it's mainly about : 
- naming : changing "client" by "foreign peer".
- moving amendPrincipal to `sessionEstablished()` in  handshaker ?

sophokles73:
@sbernard31 I'll give it a try ...

sophokles73:
@sbernard31 WDYT?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1118,4 +1125,28 @@ public void verifyCertificate(CertificateMessage message) throws HandshakeExcept
...
+	/**
+	 * Amends the peer principal with additional application level information.
+	 */
+	private void amendPeerPrincipal() {
+
+		Principal peerIdentity = session.getPeerIdentity();
+		if (peerIdentity instanceof ExtensiblePrincipal) {
+			// amend the client principal with additional application level information
+			@SuppressWarnings("unchecked")
+			ExtensiblePrincipal<? extends Principal> extensibleClientIdentity = (ExtensiblePrincipal<? extends Principal>) peerIdentity;
```
sbernard31: // amend the **peer** principal with additional application level information
@SuppressWarnings("unchecked")
ExtensiblePrincipal<? extends Principal> **extensibleIdentity** = (ExtensiblePrincipal<? extends Principal>) peerIdentity;

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -56,16 +56,21 @@
...
 import java.util.List;
+import java.util.Map;
 
+import org.eclipse.californium.elements.auth.ExtensiblePrincipal;
 import org.eclipse.californium.elements.auth.PreSharedKeyIdentity;
 import org.eclipse.californium.elements.auth.RawPublicKeyIdentity;
 import org.eclipse.californium.elements.auth.X509CertPath;
 import org.eclipse.californium.elements.util.DatagramWriter;
 import org.eclipse.californium.elements.util.StringUtil;
+import org.eclipse.californium.scandium.auth.ApplicationLevelInfoSupplier;
```
sbernard31: maybe some unused import in this class ?

scandium-core/src/test/java/org/eclipse/californium/scandium/ConnectorHelper.java
```diff
@@ -278,6 +281,15 @@ LatchDecrementingRawDataChannel givenAnEstablishedSession(DTLSConnector client,
 		return clientChannel;
 	}
 
+	static void assertClientPrincipalHasAdditionalInfo(Principal clientIdentity, String expectedName) {
```
sbernard31: assertPrincipalHasAdditionalInfo ?

boaks:
@sophokles73 
Please rebase to "master" and create a new PR towards master.
The "conflict" is in the unit tests. Should be easy to fix. If not, remove my new tests to resolve these conflicts, I will then add them again after 2.0.0-M16

sophokles73:
@boaks @sbernard31 I have created class `AdditionalInfo` which encapsuates a `Map<String, Object>` and provides a typed getter for the data. I have also rebased to master. Hope this works for everybody now :-)

boaks:
The PR is towards 2.0.x, which is since last Friday, "history".
Please create it towards "master".

boaks:
I checked the commits. they are already rebased on "master". Therefore the diff to 2.0.x get's large.
Please close this PR and open a new one towards master.

sophokles73:
replaced by #1014

PR #997, opened 2019-07-02
Add cbor and source address to receive test.

boaks:
Add argument to select configured or ephemeral port for client
endpoints.


sbernard31:
no objection

PR #998, opened 2019-07-03
Update to assigned IANA code points.

boaks:
Change TLS ExtensionType, connection_id value from 52 to 53.
Update the javadocs.


sbernard31:
:+1:

PR #999, opened 2019-07-03
Add warnings

boaks:
Improve dtls diagnose.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
@@ -112,7 +112,7 @@ public class Record {
 	/** The connection id. */
 	private ConnectionId connectionId;
 
-	/** padding to be used, if cid is used */
+	/** Padding to be used with cid */
 	private int padding;
 
 	/** The DTLS session. */
@@ -187,11 +187,11 @@ public class Record {
 	 *            Otherwise use {@code null} as connection id
 	 * @param pad if cid is enabled, pad could be used to add that number of
 	 *            zero-bytes as padding to the payload to obfuscate the payload
-	 *            length
+	 *            length.
 	 * @throws IllegalArgumentException if the given sequence number is longer
-	 *             than 48 bits or less than 0, the given epoch is less than
-	 *             0, the provided type is not supported or the fragment could
-	 *             not be converted into bytes.
+	 *             than 48 bits or less than 0, the given epoch is less than 0,
+	 *             the provided type is not supported or the fragment could not
+	 *             be converted into bytes.
 	 * @throws NullPointerException if the given type, fragment or session is
 	 *             {@code null}.
 	 * @throws GeneralSecurityException if the message could not be encrypted,
@@ -342,16 +342,21 @@ public class Record {
 			if (type == ContentType.TLS12_CID.getCode()) {
 				if (cidGenerator == null) {
 					LOGGER.debug("Received TLS_CID record, but cid is not supported. Discarding ...");
-					continue;
+					return records;
 				} else if (cidGenerator.useConnectionId()) {
-					connectionId = cidGenerator.read(reader);
-					if (connectionId == null) {
-						LOGGER.debug("Received TLS_CID record, but cid is not matching. Discarding ...");
-						continue;
+					try {
+						connectionId = cidGenerator.read(reader);
+						if (connectionId == null) {
+							LOGGER.debug("Received TLS_CID record, but cid is not matching. Discarding ...");
+							return records;
+						}
+					} catch (RuntimeException ex) {
+						LOGGER.debug("Received TLS_CID record, failed to read cid. Discarding ...", ex.getMessage());
+						return records;
 					}
 				} else {
 					LOGGER.debug("Received TLS_CID record, but cid is not used. Discarding ...");
-					continue;
+					return records;
 				}
 			}
 			int length = reader.read(LENGTH_BITS);
@@ -451,15 +456,15 @@ public class Record {
 		case NULL:
 			// do nothing
 			break;
-			
+
 		case AEAD:
 			result = decryptAEAD(ciphertextFragment, currentReadState);
 			break;
-			
+
 		case BLOCK:
 			result = decryptBlockCipher(ciphertextFragment, currentReadState);
 			break;
-			
+
 		case STREAM:
 			// Currently, Scandium does not support any stream ciphers
 			// RC4 is explicitly ruled out from being used in DTLS
@@ -521,14 +526,14 @@ public class Record {
 		plaintext.writeBytes(getBlockCipherMac(writeState, compressedFragment));
 
 		// determine padding length
-		int ciphertextLength = compressedFragment.length + writeState.getCipherSuite().getMacLength() + 1;
-		int smallestMultipleOfBlocksize = writeState.getRecordIvLength();
-		while ( smallestMultipleOfBlocksize <= ciphertextLength) {
-			smallestMultipleOfBlocksize += writeState.getRecordIvLength();
+		int ciphertextLength = compressedFragment.length + writeState.getMacLength() + 1;
+		int blocksize = writeState.getRecordIvLength();
+		int paddingLength = ciphertextLength % blocksize;
+		if (0 < paddingLength) {
+			// padding to fill last block
+			paddingLength = blocksize - paddingLength;
 		}
-		int paddingLength = smallestMultipleOfBlocksize % ciphertextLength;
-
-		// create padding
+ 		// create padding
 		byte[] padding = new byte[paddingLength + 1];
 		Arrays.fill(padding, (byte) paddingLength);
 		plaintext.writeBytes(padding);
@@ -576,7 +581,11 @@ public class Record {
 			throw new NullPointerException("Current read state must not be null");
 		} else if (ciphertextFragment == null) {
 			throw new NullPointerException("Ciphertext must not be null");
+		} else if (ciphertextFragment.length < currentReadState.getRecordIvLength() + currentReadState.getMacLength()
+				+ 1) {
+			throw new GeneralSecurityException("Ciphertext too short!");
 		}
+
 		/*
 		 * See http://tools.ietf.org/html/rfc5246#section-6.2.3.2 for
 		 * explanation
@@ -593,10 +602,10 @@ public class Record {
 		int fragmentLength = plaintext.length
 				- 1 // paddingLength byte
 				- paddingLength
-				- currentReadState.getCipherSuite().getMacLength();
+				- currentReadState.getMacLength();
 
 		reader = new DatagramReader(plaintext);
-		byte[] content = reader.readBytes(fragmentLength);			
+		byte[] content = reader.readBytes(fragmentLength);
 		byte[] macFromMessage = reader.readBytes(currentReadState.getCipherSuite().getMacLength());
 		byte[] mac = getBlockCipherMac(currentReadState, content);
 		if (Arrays.equals(macFromMessage, mac)) {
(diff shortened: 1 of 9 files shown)


boaks:
I would like to merge this soon and release 2.0.0-M16 begin of next week.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -297,10 +308,11 @@ private Record(ProtocolVersion version, int epoch, long sequenceNumber) {
 	 * @param peerAddress the IP address and port of the peer from which the bytes have been
 	 *           received
 	 * @param cidGenerator the connection id generator. May be {@code null}.
+	 * @param receiveNanos uptime nanoseconds of receiving this record
 	 * @return the {@code Record} instances
 	 * @throws NullPointerException if either one of the byte array or peer address is {@code null}
 	 */
-	public static List<Record> fromByteArray(byte[] byteArray, InetSocketAddress peerAddress, ConnectionIdGenerator cidGenerator) {
+	public static List<Record> fromByteArray(byte[] byteArray, InetSocketAddress peerAddress, ConnectionIdGenerator cidGenerator, long receiveNanos) {
```
sbernard31: A solution with less modification could have been to :
- not add this new receiveNanos argument.
- begin the method with a `ClockUtil.nanoRealtime()`

But it's ok like that too.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -330,16 +342,16 @@ private Record(ProtocolVersion version, int epoch, long sequenceNumber) {
...
 					connectionId = cidGenerator.read(reader);
 					if (connectionId == null) {
 						LOGGER.debug("Received TLS_CID record, but cid is not matching. Discarding ...");
-						continue;
+						return records;
 					}
 				} else {
 					LOGGER.debug("Received TLS_CID record, but cid is not used. Discarding ...");
-					continue;
+					return records;
```
sbernard31: was it a bug ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -521,12 +525,13 @@ private Record(ProtocolVersion version, int epoch, long sequenceNumber) {
 		plaintext.writeBytes(getBlockCipherMac(writeState, compressedFragment));
 
 		// determine padding length
-		int ciphertextLength = compressedFragment.length + writeState.getCipherSuite().getMacLength() + 1;
```
sbernard31: writeState.getCipherSuite().getMacLength() and writeState.getMacLength() should contains the same value right ? 

(it's just to make the code clearer?)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -521,12 +525,13 @@ private Record(ProtocolVersion version, int epoch, long sequenceNumber) {
...
-		int smallestMultipleOfBlocksize = writeState.getRecordIvLength();
+		int ciphertextLength = compressedFragment.length + writeState.getMacLength() + 1;
+		int blocksize = writeState.getRecordIvLength();
+		int smallestMultipleOfBlocksize = blocksize;
 		while ( smallestMultipleOfBlocksize <= ciphertextLength) {
-			smallestMultipleOfBlocksize += writeState.getRecordIvLength();
+			smallestMultipleOfBlocksize += blocksize;
 		}
-		int paddingLength = smallestMultipleOfBlocksize % ciphertextLength;
+		int paddingLength = smallestMultipleOfBlocksize - ciphertextLength;
```
sbernard31: was it a bug ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -447,27 +447,31 @@ private Record(ProtocolVersion version, int epoch, long sequenceNumber) {
...
+				// Currently, Scandium does not support any stream ciphers
+				// RC4 is explicitly ruled out from being used in DTLS
+				// see http://tools.ietf.org/html/rfc6347#section-4.1.2.2
+				break;
+
+			default:
+				break;
+			}
+		} catch(RuntimeException ex) {
+			throw new GeneralSecurityException(ex.getMessage(), ex);
```
sbernard31: Should we add a LOG with warn because this should not happened ?

I mean the idea would be to only raise "GeneralSecurityException" for case we understand (anticipated case).
A runtime would be an unexpected error, we don't want this error terminate the connection but we want to know when this happened to fix the code and  handle it in a more proper way.

boaks:
@sbernard31 

I changed the commits. the main part is a added unit test, which check the processing of damaged records.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks: yes.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks: No. The smallestMultipleOfBlocksize is just a little larger than ciphertextLength.

ctl = 103;
blockssize = 7 * 16 = 112

112 % 103 => 9
112 - 103 => 9

(I was surprised by this calculation. I will go to remove the loop, but later ...).  
e.g. 16*16

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks: I added an extensive test for malformed records. 
Therefore I removed the exception conversion.

boaks:
I added also an address check for the send.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -521,14 +526,14 @@ private Record(ProtocolVersion version, int epoch, long sequenceNumber, long rec
 		plaintext.writeBytes(getBlockCipherMac(writeState, compressedFragment));
 
 		// determine padding length
-		int ciphertextLength = compressedFragment.length + writeState.getCipherSuite().getMacLength() + 1;
-		int smallestMultipleOfBlocksize = writeState.getRecordIvLength();
-		while ( smallestMultipleOfBlocksize <= ciphertextLength) {
-			smallestMultipleOfBlocksize += writeState.getRecordIvLength();
+		int ciphertextLength = compressedFragment.length + writeState.getMacLength() + 1;
```
sbernard31: the `+1` is the paddingLenght byte ?

scandium-core/src/test/java/org/eclipse/californium/scandium/dtls/RecordDecryptTest.java
```diff
@@ -0,0 +1,374 @@
...
+	private void testEncryptDecrypt(byte[] payload) throws GeneralSecurityException, HandshakeException {
+		Record record = new Record(ContentType.APPLICATION_DATA, EPOCH, 1,
+				new ApplicationMessage(payload, session.getPeer()), session, true, 0);
+		byte[] raw = record.toByteArray();
+		List<Record> list = Record.fromByteArray(raw, session.getPeer(), null, ClockUtil.nanoRealtime());
+		assertFalse("failed to decode raw message", list.isEmpty());
+		for (Record recv : list) {
+			recv.setSession(session);
+			DTLSMessage message = recv.getFragment();
+			assertArrayEquals("decrpted payload differs", payload, message.toByteArray());
```
sbernard31: typo : "decrypted"

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks: yes, I also did a test against to old implementation. It shows, that the old has in some cases extra 32 bytes (caused by the `<=` instead `<`). The added test shows, that it works not too bad :-)

scandium-core/src/test/java/org/eclipse/californium/scandium/dtls/RecordDecryptTest.java
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -521,14 +526,14 @@ private Record(ProtocolVersion version, int epoch, long sequenceNumber, long rec
...
+		int blocksize = writeState.getRecordIvLength();
+		int paddingLength = ciphertextLength % blocksize;
+		if (0 < paddingLength) {
```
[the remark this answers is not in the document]
boaks: To be frank: nobody cares for years. So I don't think, this detail is worth to do much more :-).

boaks:
> There is 2 previous comments, you maybe missed ?

> Could we change log level in DTLSConnector ?
> I would change from INFO to WARN for all RuntimeException catched which terminate the connection, generally with a message like : "Unexpected error occurred ..."
> (I think there is 3 log like this)

That's one comment, or? What was the other?
I didn't answer because I was really too busy in the last days.
Yes we can change the log level.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks: It's more a optimization.
Failing to read the record to the end will fail the next record as well. So the result will close to almost be the same.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks: There may be some additional PR in the future. That is just a part form the PR.
It propagates the times of sending a response and receiving a repeated request to optimize sending the response again. With that, the receive time will be passed around some more functions and so I put it in the record. 
But not for now.

sbernard31:
> That's one comment, or? What was the other?

I don't get it.

> I didn't answer because I was really too busy in the last days.

No problem, I just highlighted it in case you missed it.

> Yes we can change the log level.

You will do it in this PR or you prefer I do it ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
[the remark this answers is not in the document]
sbernard31: My feeling, I blocked when reading this part of the code just because of the variable naming.
When I faced that, generally I try to fix it.
(I often face that when I read my code long time after. ^^)

I you want I can do a PR just with this little change of code and the changes about info/warning log ?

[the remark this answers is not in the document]
boaks: OK,
May be remainingBytes := lastBlocksBytes but you may go for what ever you want :-).

boaks:
> I don't get it.

You wrote 

> There is 2 previous comments, you maybe missed ?

Yes, but then you only repeated one.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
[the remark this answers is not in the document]
sbernard31: OK pour lastBlocksBytes ! 
I wait you merge this PR, then create a new one.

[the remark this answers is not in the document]
boaks: OK, so LGTM?

sbernard31:
LGTM :)

Issue #990, opened 2019-06-28
OSCORE: remove single instance pattern from HashMapCtxDB 


boaks:
Add the `CoapThreadsRule` I got aware, that the current "single instance" approach for the `OSCoreObserveTest` has its drawbacks. I would prefer, if a `HashMapCtxDB` is passed into the `CoapStack` by the `OSCoreCoapStackFactory` and then stored in the `ObjectSecurityLayer`. The changes seems to be have limited scope. I would add also a `setHashMapCtxDB` to the    `OSCoreCoapStackFactory`, so that shared `HashMapCtxDB` will still be possible.

If agreed, I can implement that next week.

@rikard-sics
@LudwigSeitz 

Any opinion?

rikard-sics:
So then the following:
`private final static HashMapCtxDB db = HashMapCtxDB.getInstance();`
Would instead become something like?
`private final static HashMapCtxDB db = ObjectSecurityLayer.getHashMapCtxDBInstance();`

And sharing the DB between a client and server like in the OSCoreObserveTest can be done with:
`OSCoreCoapStackFactor.setHashMapCtxDB(db);`
(I suppose sharing a DB like is done in the OSCoreObserveTest is not really needed, it was just for convenience to have one DB for the whole JUnit class. Normally the client and server would be two separate applications.)

The main point for me would be to keep it possible to get and modify the DB from different parts of an application in a convenient way. With your proposed changes one application would in general still have a single DB? And it would be accessible via some method both from the OSCORE code in Californium and inside code of an application using Californium?

Currently getting and modifying the DB is used in many places with:
`OSCoreCtxDB db = HashMapCtxDB.getInstance();`
So keeping it possible to do that somehow in a convenient way from different parts of the code would be good.

boaks:
The complete static stuff around the `HashMapCtxDB` must be removed.
The `ObjectSecurityLayer` holds it's (optionally shared) `HashMapCtxDB` and use that for all operations and pass it into the other classes, which access the `HashMapCtxDB`.

boaks:
> in many places

Hm, my feeling was, it's mainly the `ObjectSecurityLayer` and a few encrypt/decrypt functions.

rikard-sics:
> The complete static stuff around the `HashMapCtxDB` must be removed.
> The `ObjectSecurityLayer` holds it's (optionally shared) `HashMapCtxDB` and use that for all operations and pass it into the other classes, which access the `HashMapCtxDB`.

[...]

PR #1000, opened 2019-07-04
Add argument for custom coap stack.

boaks:
Issue #990
Replace HashMapCtxDB singleton by the introduced argument.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -79,6 +79,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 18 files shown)


boaks:
@rikard-sics
@LudwigSeitz 

Please feel free to comment, and, if possible, check, if these changes works for you.
I added a default ctx db to the factory and a custom stack argument to the CoapEndpoint.Builder.

rikard-sics:
Yes that looks like it should work for me. Then it is the responsibility of the application to create the context DB (which it could for instance create as a singleton) and set it when setting the default CoAP stack factory with OSCoreCoapStackFactory.useAsDefault(db). It is also possible to use multiple context DB:s and specify which one to use when building a CoapEndpoint.

Some thoughts I had.
1. For the context re-derivation functionality I was planning some changes so the client can set a flag in the actual context to indicate it should be re-derived. I already have some flags in the context like includeContextId (include the context id in messages generated using this context). I will make a PR for this and using the approach you suggested with an asynchronous client sending the re-derivation message.

2. I know Ludwig is using this code in some work he is doing where he needs to retrieve contexts in various situations. Hopefully this approach will work for his use case.

3. Should it be possible to retrieve the currently set default context DB from the OSCoreCoapStackFactory? For instance OSCoreCoapStackFactory.getDefaultCtxDb() which can return the value of defaultCtxDb?

boaks:
2. and 3.

In my opinion, if the application requires access to the  context DB, it's totally OK, if the application provides a separate application level singleton approach. Doing so obsoletes the access to the default context db of the factory. Currently the default of the factory could be changed (due to the requirements of the unit tests), so accessing it may lead to unexpected things.

rikard-sics:
>     1. and 3.
> 
> 
> In my opinion, if the application requires access to the context DB, it's totally OK, if the application provides a separate application level singleton approach. Doing so obsoletes the access to the default context db of the factory. Currently the default of the factory could be changed (due to the requirements of the unit tests), so accessing it may lead to unexpected things.

Alright, also hopefully the implementation of the OSCORE peer identity will help in Ludwig's case so he does not need access to the context DB as much.

PR #1002, opened 2019-07-08
OSCORE: Implements setting of peer identity and Principal for OSCORE

rikard-sics:
This commit adds a new Principal for endpoint context information when using OSCORE. It will be set in the endpoint context when communicating using OSCORE.

Currently it will only update the endpoint context if it is of type AddressEndpointContext and if the Principal is of type OSCorePeerIdentity.This means that it will not overwrite Principal information set by Scandium when using DTLS. So if both DTLS and OSCORE are used DTLS peer identity information will take precedence.

This should resolve the following issue:
https://github.com/eclipse/californium/issues/974

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
@@ -17,6 +17,8 @@
 package org.eclipse.californium.oscore;
 
 import org.eclipse.californium.core.coap.Message;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.elements.AddressEndpointContext;
 import org.eclipse.californium.elements.EndpointContext;
 import org.eclipse.californium.elements.auth.OSCorePeerIdentity;
@@ -27,29 +29,68 @@ import org.eclipse.californium.elements.auth.OSCorePeerIdentity;
  * It provides functionality to set an appropriate OSCORE peer identity
  * for an EndpointContext.
  *
+ * Will only add a peer identity for endpoint contexts that have
+ * an empty or OSCORE peer identity. If for instance Scandium has
+ * set another type of Principal as peer identity it will not be updated.
+ *
  */
 public class OSCorePeerIdentityHandler {
 
 	/**
-	 * Updates the currently set endpoint contexts adding an OSCORE peer identity.
-	 * Will only update endpoint contexts that are set as type AddressEndpointContext
-	 * and have an empty or OSCORE peer identity.
-	 * 
-	 * If for instance Scandium has set another type of EndpointContext or another
-	 * type of Principal as peer identity it will not be updated.
-	 * 
+	 * Sets destination endpoint context for outgoing requests with an OSCORE peer identity.
+	 *
 	 * @param oscoreCtx the OSCORE context used for this message
-	 * @param message the message to set the endpoint context for
+	 * @param request the request to set the endpoint context for
 	 */
-	public static void setPeerIdentity(OSCoreCtx oscoreCtx, Message message) {
+	public static void sendingRequest(OSCoreCtx oscoreCtx, Request request) {
+		setPeerIdentityOutgoing(oscoreCtx, request);
+	}
 
+	/**
+	 * Sets source endpoint context for incoming requests with an OSCORE peer identity.
+	 *
+	 * @param oscoreCtx the OSCORE context used for this message
+	 * @param request the request to set the endpoint context for
+	 */
+	public static void receivingRequest(OSCoreCtx oscoreCtx, Request request) {
+		setPeerIdentityIncoming(oscoreCtx, request);
+	}
+
+	/**
+	 * Sets destination endpoint context for outgoing responses with an OSCORE peer identity.
+	 *
+	 * @param oscoreCtx the OSCORE context used for this message
+	 * @param response the response to set the endpoint context for
+	 */
+	public static void sendingResponse(OSCoreCtx oscoreCtx, Response response) {
+		setPeerIdentityOutgoing(oscoreCtx, response);
+	}
+
+	/**
+	 * Sets destination endpoint context for incoming responses with an OSCORE peer identity.
+	 *
+	 * @param oscoreCtx the OSCORE context used for this message
+	 * @param response the response to set the endpoint context for
+	 */
+	public static void receivingResponse(OSCoreCtx oscoreCtx, Response response) {
+		setPeerIdentityIncoming(oscoreCtx, response);
+	}
+
+	/**
+	 * Updates the currently set destination endpoint context for outgoing
+	 * messages adding an OSCORE peer identity.
+	 *
+	 * @param oscoreCtx the OSCORE context used for this message
+	 * @param message the message to set the destination endpoint context for
+	 */
+	private static void setPeerIdentityOutgoing(OSCoreCtx oscoreCtx, Message message) {
 		//Create a new OSCORE peer identity
 		OSCorePeerIdentity newPeerIdentity = new OSCorePeerIdentity(oscoreCtx.getUri(),
 				oscoreCtx.getIdContext(), oscoreCtx.getRecipientId(), oscoreCtx.getSenderId());
 
 		//Set updated destination context with new OSCORE peer identity
 		EndpointContext dstContext = message.getDestinationContext();
-		if(dstContext instanceof AddressEndpointContext &&
+		if(dstContext != null &&
 				(dstContext.getPeerIdentity() == null || dstContext.getPeerIdentity() instanceof OSCorePeerIdentity)) {
 
 			AddressEndpointContext newContext = new AddressEndpointContext(dstContext.getPeerAddress(),
@@ -57,10 +98,23 @@ public class OSCorePeerIdentityHandler {
 
 			message.setDestinationContext(newContext);
 		}
+	}
+
+	/**
+	 * Updates the currently set source endpoint context for incoming
+	 * messages adding an OSCORE peer identity.
+	 *
+	 * @param oscoreCtx the OSCORE context used for this message
+	 * @param message the message to set the source endpoint context for
+	 */
+	private static void setPeerIdentityIncoming(OSCoreCtx oscoreCtx, Message message) {
+		//Create a new OSCORE peer identity
+		OSCorePeerIdentity newPeerIdentity = new OSCorePeerIdentity(oscoreCtx.getUri(),
+				oscoreCtx.getIdContext(), oscoreCtx.getRecipientId(), oscoreCtx.getSenderId());
 
 		//Set updated source context with new OSCORE peer identity
 		EndpointContext srcContext = message.getSourceContext();
-		if(srcContext instanceof AddressEndpointContext &&
+		if(srcContext != null &&
 				(srcContext.getPeerIdentity() == null || srcContext.getPeerIdentity() instanceof OSCorePeerIdentity)) {
 
 			AddressEndpointContext newContext = new AddressEndpointContext(srcContext.getPeerAddress(),
@@ -68,7 +122,5 @@ public class OSCorePeerIdentityHandler {
 
 			message.setSourceContext(newContext);
 		}
-
 	}
-
 }
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
@@ -1,0 +1,74 @@
+/*******************************************************************************
+ * Copyright (c) 2019 RISE SICS and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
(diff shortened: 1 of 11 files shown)


rikard-sics:
As we discussed in the reported issue regarding this I can update the code further if it is better to add functionality to modify the peer identity of existing endpoint contexts. Currently I create new ones with the same content except updating the Principal.

LudwigSeitz:
I don't quite like your take on Principal.getName(). Basically you return a string representation of some parts of the OSCORE context, where I would expect the RecipientId (which is the name of the other party linked to this OSCORE context). This will make my code quite ugly since I really need the Principal's name and not anything else to decide what access control rules apply. Now I'm left to either parse that name you generated, or do an <instanceof> on every Principal to check if it's an OSCorePeerIdentity.

rikard-sics:
> I don't quite like your take on Principal.getName(). Basically you return a string representation of some parts of the OSCORE context, where I would expect the RecipientId (which is the name of the other party linked to this OSCORE context). This will make my code quite ugly since I really need the Principal's name and not anything else to decide what access control rules apply. Now I'm left to do an "instanceof" on every Principal to check if it's an OSCorePeerIdentity.

Thanks for the feedback, I see your point

I could definitely change the code so that getName() returns a string representation of the other party's sender ID (which would be your recipient ID).

As an alternative I could perhaps make the structure of the string returned from getName() such that the information you need is easily parsable from the string.

LudwigSeitz:
If getName() has a structure that I need to parse that is a problem, since I don't readily know I need to parse it (I don't in DTLS). I can do a "if (peerIdentity instanceof OsCorePeerIdentity)" every time I use the identity, but it doesn't feel right and would probably be prone to errors when others use my ACE-library.

rikard-sics:
> If getName() has a structure that I need to parse that is a problem, since I don't readily know I need to parse it (I don't in DTLS). I can do a "if (peerIdentity instanceof OsCorePeerIdentity)" every time I use the identity, but it doesn't feel right and would probably be prone to errors when others use my ACE-library.

Sure, if it is alright from the Californium side and Achim I can change getName() to return a string representation of the sender ID of the other party. Potentially combined with the context ID if you need that also?

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
```diff
@@ -0,0 +1,74 @@
...
+	 * Will only update endpoint contexts that are set as type AddressEndpointContext
+	 * and have an empty or OSCORE peer identity.
+	 * 
+	 * If for instance Scandium has set another type of EndpointContext or another
+	 * type of Principal as peer identity it will not be updated.
+	 * 
+	 * @param oscoreCtx the OSCORE context used for this message
+	 * @param message the message to set the endpoint context for
+	 */
+	public static void setPeerIdentity(OSCoreCtx oscoreCtx, Message message) {
```
boaks: Are you sure, that changing the destination and source context of the same message makes sense? FMPOV, that is some stuff related to security. So I would prefer to have sendMessage and receiveMessage (or may be sendRequest, sendResponse, receiveRequest, and receiveResponse)

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
```diff
@@ -0,0 +1,74 @@
...
+
+		//Create a new OSCORE peer identity
+		OSCorePeerIdentity newPeerIdentity = new OSCorePeerIdentity(oscoreCtx.getUri(),
+				oscoreCtx.getIdContext(), oscoreCtx.getRecipientId(), oscoreCtx.getSenderId());
+
+		//Set updated destination context with new OSCORE peer identity
+		EndpointContext dstContext = message.getDestinationContext();
+		if(dstContext instanceof AddressEndpointContext &&
```
boaks: That would change a MapBasedEndpointContext into a AddressEndpointContext. 
A design, using a method in the context, may be overriden and therefore can keep the MapBasedEndpointContext or AddressEndpointContext.

boaks: My feeling is, waiting for the other PR #996 would make a implementation easier.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
rikard-sics: My understanding was that by the time a request or response reaches the ObjectSecurityLayer and the OSCORE code it will already have gotten a source or destination endpoint context set. So therefore I am changing it.

But in practice I create the new one with the exact same values as the old one except with an OSCorePeerIdentity as Principal. Basically if an endpoint context is not null I re-create it with the same values but setting a peer identity for OSCORE. 

For every message I will only change either the source or destination endpoint context (depending on which is set) because one of them will be null. At least that is how I understand things from testing the current code.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
rikard-sics: > That would change a MapBasedEndpointContext into a AddressEndpointContext.
> A design, using a method in the context, may be overriden and therefore can keep the MapBasedEndpointContext or AddressEndpointContext.

I check that the type of EndpointContext is AddressEndpointContext, otherwise I do not change it. This is to respect the values Scandium sets for DTLS (I also saw that there are different types of EndpoinContext:s that Scandium uses).

rikard-sics: > My feeling is, waiting for the other PR #996 would make a implementation easier.

Sure, I can wait for that and update my code when it is merged. If there would be a method setPrincipal or similar in the EndpointContext I can use that and not re-create the context with the same values plus a Principal.

Then I could also more easily change the peer identity for any kind of endpoint context type. But I was also worried that if I change it for a DTLS connection it may break some functionality in Scandium.

Currently I only update the context if it is of type AddressEndpointContext and if the existing peer identity is null or of type OSCorePeerIdentity.

boaks: > I check that the type of EndpointContext is AddressEndpointContext

Though `MapBasedEndpointContext extends AddressEndpointContext` a MapBasedEndpointContext will pass that "instanceof" test.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
boaks: If a message is sent, the destination context is used, the source will be null.
Now you fill it.
If a message is received, the source context is used, the destination will be null.
Now you fill it.
So, please don't fill the context just because it's null. use the destination for sending and source for receiving. That is distinguished to make it clear, what is used/checked and what not.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
rikard-sics: > Though `MapBasedEndpointContext extends AddressEndpointContext` a MapBasedEndpointContext will pass that "instanceof" test.

I see, I did not notice that. Thanks for pointing it out.

I suppose it could be fixed in code to check that it is exactly an AddressEndpointContext but as you mention probably it is better to wait for PR https://github.com/eclipse/californium/pull/996

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
rikard-sics: > If a message is sent, the destination context is used, the source will be null.
> Now you fill it.
> If a message is received, the source context is used, the destination will be null.
> Now you fill it.
> So, please don't fill the context just because it's null. use the destination for sending and source for receiving. That is distinguished to make it clear, what is used/checked and what not.

Let me elaborate. I do not fill or touch the context if it is null, due to:
`if(dstContext instanceof AddressEndpointContext &&`
(Which will fail if dstContext is null)
So I actually only update it when it is not null. But I want the peer identity principal in the context to either be null (unset) or OSCorePeerIdentity (set by OSCORE) for me to update it. (To avoid changing things set by Scandium.)

I can certainly change the structure of things so they are more clear what is happening during sending or receiving.

boaks: OK. I was mislead/mixed up by the null check for the principal.

rikard-sics:
> So I would prefer to have sendMessage and receiveMessage (or may be sendRequest, sendResponse, receiveRequest, and receiveResponse)

I made a new commit now that splits the functionality so there are separate methods for setting the peer identity in the (source or destination) endpoint context when sending a request, receiving a request, sending a response and receiving a response.
https://github.com/eclipse/californium/pull/1002/commits/4bfe85d49831a1e95fe17c6e0d2ad370646975fd

As we discussed I will wait for https://github.com/eclipse/californium/pull/996 and amend the code further after that. And as I discussed with Ludwig I could change what getName() returns for the OSCorePeerIdentity Principal.

rikard-sics:
Should I change this PR to point to the master branch instead?

boaks:
> Should I change this PR to point to the master branch instead?

Yes, please do so. 
I'm still waiting for PR #996 to continue with this PR.
Is it important for you to have this PR in 2.0.0-M16, or would it also be OK to do that afterwards?

rikard-sics:
> Yes, please do so.

Alright, I will rebase it against master and then change the PR to point to master.

> I'm still waiting for PR #996 to continue with this PR.

Sure. Then would the idea be to use the amend method to add information for OSCORE if an endpoint context has an existing Principal?
But in the case no principal is set in an endpoint context I would need some method for modifying the peerIdentity, or keep the way I do it now re-creating the context (which is not ideal since the context can be many different types and now I always re-create it as an AddressEndpointContext). (Although I could check the exact type and make sure to re-create it as the exact type it was.)

To enable OSCORE to set a peer identity when none is set I could add a setPeerIdentity method in the AddressEndpointContext. That would be sufficient for this PR to allow me to set the peer identity when it is null (and simply not set it if already set by Scandium for DTLS). Then when https://github.com/eclipse/californium/pull/996 is merged I can add code so OSCORE also sets its Principal but as additional information if an endpoint context already has a principal set.

> Is it important for you to have this PR in 2.0.0-M16, or would it also be OK to do that afterwards?

It may make things more difficult for implementers if 2.0.0-M16 has the context db no longer as a singleton but yet no easy access to OSCORE information via the peer identity. I know this could potentially affect Ludwig's work. On my side I am doing some integration of OSCORE in Leshan and it would be nice to have access to OSCORE-related information for an incoming/outgoing message without needing a reference to the context db. (Currently there is a trick where OSCORE-related information can be retrieved from the OSCORE context via the Token of a message).

boaks:
> It may make things more difficult for implementers if 2.0.0-M16 has the context db no longer as a singleton but yet no easy access to OSCORE information via the peer identity. 

I'm not sure, if we have a common understanding.
For a library, potentially used by different applications and other libraries, a singleton, if not required by something, is mostly no good idea, because it would force all applications and libraries into use-cases, where the "not-required" singleton fits. Therefore, I tried to change the library part of OSCORE.
For an application or an other library, which **requires** a singleton, it's totally OK to have that and maintain that on it's own. It could be passed in as default argument to the factory. So, I don't see a bigger issue than moving some references from the "old library singleton" to a "new application (or specific library) singleton". That should be "just" some typing.  
Or do I oversee something?

LudwigSeitz:
I can do the necessary foo to get the OSCORE context in my application (by creating a singleton class on my application level that hold the context database that I pass into all californium EnpointBuilders that create an OSCORE stack). So I can live with #1002 not getting into M16.

boaks:
Thanks! 2.0.0-M16 will not be the last ... we will integrate this PR after that milestone.

rikard-sics:
> For an application or an other library, which **requires** a singleton, it's totally OK to have that and maintain that on it's own. It could be passed in as default argument to the factory. So, I don't see a bigger issue than moving some references from the "old library singleton" to a "new application (or specific library) singleton".

Yes I see your point, that sounds very reasonable. 

Ludwig also confirms it's no issue for him.

Then we can definitely wait until after 2.0.0-M16 for this PR.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCorePeerIdentityHandler.java
```diff
@@ -0,0 +1,126 @@
...
+	 * messages adding an OSCORE peer identity.
+	 *
+	 * @param oscoreCtx the OSCORE context used for this message
+	 * @param message the message to set the destination endpoint context for
+	 */
+	private static void setPeerIdentityOutgoing(OSCoreCtx oscoreCtx, Message message) {
+		//Create a new OSCORE peer identity
+		OSCorePeerIdentity newPeerIdentity = new OSCorePeerIdentity(oscoreCtx.getUri(),
+				oscoreCtx.getIdContext(), oscoreCtx.getRecipientId(), oscoreCtx.getSenderId());
+
```
boaks: `OSCorePeerIdentity(String host, byte[] contextId, byte[] senderId, byte[] recipientId)` sender and recipient id seems to be provided in exchanged order?

rikard-sics: Yes, I did this on purpose with the intention that the peer identity contains the settings of that peer. In other words the sender ID of the peer is my recipient ID, and the recipient ID of the peer is my sender ID. That's why the parameters taken from the local context are switched.

rikard-sics:
Closing this PR to create a new one based on discussion in https://github.com/eclipse/californium/issues/974

PR #1008, opened 2019-07-11
Extend CountingMessageObserver and use it in unit tests.

boaks:


boaks:
Exchanged with PR to master.

PR #1012, opened 2019-07-13
Update address only for newest tls12_cid records.

boaks:
Fix updates with already deprecated addresses.


boaks:
I've updated the implementation, because `connectionStore.update(connection, newAddress);` is also required to prevent, that the connection is evicted.

There is one issue left:
If a connection (association) has no address (because it's initial address was used for an other connection), and a deprecated record is processed, it's not possible to send a message back.

boaks:
In the meantime, I think, that this requires a extension of the test NAT to explicitly reorder and delay messages to verify the function. But not before the weekend, so this will come next week :-).

PR #1014, opened 2019-07-16
Support storing application level info with authenticated peer's principal

sophokles73:


sophokles73:
@boaks please wait with merging, I need to check something first ...

sophokles73:
@boaks ok, all set. I forgot to commit my latest changes ;-)
@sbernard31 does this work for you as well?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2018 Institute for Pervasive Computing, ETH Zurich and others.
+ * Copyright (c) 2015, 2019 Institute for Pervasive Computing, ETH Zurich and others.
```
sbernard31: I feel this class should not be modified but you find a unused import so why not :)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -984,6 +991,7 @@ protected final void handshakeStarted() throws HandshakeException {
 	}
 
 	protected final void sessionEstablished() throws HandshakeException {
+		amendPeerPrincipal();
```
sbernard31: I don't kown if `amendPeerPrincipal` shoud be in or out the `if block` :thinking:

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -63,10 +63,12 @@
 import java.security.GeneralSecurityException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.security.Principal;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
+import java.util.Collections;
```
boaks: Not used.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: I assume, that `amendPeerPrincipal` should only be called once per handshake. To ensure that, consider to move it after

```
if (!sessionEstablished) {
	sessionEstablished = true;
```

boaks:
Is there a intended use-case, which requires the public `Principal`'s constructors with the `AdditionalInfo` parameter? If not, I would remove them. Beside of that and the "small" comments above, LGTM.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sophokles73: just out of curiosity: under what circumstances will *sessionEstablished* be already be `true` when this method is invoked?

sophokles73:
> Is there a intended use-case, which requires the public Principal's constructors with the AdditionalInfo parameter? If not, I would remove them.

No, there isn't. Thanks for pointing this out. Will make them private ...

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: Your right. It was introduced to ensure, that this doesn't happen accidentally (may be in the future).
May be we convert that into an exception, but after 2.0.0-M16 .

boaks:
The most could be removed, only the one private with the most parameter is required.
For `PreSharedKeyIdentity` you may consider to remove the second private one (that with the additional "name").

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sophokles73: Does this mean that I should keep the method invocation at its current position?

boaks: If you want, you could keep it there. 
I would still prefer to move it.

sophokles73: ok, will do

Issue #905, opened 2019-03-12
Scope of release 2.0 - 2019


boaks:
Since issue #556 of last year californium 2.0 (with java 1.7 promise) made much progress in stability and some in functionality. So I think, it's time to ask again for the "final set" of fixes and features for a
2.0 release.

One of the major topics seems to be the stability of the APIs.
From the experience of the last two year, the APIs will still slightly change.
I don't see, that this damages too much.

My open topics/issue:
- CoapClient - observer/notify - since the clustered-observe this seems to be somehow "in between".
- Deduplicator - those who tried the "extplugtest" benchmarks may have recognized, that a lot of heap is required for message-deduplication. One part may be to reduce the kept data, the other part will be to add an additional deduplicator based on `SoftReference` or something else, which will remove older messages even before the exchange time expires in order to be able to process new messages. Potentially we can also open the implementation to support "custom deduplicators".
- cleanup  of scandium, still in progress, may be postponed after the release.

Feel free to write your feedback, opinion,  or your requirements wish-list.

And no, 
- no DTLS 1.3 support for californium 2.0. I think, that should be moved into a 3.0 with java 11. There we have the basic crypto-functions to do so. 
- no LWM2M-1.1 stuff, the 2.0 is a base for leshan 1.0, and that's LWM2M 1.0. Any extensions, if contributed, will be applied to a later version.

sbernard31:
> Potentially we can also open the implementation to support "custom deduplicators"

It could be a good approach.

About feature, I need to think more about https://github.com/eclipse/californium/issues/617

boaks:
> About feature, I need to think more about #617

Please, consider to make it configurablet :-).

boaks:
My current state:

> CoapClient - observer/notify - since the clustered-observe this seems to be somehow "in between".

I would prefer to go back to the simple old solution in the client before the clustered observe. I don't see easy implementations and no one else implemented something.
The solution with "many" handlers, each called to check it's token, has too many time penalties.

> Deduplicator

Still to be done.

> cleanup of scandium 

At least my open stuff may be merged at the end of next week.

So anything else open/wanted?

boaks:
> Potentially we can also open the implementation to support "custom deduplicators".

I found, that this was already implemented in the `DeduplicatorFactory`.

boaks:
[Forgotten issues in CoapClient](https://github.com/eclipse/californium/pull/532#discussion_r163281516)

I still think, that the javadoc fix would be an improvement.

PR #1020, opened 2019-07-25
Fix CBC padding check.

boaks:
Check all padding bytes for the right values.
 
### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
@@ -579,6 +579,8 @@ public class Record {
 			throw new NullPointerException("Current read state must not be null");
 		} else if (ciphertextFragment == null) {
 			throw new NullPointerException("Ciphertext must not be null");
+		} else if (ciphertextFragment.length % currentReadState.getRecordIvLength() != 0) {
+			throw new GeneralSecurityException("Ciphertext doesn't fit block size!");
 		} else if (ciphertextFragment.length < currentReadState.getRecordIvLength() + currentReadState.getMacLength()
 				+ 1) {
 			throw new GeneralSecurityException("Ciphertext too short!");
@@ -596,15 +598,22 @@ public class Record {
 				new IvParameterSpec(iv));
 		byte[] plaintext = blockCipher.doFinal(reader.readBytesLeft());
 		// last byte contains padding length
-		int paddingLength = plaintext[plaintext.length - 1];
+		int paddingLength = plaintext[plaintext.length - 1] & 0xff;
 		int fragmentLength = plaintext.length
 				- 1 // paddingLength byte
 				- paddingLength
 				- currentReadState.getMacLength();
-
-		reader = new DatagramReader(plaintext);
-		byte[] content = reader.readBytes(fragmentLength);
-		byte[] macFromMessage = reader.readBytes(currentReadState.getCipherSuite().getMacLength());
+		if (fragmentLength < 0) {
+			throw new InvalidMacException();
+		}
+		for (int index = fragmentLength + currentReadState.getMacLength(); index < plaintext.length; ++index) {
+			if (plaintext[index] != (byte) paddingLength) {
+				throw new InvalidMacException();
+			}
+		}
+		byte[] content = Arrays.copyOf(plaintext, fragmentLength);
+		byte[] macFromMessage = Arrays.copyOfRange(plaintext, fragmentLength,
+				fragmentLength + currentReadState.getCipherSuite().getMacLength());
 		byte[] mac = getBlockCipherMac(currentReadState, content);
 		if (Arrays.equals(macFromMessage, mac)) {
 			return content;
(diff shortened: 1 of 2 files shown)


boaks:
New approach:
Decouple used compare and MAC calculations from padding length.
Always call padding compare twice, with 256 compares in sum.
Also calculate extra MAC blocks to have a fixed sum of blocks independent from the padding length.

sophokles73:
@boaks if you want me to review this, I'll need an introduction to the ideas behind it first :-)

boaks:
[page 16-17](http://www.isg.rhul.ac.uk/tls/TLStiming.pdf)

But that specification seem to be based on 13 bytes application data (step 5 on page 17), which is not valid for TLS12_CID (which includes the cid in the additional data). And using always the 256 bytes for "padding comparison" even if the "plain text/plen" is smaller than 256 (step 3 on page 16), seems for me currently also not be the right choice.

If "very similar processing times" should be achieved with that, I think we need also some test to verify that. But I have doubts, that a shared "jenkins" will be able to run such tests. With that, I'm got aware, that it will take more time to complete this work. 
And so for now, I switched back to other tasks (as PR #1002) or my "scandium clean up work" (issue #905),

boaks:
I personally prefer, to just merge the 1. commit and leave the other changes unmerged.
The CBC attacks assumes, that a series of records is also processed as series, and the last record is answered by a response.
With scandium 2.0 the records of different associations are processed in parallel, so the series will not be kept.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/InvalidMacException.java
```diff
@@ -18,25 +18,46 @@
...
+
+	/**
+	 * Create invalid MAC exception without actual MAC values.
+	 * 
+	 * @param msg message
+	 */
+	public InvalidMacException(String msg) {
+		super(msg);
+		this.expected = Bytes.EMPTY;
+		this.actual = Bytes.EMPTY;
```
sbernard31: You don't use this constructor is it intended ? (maybe you planned to add a more specific message for padding error above ?)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -596,15 +598,22 @@ private Record(ProtocolVersion version, int epoch, long sequenceNumber, long rec
 				new IvParameterSpec(iv));
 		byte[] plaintext = blockCipher.doFinal(reader.readBytesLeft());
 		// last byte contains padding length
-		int paddingLength = plaintext[plaintext.length - 1];
+		int paddingLength = plaintext[plaintext.length - 1] & 0xff;
 		int fragmentLength = plaintext.length
```
sbernard31: Just a detail but this is no more the fragment length ? (but I can find a better name)

Maybe we can keep :
```java
int fragmentLength = plaintext.length
				- 1 // paddingLength byte
				- paddingLength
				- currentReadState.getMacLength();
and check if 
``` java
if (fragmentLength < 0) {
	throw new InvalidMacException();
}
```

boaks:
That's from the description of the attack:

"In DTLS, there are no error messages, but the techniques of [1] can be applied to solve this problem. There, the authors send a packet containing a ciphertext C closely followed by a DTLS message, with the latter always provoking a response message."

So, for at least the "multi-threaded" scandium 2.0, I can't see, that this attack makes sense, because the series of records is processed in parallel by multiple threads.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks: If we agree to go for the  1 commit only, yes.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/InvalidMacException.java
boaks: It's used by

```
public InvalidMacException() {
	this(DEFAULT_MESSAGE);
}
```

(even if that may be replaced.)

sbernard31:
My general feeling : I understand why you are perplex about the timing attack (in multithread/cluster environement)  but I'm not so confident about my skills in security, so it's hard to me to bypass security expert recommendation.

Some concerns : 
1. scandium 2.0 is multi-threaded but as all traffic from the same peer is executed sequentially, so maybe it is possible to use the tricks described above ?
2. in cluster environment, I don't know if everybody to this, but as scandium does not have shared state for DTLS connection, we generally try to stick 1 client with 1 cluster instance. So the cluster in this case will not bring so much variability ?

sbernard31:
(@ic0ns, @jurajsomorovsky just in case, you are interested by this :slightly_smiling_face:)

boaks:
"same peer"
So your "on-path-attacker" sees a sequence of records, manipulates some and some not, and waits for the responses? And hope, that no other ongoing traffic influences the times in a "non-filterable" way? In my opinion, very optimistic.

So I still think:
Inform the user about the thread and offer alternatives.
But don't "overstate" the thread by only mention a crypto detail.

PR #1025, opened 2019-08-01
Introduce pre-operator for onContextEstablished.

boaks:
Address comments on issue #974


boaks:
PR #1027 requires this.

PR #1027, opened 2019-08-07
OSCORE: Implements setting OSCORE information in endpoint contexts

rikard-sics:
This pull request creates functionality that adds strings with information about the OSCORE context used to the source or destination endpoint context for a CoAP message. These strings will be set in the endpoint context when communicating using OSCORE.

In certain situations an application may need to know OSCORE-related information about a CoAP message. For instance it may need to know which OSCORE Sender ID a message was transmitted from. The MapBasedEndpointContext contains a map of strings that can be used to add arbitrary information. This pull request makes it so that OSCORE sets some relevant strings in endpoint contexts when OSCORE is used. These strings can then be used by the application to retrieve information related to OSCORE from the endpoint context of a message.

See the new JUnit tests in this pull request for a demonstration of how this functionality can be used. Note that outgoing requests are handled in a special way with the new EndpointContextOperator functionality.

This pull request can help resolve the following issue:
https://github.com/eclipse/californium/issues/974
The previous approach of setting the PeerIdentity with a Principal for OSCORE (see https://github.com/eclipse/californium/pull/1002) was changed after discussion in that thread.

Other small changes in this pull request:
- Various formatting and comment fixes
- Moved flag indicating if responses will contain a Partial IV to OSCoreCtx
- Set content format for error messages to text/plain
- Removed some code related to the old singleton from HashMapCtxDB


rikard-sics:
Some further thoughts.

The PR turned out fairly big around 900 lines but a lot is in OSCoreEndpointContextInfo & EndpointContextInfoTest.

If adding these strings for every message is performance intensive I can add a flag to the OSCORE context that must be set for the strings to be added.

The functionality in this PR may not be sufficient for applications that need custom strings. I believe Ludwig uses something like the context ID Base64 encoded concatenated with the sender ID as a string to identify an OSCORE sender in ACE.
I could expand the functionality to add the possibility for applications to set a custom string in the OSCORE context that they would like to be included in any endpoint contexts created by communication using that OSCORE context.
Or even allow them to set an object (implementing an interface) in the OSCORE context with a method that will be called to generate the custom string. Then this string can be added with key "OSCORE_APPLICATION_ID" or similar.
If this is needed I could do that in a future PR. Or the functionality in the current PR may be sufficient if the applications can perform some conversion based on the existing strings and generate the specific combination of OSCORE information they need themselves.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
```diff
@@ -27,6 +27,7 @@
 
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
+import javax.xml.bind.DatatypeConverter;
```
boaks: [requires "--add-modules java.xml.bind" for newer jvm](https://github.com/http-kit/http-kit/issues/356)

Please consider to replace that with 
[StringUtil](https://github.com/eclipse/californium/blob/master/element-connector/src/main/java/org/eclipse/californium/elements/util/StringUtil.java)

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
```diff
@@ -0,0 +1,207 @@
...
+     * Adds strings with information about the OSCORE context used to this
+     * destination endpoint context for outgoing messages.
+     *
+     * @param oscoreCtx the OSCORE context used for this message
+     * @param message the message to set information in the destination endpoint context for
+     */
+    private static void setInfoOutgoing(OSCoreCtx oscoreCtx, Message message) {
+
+        //Do nothing if this is not an instance of MapBasedEndpointContext (must cast to it)
+        if (message.getDestinationContext() instanceof MapBasedEndpointContext == false) {
```
boaks: Please, don't introduce a dependency on `MapBasedEndpointContext` instances. 
That dependency is not required!

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
```diff
@@ -0,0 +1,207 @@
...
+            return;
+        }
+
+        //Create new MapBasedEndpointContext for destination endpoint context with string values for OSCORE added
+        MapBasedEndpointContext endpointContext = (MapBasedEndpointContext) message.getDestinationContext();
+
+        MapBasedEndpointContext newEndpointContext = setInfo(oscoreCtx, endpointContext);
```
boaks: The rest of californium didn't alter the destination endpoint context!
The idea is, to keep the endpoint context provided by the application and report the context in the callback.
The Response may also carry the "extended endpoint context.

So, why should the Request be changed?

boaks: Depending on the reasons to change the destination context, consider to use:

```
EndpointContext newEndpointContext = setInfo(oscoreCtx, message.getDestinationContext());
message.setDestinationContext(newEndpointContext);
```

That should do it. No cast nor "instanceof" check.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
```diff
@@ -0,0 +1,207 @@
...
+        if (message.getSourceContext() instanceof MapBasedEndpointContext == false) {
+            return;
+        }
+
+        //Create new MapBasedEndpointContext for source endpoint context with string values for OSCORE added
+        MapBasedEndpointContext endpointContext = (MapBasedEndpointContext) message.getSourceContext();
+
+        MapBasedEndpointContext newEndpointContext = setInfo(oscoreCtx, endpointContext);
+
+        message.setSourceContext(newEndpointContext);
```
boaks: Consider to use:

```
EndpointContext newEndpointContext = setInfo(oscoreCtx, message.getSourceContext());
message.setSourceContext(newEndpointContext);
```

That should do it. No cast nor "instanceof" check.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
```diff
@@ -0,0 +1,207 @@
...
+     * @param endpointContext the original endpoint context to set information based on
+     *
+     * @return the new endpoint context
+     */
+    private static MapBasedEndpointContext setInfo(OSCoreCtx oscoreCtx, EndpointContext endpointContext) {
+
+        //Create new MapBasedEndpointContext for this endpoint context with string values for OSCORE added
+        MapBasedEndpointContext newEndpointContext = MapBasedEndpointContext.addEntries(
+                endpointContext,
+                OSCORE, Boolean.TRUE.toString(),
```
boaks: Consider to remove that "OSCORE, Boolean.TRUE.toString()" entry.
E.g. using `get(OSCORE_URI) != null` will do the same, or?

cf-oscore/src/test/java/org/eclipse/californium/oscore/EndpointContextInfoTest.java
```diff
@@ -0,0 +1,406 @@
...
+	 */
+	@Test
+	public void testEndpointContextInfo() throws Exception {
+	    createSimpleServer();
+
+	    //Set up OSCORE context information for request (client)
+	    byte[] sidClient = new byte[] { 0x77, 0x66, 0x55, 0x44 };
+	    byte[] ridClient = new byte[] { 0x01, 0x02, 0x03, 0x04 };
+	    OSCoreCtx ctx = new OSCoreCtx(master_secret, true, alg, sidClient, ridClient, kdf, 32, null, context_id);
+	    String serverUri = "coap://" + serverEndpoint.getAddress().getAddress().getHostAddress() + ":" + serverEndpoint.getAddress().getPort();
```
boaks: Did you check, if 
`serverEndpoint.getUri().toASCIIString()`
works for you?

cf-oscore/src/test/java/org/eclipse/californium/oscore/EndpointContextInfoTest.java
```diff
@@ -0,0 +1,406 @@
...
+	    request.send();
+	    System.out.println("client sent request");
+
+	    //Check original request destination context after transmission
+	    //This is a special case as it should be handled using a callback,
+	    //see testEndpointContextInfoSendingRequest
+	    assertNull(request.getSourceContext());
+	    EndpointContext requestDestinationContext = request.getDestinationContext();
+	    System.out.println("Client: Request destination context type: " + requestDestinationContext.getClass());
+	    assertEquals(AddressEndpointContext.class, requestDestinationContext.getClass());
```
boaks: If you consider the implementation class of an `EndpointContext` to be significant, it will get much more complexer, than it was intended.

cf-oscore/src/test/java/org/eclipse/californium/oscore/EndpointContextInfoTest.java
```diff
@@ -0,0 +1,406 @@
...
+	    serverEndpoint = builder.build();
+	    server = new CoapServer();
+	    server.addEndpoint(serverEndpoint);
+	    server.setMessageDeliverer(new MessageDeliverer() {
+	        @Override
+	        public void deliverRequest(Exchange exchange) {
+	            System.out.println("server received request");
+
+	            //Check request source context after reception
+	            assertNull(exchange.getRequest().getDestinationContext());
```
boaks: Just to mention:
"assert" must be executed from the "main testing thread", or, at least the exception thrown by that assert must be forwarded and (re-)thrown by the "main testing thread".
Calling "assert" on a callback usually doesn't work.
For the `OSCoreObserveTest`I fixed that by introducing the `CountingCoapHandler`.
For the `MessageDeliverer` it is required to forward the exception also.

cf-oscore/src/test/java/org/eclipse/californium/oscore/EndpointContextInfoTest.java
boaks: And just to mention:
If you modify the destination context in the coap-stack, that's execute by an other thread and the change may not be visible after starting the sending by `request.send();`. 
That "race condition" is one reason, why currently we don't change the destination context.

boaks:
Just to mention:
If you want, it's also possible to prepare two commits, one with the "clean up and formatting stuff" and one with the functional changes. If so, we can keep the two commits when merging.
But it's just a proposal, if you want to reduce the lines of relevant changes.
Sometimes a "clean up PR" is also easier, because it could be merged "very fast".

In general the PR looks good. Changing the "destination context" is the only major topic, I wold like to understand, before we should go for it.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
rikard-sics: True, that should provide the same functionality.

rikard-sics:
Thanks for the feedback. Perhaps I had misunderstood some things regarding setting of the destination endpoint context. I will start going through the feedback and amend the code. I will come back with further commits and comments.

> Just to mention:
> If you want, it's also possible to prepare two commits, one with the "clean up and formatting stuff" and one with the functional changes. If so, we can keep the two commits when merging.
> But it's just a proposal, if you want to reduce the lines of relevant changes.
> Sometimes a "clean up PR" is also easier, because it could be merged "very fast".

Yes, you are right. It could have been better to split things like that. I had some changes from earlier pending and found some things while working on the code now that I decided to change right away. Although I think the cleanup lines did not turn out to be that many in the end. For the future I can make sure to isolate such "cleanup" changes to their own separate PRs.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
rikard-sics: Yes, that worked just the same. Fixed in commit 81c1fdc07c6c06cf58e8bddb3e05be99029e1bf7

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
rikard-sics: Yes, as you wrote checking the existence of one of the other keys should provide the same functionality. I removed the OSCORE key in commit cbb3f2ed80dd3452f2b912d5cb1835cfc50120d4

cf-oscore/src/test/java/org/eclipse/californium/oscore/EndpointContextInfoTest.java
rikard-sics: Yes, thanks for the tip. Changed in commit 104bb408c84bd45b35980d0454472a833ec4dff5

cf-oscore/src/test/java/org/eclipse/californium/oscore/EndpointContextInfoTest.java
rikard-sics: > If you consider the implementation class of an `EndpointContext` to be significant, it will get much more complexer, than it was intended.

I have removed any checks for the explicit class type from the EndpointContextInfoTest. Done in commit 5bbb68aab3d52f095505eb3990090edfcfac71c5

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
rikard-sics: Yes, thanks for the suggestion. That did it with no cast or instanceof. I just had to add a check for null in the beginning of setInfo, otherwise I get a null pointer exception for some JUnit tests at https://github.com/eclipse/californium/blob/11ccb133c0356d97816e2ab1b4ba7d6dc7ece24c/element-connector/src/main/java/org/eclipse/californium/elements/MapBasedEndpointContext.java#L211

Done in commit d0c4594169ec6a56545192005ef395b2f69cf501

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
rikard-sics: > Depending on the reasons to change the destination context, consider to use:
> 
> ```
> EndpointContext newEndpointContext = setInfo(oscoreCtx, message.getDestinationContext());
> message.setDestinationContext(newEndpointContext);
> ```
> 
> That should do it. No cast nor "instanceof" check.

Update accordingly in commit d0c4594169ec6a56545192005ef395b2f69cf501

But as you mention it is better if I do not change the destination context so I leave this point open.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
rikard-sics: This should be fixed in commit d0c4594169ec6a56545192005ef395b2f69cf501

cf-oscore/src/test/java/org/eclipse/californium/oscore/EndpointContextInfoTest.java
rikard-sics: Yes I will try to solve this in a nicer way. The way I tried to do it now was that if an assert fails  when the server is receiving a request it will not reply to the message sent by the client. So if I add assertEquals(1, 2) there the tests will fail (but with a Client received no response null pointer exception). For the MessageObserver for outgoing requests I tried using the passed and status variables.

rikard-sics:
> In general the PR looks good. Changing the "destination context" is the only major topic, I would like to understand, before we should go for it.

As for changing the destination context, that is not a must for me at all. As I mentioned perhaps I had a wrong or incomplete picture of how endpoint contexts were handled in Californium.

So if I remove the method setInfoOutgoing (that sets the destination contexts) and remove anything that calls it, it will resolve this?
In fact it seems my JUnit tests in EndpointContextInfoTest still work without ever setting the destination contexts with that method.

The way the contexts for outgoing requests are handled now with the operator is okay?

boaks:
> So if I remove the method setInfoOutgoing (that sets the destination contexts) and remove anything that calls it, it will resolve this?

It solves my irritation and potential race-conditions.

> Also, is the way the contexts for outgoing requests are handled now with the operator okay?

Yes

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
```diff
@@ -0,0 +1,193 @@
...
+     *
+     * @param oscoreCtx the OSCORE context used
+     * @param endpointContext the original endpoint context to set information based on
+     *
+     * @return the new endpoint context
+     */
+    private static MapBasedEndpointContext setInfo(OSCoreCtx oscoreCtx, EndpointContext endpointContext) {
+
+    	//If endpoint context is not set, keep it unset
+    	if (endpointContext == null) {
```
boaks: I'm OK with that check here in `setInfo`.  

Just to mention:
That context is set during sending in `CoapEndpoint`:

```
@Override
public void sendRequest(final Request request) {
	if (!started) {
		request.cancel();
		return;
	}
	// create context, if not already set
	request.prepareDestinationContext();
```

and the connector provides it in the RawData for received Messages.

If the unit tests creates there own Messages and want to use the contexts, they must ensure, that the contexts are set before processing the Messages in the CoapStack..

rikard-sics:
> It solves my irritation and potential race-conditions.

Alright, sounds good. I will proceed accordingly, sorry for any irritation caused.

> Yes

Okay, good to hear.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
rikard-sics: Okay, thanks for the information. I will update the JUnit tests to set an endpoint context properly, since it seems some of them do not do this currently. They are calling methods from the ObjectSecurityLayer like prepareReceive() which will end up calling the method for adding information to the endpoint context, but none has been set which would cause the null pointer exception. If I update the JUnit tests to solve this is it alright to make that into a new PR?

boaks: FMPOV, you have been asked to change this PR so often, that I think we should go with it now. If you want to spend your time into adjust the unit tests, you may do this in a future PR.

boaks:
In some classes (OSCoreEndpointContextInfo or EndpointContextInfoTest) tabs and spaces seems to be used without applying the californium formatter. Usually we use tabs.

boaks:
Running the unit tests I get:

java.lang.NullPointerException: 3. value is null
	at org.eclipse.californium.elements.MapBasedEndpointContext.createMap(MapBasedEndpointContext.java:148)
	at org.eclipse.californium.elements.MapBasedEndpointContext.addEntries(MapBasedEndpointContext.java:210)
	at org.eclipse.californium.oscore.OSCoreEndpointContextInfo.setInfo(OSCoreEndpointContextInfo.java:184)
	at org.eclipse.californium.oscore.OSCoreEndpointContextInfo.setInfoIncoming(OSCoreEndpointContextInfo.java:163)
	at org.eclipse.californium.oscore.OSCoreEndpointContextInfo.receivingRequest(OSCoreEndpointContextInfo.java:116)

boaks:
With some fixes for the "null" attributes, the unit test will succeed.
When that is reached, (and the source formatting may be applied), then please:

VERY, VERY, FIRST, never use "master" or any other "eclipse" repository branch!
To fix that:
Squash your commits of your local master into one commit.
Create a "feature" branch from your local "master" branch with your single commit at the head.
Then checkout your local master branch again as master and reset your local master back to the eclipse master commit.
Fetch the eclipse remote and merge your local master (it will be a fast forward merge) to the eclipse master.
After that checkout your feature branch and rebase it towards your local master (which should be the same as the most current eclipse master). That rebase will report conflicts, please fix them.
Then close this PR (again)
Create a new PR with the new feature branch.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
rikard-sics: > FMPOV, you have been asked to change this PR so often, that I think we should go with it now. If you want to spend your time into adjust the unit tests, you may do this in a future PR.

Yes, if I work on updating the JUnit tests to make sure they set endpoint contexts it will be in a new PR.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreEndpointContextInfo.java
rikard-sics: > The rest of californium didn't alter the destination endpoint context!
> The idea is, to keep the endpoint context provided by the application and report the context in the callback.
> The Response may also carry the "extended endpoint context.
> 
> So, why should the Request be changed?

I have now removed the code that change the destination endpoint context in commit b4dd950c85f774f38aa96f19f57ac9c3e264de2f

cf-oscore/src/test/java/org/eclipse/californium/oscore/EndpointContextInfoTest.java
rikard-sics: > And just to mention:
> If you modify the destination context in the coap-stack, that's execute by an other thread and the change may not be visible after starting the sending by `request.send();`.
> That "race condition" is one reason, why currently we don't change the destination context.

Yes I understand the problem it causes. Changing of destination context has been removed in commit b4dd950c85f774f38aa96f19f57ac9c3e264de2f

rikard-sics:
> In some classes (OSCoreEndpointContextInfo or EndpointContextInfoTest) tabs and spaces seems to be used without applying the californium formatter. Usually we use tabs.

Okay, I will run the Californium formatter on those files. I will try to be more careful about making sure to use it in the future.

> May be you add a "String toHex(byte[] bytes)" to your OSCoreCtx, which checks for null and empty and returns a "" then. And for none-empty byte[] then call StringUtil.byteArray2Hex.
> 
> And the OSCoreCtx.uri may be null.
> Either init it with "", or consider an approach as:

Yes, I will take care of this. Thanks for the suggestions on approaches.

> With some fixes for the "null" attributes, the unit test will succeed.
> When that is reached, (and the source formatting may be applied), then please:
> 
> VERY, VERY, FIRST, never use "master" or any other "eclipse" repository branch!
> To fix that:

Thanks, I see. I will make a new branch as you instruct and create a new PR from it.


One last question. The issue you mentioned with calling of assert from outside the main testing thread. Should I take care of this before I make the new branch or after I have made the new PR? One approach I was considering was doing it like some other tests are doing and calling the methods from ObjectSecurityLayer directly, "simulating" message reception. Then it would be easier to handle the asserts, especially for the server side.

boaks:
> One last question. The issue you mentioned with calling of assert from outside the main testing thread. Should I take care of this before I make the new branch or after I have made the new PR?

Please new PR. this one is large enough :-).
the only thing, I would like to have here, are "successful running unit tests". so the proposed stuff in `setInfo(OSCoreCtx oscoreCtx, EndpointContext endpointContext)` is required, the initialization in `OSCoreCtx` may be implemented. And then format, squash, fix master, rebase and new PR :-). 

> One approach I was considering was doing it like some other tests are doing and calling the methods from ObjectSecurityLayer directly, "simulating" message reception. Then it would be easier to handle the asserts, especially for the server side.

You may have a look at [CountingCoapHandler](https://github.com/eclipse/californium/blob/master/californium-core/src/test/java/org/eclipse/californium/core/test/CountingCoapHandler.java#L61)

so the exceptions from the assert are catched, stored, and forwarded to the main-testing-thread when that is calling some methods. If this PR is "finally merged", I will have a closer look on it and propose more details for a solution.

rikard-sics:
> Please new PR. this one is large enough :-).

True :)

> the only thing, I would like to have here, are "successful running unit tests". so the proposed stuff in `setInfo(OSCoreCtx oscoreCtx, EndpointContext endpointContext)` is required, the initialization in `OSCoreCtx` may be implemented.

I have now added the proposal you gave for additions in setInfo. I also added the initialization and the new method for byte array to string conversion in OSCoreCtx. This was done in commit d3f7d46602ff7400042f1500eeae5c7a6d27e360

BTW. sorry I missed the failing JUnit test. I was running the tests from AllJUnitTests in Eclipse and I realize now that the ContextRederivationTest was not included there.
The issue is that in RequestDecryptor I added the new re-derived context to the DB without setting the URI for the new context.
https://github.com/eclipse/californium/blob/e4b93e0fc06cc9462e4f7934beff713862853e8c/cf-oscore/src/main/java/org/eclipse/californium/oscore/RequestDecryptor.java#L168
I will fix this in a coming PR also.

rikard-sics:
>  And then format,

Formatting of OSCoreEndpointContextInfo and EndpointContextInfoTest has been done in commit 2973dcff9aa2162b4f6b6181f1f99def1082a958

>  squash, fix master, rebase and new PR :-).

Yes, I will soon close this PR and proceed with creating a new one according to the points you listed above.


> You may have a look at [CountingCoapHandler](https://github.com/eclipse/californium/blob/master/californium-core/src/test/java/org/eclipse/californium/core/test/CountingCoapHandler.java#L61)
> 
> so the exceptions from the assert are catched, stored, and forwarded to the main-testing-thread when that is calling some methods. If this PR is "finally merged", I will have a closer look on it and propose more details for a solution.

Yes I recall you mentioned that one earlier also. Thanks for the tip and help in finding a solution.

rikard-sics:
Closing this PR to create a new one as discussed.

PR #1028, opened 2019-08-07
Cleanup handshaker

boaks:
Cleanup handshaker.

Process records only, if they match the current epoch.
Remove duplicate epoch checks and applySession.
Generally apply the record sequence number filter.
Forward deferred records (next epoch) with CCS to the record layer to
apply general processing and decryption with the new session.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -1065,7 +1065,8 @@ public class DTLSConnector implements Connector, RecordLayer {
 	 * @param record received record.
 	 * @param connection connection to process record.
 	 */
-	private void processRecord(Record record, Connection connection) {
+	@Override
+	public void processRecord(Record record, Connection connection) {
 
 		try {
 			// ensure, that connection is still related to record 
@@ -1080,6 +1081,28 @@ public class DTLSConnector implements Connector, RecordLayer {
 			LOGGER.trace("Received DTLS record of type [{}], length: {}, [epoche:{},reqn:{}]", 
 					record.getType(), record.getFragmentLength(), epoch, record.getSequenceNumber());
 
+			DTLSSession session = connection.getSession(epoch);
+			if (session == null) {
+				Handshaker handshaker = connection.getOngoingHandshake();
+				if (handshaker != null && handshaker.getSession().getReadEpoch() == 0 && epoch == 1) {
+					// future records, apply session after handshake finished.
+					handshaker.addRecordsForDeferredProcessing(record);
+				} else {
+					LOGGER.debug("Discarding {} record received from peer [{}] without an active session for epoch {}",
+							record.getType(), record.getPeerAddress(), epoch);
+				}
+				return;
+			}
+
+			// The DTLS 1.2 spec (section 4.1.2.6) advises to do replay detection
+			// before MAC validation based on the record's sequence numbers
+			// see http://tools.ietf.org/html/rfc6347#section-4.1.2.6
+			if (useFilter && (session != null) && !session.isRecordProcessable(record.getEpoch(), record.getSequenceNumber(), useWindowFilter)) {
+				LOGGER.debug("Discarding duplicate {} record received from peer [{}]",
+						record.getType(), record.getPeerAddress());
+				return;
+			}
+
 			boolean useCid = connectionIdGenerator != null && connectionIdGenerator.useConnectionId();
 			if (record.getType() == ContentType.TLS12_CID) {
 				// !useCid already dropped in Record.fromByteArray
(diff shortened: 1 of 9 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1403,93 +1380,42 @@ private void processChangeCipherSpecRecord(Record record, Connection connection)
...
-				// negotiate a new connection
-				// in epoch 0 no crypto params have been established yet, thus
-				// we do not need to set a session
-			} else {
-				LOGGER.debug(
-						"Discarding HANDSHAKE message [epoch={}] from peer [{}] which does not match expected epoch(s) [{}]",
-						record.getEpoch(), record.getPeerAddress());
-				return;
+			if (record.isNewClientHello()) {
+				throw new IllegalArgumentException("new CLIENT_HELLO must be processed by processClientHello!");
```
sbernard31: is this check more or less the same as the one below (`else` part)?
```java
if (record.getEpoch() > 0) {
	send(new AlertMessage(AlertLevel.WARNING, AlertDescription.NO_RENEGOTIATION, record.getPeerAddress()),session);
} else {
	LOGGER.error("Unexpected CLIENT_HELLO {}", record.getPeerAddress(),	new IllegalArgumentException("Unexpected CLIENT_HELLO"));
}
```

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -361,18 +362,20 @@ DTLSMessage getNextMessage() throws GeneralSecurityException, HandshakeException
...
-						getPeerAddress(), epoch, session.getReadEpoch());
-				return null;
-			} else if (epoch == session.getReadEpoch()) {
-				candidate.applySession(session);
+			int recordEpoch = candidate.getEpoch();
+			int sessionEpoch = session.getReadEpoch();
+			if (recordEpoch == sessionEpoch) {
 				DTLSMessage fragment = candidate.getFragment();
+				if (fragment == null) {
+					LOGGER.debug("Cannot process message of type [{}], discarding...", candidate.getType());
```
sbernard31: I'm not sure but  if this  happened, this means there is a bug at scandium side ? 

If yes, should we log ERROR or WARN ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -361,18 +362,20 @@ DTLSMessage getNextMessage() throws GeneralSecurityException, HandshakeException
...
+					return null;
+				}
+				if (fragment.getContentType() != ContentType.HANDSHAKE
+						&& fragment.getContentType() != ContentType.CHANGE_CIPHER_SPEC) {
+					LOGGER.debug("Drop message of type [{}], discarding...", candidate.getType());
```
sbernard31: (same here)
I'm not sure but if this happened, this means there is a bug at scandium side ?

If yes, should we log ERROR or WARN ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -442,58 +443,71 @@ public final void processMessage(Record record) throws HandshakeException {
...
+					doProcessMessage(messageToProcess);
+				}
 
-					// process next expected message (if available yet)
-					messageToProcess = inboundMessageBuffer.getNextMessage();
+				// process next expected message (if available yet)
+				messageToProcess = inboundMessageBuffer.getNextMessage();
+			}
+			session.markRecordAsRead(epoch, record.getSequenceNumber());
+			if (epoch < session.getReadEpoch()) {
```
sbernard31: I believe I often take few second to get some of your if condition, just because I would have written it reversely :sweat_smile: 

E.g. here in my mind, this is something like : "after we process the message, Is the session readEpoch greater than before ?" <=> if (session.getReadEpoch() > epoch)

Never mind, it's clearly not a big deal :)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -442,58 +443,71 @@ public final void processMessage(Record record) throws HandshakeException {
...
+				final SerialExecutor serialExecutor = connection.getExecutor();
+				final List<Record> records = takeDeferredRecords();
+				for (Record deferredRecord : records) {
+					if (serialExecutor != null) {
```
sbernard31: What are the benefit to create a new task ?
maybe to let a chance to handle record from other connection ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1246,69 +1252,51 @@ private void processApplicationDataRecord(final Record record, final Connection
...
-				} catch (HandshakeException | GeneralSecurityException e) {
-					// this means that we could not parse or decrypt the message
-					LOGGER.debug("Discarding APPLICATION_DATA record received from peer [{}]",
-							record.getPeerAddress(), e);
-					discardRecord(record, e);
+			// APPLICATION_DATA can only be processed within the context of
+			// an established, i.e. fully negotiated, session
+			ApplicationMessage message = (ApplicationMessage) record.getFragment();
+			InetSocketAddress newAddress = record.getPeerAddress();
+			if (connectionStore.get(newAddress) == connection) {
```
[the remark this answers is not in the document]
boaks: The check is by address, with cid the previous check was by cid.

"Parallel" Records may have changed the connection store, but will only update the connection's address within the serial execution. The check here helps to keep that address up-to-date.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: Seems so.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: That may be caused by invalid records.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: That's the same solution as in some other locations e.g. `receiveNextDatagramFromNetwork`.

The intention is to create smaller jobs, that the execution service doesn't block a thread for longer, and so, yes, have "chance to handle record from other connection".

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: does it mean we should remove one of them ?

boaks: Done.

boaks:
The comments are addressed.

sbernard31:
LGTM

PR #1029, opened 2019-08-07
Add marker for recommended cipher suites.

boaks:
Mark CBC cipher suites as not recommended.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -804,6 +804,7 @@ public final class DtlsConnectorConfig {
 
 		private DtlsConnectorConfig config;
 		private boolean clientOnly;
+		private boolean recommendedCipherSuitesOnly = true;
 
 		/**
 		 * Creates a new instance for setting configuration options
@@ -875,6 +876,22 @@ public final class DtlsConnectorConfig {
 		 */
 		public Builder setEnableAddressReuse(boolean enable) {
 			config.enableReuseAddress = enable;
+			return this;
+		}
+
+		/**
+		 * Set usage of recommended cipher suites.
+		 * 
+		 * @param recommendedCipherSuitesOnly {@code true} allow only
+		 *            recommended cipher suites, {@code false}, also allow not
+		 *            recommended cipher suites. Default value is {@code true}
+		 * @return this builder for command chaining
+		 */
+		public Builder setRecommendedCipherSuitesOnly(boolean recommendedCipherSuitesOnly) {
+			this.recommendedCipherSuitesOnly = recommendedCipherSuitesOnly;
+			if (recommendedCipherSuitesOnly && config.supportedCipherSuites != null) {
+				verifyRecommendedCipherSuitesOnly(config.supportedCipherSuites);
+			}
 			return this;
 		}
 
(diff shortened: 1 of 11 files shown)


boaks:
After that, we should add more comments/warnings about the CBC implementation.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -2008,14 +2029,14 @@ private void determineCipherSuitesFromConfig() {
 			// try to guess his intentions from properties he has set
 			List<CipherSuite> ciphers = new ArrayList<>();
 			boolean certificates = isConfiguredWithKeyPair() || config.trustCertificateTypes != null;
-
+			boolean recommendedOnly = Boolean.FALSE.equals(recommendedCipherSuitesOnly);
```
sbernard31: I would prefer that we define a default value for `recommendedCipherSuitesOnly` before to call `determineCipherSuitesFromConfig` instead of this test here.

boaks: If the value is initialized, the checks for explicit cipher suites selection will be either not that easy (currently the check is only for ! null), or the verification logic must be changed.

sbernard31: Maybe I don't get your point, but I don't see any issue if we do somthing like this in `build()`: 

```java
if (config.supportedCipherSuites == null || config.supportedCipherSuites.isEmpty()) {
    if (recommendedCipherSuitesOnly == null){
        recommendedCipherSuitesOnly = true;
    }
    determineCipherSuitesFromConfig();
}
```

boaks: OK, I was mislead ...

boaks:
Alternative config logic, just for completeness:

`recommendedCipherSuitesOnly` is also used for the `setSupportedCipherSuites`. that would ensure, that everyone gets aware of the "CBC not recommended", even if the cipher suties are set explicit. 
If CBC is the really required, `setRecommendedCipherSuitesOnly` must be set to `false` before calling `setSupportedCipherSuites`.

sbernard31:
This alternative is maybe better.
This way we can raise a WARNING only if recommendedCipherSuitesOnly = true and CBC is explicitly added ?

boaks:
> This way we can raise a WARNING only if recommendedCipherSuitesOnly = true and CBC is explicitly added ?

```
throw new IllegalStateException("Not recommended cipher suites " + message
						+ " used! (Requires to set recommendedCipherSuitesOnly to false.)");
```

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -804,6 +804,7 @@ protected Object clone() {
 
 		private DtlsConnectorConfig config;
 		private boolean clientOnly;
+		private boolean recommendedCipherSuitesOnly = true;
```
sbernard31: Why those 2 attributes (clientOnly, recommendedCipherSuitesOnly) are builder attributes ? 

I suspect this could be an issue for use case with  builder.getIncompleteConfig(); as we can not know if this boolean was modified or not. 
(E.g : [LeshanServerBuilder](https://github.com/eclipse/leshan/blob/leshan-1.0.0-M11/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/LeshanServerBuilder.java#L380)

boaks: > Just to let you know for leshan we probably don't need it. so it's just hypothetical.

Then we will see, if someone else is using the "incompleteConfig" :-).
(If you're interested, even if leshan don't need it, feel free to correct it.)

sbernard31: It sounds good. :)

boaks: Ok, so merge now?

sbernard31: Yep !

PR #1032, opened 2019-08-12
Add general thread local crypto functions.

boaks:


sbernard31:
@boaks is this a kind of performance improvement or did you try to fix an issue or maybe this just about cleaning code ?

boaks:
"Cleaning Code".
I'm not sure, if  performance improvement could be measured, thought currently the handshakes itself are not included in the benchmarks.

Issue #1034, opened 2019-08-14
DTLS retransmission question


sbernard31:
This a scenario inspired from a real world capture:

```
Time      Source      Destination      Info 
2.127170  client      server      Client Hello
2.287418  server      client      Hello Verify Request
2.315333  client      server      Client Hello
2.478780  server      client      Server Hello, Server Key Exchange, Server Hello Done
-----------------------------------------
No answer from client after 1s, we retransmit.
------------------------------------------
3.479602  server      client      Server Hello, Server Key Exchange, Server Hello Done
-----------------------------------------
Still no answer from client after 2s, we retransmit again.
-----------------------------------------
5.479355  server      client      Server Hello, Server Key Exchange, Server Hello Done
-----------------------------------------
The client recieved all messages but it sounds it was busy or not able to answer in time.
Now it unqueue message and answer one by one to each retransmission in a very short delay.
-----------------------------------------
5.673280  client      server      Client Key Exchange, Change Cipher Spec, Encrypted Handshake Message
5.683168  client      server      Client Key Exchange, Change Cipher Spec, Encrypted Handshake Message
5.705030  client      server      Client Key Exchange, Change Cipher Spec, Encrypted Handshake Message
-----------------------------------------
At server side we also recieve all messages and answer to each retransmission.
-----------------------------------------
5.841709  server      client      Change Cipher Spec, Encrypted Handshake Message
5.841823  server      client      Change Cipher Spec, Encrypted Handshake Message
5.863473  server      client      Change Cipher Spec, Encrypted Handshake Message
```

My first impression was "this is a pretty strange capture why all messages are sent 3 times ?".
After analyzing it, it seems to me there is nothing which could help to prevent this in the DTLS RFC. So I would say this is the "expected" behavior.

And I can not see any good way to avoid this...

WDYT ?

boaks:
Usually a peer should ignore the "retransmissions", if the handshake sequence number is the same as already received. So, do you have a wireshark capture?

boaks:
[RFC6347, 4.2.2](https://tools.ietf.org/html/rfc6347#page-18)

> If the sequence number is less than next_receive_seq, the message MUST be discarded. 

[RFC6347, 4.2.4](https://tools.ietf.org/html/rfc6347#page-20)

> 2. The implementation reads a retransmitted flight from the peer: the
>       implementation transitions to the SENDING state, where it
>       retransmits the flight, resets the retransmit timer, and returns
[...]

PR #1035, opened 2019-08-15
More cleanup

boaks:
More handshake cleanup.
Small fixes for connection.
Check server_hello parameter.
 
### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
@@ -519,8 +519,12 @@ public final class Connection {
 		if (peerAddress != null) {
 			builder.append(", ").append(peerAddress);
 			if (getOngoingHandshake() != null) {
-				String id = getOngoingHandshake().getSession().getSessionIdentifier().getAsString().substring(0,  8);
-				builder.append(", ongoing handshake ").append(id);
+				builder.append(", ongoing handshake ");
+				SessionId id = getOngoingHandshake().getSession().getSessionIdentifier();
+				if (id != null) {
+					// during handshake this may by not already set
+					builder.append(id.getAsString().substring(0,  8));
+				}
 			}
 			if (isResumptionRequired()) {
 				builder.append(", resumption required");
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
@@ -266,7 +266,6 @@ public class ResumingClientHandshaker extends ClientHandshaker {
 		handshakeHash = mdWithServerFinish.digest();
 		Finished finished = new Finished(session.getCipherSuite().getThreadLocalPseudoRandomFunctionMac(), session.getMasterSecret(), isClient, handshakeHash, message.getPeer());
 		wrapMessage(flight, finished);
-		state = HandshakeType.FINISHED.getCode();
 
 		flight.setRetransmissionNeeded(false);
 		// store, if we need to retransmit this flight, see
@@ -290,7 +289,6 @@ public class ResumingClientHandshaker extends ClientHandshaker {
 		addMaxFragmentLength(message);
 		addServerNameIndication(message);
 
-		state = message.getMessageType().getCode();
 		clientHello = message;
 
 		flightNumber = 1;
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
@@ -159,6 +159,7 @@ public class ResumingClientHandshaker extends ClientHandshaker {
 					// Server refuse to resume the session, go for a full handshake
 					fullHandshake  = true;
 					super.receivedServerHello(serverHello);
+					incrementNextReceiveSeq();
 					return;
 				} else if (!serverHello.getCompressionMethod().equals(session.getCompressionMethod())) {
 					throw new HandshakeException(
--- a/scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
+++ b/scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
@@ -55,7 +55,7 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.security.GeneralSecurityException;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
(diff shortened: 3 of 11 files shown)


boaks:
@sbernard31 

May be you can have a fast try with this and the client of issue #1034

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
```diff
@@ -527,7 +530,9 @@ public void testAcceptClientHelloAfterIncompleteHandshake() throws Exception {
 		client = new DTLSConnector(clientConfig, clientConnectionStore);
 
 		// THEN assert that the handshake succeeds and a session is established
-		givenAnEstablishedSession();
+		// sometime the retransmission of the previous incomplete handshake
+		// fails the new handshake. Therefore a retry is used.
+		givenAnEstablishedSessionWithRetry();
```
sbernard31: > sometime the retransmission of the previous incomplete handshake fails the new handshake.

I'm curious : how does it fail ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
```diff
@@ -159,6 +159,7 @@ protected void doProcessMessage(DTLSMessage message) throws HandshakeException,
 					// Server refuse to resume the session, go for a full handshake
 					fullHandshake  = true;
 					super.receivedServerHello(serverHello);
+					incrementNextReceiveSeq();
```
sbernard31: Good catch !
Any reason to not just remove the `return;` ?

scandium-core/src/test/java/org/eclipse/californium/scandium/dtls/AdversaryClientHandshaker.java
```diff
@@ -0,0 +1,239 @@
...
+ *                                                    trustStore := null, disable x.509
+ *                                                    trustStore := [], enable x.509, trust all
+ *    Achim Kraus (Bosch Software Innovations GmbH) - fix usage of literal address
+ *                                                    with enabled sni
+ *    Vikram (University of Rostock) - added ECDHE_PSK mode
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add handshake parameter available to
+ *                                                    process reordered handshake messages
+ *    Achim Kraus (Bosch Software Innovations GmbH) - add dtls flight number
+ *    Achim Kraus (Bosch Software Innovations GmbH) - redesign DTLSFlight and RecordLayer
+ ******************************************************************************/
```
sbernard31: is it intended to keep all this contributors in the header of this new class ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -1103,7 +1105,15 @@ private DTLSMessage handshakeMessageFromByteArray(byte[] decryptedMessage) throw
...
-		return HandshakeMessage.fromByteArray(decryptedMessage, parameter, getPeerAddress());
+		HandshakeMessage handshakeMessage = HandshakeMessage.fromByteArray(decryptedMessage, parameter, getPeerAddress());
+		if (handshakeMessage.getMessageType() == HandshakeType.FINISHED && epoch == 0) {
+			LOGGER.debug("FINISH with epoch 0 from peer [{}]!",
+					peerAddress);
+			AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.UNEXPECTED_MESSAGE,
+					peerAddress);
+			throw new HandshakeException("FINISH with epoch 0!", alert);
+		}
+		return handshakeMessage;
```
sbernard31: It sounds like Handshaker logic so I would move it in an `Handshaker` :thinking:

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
```diff
@@ -519,8 +519,12 @@ public String toString() {
 		if (peerAddress != null) {
 			builder.append(", ").append(peerAddress);
 			if (getOngoingHandshake() != null) {
-				String id = getOngoingHandshake().getSession().getSessionIdentifier().getAsString().substring(0,  8);
-				builder.append(", ongoing handshake ").append(id);
+				builder.append(", ongoing handshake ");
+				SessionId id = getOngoingHandshake().getSession().getSessionIdentifier();
+				if (id != null) {
+					// during handshake this may by not already set
+					builder.append(id.getAsString().substring(0,  8));
```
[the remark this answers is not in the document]
boaks: To have shorter logging messages.

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
boaks: First handshake is in progress. Potential, the last pending flight is retransmitted.
Second handshake sends the 2. client_hello. If now the server_hello from the first is received and processed, the handshake fails. There is a very small window, when the last pending flight is resend and the new client_hello is received on the server side. You may use the repeated test runner to see, that without the retry, the test fails rarely.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
boaks: With the return, it's clear the the control flow of the full handshake is used. 
Even if the current implementation would not differ.

scandium-core/src/test/java/org/eclipse/californium/scandium/dtls/AdversaryClientHandshaker.java
boaks: Uups, copy&paste

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
boaks: My idea was, that a FINISH epoch 0 is a invalid record. But I can move it into the handshaker.

boaks:
Remove the wrong header license.
Move FINISH epoch 0 to Handshaker

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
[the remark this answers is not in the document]
sbernard31: Personally I would keep the complete sessionId but it's OK like this too.

scandium-core/src/test/java/org/eclipse/californium/scandium/DTLSConnectorTest.java
sbernard31: Ok I see.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
sbernard31: Ok but here we just want to increment and log we receive a message and this is was is done later :)

sbernard31:
LGTM.

By the way, I don't have any devices available so I can not really test it. But I would let you know if I find a way to get feedback about that.

PR #1037, opened 2019-08-16
Cbc hash compression

boaks:
Add message digest updates to equalize calculation times.
Reduce time side channels.
Add "manual" unit test with benchmarks to demonstrate the improvements of equal calculation times. 

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CbcBlockCipher.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CbcBlockCipher.java
@@ -13,11 +13,14 @@
  * Contributors:
  *    Bosch Software Innovations GmbH - initial API and implementation
  *                                      split from Record
+ *    Achim Kraus (Bosch Software Innovations GmbH) - reduce timing side channel
  *******************************************************************************/
 package org.eclipse.californium.scandium.dtls.cipher;
 
 import java.security.GeneralSecurityException;
 import java.security.InvalidKeyException;
+import java.security.MessageDigest;
+import java.security.SecureRandom;
 import java.util.Arrays;
 
 import javax.crypto.Cipher;
@@ -25,7 +28,7 @@ import javax.crypto.Mac;
 import javax.crypto.SecretKey;
 import javax.crypto.spec.IvParameterSpec;
 
-import org.eclipse.californium.elements.util.DatagramReader;
+import org.eclipse.californium.elements.util.Bytes;
 import org.eclipse.californium.elements.util.DatagramWriter;
 import org.eclipse.californium.scandium.dtls.Record;
 
@@ -33,6 +36,9 @@ import org.eclipse.californium.scandium.dtls.Record;
  * A cbc block cipher.
  */
 public class CbcBlockCipher {
+
+	// byte used to fill-up plaintext for extra message digest compression
+	private static final byte[] FILLUP = Bytes.createBytes(new SecureRandom(), 256);
 
 	// Static methods /////////////////////////////////////////////////
 
@@ -70,46 +76,68 @@ public class CbcBlockCipher {
 	 */
 	public static byte[] decrypt(CipherSuite suite, SecretKey key, SecretKey macKey, byte[] additionalData,
 			byte[] ciphertext) throws GeneralSecurityException {
-
 		/*
-		 * See http://tools.ietf.org/html/rfc5246#section-6.2.3.2 for
-		 * explanation
+		 * See http://tools.ietf.org/html/rfc5246#section-6.2.3.2 for explanation
 		 */
-		DatagramReader reader = new DatagramReader(ciphertext);
-		byte[] iv = reader.readBytes(suite.getRecordIvLength());
+		// extend/oversize the plaintext for MAC compensation and 256 padding checks 
+		byte[] plaintextOversized = new byte[ciphertext.length + Math.max(suite.getMacMessageBlockLength(), 256)];
+		int ivlength = suite.getRecordIvLength();
 		Cipher blockCipher = suite.getThreadLocalCipher();
-		blockCipher.init(Cipher.DECRYPT_MODE,
-				key,
-				new IvParameterSpec(iv));
-		byte[] plaintext = blockCipher.doFinal(reader.readBytesLeft());
-		// last byte contains padding length
+		blockCipher.init(Cipher.DECRYPT_MODE, key, new IvParameterSpec(ciphertext, 0, ivlength));
+		int plaintextLength = blockCipher.doFinal(ciphertext, ivlength, ciphertext.length - ivlength,
+				plaintextOversized);
+		// fill up plaintext for MAC compensation
+		System.arraycopy(FILLUP, 0, plaintextOversized, plaintextLength, suite.getMacMessageBlockLength());
 		int macLength = suite.getMacLength();
-		int paddingLength = plaintext[plaintext.length - 1] & 0xff;
-		int fragmentLength = plaintext.length
-				- 1 // paddingLength byte
-				- paddingLength
-				- macLength;
-		if (fragmentLength < 0) {
+		// last byte contains padding length
+		int paddingLength = plaintextOversized[plaintextLength - 1] & 0xff;
+		// -1 := padding length byte
+		int fullLength = plaintextLength - macLength - 1;
+		int leftLength = fullLength - paddingLength;
+		int fragmentLength;
+		if (leftLength < 0) {
+			// padding length byte wrong
+			fragmentLength = fullLength;
+			paddingLength = 0;
+		} else {
+			fragmentLength = leftLength;
+		}
+		if (!checkPadding(paddingLength, plaintextOversized, fragmentLength + macLength)) {
+			fragmentLength = fullLength;
+			paddingLength = 0;
+		}
+		// adjust fragment length
+		int additionalIndex = additionalData.length - (Record.LENGTH_BITS / 8);
+		additionalData[additionalIndex] = (byte) ((fragmentLength >> 8) & 0xff);
+		additionalData[additionalIndex+1] = (byte)(fragmentLength & 0xff);
+
+		MessageDigest md = suite.getThreadLocalMacMessageDigest();
+		md.reset();
+		byte[] mac = getBlockCipherMac(suite.getThreadLocalMac(), macKey, additionalData, plaintextOversized,
+				fragmentLength);
+
+		// estimate additional MAC Hash compressions to decouple calculation
+		// times from padding. The MAC Hash compressions are done in blocks,
+		// appending the message length as extra data.
+		int macMessageLengthBytes = suite.getMacMessageLengthBytes();
+		int macMessageBlockLength = suite.getMacMessageBlockLength();
+		// add all bytes passed to MAC
+		int macBytes = additionalData.length + fragmentLength + macMessageLengthBytes;
+		// MAC blocks for all bytes including padding
+		int macBlocks1 = (macBytes + paddingLength) / macMessageBlockLength;
+		// MAC blocks for all bytes without padding
+		int macBlocks2 = macBytes / macMessageBlockLength;
+		int blocks = (macBlocks1 - macBlocks2);
+		// calculate extra compression to compensate timing differences
+		// caused by different padding
+		// extra byte, to ensure, that the final compression is triggered
+		md.update(plaintextOversized, fragmentLength, (blocks * macMessageBlockLength) + 1);
+		md.reset();
+		byte[] macFromMessage = Arrays.copyOfRange(plaintextOversized, fragmentLength, fragmentLength + macLength);
+		if (!MessageDigest.isEqual(macFromMessage, mac)) {
 			throw new InvalidMacException();
 		}
-		for (int index = fragmentLength + macLength; index < plaintext.length; ++index) {
-			if (plaintext[index] != (byte) paddingLength) {
-				throw new InvalidMacException();
-			}
-		}
-		byte[] content = Arrays.copyOf(plaintext, fragmentLength);
-		byte[] macFromMessage = Arrays.copyOfRange(plaintext, fragmentLength, fragmentLength + macLength);
-
-		// adjust fragment length
-		int additionalIndex = additionalData.length - (Record.LENGTH_BITS / Byte.SIZE);
-		additionalData[additionalIndex] = (byte) ((fragmentLength >> Byte.SIZE) & 0xff);
-		additionalData[additionalIndex+1] = (byte)(fragmentLength & 0xff);
-		byte[] mac = getBlockCipherMac(suite.getThreadLocalMac(), macKey, additionalData, plaintext, fragmentLength);
-		if (Arrays.equals(macFromMessage, mac)) {
-			return content;
-		} else {
-			throw new InvalidMacException(mac, macFromMessage);
-		}
+		return Arrays.copyOf(plaintextOversized, fragmentLength);
 	}
 
 	/**
@@ -137,25 +165,25 @@ public class CbcBlockCipher {
 	 * @param key encryption key
 	 * @param macKey mac key
 	 * @param additionalData additional data
-	 * @param plaintext message to encrypt
+	 * @param payload message to encrypt
 	 * @return encrypted message including initial vector
 	 * @throws GeneralSecurityException if the plaintext could not be encrypted
 	 */
 	public static byte[] encrypt(CipherSuite suite, SecretKey key, SecretKey macKey, byte[] additionalData,
-			byte[] plaintext) throws GeneralSecurityException {
+			byte[] payload) throws GeneralSecurityException {
 		/*
 		 * See http://tools.ietf.org/html/rfc5246#section-6.2.3.2 for
 		 * explanation
 		 */
-		DatagramWriter message = new DatagramWriter();
-		message.writeBytes(plaintext);
+		DatagramWriter plainMessage = new DatagramWriter();
+		plainMessage.writeBytes(payload);
 
 		// add MAC
-		message.writeBytes(
-				getBlockCipherMac(suite.getThreadLocalMac(), macKey, additionalData, plaintext, plaintext.length));
+		plainMessage.writeBytes(
+				getBlockCipherMac(suite.getThreadLocalMac(), macKey, additionalData, payload, payload.length));
 
 		// determine padding length
-		int ciphertextLength = plaintext.length + suite.getMacLength() + 1;
+		int ciphertextLength = payload.length + suite.getMacLength() + 1;
 		int blocksize = suite.getRecordIvLength();
 		int lastBlockBytes = ciphertextLength % blocksize;
 		int paddingLength = lastBlockBytes > 0 ? blocksize - lastBlockBytes : 0;
@@ -163,15 +191,15 @@ public class CbcBlockCipher {
 		// create padding
 		byte[] padding = new byte[paddingLength + 1];
 		Arrays.fill(padding, (byte) paddingLength);
-		message.writeBytes(padding);
+		plainMessage.writeBytes(padding);
+
 		Cipher blockCipher = suite.getThreadLocalCipher();
 		blockCipher.init(Cipher.ENCRYPT_MODE, key);
-
-		// create GenericBlockCipher structure
-		DatagramWriter result = new DatagramWriter();
-		result.writeBytes(blockCipher.getIV());
-		result.writeBytes(blockCipher.doFinal(message.toByteArray()));
-		return result.toByteArray();
+		byte[] iv = blockCipher.getIV();
+		byte[] plaintext = plainMessage.toByteArray();
+		byte[] message = Arrays.copyOf(iv,  iv.length+  plaintext.length);
+		blockCipher.doFinal(plaintext, 0,  plaintext.length, message, iv.length);
+		return message;
 	}
 
 	/**
@@ -193,5 +221,39 @@ public class CbcBlockCipher {
 		hmac.update(additionalData);
 		hmac.update(content, 0, length);
 		return hmac.doFinal();
+	}
+
+	/**
+	 * Check padding.
+	 * 
+	 * The check is implemented using a "time constant" approach by always
+	 * comparing 256 bytes.
+	 * 
+	 * @param padding padding to be checked
+	 * @param data data to be checked. Must contain at least 256 + 1 bytes from the
+	 *            offset on. The value of the last byte will be changed!
+	 * @param offset offset of the padding field.
+	 * @return {@code true}, if padding bytes in data from the offset on
+	 *         contains the value of the padding byte.
+	 * @throws IllegalArgumentException if the data array doesn't contain
+	 *             257 bytes after the offset.
+	 */
+	public static boolean checkPadding(int padding, byte[] data, int offset) {
+		if (data.length < offset + 257) {
+			throw new IllegalArgumentException("data must contain 257 bytes from offset on!");
+		}
+		byte result1 = 0;
+		byte result2 = 0;
+		byte pad = (byte) padding;
+		for (int index = 0; index <= padding; ++index) {
+			result1 |= pad ^ data[offset + index];
+		}
+		for (int index = padding + 1; index < 256; ++index) {
+			result2 |= pad ^ data[offset + index];
+		}
+		// apply result2 at the "oversize" dummy data to ensure,
+		// that the dummy loop is not skipped by optimization
+		data[data.length - 1] ^= result2;
+		return result1 == 0;
 	}
 }
--- a/scandium-core/src/test/java/org/eclipse/californium/scandium/dtls/RecordCbcValidationTest.java
+++ b/scandium-core/src/test/java/org/eclipse/californium/scandium/dtls/RecordCbcValidationTest.java
@@ -1,0 +1,937 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial implementation
+ ******************************************************************************/
+package org.eclipse.californium.scandium.dtls;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assume.assumeTrue;
+
+import java.security.GeneralSecurityException;
+import java.security.MessageDigest;
+import java.security.SecureRandom;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+import javax.crypto.Mac;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+
+import org.eclipse.californium.elements.util.Bytes;
+import org.eclipse.californium.elements.util.DatagramReader;
+import org.eclipse.californium.elements.util.DatagramWriter;
+import org.eclipse.californium.scandium.dtls.cipher.CbcBlockCipher;
+import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
+import org.eclipse.californium.scandium.dtls.cipher.InvalidMacException;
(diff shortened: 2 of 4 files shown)


sbernard31:
Did you use any documentation to inspire yourself ?

What is the good way to launch tests ? How long does it take approximately ?

boaks:
> Did you use any documentation to inspire yourself ?

Sure [TLStiming.pdf](http://www.isg.rhul.ac.uk/tls/TLStiming.pdf), page 17, point 5.

If Ln must subtract 1 for the padding length and if the "ceiling" is used, seems to be not consistent. Thought we need to support additional data with other length and also SHA with more than 8 bytes length added, I use that way to calculate the hash-compression cylces.
The rest on page 16 seems to be left overs from an former approach, where the padding is verified in order to prevent calculating different hash-compression cycles. That introduced the again a time side channel, as mentioned in that document. 
My proposal is to focus in calculate always equal hash-compressions cycles. The padding length is included in the additional data (as fragment length) and anyway is passing the MAC verification with wrong padding not too risky.

boaks:
> What is the good way to launch tests ? How long does it take approximately ?

I tried to exclude the test by "Manual" but executing them in eclipse ide then starts them again. Currently I remove the Ignore at class level and  run the test from the eclipse ide. 
It should show very close times for the calculation without "ceiling".

boaks:
> 46 of 80: 1-1=>0
> HmacSHA256-fixed verify-40-7: 9687
> 46 of 80: 1-1=>1
> HmacSHA256-fixed verify-40-7++: 11070
> 40 of 80: 1-0=>1
> 39 of 80: 1-0=>1
> 46 of 80: 1-1=>0
> 46 of 80: 1-1=>0
> HmacSHA256-fixed verify-40-7: 9403 9299 9198
> 41 of 80: 1-0=>1
> 40 of 80: 1-0=>1
> 46 of 80: 1-1=>0
> 46 of 80: 1-1=>0
> HmacSHA256-fixed verify-41-6: 9317 9209 9111
> 42 of 80: 1-0=>1
> 41 of 80: 1-0=>1
> 46 of 80: 1-1=>0
> 46 of 80: 1-1=>0
> HmacSHA256-fixed verify-42-5: 9263 9437 9244
> 43 of 80: 1-1=>0
> 42 of 80: 1-0=>1
> 46 of 80: 1-1=>0
> 46 of 80: 1-1=>0
> HmacSHA256-fixed verify-43-4: 9230 9212 9245
> 44 of 80: 1-1=>0
> 43 of 80: 1-1=>0
> 46 of 80: 1-1=>0
> 46 of 80: 1-1=>0
> HmacSHA256-fixed verify-44-3: 9310 9142 9223
> 45 of 80: 1-1=>0
> 44 of 80: 1-1=>0
> 46 of 80: 1-1=>0
> 46 of 80: 1-1=>0
> HmacSHA256-fixed verify-45-2: 9406 9222 9267
> 46 of 80: 2-2=>0
> HmacSHA256-fixed ceiling verify-40-7: 9240
> 46 of 80: 2-2=>1
> HmacSHA256-fixed ceiling verify-40-7++: 11319
> 40 of 80: 2-1=>1
> 39 of 80: 2-1=>1
> 46 of 80: 2-2=>0
> 46 of 80: 2-2=>0
> HmacSHA256-fixed ceiling verify-40-7: 9476 9325 9210
> 41 of 80: 2-1=>1
> 40 of 80: 2-1=>1
> 46 of 80: 2-2=>0
> 46 of 80: 2-2=>0
> HmacSHA256-fixed ceiling verify-41-6: 9373 9297 9215
> 42 of 80: 2-1=>1
> 41 of 80: 2-1=>1
> 46 of 80: 2-2=>0
> 46 of 80: 2-2=>0
> HmacSHA256-fixed ceiling verify-42-5: 9492 9316 9172
> 43 of 80: 2-1=>1
> 42 of 80: 2-1=>1
> 46 of 80: 2-2=>0
> 46 of 80: 2-2=>0
> HmacSHA256-fixed ceiling verify-43-4: 11004 9365 9401
> 44 of 80: 2-2=>0
> 43 of 80: 2-1=>1
> 46 of 80: 2-2=>0
> 46 of 80: 2-2=>0
> HmacSHA256-fixed ceiling verify-44-3: 9162 10796 9295
> 45 of 80: 2-2=>0
> 44 of 80: 2-2=>0
> 46 of 80: 2-2=>0
> 46 of 80: 2-2=>0
> HmacSHA256-fixed ceiling verify-45-2: 9456 9292 9361

The first two values of each variant are "warm up rounds". They show, how much time a additional hash-compression requires.
The time followings show in the case without ceiling, that there is no such difference. With ceiling you see the hits in the 4. and 5. values.
As I already mentioned, java blurrs a lot. 5000 cycles show differences, even if you use always the same record. But the "macro" difference of an additional hash-compressin is visible clearly.

sbernard31:
OK reading [TLStiming.pdf](http://www.isg.rhul.ac.uk/tls/TLStiming.pdf), another warning about trying to implement CBC correctly :

> however, we believe that implementers will find it difficult to do this in a way that eliminates
all significant timing channels (especially for DTLS).

So I would still inclined to not invest to many effort on this and if we really want CBC I would rather invest my time in the [Encrypt-then-MAC RFC 7366](https://tools.ietf.org/html/rfc7366) solution. (which is by the way the LWM2M recommended way)

I will try to look at this PR in details but I'm not sure to be skillful enough.

@ic0ns, if you would like you could maybe have a look at this ?
the coon signal

boaks:
@sbernard31 

I don't get it.

> I would rather invest my time in the Encrypt-then-MAC RFC 7366 solution

Just do it ...

FMPOV:
Believe what someone writes or says without testing it, is mostly much more dangerous.
Therefore I isolated the code in a simple benchmark test.

sbernard31:
I didn't want to say that "I will implement RFC 7366".

My position is more :

Currently I think supporting CBC is not so important.
Of course if we (californium project) have it, this would be better (mainly because this is mandatory for LWM2M) but we can live without it.
From my point of view for v2.0.0 just depreciate CBC is already a good start.
Maybe you want/need a "clean" CBC implementation for 2.0.0 ?
If we (californium project) really want it,  I advice to rather focus on implementing RFC7366 instead of fixing the current implementation which seems to be not so easy.

Anyway, this is just my position and I just try to share it in case it also makes sense to you. This way maybe it helps you to make a "better" choice.  If you think I'm wrong of course you're free to invest your time like you want.


> Believe what someone writes or says without testing it, is mostly much more dangerous.

Here we go in philosophic/epistemological discussion, I totally agree that being skeptical is important.
But you can not test/understand/reproduce every human knowledge and so in a complex world put your trust in expert is often a good position. (I'm not able to test if there is a climate change caused by human activity but I believe it)

Anyway in the case of computer science security especially cryptography stuff, I really don't consider myself skillful enough. I try to understand but I guess I just really understand it superficially. So I generally trust experts, especially when following experts is not so costly in term of usability.

Typically in this case :
- We need to be really confident to bypass expert recommendation. (Maybe you are confident enough but you can not blame other people to not be able or to not want to perfectly understand the problem and so to just trust expert)
- It seems a lot of developers try to do that and failed several time. (I'm pretty sure I'm not better than us)
- We need to test but I'm not even sure if I know how to test this correctly.

All of this makes me feel this way could be riskier and costlier. (again maybe I'm wrong)

boaks:
This PR is not about removing the "not recommended from CBC".
It's about improving a little a known "timing side channels".
Maybe, if it helps, I make it configurable :-).

boaks:
> What would be the reason ? (not stable enough ? performance issue ?)

Only your doubts :-)

sbernard31:
> Only your doubts :-)

Ok  so don't do it configurable ^^

sbernard31:
OK I finally take time to look at this.

First, this PR contains a little refactoring and the real change.
As this is a tricky part of code maybe, separating both modification could help for reviewing.

Reading [TLStiming](TLStiming.pdf) your code and  your explanation above.
I understand that you don't really implement the procedure in TLStiming (page 16/17) but you rather focus trying to have a constant time MAC calculation.
To do that, you pick the idea in TLStiming about executing message digest (also called MAC compression function) on dummy data (dummy data size equals to padding length).

As you said the idea is just to reduce calculation time difference to make the attack more difficult by increasing the needed number of data to collect.

> The rest on page 16 seems to be left overs from an former approach, where the padding is verified in order to prevent calculating different hash-compression cycles.

I didn't get the point here. Point 1, 3 and 4 (page 16) seems to be mainly sanity check.
It makes senses to me to verify those 3 points. (or maybe I missed something?)

If you decide to integrate this kind of code, I think we need to explain more what we try to achieve in the code (with reference to TLStiming.pdf).

> BouncyCastle addressed the attacks in version 1.48 of the Java
library,  released 10/02/2013.   The C# version of BouncyCas-
tle was fixed in CVS at a similar time, and will be included in
release 1.8 at a later date

I tried to look at [the code](https://github.com/bcgit/bc-java) but I didn't find anything about that.

About tests, I begin to look at this but it's still not clear to me. (so no feedbacks for now)

boaks:
> I didn't get the point here. Point 1, 3 and 4 (page 16) seems to be mainly sanity check.
It makes senses to me to verify those 3 points. (or maybe I missed something?)

Point 3 checks the padding, and if it fails, still

> Then let P 0 denote the first plen − t
> bytes of P , compute a MAC on SQN||HDR||P 0 and do a
> constant-time comparison of the computed MAC with the
> last t bytes of P . Return fatal error.
> 

So the MAC is always calculated and always compared. 

My variant also always calculates the MAC (using equal numbers of compressions). I always compare this calculated MAC to the one in the record. In my opinion, that MAC check obsoletes a padding check. I don't see, that the MAC could be verified, but the padding would show defects. Basically it's easier to just call `MessageDigest.isEqual`, than to implement and timing side channel free padding check.

If you prefer, I can also try to get some feedback from the tls mailing list. But though CBC is deprecated anyway, and all people seems to work on new stuff, I guess, I will not receive too many useful answer :-).

boaks:
> I tried to look at the code but I didn't find anything about that.
See
[BlockCipher-HMac](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/tls/TlsBlockCipher.java#L239)

[]padding check](https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/tls/TlsBlockCipher.java#L336) in my opinion, check the MAC obsoletes that.

boaks:
[RFC7366 MAC](https://tools.ietf.org/html/rfc7366#section-3)

> Decryption reverses this processing.  The MAC SHALL be evaluated
> before any further processing such as decryption is performed, and if
> the MAC verification fails, then processing SHALL terminate
> immediately.  For TLS, a fatal bad_record_mac MUST be generated [2].
> For DTLS, the record MUST be discarded, and a fatal bad_record_mac
> MAY be generated [4].  This immediate response to a bad MAC
> eliminates any timing channels that may be available through the use
> of manipulated packet data.

**This immediate response to a bad MAC eliminates any timing channels that may be available through the use of manipulated packet data.**

That is close to what I think: the MAC protects enough, and failing fast means simply, nothing additional introduce something wrong again.

boaks:
@sbernard31 

First at all: I don't plan to merge this soon. You have the time you want, to make up your own mind.  

I update the test and add more explanations to the test. Though the test takes too long, I added a `main` to start that test as application and added `@Ignore` to the test.

The hard issue is to differentiate between timing differences of the implementation itself and the "common java blurr". That makes the test consuming so much time, because the tests must be repeated many times (as an attacker would also need many records, even more, if the transmission time-blurr must also be eliminated.) That the timing depends also on other things, may also be the root for "false positive" of lucky 13, even if scandium's implementation (before this PR) is sure "true positive".

Usually some "advanced filter" are also mentioned to improve the "lucky 13 performance", but I didn't found more about that. So, for now, it's the processing time consuming benchmark for all values :-).

boaks:
@sbernard31

I would like to split the introduction of `CbcBlockCipher`, moving crypt/decrypt out of the `Record`, keeping the old decrypt, from this PR. With that I can reduce some potential merge conflicts, when `Record` is modified for other changes. I will go for that split tomorrow, if you have no general objection introducing such a `CbcBlockCipher`.

sbernard31:
Thx @boaks for all this explanation. :pray: 

> First at all: I don't plan to merge this soon. You have the time you want, to make up your own mind.

Thx to let me know that :)

> I will go for that split tomorrow, if you have no general objection introducing such a CbcBlockCipher.

This sounds good to me :+1:

boaks:
[Lucky 13, BEAST, CRIME,... Is TLS dead, or just resting?](https://datatracker.ietf.org/meeting/89/materials/slides-89-irtfopen-1)

Page 18, ff, 
Page 31-32: describes the lucky 13 timing side channel, which is caused by a different number of compressions (without extra dummy compressions).

==> but this PR eliminates that by adding dummy compressions, so the MAC calculation time gets independent from the padding.

boaks:
I raised a question about the mitigation in the IETF TLS mailing list:
[Lucky 13 mitigation](https://mailarchive.ietf.org/arch/msg/tls/lvjMhZl1RWsvyROyaql_zKV1k8A)

boaks:
Skip the "padding check" itself, creates a vulnerability for POODLE.
Therefore I added the padding check with always 256 compares.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CbcBlockCipher.java
```diff
@@ -194,4 +222,37 @@
...
+		byte result2 = 0;
+		byte pad = (byte) padding;
+		for (int index = 0; index <= padding; ++index) {
+			result1 |= pad ^ data[offset + index];
+		}
+		for (int index = padding + 1; index < 256; ++index) {
+			result2 |= pad ^ data[offset + index];
+		}
+		// apply result2 at the "oversize" dummy data
+		data[data.length - 1] ^= result2;
```
sbernard31: what is the purpose of this line ?

boaks: Ensure, that "optimization" doesn't skip the dummy compare loop.

sbernard31: (could we add this as comment ?)

boaks: Sure.

boaks: Done.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CbcBlockCipher.java
```diff
@@ -70,46 +76,68 @@
...
-		blockCipher.init(Cipher.DECRYPT_MODE,
-				key,
-				new IvParameterSpec(iv));
-		byte[] plaintext = blockCipher.doFinal(reader.readBytesLeft());
-		// last byte contains padding length
+		blockCipher.init(Cipher.DECRYPT_MODE, key, new IvParameterSpec(ciphertext, 0, ivlength));
+		int plaintextLength = blockCipher.doFinal(ciphertext, ivlength, ciphertext.length - ivlength,
+				plaintextOversized);
+		// fill up plaintext for MAC compensation
+		System.arraycopy(FILLUP, 0, plaintextOversized, plaintextLength, suite.getMacMessageBlockLength());
```
sbernard31: Is it normal that we initialize : 
`byte[] plaintextOversized = new byte[ciphertext.length + Math.max(suite.getMacMessageBlockLength(), 256)];`
And here we just use `suite.getMacMessageBlockLength()`

Is `plaintextLength == ciphertext.length` ?

boaks:
> About L1/L2 stuff, I read the §2.1 and I'm not sure to understand why we don't do that ?

Simply, because sticking to this (derived) formula, will cause troubles, if additional data is not 13 bytes (DTLS 1.2 CID is used), or SHA378 is used . 

See [Lucky 13, 2018 - Pseudo Constant Time Implementations of TLS
Are Only Pseudo Secure](https://eprint.iacr.org/2018/747.pdf), "3  Implementation Bugs in Lucky13 Countermeasures" for details. The derivation of this formula from the basic parameter is missing, and so changing these parameters results in errors.   

Generally it's about estimating the compress calculations. Therefore I added the benchmarks for them. You may also check the source e.g. [SHA5](https://github.com/frohoff/jdk8u-dev-jdk/blob/master/src/share/classes/sun/security/provider/SHA5.java).

My formula is based on the amount of data (including the additional data), the parameters of the message digest (block + length bytes), and the possible variation based on padlen.

boaks:
> Should we add a link to http://www.isg.rhul.ac.uk/tls/TLStiming.pdf somewhere in CbcBlockCipher.decrypt ?

I added that already to the wiki.

boaks:
> ('l'−55)/64 + 4, where 'l' is the message length in bytes

and page 17 uses 

> ceiling[(L−55)/64], L = 13 + plen − padlen − 1 − t

Therefore, I didn't stick to the (derived) formula and instead replaced it by my own calculation and benchmarks.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CbcBlockCipher.java
boaks: > Is plaintextLength == ciphertext.length ?

No, the ciphertext contains also the nonce so plaintext.lenght < ciphertext.length.
The array is a little more oversized. 
The `FILLUP` is filled in after the plaintext for the message digest compensation (Mac Block Length).
For the padding check, the bytes are anyway ignored (256).

boaks: Just to mention:
If `max(plen, 256)` is used for the number of compares, the oversize will be the mac block size.

sbernard31:
Thx a lot @boaks, now I think I understand the code and the idea behind.

I'm currently looking at the tests, javadoc helps to understand a bit  but there is nothing which explains "numbers", e.g. : HmacSHA256-MAC compensation-**40**-**07**-same: **1642[253]** < **1671** < **2143[103]**:  **496**

Given :  `HmacSHA256-MAC compensation-aa-bb-same: cccc[ddd] < eeee < ffff[ggg]:  hhh`
`aa`: max key length. 
`bb`: max value length. 
`cccc`: shortest mac calculation
`ddd`: index of this shortest mac calculation
`eeee` : average mac calculation
`ffff`: longest mac calculation
`ggg`: index of this longest mac calculation
`hhh`: `medianMaxMinDeviation`, I'm not sure to get exactly what is this deviation ? 

I'm a bit confused here, especially for (aa, bb, hhh), could you detailed the idea here ?
What kind of information should show (ddd, ggg)  indexes  ? 

I running the tests on a "pretty old" machine, here are my results : 
(I'm not yet able to interpret this correctly :sweat_smile:)
```
statistic:
 * HmacSHA256-MAC compensation-40-07-same: 1642[253] < 1671 < 2143[103]:  496
 * HmacSHA256-MAC none compen.-41-06-same: 1268[022] < 1388 < 1820[109]:  537
 * HmacSHA256-MAC none compen.-41-70-same: 1269[016] < 1414 < 1938[071]:  645
 * HmacSHA256-MAC none compen.-42-05-same: 1268[031] < 1400 < 1935[066]:  651
 * HmacSHA256-MAC none compen.-40-07-same: 1266[010] < 1411 < 1988[182]:  703
 * HmacSHA256-MAC none compen.-42-69-same: 1268[007] < 1439 < 2082[176]:  772
 * HmacSHA256-MAC none compen.-40-71-same: 1269[037] < 1421 < 2082[120]:  790
 * HmacSHA256-MAC compensation-41-06-same: 1642[003] < 1841 < 2573[119]:  878
 * HmacSHA256-MAC compensation-42-05-same: 1651[232] < 1973 < 2846[131]:  913
 * HmacSHA256-MAC none compen.-43-68-same: 2511[024] < 2738 < 3500[136]:  932
 * HmacSHA256-MAC compensation-45-02-all*: 2626[052] < 2927 < 3797[039]:  945
 * HmacSHA256-MAC none compen.-44-03-same: 2510[030] < 2705 < 3482[140]:  946
 * HmacSHA256-MAC none compen.-45-02-same: 2507[009] < 2711 < 3496[183]:  962
 * HmacSHA256-MAC none compen.-45-66-same: 2508[021] < 2766 < 3641[205]:  974
 * HmacSHA256-MAC none compen.-44-67-same: 2510[010] < 2738 < 3604[181]: 1004
 * HmacSHA256-MAC compensation-43-04-same: 2630[132] < 2980 < 3943[142]: 1027
 * HmacSHA256-MAC compensation-44-03-all*: 2624[029] < 2949 < 3932[047]: 1033
 * HmacSHA256-MAC compensation-41-06-all*: 1700[006] < 2641 < 3125[249]: 1047
 * HmacSHA256-MAC compensation-44-03-same: 2629[050] < 2978 < 3993[071]: 1068
 * HmacSHA256-MAC compensation-40-71-same: 1917[005] < 2300 < 3353[031]: 1106
 * HmacSHA256-MAC compensation-42-69-same: 1918[242] < 2253 < 3279[095]: 1108
 * HmacSHA256-MAC compensation-45-02-same: 2632[058] < 3016 < 4071[150]: 1121
 * HmacSHA256-MAC none compen.-43-04-same: 2509[011] < 2742 < 3693[193]: 1125
 * HmacSHA256-MAC compensation-40-07-all*: 1644[007] < 2635 < 3189[048]: 1128
 * HmacSHA256-MAC compensation-44-67-all*: 2916[010] < 3455 < 4443[112]: 1144
 * HmacSHA256-MAC compensation-44-67-same: 2904[061] < 3424 < 4435[165]: 1149
 * HmacSHA256-MAC compensation-41-70-same: 1920[204] < 2237 < 3423[134]: 1233
 * HmacSHA256-MAC compensation-45-66-same: 2905[122] < 3413 < 4567[150]: 1274
 * HmacSHA256-MAC compensation-41-70-all*: 2572[070] < 3338 < 4369[179]: 1285
 * HmacSHA256-MAC compensation-42-69-all*: 2381[069] < 3412 < 4356[076]: 1396
 * HmacSHA256-MAC compensation-42-05-all*: 1954[005] < 2924 < 3961[148]: 1419
 * HmacSHA256-MAC compensation-40-71-all*: 2144[071] < 3261 < 4162[131]: 1434
 * HmacSHA256-MAC compensation-43-68-same: 2908[248] < 3475 < 4779[062]: 1465
 * HmacSHA256-MAC compensation-45-66-all*: 2910[220] < 3444 < 4962[038]: 1630
 * HmacSHA256-MAC compensation-43-04-all*: 2620[010] < 2928 < 4539[206]: 1673
 * HmacSHA256-MAC compensation-43-68-all*: 2910[030] < 3474 < 5201[137]: 1855
 * HmacSHA256-MAC none compen.-40-07-all*:  683[251] <  930 < 2596[003]: 1881
 * HmacSHA256-MAC none compen.-42-05-all*:  670[127] <  935 < 2713[003]: 2012
 * HmacSHA256-MAC none compen.-43-04-all*:  670[234] <  912 < 2731[004]: 2025
 * HmacSHA256-MAC none compen.-43-68-all*:  673[236] < 1240 < 3237[062]: 2202
 * HmacSHA256-MAC none compen.-45-02-all*:  670[214] <  909 < 2945[001]: 2246
 * HmacSHA256-MAC none compen.-41-70-all*:  669[180] < 1210 < 3275[067]: 2274
 * HmacSHA256-MAC none compen.-44-03-all*:  673[090] <  911 < 2988[001]: 2283
 * HmacSHA256-MAC none compen.-41-06-all*:  671[074] <  917 < 2994[001]: 2292
 * HmacSHA256-MAC none compen.-40-71-all*:  669[218] < 1235 < 3305[001]: 2317
 * HmacSHA256-MAC none compen.-45-66-all*:  669[233] < 1231 < 3352[003]: 2382
 * HmacSHA256-MAC none compen.-42-69-all*:  669[143] < 1223 < 3463[065]: 2449
 * HmacSHA256-MAC none compen.-44-67-all*:  670[196] < 1225 < 3558[000]: 2627
time: 116 [min]
```

boaks:
> aa: payload length
> bb: padding length


> hhh: 

median: the value in the middle of the sorted value-list. This is more "stable" against single out-of-range values than the average.
deviation: the square-root of the squares of the differences between the values and the madian (or average) divided by samples - 1.  That's a metric for the "value range", and should indicate, if values are more out of range.

> ddd
> ggg

The idea was to see, if lucky13 hits the real values. But for lucky13 the padding check must be added to the verifyWithoutMacCompensation.

boaks:
Your times are matching my experience as well.

>  *  HmacSHA256-MAC compensation-43-68-same: 2908[248] < 3475 < 4779[062]: 1465
>  * HmacSHA256-MAC compensation-45-66-all*: 2910[220] < 3444 < 4962[038]: 1630
>  * HmacSHA256-MAC compensation-43-04-all*: 2620[010] < 2928 < 4539[206]: 1673
>  * HmacSHA256-MAC compensation-43-68-all*: 2910[030] < 3474 < 5201[137]: 1855

Even with "compensation", the time differences are rather large. And if the "same" padding is used, the shown difference is not caused by the padding. Therefore I mentioned, that the time measurement is not too easy. An "on-path-attacker" will also have some additional blurr from the network. So, in theory, for lucky13, 64k different 2-bytes-padding must be checked, multiplied by e.g 1000 messages to compensate the blurr. And if there is a load-change within that time, the measurement may be in vain.

scandium-core/src/test/java/org/eclipse/californium/scandium/dtls/RecordCbcValidationTest.java
```diff
@@ -0,0 +1,937 @@
...
+ * differences, and it seems to be not related to "same" or "all*".
+ * 
+ * Remark: This test contains some more stuff; different padding operators and
+ * different statistic functions. Feel free to create your own benchmarks!
+ */
+@RunWith(Parameterized.class)
+@Ignore
+public class RecordCbcValidationTest {
+
+	private static final int LOOPS = 500;
```
boaks: You may reduce this number of loops for slower machines.
The benchmark uses 256*LOOPS decrypt calls to measure the time.

sbernard31:
About hhh, I see what you mean but this sounds not to look like to `medianMaxMinDeviation` which seems to only  compare the min and max value  (not all the value range) ?

boaks:
```
Statistic overall = new Statistic(key, 256);
Statistic[] allStatistics = new Statistic[256];
for (int p = 0; p < 256; ++p) {
	String k = String.format("%s-%d-%03d", text, size, p);
	Statistic sample = run(plaintext, k, new PadSet(p), verify);
	allStatistics[p] = sample;
	overall.add(sample.avg);
	if (p % 32 == 0) {
		System.out.println(sample);
	}
}
allResults.put(overall, overall.medianMaxMinDeviation());
```

It is based one the average of each "padlen" test.
That average is sorted, and the median is then used to calculate a deviation.

What doesn't fit for you?

boaks:
Just to mention:
You spend as many PC time you want to calculate different statistics to see, if you get different indicators. For me it got too time consuming.

boaks:
I added a configuration to see lucky 13 effects for sha-1. For the "hit" , I prepare a proper "1" padding and leave the other paddings invalid.
I didn't add stuff for sha-256, because the 2^48 tries multiplied by repeats to compensate the java blurr, would take too long on my machine :-).

sbernard31:
> This PR is not about removing the "not recommended from CBC".

We agree about that.

> I don't insist, that this is merged, but I would also not spend my time in alternative approaches.

The tests seem to show that this new implementation is closer to a "constant time" implementation than before. So I don't see any reason to not integrate this.

I think you're right about not spending much more time one this.

This issue is stimulating (very interesting issue) and also very frustrating to me (because it's not to rational to take much more time on a deprecated cipher and also because I'm far to understand it correctly !)

sophokles73:
@boaks @sbernard31 

I haven't found the time to read through all of the comments on this PR. However, my feeling is that you already have spent more time on this than it deserves. It might be an interesting challenge to implement CBC in a better way, but I think that we should simply advise people not to use CBC in production but still keep it in order to comply with the LWM2M spec. If @boaks has improved the existing CBC code, even better. But I do not think that you should try to find the best possible solution ...
If people want to use it they have been warned. Otherwise, I think you should focus on getting everything prepared for a 2.0 release.

Just my 2 cents.

boaks:
> This issue is stimulating (very interesting issue) and also very frustrating to me

Yea, you can try to implement it in a compensated way, but the java blurr will make it close to impossible to show that, because the timing variance from java seems to be in the same "order". Therefore all test has to be "amplified" by many repeats and so the test run very long.

> The tests seem to show that this new implementation is closer to a "constant time" implementation than before. So I don't see any reason to not integrate this.

OK, so LGTM, or do you want to spent "your time" in an other statistic evaluation approach?

sbernard31:
LGTM

boaks:
> However, my feeling is that you already have spent more time on this than it deserves. 

Luckily, it's mainly only the time of the machine, not of me :-).
But, though I'm not a statistic nor filter expert, I'm more or less "done".

boaks:
I prefer without the last commit, OK?

sbernard31:
> the timing variance from java seems to be in the same "order".

This surprising me a lot, I didn't expect this. When I played with jmh by the past, I already see some code with not so much variation. 

> I prefer without the last commit, OK?

OK

boaks:
> This surprising me a lot, I didn't expect this. When I played with jmh by the past, I already see some code with not so much variation.

My feeling is, that this is partially caused by the GC.

PR #1038, opened 2019-08-17
Handshake validation

boaks:


boaks:
Based PR #1035

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -511,6 +531,47 @@ public void run() {
...
+						+ " handshake message, no more expected!", alert);
+			}
+			HandshakeState expectedState = states[statesIndex];
+			boolean expected = expectedState.expect(message);
+			if (!expected && expectedState.isOptional()) {
+				if (statesIndex + 1 < states.length) {
+					HandshakeState nextExpectedState = states[statesIndex + 1];
+					if (nextExpectedState.expect(message)) {
+						++statesIndex;
+						expectedState = nextExpectedState;
```
sbernard31: why setting expectedState ? (it seems used only if expected is false)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -470,7 +477,14 @@ private void receivedClientHello(final ClientHello clientHello) throws Handshake
...
 
-		createCertificateRequest(clientHello, flight);
+		boolean clientCertificate = createCertificateRequest(clientHello, flight);
+
+		if (clientCertificate) {
+			states = CLIENT_CERTIFICATE;
+		} else {
+			states = NO_CLIENT_CERTIFICATE;
+		}
+		statesIndex = -1;
```
sbernard31: Should we add CLIENT_HELLO to CLIENT_CERTIFICATE and NO_CLIENT_CERTIFICATE initialize `states` with CLIENT_CERTIFICATE (or NO_CLIENT_CERTIFICATE)
And just switch the `states` here if needed ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks: Though the CLIENT_HELLO is validates ahead, FMPOV, I would not "double check" it here.

sbernard31: OK you mean that CLIENT_HELLO is already checked in DTLSConnector ?

I get it, I still feel regarding in term of separation of concerns and readability (avoid the -1 initialization), it could make sense to let ServerHandshaker check CLIENT_HELLO too, but I can understand you want to avoid the double check. :)

sbernard31:
we can merge it ! :+1:

PR #1039, opened 2019-08-17
Add aes256gcm

boaks:


boaks:
Based on #1035 and #1038

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
```diff
@@ -85,16 +85,26 @@
 
 	// PSK cipher suites, ordered by default preference, see getPskCiperSuites ///
 	TLS_PSK_WITH_AES_128_GCM_SHA256(0x00A8, CertificateKeyAlgorithm.NONE, KeyExchangeAlgorithm.PSK, CipherSpec.AES_128_GCM, MACAlgorithm.NULL, true),
+	TLS_PSK_WITH_AES_2568_GCM_SHA378(0x00A9, CertificateKeyAlgorithm.NONE, KeyExchangeAlgorithm.PSK, CipherSpec.AES_256_GCM, MACAlgorithm.NULL, true, PRFAlgorithm.TLS_PRF_SHA384),
```
sbernard31: 2568 => 256 ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
```diff
@@ -85,16 +85,26 @@
...
 	TLS_PSK_WITH_AES_256_CCM(0xC0A5, CertificateKeyAlgorithm.NONE, KeyExchangeAlgorithm.PSK, CipherSpec.AES_256_CCM, MACAlgorithm.NULL, true),
+
+	/**See <a href="https://tools.ietf.org/html/rfc8442#section-3">RFC 8442</a> for details*/
+	/**Note: compatibility not tested! openssl 1.1.1 seems not supporting them */
+	TLS_ECDHE_PSK_WITH_AES_128_GCM_SHA256(0xD001, CertificateKeyAlgorithm.NONE, KeyExchangeAlgorithm.ECDHE_PSK, CipherSpec.AES_128_GCM, MACAlgorithm.NULL, true),
+	TLS_ECDHE_PSK_WITH_AES_256_GCM_SHA378(0xD002, CertificateKeyAlgorithm.NONE, KeyExchangeAlgorithm.ECDHE_PSK, CipherSpec.AES_256_GCM, MACAlgorithm.NULL, true, PRFAlgorithm.TLS_PRF_SHA384),
+	TLS_ECDHE_PSK_WITH_AES_128_CCM_SHA256(0xD005, CertificateKeyAlgorithm.NONE, KeyExchangeAlgorithm.ECDHE_PSK, CipherSpec.AES_128_CCM, MACAlgorithm.NULL, true),
+	/**not recommended, see <a href="https://tools.ietf.org/html/rfc8442#section-4">RFC 8442</a> for details*/
+	TLS_ECDHE_PSK_WITH_AES_128_CCM_8_SHA256(0xD003, CertificateKeyAlgorithm.NONE, KeyExchangeAlgorithm.ECDHE_PSK, CipherSpec.AES_128_CCM_8, MACAlgorithm.NULL, false),
```
sbernard31: Regarding iana : https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4
Even CoAP mandatory cipher (based on CCM_8) is not recommended :thinking: 
It seems that recommended does not necessary means "flawed", see https://tools.ietf.org/html/rfc8447#section-5

For CCM8 in particular, they say : 
> CCM_8 cipher suites are not marked as "Recommended".  These
cipher suites have a significantly truncated authentication tag
that represents a security trade-off that may not be appropriate
for general environments.

@ic0ns Do you know if there is known issue with cipher present here (except CBC) ?

By the way does the way we class cipher suite makes more or less sense ? (see https://github.com/eclipse/californium/blob/2.0.0-M16/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java#L61)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
boaks: Typo

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
boaks: At least for the GCM/CCM the size of the SHA doesn't make the record larger.

boaks: Should I can change the `false` to `true` and just leave the comment?

ic0ns: They are fine. The only known attack against gcm in tls is nonce reuse. If you dont do that you are good.

boaks: > /**not recommended, see <a href="https://tools.ietf.org/html/rfc8442#section-4">RFC 8442</a> for details*/
> 	TLS_ECDHE_PSK_WITH_AES_128_CCM_8_SHA256

I guess, @sbernard31 mainly want to know, if that could be recommended for usage with constrained devices.

sbernard31: > I guess, @sbernard31 mainly want to know, if that could be recommended for usage with constrained devices.

That's the idea ! I would mainly know : 
1. If there is issue with CCM8 in a general way ? (The question is mainly driven by the IANA comment above) => regarding @ic0ns comment I understand that CCM8 is OK.
2. I would like to know if the [comment](https://github.com/eclipse/californium/blob/2.0.0-M16/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java#L61) which explains how we order cipher makes sense because I'm not so confident about what I written there.

> Should I  change the false to true and just leave the comment?

From what I understand for now, there is no reason to treat CCM8 ciphers differently.
So if we make  TLS_ECDHE_PSK_WITH_AES_128_CCM_8_SHA256 "not recommended", we should make all the others not recommended.
Same thing regarding the addition of a comment.

From what "the coon" said and what I read until know, I feel that CCM8 is OK, so I would go for :
- not adding any comments.
- making TLS_ECDHE_PSK_WITH_AES_128_CCM_8_SHA256 a recommended cipher.
- putting TLS_ECDHE_PSK_WITH_AES_128_CCM_8_SHA256 before TLS_ECDHE_PSK_WITH_AES_128_CCM_SHA256 and putting TLS_ECDHE_PSK before TLS_PSK (regarding [comment](https://github.com/eclipse/californium/blob/2.0.0-M16/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java#L61)  above)

So something like this : 
```java
TLS_ECDHE_PSK_WITH_AES_128_GCM_SHA256
TLS_ECDHE_PSK_WITH_AES_256_GCM_SHA378
TLS_ECDHE_PSK_WITH_AES_128_CCM_8_SHA256
TLS_ECDHE_PSK_WITH_AES_128_CCM_SHA256
TLS_PSK_WITH_AES_128_GCM_SHA256
TLS_PSK_WITH_AES_256_GCM_SHA378
TLS_PSK_WITH_AES_128_CCM_8
TLS_PSK_WITH_AES_256_CCM_8
TLS_PSK_WITH_AES_128_CCM
TLS_PSK_WITH_AES_256_CCM
```

sbernard31:
LGTM

PR #1042, opened 2019-08-21
More cleanup 2

boaks:
Move general message processing to base handshaker.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
@@ -166,6 +166,8 @@ public abstract class Handshaker {
 	 */
 	private int nextReceiveMessageSequence = 0;
 
+	private boolean lastFlight;
+
 	/** Realtime nanoseconds of last sending a flight */
 	private long flightSendNanos;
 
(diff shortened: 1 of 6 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -506,7 +506,11 @@ public final void processMessage(Record record) throws HandshakeException {
 						}
 					}
 					expectMessage(messageToProcess);
-					doProcessMessage(messageToProcess);
+					if (messageToProcess.getContentType() == ContentType.CHANGE_CIPHER_SPEC) {
+						setCurrentReadState();
```
sbernard31: I don't know if this is an issue but I think we remove the debug log about receiving a CHANGE_CIPHER_SPEC

boaks:
> maybe we don't need expectChangeCipherSpecMessage is not needed and we can use states for isChangeCipherSpecMessageExpected

My intention was to check, if the handshake state validation works. Currently it's configurable, not that it causes serious issues with "exotic clients". But, if no issues are reported, that could be cleaned up.

boaks:
> add a receivedServerHello in ResumingClientHandshaker

Done.

sbernard31:
LGTM
(just the commit name issue `receivedServerFinished` instead of `receivedServerHello`)

PR #1043, opened 2019-08-21
Fixes for cf_secure demo.

boaks:
Default arguments didn't work.
Client prints now peer and cipher suite.


sbernard31:
no objection

PR #1046, opened 2019-08-23
Update to netty 4.1.39

boaks:
Extend demo certificates
Fix hang on close.

### DIFF
--- a/element-connector-tcp/pom.xml
+++ b/element-connector-tcp/pom.xml
@@ -16,7 +16,7 @@
 	<description>Element connector implementation for TCP/TLS</description>
 	
 	<properties>
-		<netty.version>4.1.25.Final</netty.version>
+		<netty.version>4.1.39.Final</netty.version>
 		<netty.version.lowerbound>4.1</netty.version.lowerbound>
 		<netty.version.upperbound>5</netty.version.upperbound>
 	</properties>
--- a/element-connector-tcp/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
+++ b/element-connector-tcp/src/main/java/org/eclipse/californium/elements/tcp/TcpClientConnector.java
@@ -46,6 +46,7 @@ import io.netty.channel.pool.AbstractChannelPoolHandler;
 import io.netty.channel.pool.AbstractChannelPoolMap;
 import io.netty.channel.pool.ChannelPool;
 import io.netty.channel.pool.FixedChannelPool;
+import io.netty.channel.pool.SimpleChannelPool;
 import io.netty.channel.socket.nio.NioSocketChannel;
 import io.netty.handler.timeout.IdleStateHandler;
 import io.netty.util.concurrent.Future;
@@ -74,6 +75,7 @@ import org.slf4j.LoggerFactory;
  * accept new incoming connections.
  */
 public class TcpClientConnector implements Connector {
+	private static final boolean USE_FIXED_CONNECTION_POOL = false;
 
 	protected final Logger LOGGER = LoggerFactory.getLogger(getClass().getName());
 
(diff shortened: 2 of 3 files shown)


boaks:
Using netty newer than 4.1.25 leads to fails on close (blocks execution). Using asnycClose reduces these failures, but using the repeated test runners shows, that it is still not stable (at least on unix).

"Last warning -before it blocks"

```
WARN io.netty.util.concurrent.SingleThreadEventExecutor - An event executor terminated with non-empty task queue (1)
```


There is already a netty issue, very close to my observation:
[netty - FixedChannelPool#close()](https://github.com/netty/netty/issues/8398)

And [PR](https://github.com/netty/netty/pull/8408) for that.

So, using netty 4.1.39 may cause issues, when shutdown. Should we go anyway for it?

sbernard31:
> So, using netty 4.1.39 may cause issues, when shutdown. Should we go anyway for it?

Good question.
Netty release very often, so maybe we should wait ? On the other hand, the issue sounds pretty old (oct 2018) and there is not so much activities on the PR (nov/dec 2018) and If I well understand the issue not so impacting.

Is there any reason about moving to 4.1.39 ? (I mean other than using the last version)

boaks:
@sophokles73 asked about. Currently hono uses only coap (over-udp) experimentally. But with a Californium 2.0 release, hono can integrate the coap support. If someone wants to continue the work on the "coap-over-tcp" stuff, the netty version must be in line.
From my feeling, the most expect that this is done by someone else. And though no one is contributing it, it stays "in between". 
For me it's mainly used for tests, if tls or dtls performs somehow better in some scenarios. But therefore no complete coap-over-tcp is required. And being "always short in time", I don't think fixing netty and implement "coap-over-tcp" is an job for me.

sbernard31:
> @sophokles73 asked about.

@sophokles73, what was the reason ?

sophokles73:
Well, for Hono it was important because of vulnerabilities in the HTTP/2 support in Netty. For Californium, Netty is only used in the coap-over-tcp stuff IIRC. Since the coap-over-tcp implementation is not really useful anyway as long as it is does not implement the final RFC, we could think about removing it altogether from Cf and get rid of Netty along the way. On the other hand, if we continue to support coap-over-tcp, IMHO we should use a recent version of Netty in order to take advantage of recent fixes and improvements. If something serious comes up in Netty which we are also affected by (i.e. not just HTTP/2) related, then I believe that is unlikely that a bugfix will be back-ported to a not so recent version of Netty ...

boaks:
Instead of removing it, which would make it some harder, if someone else want to contribute to it (e.g. in the context of LwM2M 1.1), hono can simply select the udp stuff only (just don't use the element-connector-tcp).

Californium itself is sure not affected by the HTTP/2 stuff, though only tcp/tls is used.

sbernard31:
> Instead of removing it, which would make it some harder, if someone else want to contribute to it

:+1:! This makes me think that coap-tcp in californium is not really in a releasable state. Will we release it in the v2.0.0 ? (It's an open question, I have no strong opinion about that)

> On the other hand, if we continue to support coap-over-tcp, IMHO we should use a recent version of Netty in order to take advantage of recent fixes and improvements

:+1:, but the question here is should we use the last version knowing that this bring with a small regression or wait the fix at netty. 

In all case, we should maybe manifest ourself at netty side to let them know we are impacted. (try to revive the PR https://github.com/netty/netty/pull/8408)

sophokles73:
> Will we release it in the v2.0.0?

I do not see a problem with including it in the release. We can document it as an *experimental feature* which should indicate to potential users the maturity of the code ...

boaks:
If we keep it (even incomplete), I try to provide more information to the netty project.
It seems, that it works with the `SimpleChannelPool` instead of the `FixedChannelPool`, may be that helps to fix it ...

boaks:
I opened two issues on netty for this update.

[Question about the difference of the channel pools](https://github.com/netty/netty/issues/9604)

[Description of the unit test issues](https://github.com/netty/netty/issues/9603)

boaks:
Though `FixedChannelPool` is only used on the client side, I think we merge this and switch to the `SimpleChannelPool`. I didn't receive an answer to the question about the difference, but I guess, the one opens an other channel, it the other get's broken. Though currently closing/changing the channel is not handled, the difference for the experimental usage seems to be low.

@sbernard31 
@sophokles73 

WDYT?

sbernard31:
:+1:

PR #1047, opened 2019-08-26
Add illegal argument, if reading exceeds available bytes.

boaks:
Fix record length checks base on left bits instead of bytes.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
@@ -385,7 +385,7 @@ public class Record {
 			}
 			int length = reader.read(LENGTH_BITS);
 
-			if (reader.bitsLeft() < length) {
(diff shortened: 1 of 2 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java
```diff
@@ -385,7 +385,7 @@ private Record(ProtocolVersion version, int epoch, long sequenceNumber, long rec
 			}
 			int length = reader.read(LENGTH_BITS);
 
-			if (reader.bitsLeft() < length) {
```
sbernard31: This was a bug ? Which made that some "valid" record be discarded ?

boaks: > This was a bug ? 

Yes.

> Which made that some "valid" record be discarded ?

No, invalid records may have passed the invalid test. If the length was larger than the actual available bytes, but smaller than the available bits (8xbytes). Encrypted records would have been invalidated with that invalid MAC.

PR #1049, opened 2019-08-28
Include hello extensions into the handshake validation.

boaks:
Add range reader for that.


element-connector/src/main/java/org/eclipse/californium/elements/util/DatagramReader.java
```diff
@@ -273,6 +334,25 @@ public int bitsLeft() {
 		return (byteStream.available() * Byte.SIZE) + (currentBitIndex + 1);
 	}
 
+	public DatagramReader createRangeReader(int count) {
+		if (0 < currentBitIndex) {
```
sbernard31: `currentBitIndex > 0` ?

boaks:
Yes, with the "verification/tests about hello extensions" as main part.

PR #1050, opened 2019-08-29
Add connection listener.

boaks:


sbernard31:
:+1: 

This is mainly intended to monitor when connection is closed  ?

boaks:
yes.

PR #1052, opened 2019-08-29
Limit used resource for handshakes.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -93,6 +93,10 @@ public final class DtlsConnectorConfig {
 	 */
 	public static final int DEFAULT_MAX_FRAGMENTED_HANDSHAKE_MESSAGE_LENGTH = 8192;
 	/**
+	 * The default value for the <em>maxDeferredProcessedHandshakeRecordsSize</em> property.
+	 */
+	public static final int DEFAULT_MAX_DEFERRED_PROCESSED_HANDSHAKE_RECORDS_SIZE = 8192;
+	/**
 	 * The default value for the <em>staleConnectionThreshold</em> property.
 	 */
 	public static final long DEFAULT_STALE_CONNECTION_TRESHOLD = 30 * 60; // 30 minutes
@@ -204,7 +208,9 @@ public final class DtlsConnectorConfig {
 
 	private Integer outboundMessageBufferSize;
 
-	private Integer maxDeferredProcessedApplicationDataMessages;
+	private Integer maxDeferredProcessedOutgoingApplicationDataMessages;
+
+	private Integer maxDeferredProcessedIncomingRecordsSize;
 
 	private Integer maxConnections;
 
@@ -349,12 +355,21 @@ public final class DtlsConnectorConfig {
 	}
 
 	/**
-	 * Gets the maximum number of deferred processed application data messages.
+	 * Gets the maximum number of deferred processed outgoing application data messages.
 	 * 
-	 * @return the maximum number of deferred processed application data messages
+	 * @return the maximum number of deferred processed outgoing application data messages
 	 */
-	public Integer getMaxDeferredProcessedApplicationDataMessages() {
-		return maxDeferredProcessedApplicationDataMessages;
+	public Integer getMaxDeferredProcessedOutgoingApplicationDataMessages() {
+		return maxDeferredProcessedOutgoingApplicationDataMessages;
+	}
+
+	/**
+	 * Gets the maximum size of all deferred processed incoming records.
+	 * 
+	 * @return the maximum size of all deferred processed incoming records
+	 */
+	public Integer getMaxDeferredProcessedIncomingRecordsSize() {
+		return maxDeferredProcessedIncomingRecordsSize;
 	}
 
 	/**
@@ -799,7 +814,8 @@ public final class DtlsConnectorConfig {
 		cloned.supportedCipherSuites = supportedCipherSuites;
 		cloned.trustedRPKs = trustedRPKs;
 		cloned.outboundMessageBufferSize = outboundMessageBufferSize;
-		cloned.maxDeferredProcessedApplicationDataMessages = maxDeferredProcessedApplicationDataMessages;
+		cloned.maxDeferredProcessedOutgoingApplicationDataMessages = maxDeferredProcessedOutgoingApplicationDataMessages;
+		cloned.maxDeferredProcessedIncomingRecordsSize = maxDeferredProcessedIncomingRecordsSize;
 		cloned.maxConnections = maxConnections;
 		cloned.staleConnectionThreshold = staleConnectionThreshold;
 		cloned.connectionThreadCount = connectionThreadCount;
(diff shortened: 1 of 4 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1561,24 +1577,49 @@ public Builder setTrustCertificateTypes(CertificateType... certificateTypes) {
...
-		 *            deferred processed messages
+		 * @param maxDeferredProcessedOutgoingApplicationDataMessages maximum
+		 *            number of deferred processed messages
 		 * @return this builder for command chaining.
 		 * @throws IllegalArgumentException if the given limit is &lt; 0.
 		 */
-		public Builder setMaxDeferredProcessedApplicationDataMessages(final int maxDeferredProcessedApplicationDataMessages) {
-			if (maxDeferredProcessedApplicationDataMessages < 0) {
-				throw new IllegalArgumentException("Max deferred processed application data messages must not be negative!");
+		public Builder setMaxDeferredProcessedOutgiongApplicationDataMessages(
```
sbernard31: typo `outgiong`

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1561,24 +1577,49 @@ public Builder setTrustCertificateTypes(CertificateType... certificateTypes) {
...
+				final int maxDeferredProcessedOutgoingApplicationDataMessages) {
```
sbernard31: If we don't need the final keyword, could we remove it ?

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1561,24 +1577,49 @@ public Builder setTrustCertificateTypes(CertificateType... certificateTypes) {
...
+		 * or processed deferred. Set this to limit the maximum size of all
+		 * records, which are intended to be processed deferred. If more records
+		 * are received, these records are dropped.
+		 * 
+		 * @param maxDeferredProcessedIncomingRecordsSize maximum size of all
+		 *            deferred handshake records
+		 * @return this builder for command chaining.
+		 * @throws IllegalArgumentException if the given limit is &lt; 0.
+		 */
+		public Builder setMaxDeferredProcessedIncomingRecordsSize(final int maxDeferredProcessedIncomingRecordsSize) {
```
sbernard31: remove final too ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1034,19 +1048,31 @@ final int getNextReceiveMessageSequenceNumber() {
 	}
 
 	public void addApplicationDataForDeferredProcessing(RawData outgoingMessage) {
-		// backwards compatibility, allow on outgoing message to be stored.
-		int max = maxDeferredProcessedApplicationDataMessages == 0 ? 1 : maxDeferredProcessedApplicationDataMessages;
+		// backwards compatibility, allow one outgoing message to be stored.
+		int max = maxDeferredProcessedOutgoingApplicationDataMessages == 0 ? 1 : maxDeferredProcessedOutgoingApplicationDataMessages;
```
sbernard31: as there is a default value > 0, do we need this backwards compatibility stuff ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1056,6 +1082,9 @@ public void addRecordsForDeferredProcessing(Record incomingMessage) {
 	public List<Record> takeDeferredRecords() {
 		List<Record> records = new ArrayList<Record>(deferredRecords);
 		deferredRecords.clear();
+		for (Record record : records) {
+			deferredRecordsSize -= record.size();
```
sbernard31: This should not happened if there is not a bug but should we ensure that `deferredRecordsSize` is not less than 0 ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -292,8 +297,8 @@ protected Handshaker(boolean isClient, int initialMessageSeq, DTLSSession sessio
 		this.pskStore = config.getPskStore();
 		this.session.setMaxTransmissionUnit(maxTransmissionUnit);
 		this.applicationLevelInfoSupplier = config.getApplicationLevelInfoSupplier();
-		this.deferredApplicationData = new ArrayList<RawData>(maxDeferredProcessedApplicationDataMessages);
-		this.deferredRecords = new ArrayList<Record>(maxDeferredProcessedApplicationDataMessages);
+		this.deferredApplicationData = new ArrayList<RawData>(maxDeferredProcessedOutgoingApplicationDataMessages);
+		this.deferredRecords = new ArrayList<Record>(maxDeferredProcessedOutgoingApplicationDataMessages);
```
sbernard31: Should we initialize this arraylist with the max size ?
If I'm not wrong, with default value this makes 64KB (8192*2*4 bytes) used for each Handshaker.
My guess is that most of the time the order of magniture is most arround 10.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: Hm, my intention was for outgoing the number of messages, default 10.
For incoming my intention was the amount of bytes. But the RecordsSize is not used to initialize the arrays. So I have no idea, why it's 64K (or 819224 )?

boaks: My understanding for incoming stuff:
Max 8K for deferred records.
Max 8k for fragmented records
Sum => 16K per handshake

sbernard31: I read this too quickly, I believe that default value for `maxDeferredProcessedOutgoingApplicationDataMessages` was `8192` ... :confounded: 
(I lost myself with all those `maxDeferred` variable :sweat_smile: )

Default value is 10, so initializing arrayList to `maxDeferredProcessedOutgoingApplicationDataMessages` is OK with default value.
This will cost 80B(10*2*4bytes) by handshaker by default.

But this makes me think that if somebody choose a very large value, he could consume lot of RAM because we initialize arrayList to this size instead of just let start with default capacity (using default constructor `new ArrayList()`)

boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sbernard31: I'm not sure my proposition above was a good one, so I'm ok if we don't add anything about that.
But as you said OK, I search how you implement this and I didn't find that.
did I missed something ?

boaks: Uups, just missed it.

boaks: Now, done.

sbernard31:
LGTM

PR #1056, opened 2019-08-30
Redesign handshake hashs

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
@@ -57,7 +57,6 @@ import java.util.List;
 import org.eclipse.californium.elements.auth.RawPublicKeyIdentity;
 import org.eclipse.californium.elements.auth.X509CertPath;
 import org.eclipse.californium.elements.util.Bytes;
-import org.eclipse.californium.elements.util.DatagramWriter;
 import org.eclipse.californium.elements.util.StringUtil;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
 import org.eclipse.californium.scandium.dtls.AlertMessage.AlertDescription;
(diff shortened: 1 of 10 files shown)


boaks:
May be sometime in the future this also simplifies [RFC7627](https://tools.ietf.org/html/rfc7627)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -223,11 +208,13 @@ protected void doProcessMessage(HandshakeMessage message) throws HandshakeExcept
 						String.format("Unsupported key exchange algorithm %s", getKeyExchangeAlgorithm().name()),
 						new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, message.getPeer()));
 			}
+			handshakeMessages.add(message);
```
sbernard31: ~(clearly a details)
I feel this is a bit strange that sometime we call `handshakeMessages.add(message)` in the generic `doProcessMessage` and sometime in the specific processing method (like `receivedServerHello`).~

Fixed in next commit.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
```diff
@@ -743,12 +745,38 @@ public boolean isSupported() {
...
+		 * 
+		 * Calls {@link Mac#reset()} on access.
+		 * 
+		 * @return mac, or {@code null}, if not supported by vm.
+		 */
 		public Mac getMac() {
-			return mac == null ? null : mac.current();
+			if (mac != null) {
+				Mac current = mac.current();
+				current.reset();
```
sbernard31: was it a bug ? (the missing reset ?)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -207,7 +192,7 @@ protected void doProcessMessage(HandshakeMessage message) throws HandshakeExcept
 				break;
 
 			case PSK:
-				serverKeyExchange = (PSKServerKeyExchange) message;
+//				serverKeyExchange = (PSKServerKeyExchange) message;
```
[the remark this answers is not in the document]
boaks: Uups, no.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/CipherSuite.java
boaks: I added it for cases, where something fails before and so the `doFinal` or `digest` is not called. But I didn't recognized, that actually this happens.

boaks:
I squashed the commits not to change the same files over and over for the same.

sbernard31:
Ok

PR #1057, opened 2019-09-02
Use range reader for handshake message.

boaks:
Reduce read byte calculations.


element-connector/src/main/java/org/eclipse/californium/elements/util/DatagramReader.java
```diff
@@ -344,6 +344,19 @@ public int bitsLeft() {
 	 *             bytes
 	 */
 	public DatagramReader createRangeReader(int count) {
+		return new DatagramReader(createRangeInputStream(count));
+	}
+
+	/**
+	 * Create input stream for provided range.
+	 * 
+	 * @param count size of the range
```
sbernard31: maybe we could precise in the size is in bits or bytes ? (same above)

PR #1059, opened 2019-09-02
Add want auth option to cf-secure demo

boaks:


sbernard31:
no objection.

NO_AUTH : no client authentication required
WANT_AUTH : client authentication wanted
and nothing means : client authentication required ?

boaks:
Yes

sbernard31:
maybe this will be more intuitive to have a REQ_AUTH option but it's just a demo so it's ok like this !

PR #1060, opened 2019-09-04
Correct the anti-reply marking of handshake records.

boaks:
Sort handshake records according the handshake message number.
Only mark a record after successful processing. Remove records of
current epoch with same sequence number after processing the record.


boaks:
The last changes for M17 :-).

PR #1062, opened 2019-09-04
Add openssl interoperability tests.

boaks:
Requires installation of openssl.
Test currently not enabled for maven builds. Therefore run unit test from IDE.


boaks:
Not before M17 :-)

boaks:
Currently these test are intended to be executed on "local machines", though openssl would require "custom images" and so pipelines. Anyway, I think, they contribute to quality.

boaks:
For tests with windows, the openssl s_server 1.1.1c seems to be broken. I opened an issue for that.

https://github.com/openssl/openssl/issues/9817

sbernard31:
:+1:  

(on my debian 10 4.19.0-5-amd64/OpenSSL 1.1.1c  28 May 2019 : all is green  :heavy_check_mark: )

Issue #1064, opened 2019-09-05
Extend TokenGenerator - reintroduce address scoped tokens


boaks:
In late 2017 we introduced the `TokenGenerator` with two kind of tokens:

- long term tokens, used for clustered observe/notifies
- short term tokens, used for normal requests

With that, we changed the scope of the tokens to be no only unique per peer, we changed it to be unique per cluster or single cluster node.

FMPOV, that limits to RFC7252 confirm usage of tokens too much. 
Therefore, i would like to change this again and introduce 3 token scopes:

- long term tokens, used for clustered observe/notifies unique within the cluster
- short term tokens - node unique, used for multicast requests
- short term tokens - peer unique, used for normal requests

For the "short term tokens - peer unique",  the equals and hashCode will be extended by the ip-address:port (endpoint) of the peer. That rules are only apply for outgoing request and incoming responses. The server-side handle the tokens still opaque.
 
 To distinguish the scope, my plan is to add a Scope enum to the `TokenGenerator`, replace the `boolean longTermScope` argument by that enum, and add a `Scope getScope(Token)`. That enables the usage of custom rules creating tokens as well as support "application supplied tokens" with that scoping.

If no objection, I would like to prepare a PR for that.

sbernard31:
> FMPOV, that limits to RFC7252 confirm usage of tokens too much.

Could you explain a bit more the limitation you are talking about ?

boaks:
[RFC7 7252 - Token](https://tools.ietf.org/html/rfc7252#section-5.3.1)  

> A token is intended for use as a client-local identifier for differentiating between concurrent requests (see Section 5.3); it could have been called a "request ID".

So, if the token is provided by the application, may be even a EMPTY token (as LwM2M specified ago), the consideration may be based on that "client-local identifier" intention.

sbernard31:
I remember that token must be unique by foreign peer.
By the past I tried to introduced a [endpoint Identifier](https://github.com/eclipse/californium/pull/322). This way we could have identified a request by [token+endpoint identifier].
I still feel that this kind of abstraction  would really help. (E.g. with connectionID, I bet that we could find that address:port is not a really good identifier :thinking:)

I thought that maybe you discovered a new limitation but I understand correctly this is not the case. You finally just think this is too restrictive, am I right ?

This brings me into those questions :
 1. Do you face a use case (situation) where the current behavior is too restrictive ?
 2. Do we need this for 2.0.0 ?
[...]

PR #1066, opened 2019-09-05
Introduce key-token to support client-local tokens.

boaks:
Issue #1064

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
@@ -62,6 +62,7 @@ import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.coap.Token;
 import org.eclipse.californium.core.network.Exchange.KeyMID;
+import org.eclipse.californium.core.network.TokenGenerator.Scope;
 import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.core.network.config.NetworkConfigDefaults;
 import org.eclipse.californium.core.network.deduplication.Deduplicator;
@@ -77,7 +78,7 @@ public class InMemoryMessageExchangeStore implements MessageExchangeStore {
 	// for all
 	private final ConcurrentMap<KeyMID, Exchange> exchangesByMID = new ConcurrentHashMap<>();
 	// for outgoing
-	private final ConcurrentMap<Token, Exchange> exchangesByToken = new ConcurrentHashMap<>();
+	private final ConcurrentMap<KeyToken, Exchange> exchangesByToken = new ConcurrentHashMap<>();
 	private volatile boolean enableStatus;
 
 	private final NetworkConfig config;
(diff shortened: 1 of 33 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/RandomTokenGenerator.java
```diff
@@ -56,16 +58,49 @@ public RandomTokenGenerator(final NetworkConfig networkConfig) {
...
+			return Scope.SHORT_TERM_CLIENT_LOCAL;
+		}
+		int scope = token.getBytes()[0] & 0x3;
+		switch (scope) {
+		case 0:
+			return Scope.SHORT_TERM_CLIENT_LOCAL;
+		case 2:
+			return Scope.SHORT_TERM;
+		}
+		return Scope.LONG_TERM;
```
sbernard31: why not using the binary notation in the switch ?

californium-core/src/main/java/org/eclipse/californium/core/network/KeyToken.java
```diff
@@ -0,0 +1,112 @@
...
+		return peer;
+	}
+
+	@Override
+	public final int hashCode() {
+		return hash;
+	}
+
+	@Override
+	public final boolean equals(Object obj) {
```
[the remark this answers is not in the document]
boaks: Usually to block unintended extensions. Some believe also in better performance.
I changed it to use Object instead of InetSocketAddress. That enables implementers to provide their own identity object and have their hash/equals there.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -484,9 +489,9 @@ public Exchange find(final KeyMID messageId) {
 	@Override
 	public List<Exchange> findByToken(Token token) {
 		List<Exchange> result = new ArrayList<>();
-		if (token != null) {
+		if (token != null && tokenGenerator.getScope(token) == Scope.LONG_TERM) {
```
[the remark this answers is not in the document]
boaks: I changed this to

```
	/**
	 * Cancel observation for this request.
	 * 
	 * @param token
	 *            the key token for the original request which establishes the
	 *            observe relation to cancel.
	 * @param context
	 *            the endpoint context of the observe request. May be {@code null},
	 *            if the token has a none client-local scope
	 * @throws IllegalArgumentException if the token has a client-local scope and
	 *            no endpoint context is provided.
	 */
	void cancelObservation(Token token, EndpointContext context);
```

That should work with leshan. If "long-term tokens" are used, `null` works as well.

californium-core/src/main/java/org/eclipse/californium/core/network/RandomTokenGenerator.java
boaks: OK

boaks:
That's still more intended as base to discus,
May be you consider also [device/cloud-server](https://github.com/eclipse/californium/issues/1064#issuecomment-528718772) specific solution based on this.

californium-core/src/main/java/org/eclipse/californium/core/network/KeyToken.java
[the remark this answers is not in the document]
sbernard31: Personnally I really don't like this idea to avoid users to extends methods. (I prefer a warning in the javadoc, because maybe there is some case we don't think where this could make sense)

(I'm a bit traumatized because I really suffer by the past when I was coding eclipse plugin and I was force to copy/paste several classes just because someone decide that I MUST not extends or access to this class or this fields :sweat_smile: )

californium-core/src/main/java/org/eclipse/californium/core/network/RandomTokenGenerator.java
```diff
@@ -56,16 +58,52 @@ public RandomTokenGenerator(final NetworkConfig networkConfig) {
...
-			token[0] |= 0x1;
-		} else {
-			// set bit 0 to 0
-			token[0] &= 0xfe;
+			token[0] |= 0b1;
+			break;
+		case SHORT_TERM:
+			// set bit 1-0 to 0b10
+			token[0] &= 0b11111100;
+			token[0] |= 0b10;
```
sbernard31: <nitpicking>for mask, I feel it is a bit clearer to use the full notation like 0b00000010</nitpicking>

sbernard31:
So for now the idea it do not use this new `KeyToken` for `ObservationStore` ?

Some thoughts :
Imagine in Leshan we decide to use an identifier for KeyToken (instead of peer address).
Currently we have a kind of `Identity` object which is an abstraction of the peer identity.
Here we will need to create an `EndpointContext` from this `Identity`, then the `TokenGenerator` will be in charge to create an `Identity` (another or same kind of object) from `EndpointContext`. It sounds a bit strange.

Or maybe the solution would be to add the identity in the `EndpointContext` and so `TokenGenerator` use it when he find it or he try to create a new one from the orther `EndpointContext` fields ... :thinking:

californium-core/src/main/java/org/eclipse/californium/core/network/RandomTokenGenerator.java
boaks: That makes the compile time too long ;-)

boaks:
> So for now the idea it do not use this new KeyToken for ObservationStore ?

Yes. I'm still confident, that using only a "token" (LONG_TERM, cluster unique, so the key token will use null for peer) makes the synchronization in the cluster easier. Serialize the "unspecific" endpoint identifier would make it more complex.

sbernard31:
> Yes. I'm still confident, that using only a "token" (LONG_TERM, cluster unique, so the key token will use null for peer) makes the synchronization in the cluster easier. Serialize the "unspecific" endpoint identifier would make it more complex.

I would not say that we implement it but maybe just make the API in way it is possible to do it ?

boaks:
I replaced it by PR #1074 
Using the KeyMID and KeyToken consequently, changes a lot but in the end, simplifies also a lot. Introduce only parts will make it complexer in the "in between commits".

PR #1069, opened 2019-09-10
Make reliability layer parameters dynamic and configurable by endpoint

lhotari:
The usecase for this is being able to configure the reliability layer parameters in a dynamic way, based on the endpoint. The parameters covered are:
ACK_TIMEOUT
ACK_TIMEOUT_SCALE
ACK_RANDOM_FACTOR
MAX_RETRANSMIT
NSTART

This PR also add a generic extension for NetworkConfig for adding endpoint
  specific configuration handling so that it could be used for other similar purposes later.


californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
```diff
@@ -413,6 +415,8 @@ public NetworkConfig() {
 	public NetworkConfig(NetworkConfig config) {
 		this.properties = new Properties();
 		this.properties.putAll(config.properties);
+		this.observers.addAll(config.observers);
```
boaks: In the meantime, the `NetworkConfig` turns more into immutable one, so the observer are more to be removed.

californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
```diff
@@ -822,4 +826,24 @@ public NetworkConfig removeConfigObserver(NetworkConfigObserver observer) {
 		observers.remove(observer);
 		return this;
 	}
+
+	public <T> NetworkConfig addEndpointSpecificConfigProvider(Class<T> clazz, EndpointSpecificConfigProvider<T> endpointSpecificConfigProvider) {
+		configProviderMap.put(clazz, endpointSpecificConfigProvider);
```
boaks: We have some other configuration stuff, which may be added. But currently they are not endpoint specific.

boaks:
@sbernard31 

What do you think, should we consider to include this extension? At least for LwM2M it offers some possibilities to use device specific settings.

The second part question would be:
For the growing "CoapEndpointConfiguration" and "CoapStackConfiguration" should we go to move that stuff to the NetworkConfig? Or should we leave the item on their own, as they are now? Or should we consider to introduce a new "Configuration" class to collect these items but leaf the NetworkConfig asit is?

boaks:
@sbernard31 
> About NetworkConfig, I'm bit confuse about what should be in or not be in this class. Currently I see it like a kind of key/value (with simple primitive/serializable value). So my feeling is that this is not really the good place for setting this provider.

@boaks 
> For the growing "CoapEndpointConfiguration" and "CoapStackConfiguration" should we go to move that stuff to the NetworkConfig? Or should we leave the item on their own, as they are now? Or should we consider to introduce a new "Configuration" class to collect these items but leaf the NetworkConfig as it is?

If it's not added to a "Configuration", adding new stuff will causes always plenty API changes. So my feeling is, that at least for the CoapStack we should consider a "Configuration Container". If that is a extended `NetworkConfig` or a new `Configuration` which contains a `NetworkConfig` is not important for me, but always extending functions with separate parameters would not be my preferred design.

sbernard31:
> If it's not added to a "Configuration", adding new stuff will causes always plenty API changes. So my feeling is, that at least for the CoapStack we should consider a "Configuration Container"

For me it's sound not to be a big deal to add or remove parameters. This allow to clearly identify what is needed or not by an object. My concern with this kind of "container" classes is that finally we put everything on it and every classes depend on it.

In this particular case the "Configuration Container" will be created by `CoapEndpoint.Builder` then will get through:
- `CoapEndpoint`,
- `CoapStack`,
to finally arrive in `CoapLayer`.

I'm not sure this worth it to create a "container" just to get though 2 classes, :thinking: 
If you feel this is better go for it, but please in the "Leaf" class(CoapLayer in this case), do not pass the container but just the needed provider.

boaks:
The `CoapEndpoint.Builder` may create the "Container" from single build parameter.

> I'm not sure this worth it to create a "container" just to get though 2 classes

It may get more ...

> Is there any plan to do pretty much the same for BlockwiseLayer ?

boaks:
@sbernard31 

> So having a dynamic provider sounds good to me.

OK, so it's more about the details.

> Some questions :
> Is there any plan to do pretty much the same for BlockwiseLayer ?
> If yes, will we create another providers (like BlockwiseParametersProvider) ? or we will have only one provider for both config?

I pointed to that in my last comment. If it's extra, and configuration container is not used, it requires an next additional parameter.

FMPOV additional parameters are neither bad nor good. But passing that stuff through several calls just for passing it, makes me feel, that a container is better. So generally, If we agree to just extend the PerPeerConfig for future values, then we may go for an additional parameter.

Or, we extend the `NetworkConfig` with 

```
NetworkConfig getConfigForRemoteEndpoint(EndpointContext destinationEndpointContext);
```

The default returns `this`. It's possible to create a implementation, which provides a  `NetworkConfig` specific for remote endpoints. I would check a implementation, which only contains the specific values and read the others form the general `NetworkConfig`, if that approach would also be considered.

sbernard31:
Some other (maybe premature) thoughts :   
If several Layers needs to call "providers" for specific endpoint config, maybe it make sense if we call it only one time by message(request or response) processing ? (maybe this could just be achieved using cache)

> OK, so it's more about the details.

Correct. As I said : do not hesitate to use a container if you think it's better. (I have no strong opinion :))

About `getConfigForRemoteEndpoint`, it's an interesting idea (which could partially address my "premature" concerns above) but I'm bit disturbed because there some keys which doesn't make sense to be used "for specific endpoint"... (even If I suppose I can live with it :) )

boaks:
> Some other (maybe premature) thoughts :
If several Layers needs to call "providers" for specific endpoint config, maybe it make sense if we call it only one time by message(request or response) processing ? (maybe this could just be achieved using cache)

> About getConfigForRemoteEndpoint
> but I'm bit disturbed because there some keys which doesn't make sense to be used "for specific endpoint"... (even If I suppose I can live with it :) )

My basic idea would be to call the once and then just pass it in the `Stack`. I don't want to copy all the configuration values, instead I would like to use a internal implementation, which first tries the endpoint specific configuration, and if that is not available, falls back to the general one. I don't assume, that different NetworkConfig are provided for many devices, I think, there are just some classes (e.g. high-speed, narrow band, etc.).

boaks:
I did some benchmark.
If the `NetworkConfig` is used for each `Message` transmission, that will be about 2% performance penalty. This PR is better on that, because it uses `ReliabilityLayerParameters` without parsing the numbers and `Map.get`.

I don't really want to spend that for this feature to use a endpoint specific `NetworkConfig` :-).

But I got a next idea: extend the `Message`by `ReliabilityLayerParameters`.
That reduces the changes, and from other scenarios (e.g. temporary bad connections) it's sometimes useful, to change these settings for a single message.

WDYT?

sbernard31:
> I don't really want to spend that for this feature to use a endpoint specific NetworkConfig :-).

It understand that the penalty is mainly caused by the way data is store/retrieve in NetworkConfig but not the idea to have a PerEndpointConfig. So I will not throw the idea just because of that.
(passing a simple perEndpoindConfigProvider with a simple bean as config should still be an option)

> But I got a next idea: extend the Messageby ReliabilityLayerParameters.
That reduces the changes, and from other scenarios (e.g. temporary bad connections) it's sometimes useful, to change these settings for a single message.

FMPOV this is another feature.
I mean there are 3 features : 
 - being able to configure this at global level
 - being able to configure it at endpoint level
 - being able to configure it at message level.

For a cloud server, currently my feeling is that "by endpoint" is more adapted.
But on my side there is no urgency. (I mean we can add this feature later in 2.1 or 3.0 version)

So the question is more for @lhotari, as it seems to need this feature for the v2.0.
@lhotari, what is your opinion about that ? Does the message level fit your need ?

boaks:
> It understand that the penalty is mainly caused by the way data is store/retrieve in NetworkConfig but not the idea to have a PerEndpointConfig. So I will not throw the idea just because of that.

Yes, and therefore I only revoke my proposal using `NetworkConfig` (only).

> FMPOV this is another feature.
> I mean there are 3 features :
>  

The global stays, as it is.
And using the third (per message config) would obsolete the second (per endpoint config) and so reduces the changed lines a lot. You don't need to "pass it in", because the Message is already passed in. No factory, no change in NetworkConfig. That results in keeping `ReliabilityLayerParameters` (may be renamed into `TransmissionParameters`), add that to the `Message` and on usage, check, if the `Message` contains specific values or use the global values read from `NetworkConfig` on startup. 
FMPIOV, that simplifies the implementation of this extension.

sbernard31:
> And using the third (per message config) would obsolete the second (per endpoint config) 

I understand the idea. We could even say that "default value" + "the third" obsolete the first too.
But the API convenience is not the same. 

So my current feeling is that an easy way/API to configure "by endpoint" will probably better fit the "cloud server" use case. But I agree this feeling could maybe be biased by the way we use californium in Leshan.  And I understand we should be able to implement the second on top of third in Leshan. :thinking:

boaks:
> I understand the idea. We could even say that "default value" + "the third" obsolete the first too.

But the "first" is already implemented, though the "second" one is this PR.
So my feeling is:
Using the approach 3 reduces the changes in californium a lot!

I'm not sure, if it makes the "using-implementation" more complex. My feeling is, it's mostly the same, if you provide a "lookup handler" for `EndpointContext`, or you just apply the result of such a lookup to the `Message` on the "using-implementation" side.

sbernard31:
just to be clear here : ok for the per message way on my side.

boaks:
@lhotari 

Though we want to release 2.0.0-M18 as RC1, I would appreciate, if you can provide us your feedback for the alternative PR#1077 .

boaks:
Merged with commit 0435df87c57f34da38f90ae7cfea7554d79a5ed3

lhotari:
@boaks @sbernard31 I'm sorry that I didn't reply to your requests about feedback. 
Thank you for handling this PR and polishing it to fit in Californium. I appreciate your effort in doing that. 

Regarding per-message reliability parameters, that is an interesting idea. For the stuff I'm working on, it hasn't been set as a requirement since nobody probably thinks about that possibility, so I don't currently have a use case for that.

PR #1073, opened 2019-09-19
OSCORE: Finalizing functionality to securely re-generate contexts

rikard-sics:
This pull request adds the remaining functionality for the context re-derivation procedure. See the steps described in the OSCORE RFC:

This also includes what is described in section B.2.1 about generating certain values using an HMAC function rather that storing them.

Finally the first context re-derivation message should be automatically triggered by the client sending a request, not triggered separately.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
@@ -18,86 +18,415 @@ package org.eclipse.californium.oscore;
 
 import java.io.IOException;
 import java.security.SecureRandom;
+import java.util.Arrays;
 
-import org.eclipse.californium.core.CoapClient;
-import org.eclipse.californium.core.CoapResponse;
-import org.eclipse.californium.core.Utils;
-import org.eclipse.californium.core.coap.Request;
-import org.eclipse.californium.core.coap.CoAP.Code;
 import org.eclipse.californium.core.coap.CoAP.ResponseCode;
+import org.eclipse.californium.cose.CoseException;
 import org.eclipse.californium.elements.exception.ConnectorException;
 import org.eclipse.californium.elements.util.Bytes;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Class that implements methods for dynamic re-generation of OSCORE contexts.
+ * Methods for perform re-derivation of contexts as detailed in Appendix B.2. It
+ * uses two message exchanges together with varying the Context ID field in the
+ * OSCORE option to securely generate a new shared context. The second exchange
+ * will be for the request the client actually wants to send.
+ * 
+ * Note that the implementation requires that no additional requests are sent
+ * without first getting the response to the pending request during the 2
+ * request/response exchanges of the procedure.
  *
- * See https://tools.ietf.org/html/draft-ietf-core-object-security-16#appendix-B.2
- *
+ * See https://tools.ietf.org/html/rfc8613#appendix-B.2
  */
 public class ContextRederivation {
-	
+
+	private static SecureRandom random = new SecureRandom();
+
+	private static final String SCHEME = "coap://";
+
+	/**
+	 * The different phases of the re-derivation procedure.
+	 *
+	 */
+	public static enum PHASE {
+		INACTIVE, CLIENT_INITIATE, SERVER_PHASE_1, SERVER_PHASE_2, SERVER_PHASE_3, CLIENT_PHASE_1, CLIENT_PHASE_2, CLIENT_PHASE_3;
+	}
+
+	/**
+	 * Length of each segment (ID1, S2 and R3) in the Context ID when performing
+	 * context re-derivation. R2 will be of length 2 segments since it is
+	 * actually composed of S2 || HMAC(S2).
+	 */
+	protected static int SEGMENT_LENGTH = 8;
+
 	/**
 	 * The logger
 	 */
-	private static final Logger LOGGER = LoggerFactory.getLogger(RequestDecryptor.class.getName());
-	
+	private static final Logger LOGGER = LoggerFactory.getLogger(ContextRederivation.class.getName());
+
 	/**
-	 * Method for an application to indicate that the mutable parts of an OSCORE context has been lost.
-	 * In such case the context re-derivation procedure is triggered.
+	 * Method to indicate that the mutable parts of an OSCORE context has been
+	 * lost. In such case the context re-derivation procedure is triggered.
 	 * 
-	 * @param uri the URI associated with the context information has been lost for
+	 * @param uri the URI associated with context information has been lost for
 	 * @throws CoapOSException if re-generation of the context fails
 	 */
 	public static void setLostContext(OSCoreCtxDB db, String uri) throws CoapOSException
 	{
 		try {
-			rederive(db, uri);
+			initiateRequest(db, uri);
 		} catch (ConnectorException | IOException | OSException e) {
 			LOGGER.error(ErrorDescriptions.CONTEXT_REGENERATION_FAILED);
 			throw new CoapOSException(ErrorDescriptions.CONTEXT_REGENERATION_FAILED, ResponseCode.BAD_REQUEST);
 		}
 	}
-	
+
+	/* Client side related methods below */
+
 	/**
-	 * Perform re-derivation of contexts as detailed in Appendix B.2.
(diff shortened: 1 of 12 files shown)


rikard-sics:
Regarding the automation of sending the first context re-derivation message triggered by the client trying to find a message, as I mentioned previously I have a problem with the timing of these messages.

In the ContextRederivationTest it works well if I trigger the first message separately as in line XXX but it does not work if I try to trigger it automatically from the ObjectSecurityLayer as in line YYY.

boaks:
```
ContextRederivation.setLostContext(dbClient, serverUri);
```

That triggers a rederivation from the "main test thread", the coap endpoint executor is then processing the request while "sleep(1000)". 

```
ctx.setContextRederivationPhase(PHASE.INITIATE);
```

That triggers then from the "coap endpoint executor", that sleeps(1000) without doing anything, then send the original request and fast after that, the request created in rederivation is send. so two request are send for CLIENT_PHASE_1. 

If you add logging, you will see that.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
```diff
@@ -18,86 +18,376 @@
...
+			@Override
+			public void onError() {
+				LOGGER.info("Error while receiving context re-derivation message");
+			}
+		};
+
+		c.advanced(handler, r);
+
+		try {
+			Thread.sleep(1000);
```
boaks: In close to any case: that indicates mainly a threading issue.
With CoapHandler  the implementation pretends to be asyn, with sleep it get pseudo sync again.

boaks:
```
	static final ThreadFactory TEST_THREAD_FACTORY = new TestThreadFactory("TEST-");

	@Before
	public void initLogger() {
		System.out.println(System.lineSeparator() + "Start " + getClass().getSimpleName());
		EndpointManager.clear();
		CoapEndpoint coapEndpoint = new CoapEndpoint.Builder().build();
		ScheduledExecutorService scheduledThreadPool = ExecutorsUtil.newScheduledThreadPool(3, TEST_THREAD_FACTORY);
		coapEndpoint.setExecutors(scheduledThreadPool, scheduledThreadPool);
		EndpointManager.getEndpointManager().setDefaultEndpoint(coapEndpoint);
	}
```

Adjust your test with that, and you will see it works. I will try to provie a proposal to replace the usage of CoapClient with the CoapStack.

cf-oscore/src/test/java/org/eclipse/californium/oscore/ContextRederivationTest.java
```diff
@@ -124,6 +130,26 @@ public void rederivationTest() throws OSException, ConnectorException, IOExcepti
 		
 		assertEquals(resp.getCode(), ResponseCode.CONTENT);
```
boaks: The output of junit test are much easier to understand, if the parameters are provided in the expected order documented in the test-method. here: expected, actual.
if you provide actual, expected, that results in wired messages as 

> java.lang.AssertionError: expected:<4.04> but was:<2.05>

boaks:
Just don't try to send an additional request in `ContextRederivation.initiateRequest`.

Delete lines 110-136, and one more GET is added to the unit test.

Currently the `ObjectSecurityLayer` replaces the original `Request` by the prepared. That keeps that original `Request` untouched, without the MID and token assigned during send. I'm not sure, if that is really good, because it breaks my expectations and somehow the undocumented grants. If the MessageObserverAdapter.onReadyToSend copies these two values also to the original `Request` (as I would expect), the unit test must use always new `Request`s and cold not longer reuse already send `Request`s.  But that will also be the expected usage, never reuse a `Request`. 

The implementation seems to require NSTART-1 (no additional request without response to the pending request). I think that should be documented.

rikard-sics:
Thank you for the feedback to the PR. I will start going through your comments.

boaks:
Focus mainly on the last, delete lines 110-136 and the most is done.

boaks:
@rikard-sics 

currently I plan to release 2.0.0-M18 tomorrow.
Do you want to include this PR in M18?

rikard-sics:
> currently I plan to release 2.0.0-M18 tomorrow.
> Do you want to include this PR in M18?

Yes, that would be nice although I understand if it is not possible. Sorry for being slow on updating this PR, I have had a lot of other things ongoing.

If I update the PR according to your points in this thread until tomorrow morning, would that be alright?

cf-oscore/src/test/java/org/eclipse/californium/oscore/ContextRederivationTest.java
rikard-sics: Fixed in commit cab5b8639dc31aa6ec31b0747a631d127b50c993

boaks:
> If I update the PR according to your points in this thread until tomorrow morning, would that be alright?

Sure. Mainly consider to remove lines 110-136 .

rikard-sics:
> Sure. Mainly consider to remove lines 110-136 .

Yes, I am working on that right now.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
rikard-sics: Done in commit 89447884a6cbf853d0e3e34b0a73e4cad4a24f1f

rikard-sics:
> Please:
> add logging to ContextRederivation.
> My impression is, the PHASE implements a statemachine, which is drive by further requests.
> So, without the additional two GET request from the test, I'm not sure, what the state of the OSCoreCtx is intended to be.

I have tried adding more logging now, including the meaning of each state, in commit a9b9d485cb82e5dc573ac0663368e2c8eedbaf07

rikard-sics:
> The implementation seems to require NSTART-1 (no additional request without response to the pending request). I think that should be documented.

I have now added such a note as a comment in the code in commit 097bdfc6442778389c85654d9bf12e38a0370d35

rikard-sics:
Just to clarify, the reason I had the request being sent from the ContextRederivation class was so it could be triggered from the ObjectSecurityLayer "automatically" when the client wanted to send another request. The end goal we would like to achieve is that the client does not need to send a separate first request but that the first request in the re-derivation procedure will be sent when the client actually wants to send a normal request (which would be request 2 in the procedure).

So the client should only have to set `ctx.setContextRederivationPhase(PHASE.INITIATE);` and the next time it sends a request the context re-derivation procedure happens as part of that.

boaks:
> The end goal we would like to achieve is that the client does not need to send a separate first request but that the first request in the re-derivation procedure will be sent when the client actually wants to send a normal request (which would be request 2 in the procedure).

I didn't check the RFc, but the implementation just adds something to requests. I didn't see a difference in the request itself, rather it's the first or the second one.
When I was trying to test the function removing the lines, I just added a "random" additional request to the test, and it works.

So, why do you think, that the "first request" is something special according the application layer?

boaks:
Please squash your commits.

rikard-sics:
> I didn't check the RFC, but the implementation just adds something to requests. I didn't see a difference in the request itself, rather it's the first or the second one.
> When I was trying to test the function removing the lines, I just added a "random" additional request to the test, and it works.
> 
> So, why do you think, that the "first request" is something special according the application layer?
> Why? Is that "other request" something special? Or could any next request do the job?

Yes, any first request could do the job. They way I look at it is that the client wants to send request X, but for the re-derivation procedure to happen it first needs to send a first "dummy" request. So from the RFC the dummy request would be "request 1", answered with "response 1". These should just be "dummy" request/responses where the client should not request an actual resource and should get a 4.01 (Unauthorized) response.

Then request X will be sent as the second request in the re-derivation procedure (request 2/response 2). From a client application's point of view it would just like to send request X, and not know or want to first manually have to send the "dummy" request. It would be more convenient for it to just set a flag and then trigger sending of its actual desired request X, then the OSCORE/Californium code can take care of sending the first dummy request automatically.

It would send this dummy request before request X, and not even relay the response back to the client. It then sends request X and gives the response to the client, like the client sent only 1 request (although another was first sent in the background without the client needing to know).

rikard-sics:
> Please squash your commits.

Will do.

rikard-sics:
Okay, the commits have now been squashed.

boaks:
I forgot: 

and rebase and rerun the unit tests :-).

boaks:
> and rebase and rerun the unit tests :-).

I did it on my own.

rikard-sics:
> I did it on my own.

Thanks! :)

PR #1074, opened 2019-09-20
Introduce key-token to support client-local tokens.

boaks:
Replaces PR #1066 as fix for issue #1064

Move KeyMID into outer class.
Add EndpointContextResolver and use it for KeyToken and KeyMID.
Keep KeyToken and KeyMID in Exchange.
Replace response in observe relation by KeyMID.
Add test with principal based peer identity.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
@@ -52,10 +52,9 @@
  ******************************************************************************/
 package org.eclipse.californium.core.network;
 
-import java.net.InetSocketAddress;
-import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.ConcurrentModificationException;
-import java.util.Iterator;
+import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledFuture;
@@ -64,7 +63,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
-import org.eclipse.californium.core.Utils;
 import org.eclipse.californium.core.coap.BlockOption;
 import org.eclipse.californium.core.coap.CoAP.Type;
 import org.eclipse.californium.core.coap.EmptyMessage;
@@ -201,6 +199,19 @@ public class Exchange {
 	 * Mark exchange as notification.
 	 */
 	private final boolean notification;
+	/**
+	 * The key mid for the current request.
+	 */
+	private KeyMID currentKeyMID;
+	/**
+	 * The key token for the original request, if {@link #keepRequestInStore} is
+	 * {@code true}. {@code null} otherwise.
+	 */
+	private KeyToken originalKeyToken;
+	/**
+	 * The key token for the current request.
+	 */
+	private KeyToken currentKeyToken;
 
 	/**
 	 * The actual request that caused this exchange. Layers below the
@@ -255,13 +266,17 @@ public class Exchange {
 	// The relation that the target resource has established with the source
 	private volatile ObserveRelation relation;
 
+	/** The notifications that have been sent, so they can be removed from the Matcher */
+	private volatile List<KeyMID> notifications;
+
 	private final AtomicReference<EndpointContext> endpointContext = new AtomicReference<EndpointContext>();
+
 	private volatile EndpointContextOperator endpointContextPreOperator;
 
 	//If object security option is used, the Cryptographic context identifier is stored here
-    // for request/response mapping of contexts
-    private byte[] cryptoContextId;
-	
+	// for request/response mapping of contexts
+	private byte[] cryptoContextId;
+
 	/**
 	 * Creates a new exchange with the specified request and origin.
 	 * 
@@ -271,7 +286,7 @@ public class Exchange {
 	 * @throws NullPointerException, if request is {@code null}
 	 */
 	public Exchange(Request request, Origin origin, Executor executor) {
-		this(request, origin, executor, null, request != null && request.isObserve(), false);
+		this(request, origin, executor, null, false);
 	}
 
 	/**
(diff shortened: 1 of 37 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java
```diff
@@ -201,6 +198,19 @@
...
+	private KeyMID currentKeyMID;
+	/**
+	 * The key token for the original request, if {@link #keepRequestInStore} is
+	 * {@code true}. {@code null} otherwise.
+	 */
+	private KeyToken originalKeyToken;
+	/**
+	 * The key token for the current request.
+	 */
+	private KeyToken currentKeyToken;
```
boaks: The new idea is to save the keys after successfully put in the exchange itself. That simplifies the house keeping.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -387,12 +388,6 @@ public void run() {
 								LOGGER.debug("ignoring notify for pending cancel {}!", response);
 								return;
 							}
-							// we have received the expected response for the
-							// original request
-							idByMID = KeyMID.fromOutboundMessage(currentRequest);
-							if (exchangeStore.remove(idByMID, exchange) != null) {
-								LOGGER.debug("closed open request [{}]", idByMID);
-							}
```
boaks: That's removed by the exchange.complete anyway. May result only in some more
"ignoring response {}, exchange not longer matching!"

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ObserveLayer.java
```diff
@@ -119,14 +119,6 @@ public void sendResponse(final Exchange exchange, final Response response) {
 				relation.setNextControlNotification(null);
 			}
 
-			/*
-			 * The matcher must be able to find the NON notifications to remove
-			 * them from the exchangesByMID hashmap
-			 */
-			if (response.getType() == Type.NON) {
```
boaks: Now the KeyMID of the NON notify is save by the Exchange itself.

californium-core/src/main/java/org/eclipse/californium/core/observe/ObserveRelation.java
```diff
@@ -75,9 +73,6 @@
 	private long interestCheckTimer = System.currentTimeMillis();
 	private int interestCheckCounter = 1;
 
-	/** The notifications that have been sent, so they can be removed from the Matcher */
-	private ConcurrentLinkedQueue<Response> notifications = new ConcurrentLinkedQueue<Response>();
```
boaks: Moved to Exchange, which keep the KeyMIDs.

element-connector/src/main/java/org/eclipse/californium/elements/EndpointContextResolver.java
```diff
@@ -0,0 +1,38 @@
...
+ * 
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial implementation
+ ******************************************************************************/
+package org.eclipse.californium.elements;
+
+/**
+ * Interface for resolving the endpoint identity from a endpoint context.
+ */
+public interface EndpointContextResolver {
```
sbernard31: Should we choose a name which express that this class is responsible to extract an identifier ?

boaks:
> With new PrincipalEndpointContextMatcher(true), when you send a request, this is mandatory to use an endpointContext which contains enough information to recreate the identifier ?
> I think it's OK, because most of the time you MUST know to which you try to talk.
> (maybe we should document this)

My thought was, it's either a device client, then use the  IneTSocketAddress of the server. Or the server, then the Principal is available.
If that doesn't fit, then implement `Object getEndpointIdentifier(EndpointContext context)` according the situation.

californium-core/logback-test.xml
```diff
@@ -8,6 +8,10 @@
 		</encoder>
 	</appender>
 
+	<logger name="org.eclipse.californium.core.network.Exchange" level="TRACE" additivity="false">
+		<appender-ref ref="STDOUT" />
+	</logger>
+
```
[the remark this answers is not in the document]
boaks: I switch back.

element-connector/src/main/java/org/eclipse/californium/elements/PrincipalEndpointContextMatcher.java
```diff
@@ -24,14 +24,34 @@
...
 public class PrincipalEndpointContextMatcher implements EndpointContextMatcher {
 
+	private final boolean principalAsPeer;
+
 	public PrincipalEndpointContextMatcher() {
+		this(false);
+	}
+
+	public PrincipalEndpointContextMatcher(boolean serverSide) {
+		principalAsPeer = serverSide;
```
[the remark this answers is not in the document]
boaks: It works on the server side.

element-connector/src/main/java/org/eclipse/californium/elements/EndpointContextResolver.java
boaks: Make a proposal.

sbernard31:
:+1: 

My guess at device side is, even if inetSocketAddress is maybe enough, you surely know to which server you talk  anyway.

boaks:
Any proposal for renaming `EndpointContextResolver`?
Or LGTM?

element-connector/src/main/java/org/eclipse/californium/elements/PrincipalEndpointContextMatcher.java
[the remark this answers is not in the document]
sbernard31: For me this is not really about "server side". and I feel that "server side" could be confusing in coap because often peers play both role.

the boolean could be renamed by `usePrincipalAsIdentity` or something like this.

element-connector/src/main/java/org/eclipse/californium/elements/EndpointContextResolver.java
sbernard31: EndpointIdentifierExtractor ?
EndpointIdentifierFactory ?

boaks: With resolver I associated, that what ever is used to convert the endpoint context into the "identifier" could be applied. Potentially a micro-service or DNS lookup. 
An extractor would something, which takes a part, a factory, which creates something out of the available.
So for me, resolver comes closest to my idea.

sbernard31: In fact this is not really resolver which disturb me but the fact there is not `Identifier` in the class name.

boaks: EndpointIdentifierResolver?

boaks: EndpointIdentityResolver?

sbernard31: both are OK for me :)

element-connector/src/main/java/org/eclipse/californium/elements/PrincipalEndpointContextMatcher.java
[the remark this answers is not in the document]
sbernard31: (ping just in case you missed this comment)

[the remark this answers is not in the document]
boaks: I rename it.

sbernard31:
:+1:

PR #1075, opened 2019-09-23
Fix ignored block2

boaks:
Fix block2 warnings.

Add synch section to recheck blockstatus.
Use unmodifiable NotificationOrder to check for new notifies,
fix unintended updates of the orderer by new notifications.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
@@ -37,7 +37,7 @@ import org.eclipse.californium.core.coap.OptionSet;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.network.Exchange;
-import org.eclipse.californium.core.observe.ObserveNotificationOrderer;
+import org.eclipse.californium.core.observe.NotificationOrder;
 
 /**
  * A tracker for the blockwise transfer of a response body.
@@ -55,7 +55,7 @@ public final class Block2BlockwiseStatus extends BlockwiseStatus {
 	 * block of the notification and we still need to remember it, when the
 	 * last block arrives.
 	 */
-	private ObserveNotificationOrderer orderer;
+	private NotificationOrder order;
 	/**
 	 * Starting exchange to stop deprecated transfers. 
 	 */
@@ -108,7 +108,7 @@ public final class Block2BlockwiseStatus extends BlockwiseStatus {
 		Integer observeCount = block.getOptions().getObserve();
 		if (observeCount != null && OptionSet.isValidObserveOption(observeCount)) {
 			// mark this tracker with the observe no of the block it has been created for
-			status.orderer = new ObserveNotificationOrderer(observeCount);
+			status.order = new NotificationOrder(observeCount);
 			exchange.setNotificationNumber(observeCount);
 		}
 		if (block.getOptions().getETagCount() > 0) {
@@ -160,7 +160,7 @@ public final class Block2BlockwiseStatus extends BlockwiseStatus {
 	 *                      the body of a notification.
 	 */
 	public final synchronized boolean isNotification() {
-		return orderer != null;
+		return order != null;
 	}
 
 	/**
@@ -169,7 +169,7 @@ public final class Block2BlockwiseStatus extends BlockwiseStatus {
 	 * @return The value.
 	 */
 	final synchronized Integer getObserve() {
-		return orderer == null ? null : orderer.getCurrent();
+		return order == null ? null : order.getObserve();
 	}
 
 	/**
@@ -189,7 +189,7 @@ public final class Block2BlockwiseStatus extends BlockwiseStatus {
 				// none observe exchanges are cleared on sending request
 				return false;
 			}
-			return orderer == null || orderer.isNew(response);
+			return order == null || order.isNew(response);
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block2BlockwiseStatus.java
```diff
@@ -189,7 +189,7 @@ public final synchronized boolean isNew(final Response response) {
 				// none observe exchanges are cleared on sending request
 				return false;
 			}
-			return orderer == null || orderer.isNew(response);
+			return order == null || order.isNew(response);
```
boaks: Using the orderer, a new notify response updated the observe number, and that seems to cause issues.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -874,12 +889,15 @@ private void handleBlock2Response(final Exchange exchange, final Response respon
...
 
 		} else {
-
-			Block2BlockwiseStatus status = getInboundBlock2Status(key, exchange, response);
-
-			if (block2.getNum() == status.getCurrentNum() && (block2.getNum() == 0 || response.getToken().equals(exchange.getCurrentRequest().getToken()))) {
-
-				// check token to avoid mixed blockwise transfers (possible with observe) 
+			Block2BlockwiseStatus status;
+			synchronized (block2Transfers) {
```
boaks: That ensure, the the status is either the right one, cleared and renewed, or the response is ignored. it must be executed in a snyc block.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -874,12 +889,15 @@ private void handleBlock2Response(final Exchange exchange, final Response respon
...
+				status = getBlock2Status(key);
+				if (discardBlock2(key, status, exchange, response)) {
+					return;
+				}
+				status = getInboundBlock2Status(key, exchange, response);
+			}
+			if (block2.getNum() == status.getCurrentNum()) {
```
boaks: The token check seems to be a left over from far past :-).

sbernard31:
Could you explain difference between NotificationOrder and ObserveNotificationOrderer ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -925,8 +928,9 @@ private void handleBlock2Response(final Exchange exchange, final Response respon
 			} else {
 				// ERROR, wrong block number (server error)
 				// Canceling the request would interfere with Observe, so just ignore it
-				LOGGER.warn("ignoring block2 response with wrong block number {} (expected {}): {}", block2.getNum(),
-						status.getCurrentNum(), response);
+				ignoredBlock2.incrementAndGet();
```
sbernard31: I don't know if we should increment only if healthlogger is activated.

sbernard31:
Ok the new one is immutable and only serve to check against one observe number.
The old one is used all along the observation lifetime.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: There is one thing that I don't understand : 
 it seems we already checked before  `if (discarding ..` in `receiveResponse`. :thinking:

boaks: Yes, but between the first and the second check, a new notification may change the result :-). I still try to check, if a final "!status complete" check, before sending the next request is a further improvement ...

boaks:
> Could you explain difference between NotificationOrder and ObserveNotificationOrderer ?

The NotificationOrder doesn't change, if a new notify is applied, the ObserveNotificationOrderer updates the observe number to the new notify.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: Currently it*s not more triggered ... at least at my machine :-).

demo-apps/cf-extplugtest-client/benchmark.sh
```diff
@@ -32,13 +32,16 @@ CF_OPT="-d64 -XX:+UseG1GC -Xmx6g -Dcalifornium.statistic=M17"
 CF_HOST=localhost
 
 # adjust the multiplier according the speed of your CPU
-USE_TCP=0
+USE_TCP=1
```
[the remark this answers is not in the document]
boaks: Yes, now it works :-)

boaks:
Check it out and try it :-). You hopefully get no ignores anymore.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: Just to mention:
That seems to be race condition.

boaks:
Hi @mb590721,

I think, this PR is also interesting for you as "blockwise notify power user" :-). 
It removes some race-conditions for blockwise notifies.
If it's possible for you to test this PR and give us your feedback, I would appreciate that very much.

boaks:
@sbernard31 

Any objection left/overseen by me?
Or may I merge it tomorrow on Friday?

sbernard31:
Yes no objection !  :)

mb590721:
Hi @boaks 

I'm no longer actively working on the project which required Californium but I will let know the new team so that they give it a try :-).

I'm not sure to really understand what this PR aims to do, can you give me some pointers ? 

I think understand it improve tracking of the response when using the Observe mechanism on a request/response with Block2, but not sure.

boaks:
@mb590721 

Thanks for your update!

This PR fixes race-conditions, if blocks of old notifies transfers are interrupted by new notifies.
Before it result in log messages of "ignored blocks".

PR #1077, opened 2019-09-25
Make reliability layer parameters message specific - message specific approach.

boaks:
Based on Lari Hotari work in PR #1069 this implements the reliability layer parameters message specific.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/RemoteEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/RemoteEndpoint.java
@@ -22,7 +22,7 @@ import java.util.LinkedList;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
 
-import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.network.stack.ReliabilityLayerParameters;
 
 public class RemoteEndpoint {
 	
@@ -30,8 +30,10 @@ public class RemoteEndpoint {
 	private InetAddress Address;
 	// The port number of the remote endpoint
 	private int Port;
+	// Endpoint specific reliability layer parameters from the first message.
+	private final ReliabilityLayerParameters endpointReliabilityLayerParameters;
 	// A concurrent Hash Map that contains timestamp information for the exchanges
-	private ConcurrentHashMap<Exchange, exchangeInfo> exchangeInfoMap;
+	private ConcurrentHashMap<Exchange, ExchangeInfo> exchangeInfoMap;
 	
 	//Overall RTO, Strong RTO, Strong RTT, Strong RTTVAR, to be used to set the retransmission timeout.
 	private long[] overallRTO;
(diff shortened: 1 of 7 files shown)


boaks:
@lhotari 

during the discussion of your pull request #1069 I proposed to solve that function based on passing the parameters with the message. For me it simplifies it, though it doesn't require the "endpoint specific EndpointSpecificConfigProvider" stuff. OK, it mainly push that out of the scope of the stack into the application responsibility.

We will see, if you and the other committers agree on this or your original PR.

If this PR gets agreed, it would be great, if you change your PR into a PR containing only your `ReliabilityLayerParameters`. Please ensure, that the file has a license header. if you contribute as private person, then you may just copy the header of `ReliabilityLayerParameters` here in this PR. Otherwise, please replace the name with the name of your company into that license header. Thanks in advance!

californium-core/src/main/java/org/eclipse/californium/core/network/RemoteEndpoint.java
```diff
@@ -197,7 +199,11 @@ public void setEstimatorValues(long rto, long rtt, long rttvar, int estimatorTyp
 	public Queue<Exchange> getNonConfirmableQueue(){
 		return nonConfirmableQueue;
 	}
-	
+
+	public ReliabilityLayerParameters getConfigForEndpoint() {
+		return endpointReliabilityLayerParameters;
+	}
```
sbernard31: "ForEndpoint" is not needed ? and I find getConfig to general ?

Why not something like getReliabilityLayerParameters() ?

by the way the specficiation seems to call that : "Transmission Parameters"
(see https://tools.ietf.org/html/rfc7252#section-4.8 )

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayerParameters.java
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		public Builder applyConfig(NetworkConfig config) {
+			ackTimeout = config.getInt(NetworkConfig.Keys.ACK_TIMEOUT);
+			ackRandomFactor = config.getFloat(NetworkConfig.Keys.ACK_RANDOM_FACTOR);
+			ackTimeoutScale = config.getFloat(NetworkConfig.Keys.ACK_TIMEOUT_SCALE);
+			maxRetransmit = config.getInt(NetworkConfig.Keys.MAX_RETRANSMIT);
+			nstart = config.getInt(NetworkConfig.Keys.NSTART);
+			return this;
+		}
```
sbernard31: I still not like so much the idea to make this class dependent of NetworkConfig but this time there is not easy alternative :/

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayerParameters.java
```diff
@@ -0,0 +1,110 @@
...
+ *     Lari Hotari - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.californium.core.network.stack;
+
+import org.eclipse.californium.core.network.config.NetworkConfig;
+
+/**
+ * Contains the configuration values for {@link ReliabilityLayer} and {@link CongestionControlLayer}
+ */
+public class ReliabilityLayerParameters {
```
sbernard31: should we call this class TransmissionParameters ? and add a reference to this url in the comment https://tools.ietf.org/html/rfc7252#section-4.8 ?

boaks:
@lhotari 

Though we want to release 2.0.0-M18 as RC1, I would appreciate, if you can provide us your feedback.

boaks:
@sbernard31 

According emo-ip-team, we merge PR #1069 and then overwrite that by this PR.
Do you generally agree?

if so, I will go to  rename the function to getReliabilityLayerParameters()  according your comment and merge the PRs.

sbernard31:
Do you generally agree?

:+1:

boaks:
merged with commit 880fe965cac302a4e5b827c287b8f6324afd7b11

lhotari:
> Though we want to release 2.0.0-M18 as RC1, I would appreciate, if you can provide us your feedback.

Thank you @boaks . This looks great. It simplifies the solution a lot when the reliability parameters can be set at message level.

PR #1081, opened 2019-09-27
Healthcheck

boaks:
Add healthcheck to dtls.


boaks:
FMPOV, even if it was essential to narrow down the root-cause of [ACKs lost](https://github.com/eclipse/leshan/issues/744), in generally it's not too helpful and creates too much logs.

Issue #1082, opened 2019-10-01
warning :  "new request could not be registered!"


sbernard31:
(using californium-2.0.0-M16)

We are facing several raised by UdpMatcher: `new request could not be registered!`

Reading the code,  I understand that happens at "deduplicating" time when an MID+address is reused by a different peer. In that case the new request is not "registered" and deduplication will not be available for this one. Is that right ?

boaks:
```
// the new context doesn't match the previous.
exchangeStore.remove(idByMID, previous);
if (exchangeStore.findPrevious(idByMID, exchange) != null) {
	LOGGER.warn("new request could not be registered!");
}
```

with a closer look:
That `remove` affects the `exchangesByMID` but the `findPrevious` the `deduplicator`.
So the workaround of issue #837 is buggy ...
The intention was to remove the deprecated exchange and try the new one ...
I will fix it ...

boaks:
But your analysis of the current behaviour is right, as far as I understand. But not my intended one.

sbernard31:
The intended would be : remove/drop any information about the previous request ? and register the new one ?

boaks:
yes

sbernard31:
I suppose you will keep the warning, in that case please could you log related requests. (I would like to see which request is replaced by which)

boaks:
OK.

sbernard31:
I get it around 7 by hour. It's hard to me to know if this is a "normal" situation.

boaks:
OK. The probability **AFTER THE NEW FIX** for that message should be very small,

sbernard31:
I was talking about the request replacement itself not about the failure of the replacement. :)

boaks:
Then I didn't get, what you want to say/ask.
Should the exchange be not replaced? Do you have an other preference?
With the new KeyMID using the Principal as peer I would guess, that you will have less such warnings. And with the new "replace" in #1083 also. 
So I'm not sure, if there is something left, you want to know?

sbernard31:
No I think it's ok. :+1: 
I agree with the behavior expected.
And #1083 should implement it + add a log to see which request is replaced by which.

PR #1083, opened 2019-10-01
Fix deduplication on address changes.

boaks:
Fix issue #1082 .

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -219,9 +219,10 @@ public final class UdpMatcher extends BaseMatcher {
 			duplicate = endpointContextMatcher.isToBeSent(previousSourceContext, sourceContext);
 			if (!duplicate) {
 				// the new context doesn't match the previous.
-				exchangeStore.remove(idByMID, previous);
-				if (exchangeStore.findPrevious(idByMID, exchange) != null) {
-					LOGGER.warn("new request could not be registered!");
+				if (exchangeStore.replacePrevious(idByMID, previous, exchange)) {
+					LOGGER.debug("replaced request {} by new request {}!", previous.getCurrentRequest(), request);
+				} else {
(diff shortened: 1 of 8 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -219,9 +219,8 @@ public void receiveRequest(final Request request, final EndpointReceiver receive
 			duplicate = endpointContextMatcher.isToBeSent(previousSourceContext, sourceContext);
 			if (!duplicate) {
 				// the new context doesn't match the previous.
-				exchangeStore.remove(idByMID, previous);
-				if (exchangeStore.findPrevious(idByMID, exchange) != null) {
-					LOGGER.warn("new request could not be registered!");
+				if (!exchangeStore.replacePrevious(idByMID, previous, exchange)) {
+					LOGGER.warn("new request {} could not be registered!", request);
```
sbernard31: could we add a log about the successful replacement  with old and new request ?

californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/CropRotation.java
```diff
@@ -114,14 +113,20 @@ public Exchange findPrevious(KeyMID key, Exchange exchange) {
 		return prev;
 	}
 
+	@Override
+	public boolean replacePrevious(KeyMID key, Exchange previous, Exchange exchange) {
+		int s = second;
+		return maps[s].replace(key, previous, exchange) || maps[s].putIfAbsent(key, exchange) == null;
```
sbernard31: Maybe we don't care about CropRotation but I don't get why in findPrevious we write in f then in s and here we only write in s.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
boaks: OK, I'll add it.

sbernard31:
LGTM.

boaks:
Logging message added.

Issue #1085, opened 2019-10-02
Timestamp on request ?


sbernard31:
There is a get/setTimestamp  on `Message`.

I was expected that was the date where the request was sent. I wanted to use it to know if I was handling an "old" request.

But it seems this is never set.
Should we set this value ? or remove those setter/getter ? or keep it to let user to what ever they want with it (this one sounds a bit strange) ?

boaks:
After  reconsider the current implementation, I think to adapt the code again.
It's not only the token or mid, which are modified by different threads.
It's more "close to all", sent error, timeout, sent, sometimes cancel, ACK, RST.
So from that point of view, my intention to process the timestamp different is some how ridiculous. I would go to rename it back to timestamp, set it, and I think we need also some additional waits ( waitSent, waitCompleted).
WDYT?

sbernard31:
Go back to timestamp naming ? Just set it without taking account which thread modify it ?
==> no objection.

Not clear to me what is about waitSent waitCompleted and how this is related this this.

Bonus question, what will contain timestamp attibutes? the time we send or receive the message ?

boaks:
> Not clear to me what is about waitSent waitCompleted and how this is related this this.

Though someone may like to wait for the timestamp to contain a valid value, a Message may provide such additional waits.

> Bonus question, what will contain timestamp attibutes? the time we send or receive the message ?

Outgoing: just before send
Incoming: just after receive

sbernard31:
Wait would be mainly used for tests ?

OK sounds good.

boaks:
See PR #1092

PR #1086, opened 2019-10-02
Use message timestamp for receive time.

boaks:
Use it for improved deduplication by suppress duplicated request
received before the last response was sent out.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
@@ -41,6 +41,7 @@ package org.eclipse.californium.core.network.stack;
 import java.util.Random;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.eclipse.californium.core.coap.CoAP.Type;
 import org.eclipse.californium.core.coap.EmptyMessage;
@@ -69,6 +70,8 @@ public class ReliabilityLayer extends AbstractLayer {
 	private final float ack_timeout_scale;
 	private final int max_retransmit;
 
+	private final AtomicInteger counter = new AtomicInteger();
+	
 	/**
 	 * Constructs a new reliability layer. Changes to the configuration are
 	 * observed and automatically applied.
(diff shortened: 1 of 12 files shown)


boaks:
Not polished right now, but as answer to issue #1085,

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -1016,6 +1017,7 @@ public ExchangeCallback(final Exchange exchange, final Message message) {
 
 		@Override
 		public void onContextEstablished(EndpointContext context) {
+			exchange.setSendTimestamp(ClockUtil.nanoRealtime());
```
sbernard31: Any reason to set this onContextEstablished instead of onSent ?
(I didn't remember how this behave in case of blockwise transfer :thinking:)

(should we also setTimestamp directly on message in case exchange is null (see line 776))

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -210,6 +213,14 @@ public void run() {
 	public void receiveRequest(final Exchange exchange, final Request request) {
 
 		if (request.isDuplicate()) {
+			if (exchange.getSendTimestamp() > request.getTimestamp()) {
+				// received before response was sent
+				int c = counter.incrementAndGet();
```
sbernard31: what is the purpose of c ?
(You must not answer : "count" :p )

boaks:
I'm currently too busy with other stuff. I just want to show, why I didn't remove the timestamp. Please postpone the review.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
boaks: `onContextEstablished` is the last, before the message is sent. If a request is received before that, then it could be ignored, because it's ensured, that the response was not sent before.

`onSent` is the first, after the message is sent. If a request is received before that, then it's not sure, if the response was not sent before.

The difference may be very small ...

boaks: > (should we also setTimestamp directly on message in case exchange is null (see line 776))

I know, that the MID and token is set to the Request (even from an other thread). I don't like to alter too much in parallel. Therefore I used a timestamp within the exchange.

boaks:
After preparing PR #107, I found some time to address your comments.

sbernard31:
:thinking:  but now we lost the information about when we sent the request ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
sbernard31: ping ? :)

boaks:
> but now we lost the information about when we sent the request ?

Are you sure, that it is lost? Not, that it was never there (I didn't check the 1.0, maybe you?).
I'm still not sure, if it's a good idea, to write that time for outgoing `Messages` from an other thread, especially, because I currently don't see the use-case.
You have requested to remove the "timestamp" at all, so I considered to keep the "receive timestamp" gets closest to your view.

boaks:
> So maybe it's ok like this. If user want to know when the request was sent, he can use onSent callback ?

Yes, I think this would be my preferred way. Then it's up to the use-case, to catch the time before or after sending the message.

sbernard31:
> Are you sure, that it is lost? Not, that it was never there

It was in this PR !

Anyway LGTM.

boaks:
There was a term "timestamp", but it was only set for the receive time, therefore I renamed it.

PR #1088, opened 2019-10-04
Move Mac.reset() call.

boaks:
Caused NullPointerException in Android 4.4, if called without prior
init. Though the logic of such a MAC is to provide a secret and a
message, to miss the secret is not very probable. Therefore I moved the
reset to the using parts.


sbernard31:
no objection.

PR #1089, opened 2019-10-07
Fix deduplication of requests for local exchanges.

boaks:

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
@@ -84,7 +84,6 @@ public final class UdpMatcher extends BaseMatcher {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(UdpMatcher.class.getName());
 
-	// TODO: Multicast Exchanges: should not be removed from deduplicator
 	private final RemoveHandler exchangeRemoveHandler = new RemoveHandlerImpl();
 	private final EndpointContextMatcher endpointContextMatcher;
 
@@ -210,17 +209,23 @@ public final class UdpMatcher extends BaseMatcher {
 		boolean duplicate = previous != null;
 
 		if (duplicate) {
-			// assuming addresses changing, request could not only be deduplicated by their address and mid.
+			// assuming addresses changing, request could not be
+			// deduplicated only by their address and MID.
 			EndpointContext sourceContext = request.getSourceContext();
-			Request previousRequest = previous.getRequest();
-			EndpointContext previousSourceContext = previousRequest.getSourceContext();
-			// the previous response would be send with its previous context using 
-			// the current request context as connection context
+			Request previousRequest = previous.getCurrentRequest();
+			EndpointContext previousSourceContext;
+			if (previous.getOrigin() == Origin.REMOTE) {
+				previousSourceContext = previousRequest.getSourceContext();
+			} else {
+				previousSourceContext = previousRequest.getDestinationContext();
+			}
+			// the previous response would be send with its previous context
+			// using the current request context as connection context
 			duplicate = endpointContextMatcher.isToBeSent(previousSourceContext, sourceContext);
 			if (!duplicate) {
 				// the new context doesn't match the previous.
 				if (exchangeStore.replacePrevious(idByMID, previous, exchange)) {
-					LOGGER.debug("replaced request {} by new request {}!", previous.getCurrentRequest(), request);
+					LOGGER.debug("replaced request {} by new request {}!", previousRequest, request);
 				} else {
 					LOGGER.warn("new request {} could not be registered! Deduplication disabled!", request);
 				}
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfig.java
@@ -218,8 +218,13 @@ public final class NetworkConfig {
 		 */
 		public static final String MARK_AND_SWEEP_INTERVAL = "MARK_AND_SWEEP_INTERVAL";
 		public static final String DEDUPLICATOR_CROP_ROTATION = "DEDUPLICATOR_CROP_ROTATION";
+		/**
+		 * The interval after which the next crop run should occur (in
+		 * MILLISECONDS).
+		 */
 		public static final String CROP_ROTATION_PERIOD = "CROP_ROTATION_PERIOD";
 		public static final String NO_DEDUPLICATOR = "NO_DEDUPLICATOR";
+		public static final String DEDUPLICATOR_AUTO_REPLACE = "DEDUPLICATOR_AUTO_REPLACE";
 		public static final String RESPONSE_MATCHING = "RESPONSE_MATCHING";
 
 		public static final String HTTP_PORT = "HTTP_PORT";
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java
@@ -105,6 +105,14 @@ public class NetworkConfigDefaults {
 	 */
 	public static final long DEFAULT_EXCHANGE_LIFETIME = 247 * 1000;
 
+	public static final String DEFAULT_DEDUPLICATOR =  Keys.DEDUPLICATOR_MARK_AND_SWEEP;
+
+	public static final long DEFAULT_MARK_AND_SWEEP_INTERVAL =  10 * 1000; // 10 secs
+
+	public static final int DEFAULT_CROP_ROTATION_PERIOD = (int) DEFAULT_EXCHANGE_LIFETIME;
+
+	public static final boolean DEFAULT_DEDUPLICATOR_AUTO_REPLACE = true;
+
 	/**
 	 * The default DTLS response matcher.
 	 * 
@@ -222,9 +230,10 @@ public class NetworkConfigDefaults {
 		config.setInt(Keys.UDP_CONNECTOR_SEND_BUFFER, UDPConnector.UNDEFINED);
 		config.setInt(Keys.UDP_CONNECTOR_OUT_CAPACITY, Integer.MAX_VALUE); // unbounded
 
-		config.setString(Keys.DEDUPLICATOR, Keys.DEDUPLICATOR_MARK_AND_SWEEP);
-		config.setLong(Keys.MARK_AND_SWEEP_INTERVAL, 10 * 1000); // 10 secs
-		config.setInt(Keys.CROP_ROTATION_PERIOD, 2000);
+		config.setString(Keys.DEDUPLICATOR, DEFAULT_DEDUPLICATOR);
+		config.setLong(Keys.MARK_AND_SWEEP_INTERVAL, DEFAULT_MARK_AND_SWEEP_INTERVAL);
+		config.setInt(Keys.CROP_ROTATION_PERIOD, DEFAULT_CROP_ROTATION_PERIOD);
+		config.setBoolean(Keys.DEDUPLICATOR_AUTO_REPLACE, DEFAULT_DEDUPLICATOR_AUTO_REPLACE);
 		config.setString(Keys.RESPONSE_MATCHING, DEFAULT_RESPONSE_MATCHING);
 
 		config.setInt(Keys.HTTP_PORT, 8080);
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/CropRotation.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/CropRotation.java
@@ -32,6 +32,7 @@ import java.util.concurrent.TimeUnit;
 import org.eclipse.californium.core.network.Exchange;
 import org.eclipse.californium.core.network.KeyMID;
 import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.network.config.NetworkConfig.Keys;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -55,6 +56,7 @@ public class CropRotation implements Deduplicator {
 	private volatile int second;
 
 	private final long period;
+	private final boolean replace;
 	private final Rotation rotation;
 	private ScheduledExecutorService executor;
 
@@ -77,6 +79,7 @@ public class CropRotation implements Deduplicator {
 		first = 0;
 		second = 1;
 		period = config.getLong(NetworkConfig.Keys.CROP_ROTATION_PERIOD);
+		replace = config.getBoolean(Keys.DEDUPLICATOR_AUTO_REPLACE);
 	}
 
 	@Override
@@ -110,6 +113,16 @@ public class CropRotation implements Deduplicator {
 		if (prev != null || f == s)
 			return prev;
 		prev = maps[s].putIfAbsent(key, exchange);
+		if (replace && prev != null) {
+			if (prev.getOrigin() != exchange.getOrigin()) {
+				LOGGER.debug("replace exchange for {}", key);
+				if (maps[s].replace(key, prev, exchange)) {
+					prev = null;
+				} else {
+					prev = maps[s].putIfAbsent(key, exchange);
+				}
+			}
+		}
 		return prev;
 	}
 
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/SweepDeduplicator.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/deduplication/SweepDeduplicator.java
@@ -42,6 +42,7 @@ import java.util.concurrent.TimeUnit;
 import org.eclipse.californium.core.network.Exchange;
 import org.eclipse.californium.core.network.KeyMID;
 import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.network.config.NetworkConfig.Keys;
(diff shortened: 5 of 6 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -210,13 +209,19 @@ public void receiveRequest(final Request request, final EndpointReceiver receive
...
-			Request previousRequest = previous.getRequest();
-			EndpointContext previousSourceContext = previousRequest.getSourceContext();
-			// the previous response would be send with its previous context using 
-			// the current request context as connection context
-			duplicate = endpointContextMatcher.isToBeSent(previousSourceContext, sourceContext);
+			// assuming addresses changing, request could not be
+			// deduplicated only by their address and MID.
+			if (previous.getOrigin() != Origin.REMOTE) {
+				// accidently match MID of response
+				duplicate = false;
```
sbernard31: If this happens that means that device does not respect the specification ? 

Should we have this kind of "accident" for responce deduplication ?

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
```diff
@@ -210,13 +209,19 @@ public void receiveRequest(final Request request, final EndpointReceiver receive
...
 		if (duplicate) {
-			// assuming addresses changing, request could not only be deduplicated by their address and mid.
-			EndpointContext sourceContext = request.getSourceContext();
-			Request previousRequest = previous.getRequest();
-			EndpointContext previousSourceContext = previousRequest.getSourceContext();
-			// the previous response would be send with its previous context using 
-			// the current request context as connection context
-			duplicate = endpointContextMatcher.isToBeSent(previousSourceContext, sourceContext);
+			// assuming addresses changing, request could not be
+			// deduplicated only by their address and MID.
```
sbernard31: should this comment be moved in else block ?

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
boaks: No. It mean's, the device and the server use both coap-client and coap-server roles. while one device send CON notifies, a next device, assigned to that address sends request.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
boaks: Why? It says, that the request (message) maybe from a different peer than the previous message.

californium-core/src/main/java/org/eclipse/californium/core/network/UdpMatcher.java
sbernard31: I didn't understand this at all. 

> while one device send CON notifies, a next device, assigned to that address sends request.

So the test would generate an incorrect behavior (false negative) if device does not change its address but reuse an MID between a CON response(e.g. notify) and a request ? Should we rather use echange.getOrigin to know if we should check source or destination endpoint context ?

(:thinking: if peer identity

boaks: Next try:

device 1, ip1, is observed by the cloud server, and therefore sends CON notifies, which are considered to be deduplicated.

device 2, gets mapped to ip1 within the 247s since the last CON notify, and sends a request. Randomly (1/64k) it chooses the MID of the CON notify with the pending deduplication. 

Nothing is wrong, except, that the NAT reassigned the ip address before the 247s expired. That NAT's don't reassigng within that time is a requirement from LWM2M. And if every system complies to that, no special deduplication would be required. But, in my experience, it's hard to ensure that. And therefore I mitigated it in order not to fail the request.

> (thinking if peer identity is used all of this tests are obsolete, right ?)
  
Hope so, but I didn't spend too much time into verify that.

boaks: Just to mention:

that errors/collisions occurred during test over a couple of hours. So this will rarely hit in the reality. But if, this PR mitigates the effect.

sbernard31: That's what I was understanding (the second time).

What I tried to say is : 

With your PR : 
device 1 sends CON notify with a given MID.
device 1(same device, no peer address changing)  sends a request with the same MID. (this device acts wrongly)
We will not deduplicate as before as we will says this is not a duplicate.

So I asked my self, if this behavior change is correct.
If not, we should use exchange.getOrigin to know which endpoint context to use to check if peer changed.

boaks: FMPOV, RFC7252 considers only valid MID usage, means, if the device sends a different message with the same MID within the 247s, I don't see, that this is specified in a useful way. Therefore I would not care too much about misbehaving devices and instead focus on unaware hits by address changes.

boaks: Should we add a enpoint context check for that as well?
using the source context of the current response instead of the current request?
I don't think, this makes it better, but if you prefer it ...

sbernard31: My experience is : "world is full of devices which does not respect the specification correctly"

How this happened ? My understanding is some developers don't really understand specification they are using. They code then test against a server, if it works that's OK and devices are shipped.  
One days we fix a too flexible behavior at server side, and boom this breaks some devices which work only with this flexible/out of spec behavior. 

The idea : 
 - If server never allowed this devices will never be shipped with this bugs.
 - In another way, a server should tolerate bad citizen devices...  

That's a common dilemma between be strict or flexible with input...

My current position is probably that the best solution is :
 - be strict and add flexibility only when needed and in a configurable way and log when this flexibility is used.

So here, I don't like so much the idea to let devices not respect the MID usage transparently.

> Therefore I would not care too much about misbehaving devices and instead focus on unaware hits by address changes.

So I care about both :)

sbernard31: > I don't think, this makes it better, but if you prefer it ...

I prefer. You think this doesn't make it better because you think this useless ? or there is something else ?

boaks: I'm lost.

I don't understand, what you assume to be the way to process in this case. Maybe you specify that and best, test, if the current implementation before that PR matches your expectation.

sbernard31: > what you assume to be the way to process in this case

To know if a message is a duplicate, we use MID value + "identity/endpoint context".

So for example with : 

Case A)
device1 send a CON notify with MID=a
device1 send a request with MID=a

Case B)
device1 send a GET request with MID=b
device1 send a PUT request with MID=b

The 2nd message should be considered as a duplicate of the first one. (I think that was the case before we introduce the changing address/endpoint context check)

But now case A) and B) is not handled in a same way.
For A) 2nd request is no more considered as a duplicate and replace the notify.
For B) 2nd request is a duplicate.

Is it more clear ?

boaks: These are not the details I expect!

Case a)
device 1 send CON (Notify or response) with MID=a (e.g. token=abcd)
=> device 1 gets an ACK with MID=a back.

device 1 send a CON Get with MID=a (token=efgh).

What do you expect now to be sent to the device?
ACK with MID=a and let the request timeout (with NSTARt-1 that blocks for 247s).
Or send the CON notify (response?) MID=a token=abcd back to the device (maybe that's the current behavior, but I haven tested it)?
Or ???

So, what do you prefer?

sbernard31: This is what I expected : ACK with MID=a and let the request timeout.

And reading the code this is probably not what it will happen, so we can let the code like this...
(I will test it on my side just to know)

boaks: > like this ...

What does this ... mean? This PR?

boaks: And just to say:
What do you intend for your case B, PUT?
Only ACK? Piggyback response of the GET?

From my experience, starting with my issue #30, it cost you hours and hours to analyze a communication, where the message returned matches parts of your expectation and others not. And, if the MID usage is doubt, there is close to nothing you can do.

sbernard31: > What does this ... mean? This PR?

I wanted say : "this PR"
But regarding my analyze below I think this is maybe better to add the test source endpoint test. (but no strong opinion)

> What do you intend for your case B, PUT ?

Same ACK as GET.

------------------------------------------------------------
Just in case this interest you :

I create a dummy test in ObserveClientSideTest (+ adding a coap server to clientendpoint)
```java
@Test
public void testNotifyRequestSameMID() throws Exception {
	System.out.println("Observe with lost ACKs:");
	respPayload = generateRandomPayload(10);
	String path = "test";
	int obs = 100;

	Request request = createRequest(GET, path, server);
	SynchronousNotificationListener notificationListener = new SynchronousNotificationListener(request);
	client.addNotificationListener(notificationListener);
	request.setObserve();
	client.sendRequest(request);

	server.expectRequest(CON, GET, path).storeMID("A").storeToken("B").observe(0).go();
	server.sendEmpty(ACK).loadMID("A").go();
	Thread.sleep(50);
	server.sendResponse(CON, CONTENT).loadToken("B").payload(respPayload).mid(++mid).observe(++obs).go();
	Thread.sleep(10);
	server.sendRequest(CON, GET, new Token(new byte[] {1,1}), mid).go();
	
	Thread.sleep(1000);
}
```

**Without any endpoint context check at all (long time ago) :**
0081: CON [MID=52852, T=836EBDB1714EE02C], GET, /test, observe(0)    ----->
0089: <-----   ACK [MID=52852]
0147: <-----   CON [MID=8001, T=836EBDB1714EE02C], 2.05, observe(101)
0152: ACK [MID=8001]   ----->
0159: ERROR: Server sent CON-GET    MID= 8001, Token=0101, OptionSet={}, no payload
0169: RST [MID=8001]   ----->

RST is triggered by 
```
11:58:48.312 WARN [UdpMatcher]: error receiving request CON-GET    MID= 8001, Token=0101, OptionSet={}, no payload again! (o.e.c.c.n.UdpMatcher$1.run:240)
java.lang.IllegalStateException: Exchange[L1, complete] retransmit on local exchange not allowed!
	at org.eclipse.californium.core.network.Exchange.retransmitResponse(Exchange.java:705)
	at org.eclipse.californium.core.network.stack.ReliabilityLayer.receiveRequest(ReliabilityLayer.java:225)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveRequest(AbstractLayer.java:84)
	at org.eclipse.californium.core.network.stack.BaseCoapStack.receiveRequest(BaseCoapStack.java:133)
	at org.eclipse.californium.core.network.CoapEndpoint$1.receiveRequest(CoapEndpoint.java:250)
	at org.eclipse.californium.core.network.UdpMatcher$1.run(UdpMatcher.java:238)
	at org.eclipse.californium.elements.util.SerialExecutor$1.run(SerialExecutor.java:276)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:473)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:178)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:292)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1152)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:622)
```

**Just before this PR :**
0083: CON [MID=64170, T=2390F59C8B943ED8], GET, /test, observe(0)    ----->
0090: <-----   ACK [MID=64170]
0151: <-----   CON [MID=8001, T=2390F59C8B943ED8], 2.05, observe(101)
0156: ACK [MID=8001]   ----->
0163: ERROR: Server sent CON-GET    MID= 8001, Token=0101, OptionSet={}, no payload

No message sent because of : 
```
12:13:45.790 ERROR [CoapEndpoint]: exception in protocol stage thread: null (o.e.c.c.n.CoapEndpoint$7.run:1042)
java.lang.NullPointerException: null
	at org.eclipse.californium.elements.KeySetEndpointContextMatcher.internalMatch(KeySetEndpointContextMatcher.java:102)
	at org.eclipse.californium.elements.KeySetEndpointContextMatcher.isToBeSent(KeySetEndpointContextMatcher.java:98)
	at org.eclipse.californium.core.network.UdpMatcher.receiveRequest(UdpMatcher.java:219)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl.receiveRequest(CoapEndpoint.java:896)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl.receiveMessage(CoapEndpoint.java:832)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl.access$0(CoapEndpoint.java:823)
	at org.eclipse.californium.core.network.CoapEndpoint$InboxImpl$1.run(CoapEndpoint.java:811)
	at org.eclipse.californium.core.network.CoapEndpoint$7.run(CoapEndpoint.java:1040)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:473)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:178)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:292)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1152)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:622)
````

**With this PR**
0101: CON [MID=49028, T=6B26C34488253B90], GET, /test, observe(0)    ----->
0108: <-----   ACK [MID=49028]
0160: <-----   CON [MID=8001, T=6B26C34488253B90], 2.05, observe(101)
0161: ACK [MID=8001]   ----->
0171: ERROR: Server sent CON-GET    MID= 8001, Token=0101, OptionSet={}, no payload
0180: ERROR: Server received ACK-2.05   MID= 8001, Token=0101, OptionSet={"Content-Format":"text/plain", "Block2":"(szx=0/16, m=true, num=0)", "Size2":448}, "****************"

client is not "punished". we give the priority to the last message in this case.
**If we keep this behavior could we at least log with a warn that the device does not behave correctly ?**

**With this PR + source test** like this 
```java
EndpointContext sourceContext = request.getSourceContext();
Request previousRequest = previous.getCurrentRequest();
if (previous.getOrigin() == Origin.LOCAL) {
	EndpointContext previousSourceContext = previousRequest.getDestinationContext();
	// the previous response would be send with its previous context
	// using the current request context as connection context
	duplicate = endpointContextMatcher.isToBeSent(previousSourceContext, sourceContext);
} else {
	EndpointContext previousSourceContext = previousRequest.getSourceContext();
	// the previous response would be send with its previous context
	// using the current request context as connection context
	duplicate = endpointContextMatcher.isToBeSent(previousSourceContext, sourceContext);
}
```
0079: CON [MID=25149, T=1DBD9E5A0D5FFA14], GET, /test, observe(0)    ----->
0086: <-----   ACK [MID=25149]
0138: <-----   CON [MID=8001, T=1DBD9E5A0D5FFA14], 2.05, observe(101)
0140: ACK [MID=8001]   ----->
0152: ERROR: Server sent CON-GET    MID= 8001, Token=0101, OptionSet={}, no payload
0164: RST [MID=8001]   ----->
We get again the initial behavior with RST message.

boaks:
I introduce a new configuration DEDUPLICATOR_AUTO_REPLACE.
With that it's possible to switch between the behaviour preferred by you or me.

I'm not sure, it the behaviour in the past was good. FMPOV, it's not specified in RFC7252 and therefore I guess, it has change even unaware.

AFAIK, leshan overwrites some default configs, so I would prefer to have the "auto_replace" true as default for californium, and you may switch that to false for leshan.

sbernard31:
> I introduce a new configuration DEDUPLICATOR_AUTO_REPLACE.

Please do no hesitate to ask before, because as I said this is not a strong opinion and if if you asked for this I probably say : no.

> With that it's possible to switch between the behaviour preferred by you

I didn't even understand that you have a preferred behavior ^^ my understanding was you didn't care about behavior when device does not respect MID uniqueness. 
Just by curiosity why do you prefer this ?

My concern with auto_replace :
- I don't like the idea to accept transparently a bad client behavior. 
- Why just doing this when exchange.getOrigin differ... why not also doing this when message differ (e.g GET != PUT).

Anyway to be clear, my preference are : 
1. checking endpoint context even when echange.getOrigin differs
2. your solution.
3. make it configurable.

So do not hesitate to remove configurability or just let it like if you don't want to spend more time one this.

boaks:
I prefer 3.

sbernard31:
OK :)

I still don't get why you prefer autoreplace ?

boaks:
In my experience, more well clients got mystic messages, than bad clients send bad messages. Therefore, I prefer to send well messages back. But if you prefer to punish the bad devices, feel free to do so.

sbernard31:
when I say "punish", this is in fact letting know in a way that there is an issue and so giving it a chance to fix it.

PR #1090, opened 2019-10-07
Add CUSTOM_30 to support custom CoAP extension CoIoT.

boaks:
Used by openHAB shelly binding.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/CoAP.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/CoAP.java
@@ -394,7 +394,19 @@ public final class CoAP {
 		PATCH(6),
 
 		/** The IPATCH code. */
-		IPATCH(7);
+		IPATCH(7),
+
+		/** 
+		 * The custom code 30.
+		 * 
+		 * Support for openHAB custom CoAP extension, CoIoT, used for shelly binding.
+		 * <a href="https://shelly-api-docs.shelly.cloud/images/CoIoT%20for%20Shelly%20devices%20(rev%201.0)%20.pdf">CoIot Shelly</a>.
+		 * 
+		 * Note: though this code is not assigned byt IANA, it may cause future incompatibilities.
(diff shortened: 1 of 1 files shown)


californium-core/src/main/java/org/eclipse/californium/core/coap/CoAP.java
```diff
@@ -394,7 +394,19 @@ public static CodeClass valueOf(final int value) {
...
-		IPATCH(7);
+		IPATCH(7),
+
+		/** 
+		 * The custom code 30.
+		 * 
+		 * Support for openHAB custom CoAP extension, CoIoT, used for shelly binding.
+		 * <a href="https://shelly-api-docs.shelly.cloud/images/CoIoT%20for%20Shelly%20devices%20(rev%201.0)%20.pdf">CoIot Shelly</a>.
+		 * 
+		 * Note: though this code is not assigned byt IANA, it may cause future incompatibilities.
```
sbernard31: typo "byt" ?

boaks:
The complex thing is, that the deserialization is already using that.
To alter the approach more, seems for me not to pay off.
Let's see, what happens in the future :-).

sbernard31:
> To alter the approach more, seems for me not to pay off.

I understand that refactoring code to replace Enum by a Class is a bigger change.
But maybe just add all the enum is not so big and avoid the strange behavior about raising an exception for 8 to 29 and not for 30.

Anyway, I'm not strongly about that just a bit disturb :)

boaks:
I just don't want to encourage others to use none IANA code.

sbernard31:
> I just don't want to encourage others to use none IANA code.

I'm not sure to get the point as we add a none IANA code for Shelly. :thinking: 

You mean that we don't accept all code to avoid that someone who code a CoAP application with californium will not use none IANA code, but If  someone ask for it we add the support of it just for the concerned code.
That's sounds a bit strange to me, adding none or all custom code sounds more consistent.

But again It's clearly not a big deal, so do not hesitate to merge it like this. :)

PR #1092, opened 2019-10-07
Reintroduce timestamp for message.

boaks:
Add waitOnSent for message.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
@@ -216,7 +216,7 @@ public class ReliabilityLayer extends AbstractLayer {
 			if (exchange.getSendNanoTimestamp() > request.getNanoTimestamp()) {
 				// received before response was sent
 				int count = counter.incrementAndGet();
-				LOGGER.warn("{}: {} duplicate request {}, server sent response delayed, ignore request", count,
+				LOGGER.debug("{}: {} duplicate request {}, server sent response delayed, ignore request", count,
 						exchange, request);
 				return;
 			}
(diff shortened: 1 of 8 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -213,7 +213,7 @@ public void run() {
 	public void receiveRequest(final Exchange exchange, final Request request) {
 
 		if (request.isDuplicate()) {
-			if (exchange.getSendNanoTimestamp() > request.getReceiveNanoTimestamp()) {
+			if (exchange.getSendNanoTimestamp() > request.getNanoTimestamp()) {
```
sbernard31: Should we do that  for response too  ? I'm not sure

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -213,7 +213,7 @@ public void run() {
...
 				// received before response was sent
 				int count = counter.incrementAndGet();
```
sbernard31: I still not get why we count this :sweat_smile:

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
boaks: Maybe. I try it.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
boaks: Just to have a small statistic, if that pays off.

sbernard31:
LGTM.
(Just to let you know if you are using log.debug, this will be less visible for us when we will deploy it)

boaks:
> (Just to let you know if you are using log.debug, this will be less visible for us when we will deploy it)

Therefore I used it for the statistic. My understanding is, that you don't see too much benefit in it.

PR #1095, opened 2019-10-09
Add some logs in DTLSConnector about sending ALERT & refuse RENEGOCIAION

sbernard31:
This was triggered by https://github.com/eclipse/leshan/issues/754.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -1390,11 +1390,10 @@ public class DTLSConnector implements Connector, RecordLayer {
 			case CLIENT_HELLO:
 				// We do not support re-negotiation as recommended in :
 				// https://tools.ietf.org/html/rfc7925#section-17
-				if (record.getEpoch() > 0) {
-					DTLSSession session = connection.getEstablishedSession();
-					send(new AlertMessage(AlertLevel.WARNING, AlertDescription.NO_RENEGOTIATION, record.getPeerAddress()),
-							session);
-				}
+				LOGGER.debug("Reject re-negociation from peer {}", record.getPeerAddress());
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1391,9 +1391,12 @@ private void processHandshakeRecord(final Record record, final Connection connec
 				// We do not support re-negotiation as recommended in :
 				// https://tools.ietf.org/html/rfc7925#section-17
 				if (record.getEpoch() > 0) {
```
sbernard31: I asked my self if this `if (record.getEpoch() > 0) {` is really needed as some line above we check that this is not a new CLIENT_HELLO ?

boaks: > as some line above we check that this is not a new CLIENT_HELLO ?

do you mean 

```
public boolean isNewClientHello() {
	if (0 < epoch || type != ContentType.HANDSHAKE || 0 == fragmentBytes.length) {
		return false;
	}
	HandshakeType handshakeType = HandshakeType.getTypeByCode(fragmentBytes[0]);
	return handshakeType == HandshakeType.CLIENT_HELLO;
}
```

that is only for epoch 0 CLIENT_HELLOs. So my feeling is, this check seems to be not too bad.

boaks: OK, I got it. Yes, it look like it' obsolete in the meantime.

boaks:
LGTM1
Do you want to remove the `if` and then merge it before 2.0.0-M18?

sbernard31:
> Do you want to remove the if and then merge it before 2.0.0-M18?

Yep

sbernard31:
`if` is removed.

Do I merge it now ? or I let you merge this ?

PR #1096, opened 2019-10-11
Add maven assembly descriptor module for improved submodule support.

boaks:
Supports starting of  maven builds from submodule folders.


boaks:
This solution supports also sub-sub-folder.

PR #1097, opened 2019-10-11
Add dtls connect benchmark.

boaks:
Enable multiple RPK support for benchmark.


sbernard31:
no objection

Issue #1099, opened 2019-10-16
Memory leak in InMemoryConnectionStore ?


sbernard31:
Since we are deploying cf-2.0.0-M16 we notice that memory consumption increase significantly.
After few days (less than a week), heap is full and garbage collector use all the CPU trying to make some space ^^. (In our case, the instance is killed and redeployed automatically without real issue)

Doing some heap dump, I see some suspect values in `InMemoryConnectionStore`.

The `LeastRecentlyUsedCache` capacity is set to 100 000.
The size is around 99998. (That's ok.)

But looking at `connectionsByAddress` the size is 181883 and for `connectionsByEstablishedSession` the size is 460429.

I didn't try to reproduce this locally for now.

sbernard31:
Reading the code quickly my first bet would be :
When "a put in LRUCache" replaces a value,  `connectionByAddress` and `connectionByEstablishedSession` sounds not be cleaned.

But this should only happen on connectionId collision. :thinking:

sbernard31:
Ok my first bet was not so good.
It seems the issue is more simple than that this [`if (sessionCache != null)`](https://github.com/eclipse/californium/blob/2.0.0-M18/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java#L157) prevent to cleaning this map if we don't use `SessionCache`.

About my comment just above, I will investigate a bit more about.

sbernard31:
:warning: This is not directly linked to the memory leak and this is not really an issue :warning: 

Following the https://github.com/eclipse/californium/issues/1099#issuecomment-542669928 idea, I finally understand there is no way to do a `LRUCache.put()` which will do a replace.
But by curiosity, I do some math about connectionID generation and connectionStore size to see if we could have collision issue. :mag: 

With my use case connectionStore size = 100.000, the [default algorithm](https://github.com/eclipse/californium/blob/2.0.0-M18/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/InMemoryConnectionStore.java#L238) generates a 5 bytes conID.
This means that when the store is full, we get [~0,45%](https://www.wolframalpha.com/input/?i=1-e%5E%28-100000%5E2%2F%282*2%5E40%29%29) to get a conflict, which is pretty high but 
[...]

PR #1100, opened 2019-10-17
Fix Memory leak in InMemoryConnectionStore.

sbernard31:
(see #1099 for more detailed)


boaks:
Good catch :-). 
LGTM.

Issue #1101, opened 2019-10-17
getKey(identity) from PSKStore is never used ?


sbernard31:
I integrated cf-2.0.0-M18 in Leshan master and see some test failing.

This is because I have some pskStore which does not implement `PskStore.getKey(ServerNames, PskPublicInformation)`.
I don't think to much how to integrate SNI in Leshan, so for now this is deactivated and so this sounds not a problem to not implement this method.

But since cf-2.0.0-M18, the other `PskStore.getKey(PskPublicInformation)` method is never called. Only the `serverNames` version is called.

So my question 
1. should we fix this ?
 2. or remove the `PskStore.getKey(PskPublicInformation)` and explain in the javadoc how this should behave when SNI is deactivated ?

(the "issue" is probably in [`PskUtil`](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/PskUtil.java#L77))

boaks:
#1102 is merged, so the issue should be fixed.

boaks:
@sbernard31 

Can we close this?

sbernard31:
Yep

PR #1102, opened 2019-10-22
Fix 1101

boaks:
Fix issue #1101 

Cleanup naming virtualhost/servernames

Add ServerNames.newInstance(final String hostName) and remove
ServerNameExtension.forHostName(final String hostName).


sbernard31:
It sounds good.
(It's just not clear to me what is the difference between  hostname and ServerName)

boaks:
That [Server Name Indication](https://tools.ietf.org/html/rfc6066#section-3) defines a flexible, extendible way to indicate a server name, but actually only defines it for "hostname". 

ServerNames reflect that specification and may be extended in the future.
 
I also checked to remove the field "hostname" therefore from the DTLSSession, but it's used for the EndpointContext and always creating the hostname from bytes would FMPOV not be that optimal.

Issue #1103, opened 2019-10-22
 drops x bytes (A != A)


sbernard31:
(We still use cf-2.0.0-M16)
Today we I get some surprising logs during around 1 hour, where all began with : 
```
2019-10-22 16:15:07.000 drops 4 bytes, DTLS(166.141.45.246:25683,ID:5DAF0E) != DTLS(166.141.45.246:25683,ID:5DAF0E)
2019-10-22 16:15:07.000 drops 4 bytes, DTLS(166.141.45.246:25683,ID:5DAF0E) != DTLS(166.141.45.246:25683,ID:5DAF0E)
2019-10-22 16:15:01.000 drops 4 bytes, DTLS(166.155.37.51:25683,ID:5DAF0E) != DTLS(166.155.37.51:25683,ID:5DAF0E)
2019-10-22 16:15:01.000 drops 4 bytes, DTLS(166.155.37.51:25683,ID:5DAF0E) != DTLS(166.155.37.51:25683,ID:5DAF0E)
```
Bytes are drop when connection seems to be the same :thinking: 
But looking at this a bit more, ID is just the 6 first byte of sessionID and so maybe the whole sessionID is different.

We didn't see this kind of logs anymore and during this hour we faced a lot.

Here is the full logs.
It's surprising that all session ID starts with 5DAF :thinking: 
```
2019-10-22 17:00:10.000 drops 33 bytes, DTLS(54.171.103.37:32738,ID:5DAF16) != DTLS(54.171.103.37:32738,ID:5DAF18)
2019-10-22 16:58:28.000 drops 25 bytes, DTLS(69.48.214.14:25683,ID:5DAF16) != DTLS(69.48.214.14:25683,ID:5DAF19)
2019-10-22 16:58:28.000 drops 25 bytes, DTLS(69.48.214.14:25683,ID:5DAF16) != DTLS(69.48.214.14:25683,ID:5DAF19)
2019-10-22 16:58:18.000 drops 22 bytes, DTLS(122.56.47.49:36988,ID:5DAF16) != DTLS(122.56.47.49:36988,ID:5DAF18)
2019-10-22 16:58:17.000 drops 12 bytes, DTLS(54.171.103.37:32739,ID:5DAF16) != DTLS(54.171.103.37:32739,ID:5DAF18)
2019-10-22 16:58:14.000 drops 18 bytes, DTLS(223.224.72.128:32894,ID:5DAF16) != DTLS(223.224.72.128:32894,ID:5DAF18)
[...]

PR #1105, opened 2019-10-23
Fix 1103

boaks:
Fix issue #1103.

Enlarge truncated id length to 10 characters.
Create endpoint context string for logging using the matcher's relevant
information.


boaks:
I "finally" changed the logging infos.
And I changed the session id generation.

sbernard31:
:+1:

Issue #1104, opened 2019-10-22
DTLS server only by message ?


sbernard31:
Currently there is a way to configure DTLSConnector in a way it acts as a `client only` or a `server only` or `both`.

With LWM2M, I think a server might want to act as `both` for device using U binding and `server only` for device using UQ binding.

Should it make sense to add a way to configure this maybe by adding a parameter in endpoint context like for `KEY_RESUMPTION_TIMEOUT` ?

boaks:
Maybe I'm not sure, what the exact situation is.
Does scandium have a Connection with EndpointContext for such "UQ" clients? 
if yes, what triggers then the handshake? 
If there is no EndpointContext, but the Messages uses a EndpointContext with a principal set, isn't that already denied instead of triggering a handshake?
So, maybe, if you just set or not set the principal in the EndpointContext, you may already have the behavior. Or do I miss something?

sbernard31:
> Does scandium have a Connection with EndpointContext for such "UQ" clients?

It does not have a DTLS connection but it have a registration.

> if yes, what triggers then the handshake?

A request we need to send to it. 

> If there is no EndpointContext, but the Messages uses a EndpointContext with a principal set, isn't that already denied instead of triggering a handshake?

No with "both" mode, this will not deny. If there is no connection and a Endpointcontext set. We try to handshake and at the end of the handshake we check the endpointContext.

> So, maybe, if you just set or not set the principal in the EndpointContext, you may already have the behavior.

We always send the endpointcontext.

----------------------------------------------------
To be clearer, the case is : 
- device does an handshake.
- device does an update
- server starts to send it requests.
- device answers and before we finished to sent it all requests.
- it close the dtls connection (close notify) => this does not really respect the specification.
- server tries to send the next request
- no dtls connection, so server starts an handshake which will timeout. (device is sleeping)

This is 1 example where DTLS connection is lost. I suppose we can imagine other.
And so it seems to me that especially for QueueMode, this does not makes sense to try to initiate a handshake, I would prefer to fail fast.

Does it makes sense ?

boaks:
That behaviour seems to be the difference between:

[PrincipalEndpointContextMatcher](https://github.com/eclipse/californium/blob/master/element-connector/src/main/java/org/eclipse/californium/elements/PrincipalEndpointContextMatcher.java#L63)

[...]

PR #1107, opened 2019-10-24
Add handshake mode to endpoint context.

boaks:
Issue #1104
Correct some key names.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
@@ -47,12 +47,36 @@ public class DtlsEndpointContext extends MapBasedEndpointContext {
 	 * The name of the attribute that contains the timestamp of the last
 	 * handshake of the DTLS session.
 	 */
-	public static final String KEY_HANDSHAKE_TIMESTAMP = "KEY_HANDSHAKE_TIMESTAMP";
+	public static final String KEY_HANDSHAKE_TIMESTAMP = "DTLS_HANDSHAKE_TIMESTAMP";
+	/**
+	 * The name of the attribute that contains a handshake mode. Values see
+	 * {@link #HANDSHAKE_MODE_FORCE_FULL}, {@link #HANDSHAKE_MODE_FORCE}, and
+	 * {@link #HANDSHAKE_MODE_NONE}. Only considered, if endpoint is not
+	 * configured as "server only". If not provided, a handshake will be
+	 * started, if required and the connector is not configured to act as server
+	 * only. None critical attribute, not considered for matching.
+	 */
+	public static final String KEY_HANDSHAKE_MODE = KEY_PREFIX_NONE_CRITICAL + "DTLS_HANDSHAKE_MODE";
+
+	/**
+	 * Force full handshake before send this message. Doesn't start a handshake,
+	 * if the connector is configured to act as server only.
+	 */
+	public static final String HANDSHAKE_MODE_FORCE_FULL = "full";
+	/**
+	 * Force handshake before send this message. Doesn't start a handshake, if
+	 * the connector is configured to act as server only.
+	 */
+	public static final String HANDSHAKE_MODE_FORCE = "force";
+	/**
+	 * Don't start a handshake, even, if no session is available.
+	 */
+	public static final String HANDSHAKE_MODE_NONE = "none";
+
 	/**
 	 * The name of the attribute that contains a auto session resumption timeout
-	 * in milliseconds. {@code "0"}, force a session resumption, {@code ""},
-	 * disable auto session resumption. None critical attribute, not considered
-	 * for matching.
+	 * in milliseconds. {@code ""}, disable auto session resumption. None
+	 * critical attribute, not considered for matching.
 	 */
 	public static final String KEY_RESUMPTION_TIMEOUT = KEY_PREFIX_NONE_CRITICAL + "DTLS_RESUMPTION_TIMEOUT";
 
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -1825,13 +1825,13 @@ public class DTLSConnector implements Connector, RecordLayer {
 	 * {@inheritDoc}
 	 */
 	@Override
-	public final void send(final RawData msg) {
-		if (msg == null) {
+	public final void send(final RawData message) {
+		if (message == null) {
 			throw new NullPointerException("Message must not be null");
 		}
-		if (msg.isMulticast()) {
-			LOGGER.warn("DTLSConnector drops {} bytes to multicast {}:{}", msg.getSize(), msg.getAddress(), msg.getPort());
-			msg.onError(new MulticastNotSupportedException("DTLS doesn't support multicast!"));
+		if (message.isMulticast()) {
+			LOGGER.warn("DTLSConnector drops {} bytes to multicast {}:{}", message.getSize(), message.getAddress(), message.getPort());
+			message.onError(new MulticastNotSupportedException("DTLS doesn't support multicast!"));
 			return;
 		}
 		final Connection connection;
@@ -1840,23 +1840,32 @@ public class DTLSConnector implements Connector, RecordLayer {
 		if (!running.get()) {
 			connection = null;
 			error = new IllegalStateException("connector must be started before sending messages is possible");
-		} else if (msg.getSize() > MAX_PLAINTEXT_FRAGMENT_LENGTH) {
+		} else if (message.getSize() > MAX_PLAINTEXT_FRAGMENT_LENGTH) {
 			connection = null;
 			error = new IllegalArgumentException(
 					"Message data must not exceed " + MAX_PLAINTEXT_FRAGMENT_LENGTH + " bytes");
 		} else {
-			connection = getConnection(msg.getInetSocketAddress(), null, !serverOnly);
+			boolean create = !serverOnly;
+			if (create) {
+				create = !DtlsEndpointContext.HANDSHAKE_MODE_NONE
+						.equals(message.getEndpointContext().get(DtlsEndpointContext.KEY_HANDSHAKE_MODE));
+			}
+			connection = getConnection(message.getInetSocketAddress(), null, create);
 			if (connection == null) {
-				if (serverOnly) {
-					msg.onError(new EndpointUnconnectedException());
-					return;
-				} else {
+				if (create) {
 					error = new IllegalStateException("connection store is exhausted!");
+				} else {
+					if (serverOnly) {
+						message.onError(new EndpointUnconnectedException("server only, connection missing!"));
+					} else {
+						message.onError(new EndpointUnconnectedException("connection missing!"));
+					}
+					return;
 				}
 			}
 		}
 		if (error != null) {
-			msg.onError(error);
+			message.onError(error);
 			throw error;
 		}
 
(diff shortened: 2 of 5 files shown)


element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
```diff
@@ -47,7 +47,28 @@
...
+	 * {@link #HANDSHAKE_MODE_FORCE}, {@link #HANDSHAKE_MODE_START}, and
+	 * {@link #HANDSHAKE_MODE_NONE}. None critical attribute, not considered for
+	 * matching.
+	 */
+	public static final String KEY_HANDSHAKE_MODE = KEY_PREFIX_NONE_CRITICAL + "DTLS_HANDSHAKE_MODE";
+
+	/**
+	 * Force handshake before send this message.
+	 */
+	public static final String HANDSHAKE_MODE_FORCE = "force";
```
sbernard31: Is it the same as KEY_RESUMPTION_TIMEOUT with 0 value (but more intuitive?)

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
```diff
@@ -47,7 +47,28 @@
...
+	/**
+	 * Start handshake before send this message, if no session is available.
+	 */
+	public static final String HANDSHAKE_MODE_START = "start";
```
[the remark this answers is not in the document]
boaks: It's the default behaviour (which is the one without this extension). Therefore it's not explicitly checked. My idea was to list it there to have a list of the possibilities.
Or should we change this to "overwrite general server only"?

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
boaks: Yes, therefore I removed the javadoc for the 0 stuff :-).

boaks:
About 1:
At that location I'm generally afraid of unaware "race-conditions". 
Therefore I prefer to check it later.

About 2:
"FORCE_FULL_HANDSHAKE": OK, I will extend it.

sbernard31:
About 1, to be clearer,  I asked for maybe adding MODE_NONE test here to replace the one in `sendMessage`.  
I do not get why it is safe to check `serverOnly` at this place and not `MODE_NONE` ? :thinking:

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
sbernard31: Missing removing javadoc few line below in this class (l.52)

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
```diff
@@ -47,7 +47,32 @@
...
+	 * critical attribute, not considered for matching.
+	 */
+	public static final String KEY_HANDSHAKE_MODE = KEY_PREFIX_NONE_CRITICAL + "DTLS_HANDSHAKE_MODE";
+
+	/**
+	 * Force full handshake before send this message.
+	 */
+	public static final String HANDSHAKE_MODE_FORCE_FULL = "full";
+	/**
+	 * Force handshake before send this message.
```
sbernard31: should we add javadoc to explain that abbreviated handshake will be tried if a session is available ?

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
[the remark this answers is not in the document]
sbernard31: > should we change this to "overwrite general server only"?

My intuition : if you set `serverOnly`, you don't need to change that. :thinking: 

maybe we remove this and we add more javadoc at `KEY_HANDSHAKE_MODE` (few line above) ?
something like, when serverOnly is not used the default behavior is  to start handshake before send this message, if no session is available. It is possible to change this default behavior with following parameters ... ...

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1941,21 +1945,29 @@ private void sendMessage(final long nanos, final RawData message, final Connecti
...
+						timeout = null;
+					} else {
+						try {
+							timeout = Long.valueOf(contextTimeout);
+						} catch (NumberFormatException e) {
+						}
 					}
 				}
 			}
+			if (force) {
```
sbernard31: `else` instead of `if (force)` ?

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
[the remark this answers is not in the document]
boaks: My idea was, even if "serverOnly" defines the default behaviour, "start" could be used to override that default. In difference to other implementations, we have always to code.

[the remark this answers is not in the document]
sbernard31: > My idea was, even if "serverOnly" defines the default behaviour, "start" could be used to override that default.

I got it, but I still think that if you are using serverOnly, you don't want to do that.
Tweaking is only when you are using both (but it's just my vision and if you prefer to add it go for it)

boaks:
@sbernard31 

If I remember it, leshan uses the "endpoint context" keys to persist the endpoint context.
I would like to clean up  the names, therefore I added that to this PR.
If that causes too much trouble, we can also keep the old key names.

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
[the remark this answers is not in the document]
boaks: No, I don't want it. But I would like to keep the "start", even if it's not checked, because it's the default, if that value is not given.

sbernard31:
> If I remember it, leshan uses the "endpoint context" keys to persist the endpoint context.

This is just to [persist endpoint context](https://github.com/eclipse/leshan/blob/leshan-1.0.0-M12/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/EndpointContextSerDes.java#L54) of [observation](https://github.com/eclipse/leshan/blob/leshan-1.0.0-M12/leshan-server-redis/src/main/java/org/eclipse/leshan/server/redis/serialization/ObservationSerDes.java#L49).

As we don't use TLS and DTLS_HANDSHAKE_TIMESTAMP, I think it should be OK.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1932,19 +1941,23 @@ private void sendMessage(final long nanos, final RawData message, final Connecti
...
 		LOGGER.debug("Sending application layer message to [{}]", message.getEndpointContext());
 
+		String handshakeMode = message.getEndpointContext().get(DtlsEndpointContext.KEY_HANDSHAKE_MODE);
 		DTLSSession session = connection.getEstablishedSession();
 		SessionTicket ticket = connection.getSessionTicket();
 		if (session == null && ticket == null) {
-			if (serverOnly) {
-				message.onError(new EndpointUnconnectedException("server only, connection missing!"));
-				return;
-			}
```
sbernard31: This means, in certain condition we refuse some some rawData instead of deferred it ?
That was a but ?

boaks:
I added the unit tests.
And hopefully addressed "all comments" :-).

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1953,25 +1966,32 @@ private void sendMessage(final long nanos, final RawData message, final Connecti
...
+						} catch (NumberFormatException e) {
+						}
 					}
 				}
 			}
 			if (connection.isAutoResumptionRequired(timeout)) {
 				// create the session to resume from the previous one.
 				if (serverOnly) {
-					message.onError(new EndpointUnconnectedException("server only, resumption required!"));
+					message.onError(new EndpointUnconnectedException("server only, resumption requested failed!"));
```
sbernard31: I'm a bit lost here.
We are in the else block which begins line 1972.
In this block `session != null || ticket != null`.
What does it mean exactly ?
1. We can have a connection with a sesson established ?
2. Is it also possible to have a connection linked to an existing session (maybe from cache) but with no session established ?

Should we launch an handshake when HANDSHAKE_NONE is used for 1. or 2.  and autoresumtion is required?

boaks:
```
SessionTicket ticket = connection.getSessionTicket();
```
`session != null` => established
`ticket != null` => from cache

And sorry, I'm just lost by your questions.
FMPOV, I prefer to use "server only" instead of the NONE.
This is functionality, mainly requested by you. So, feel free to specify what you want.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: In my setup, I use "server only" for "servers only". The nature there is, that it sends only responses. And with that, it's very probable, that the response will be rejected by the matcher after the handshake succeeded. So for my use-cases this doesn't change.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: > This is functionality, mainly requested by you. So, feel free to specify what you want.

At short term we will use the NONE feature , but for now we don't plan to use autoresumption  neither session cache. So I just try to think what should be the API :-)

So, I try to be clearer.  
In case where we use both (server and client) mode and HANDSHAKE NONE  and we either have: 
1. a connection established and an autoresumption required,
2. or a session from cache (so no established session),

I ask myself if we should initialize a handshake. I would say no because if you say no handshake I suppose you don't want handshake at all but I'm not sure. :thinking:

boaks: FMPOV, NONE is used to inhibit a handshake, so inhibit ...

boaks: Hope, it's fixed now.

sbernard31: :thinking: 

It seems to me that if ticket != null && session == null and handshake NONE.
We will not raise : `message.onError(new EndpointUnconnectedException("connection missing!"));`
And I'm not sure what will be the behavior, because we will go in 
```java
// session with peer has already been established,
// use it to send encrypted message
sendMessage(message, connection, session);
```
(and in this case we don't have session established)

boaks: Next try ... :-)

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1947,65 +1961,80 @@ private void sendMessage(final long nanos, final RawData message, final Connecti
...
-				}
-			}
-			if (connection.isAutoResumptionRequired(timeout)) {
-				// create the session to resume from the previous one.
-				if (serverOnly) {
-					message.onError(new EndpointUnconnectedException("server only, resumption required!"));
+		} else {
+			if (none) {
+				if (connection.isResumptionRequired()) {
+					message.onError(new EndpointUnconnectedException("resumption requested failed!"));
```
sbernard31: "resumption requested failed!" sounds like if this is an error.
Maybe just "session without connection". ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1947,65 +1961,80 @@ private void sendMessage(final long nanos, final RawData message, final Connecti
 				initializeHandshaker(handshaker);
 				handshaker.startHandshake();
 			}
+			message.onConnecting();
```
sbernard31: I have no idea but is this `onConnecting` should be add in else block below ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: There are many reasons for "resumption required".
May be that's also a good message: "resumption required!" ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: This message is sent while the handshake. For me this is right.

PR #1108, opened 2019-10-25
Improve JavaDocs.

sophokles73:
Document *PSK* algorithm name to be used for key in PSK handshake

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/util/SecretUtil.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/util/SecretUtil.java
@@ -63,9 +63,10 @@ public class SecretUtil {
 	}
 
 	/**
-	 * Check, if secret key was destroyed.
+	 * Checks if a secret key has already been destroyed.
 	 * 
-	 * @param key secret key to check. If {@code null}, {@code true} is returned.
+	 * @param key secret key to check (may be {@code null}).
+	 * @return {@code true} if the key either is {@code null} or has been destroyed.
 	 */
 	public static boolean isDestroyed(SecretKey key) {
 		if (key != null) {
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/util/SecretUtil.java
```diff
@@ -79,43 +80,48 @@ public static boolean isDestroyed(SecretKey key) {
...
+	 *                  be used with a PSK based TLS handshake.
+	 * @return The key.
 	 */
 	public static SecretKey create(byte[] secret, String algorithm) {
 		return new DestroyableSecretKeySpec(secret, algorithm);
 	}
 
 	/**
-	 * Create secret key (with destroyable implementation).
+	 * Create a secret key.
```
sbernard31: Why removing the information about the key is destroyable ?

scandium-core/src/main/java/org/eclipse/californium/scandium/util/SecretUtil.java
```diff
@@ -79,43 +80,48 @@ public static boolean isDestroyed(SecretKey key) {
...
+	 *                  be used with a PSK based TLS handshake.
+	 * @return The key.
 	 */
 	public static SecretKey create(byte[] secret, int offset, int length, String algorithm) {
 		return new DestroyableSecretKeySpec(secret, offset, length, algorithm);
 	}
 
 	/**
-	 * Create a copy destroyable of the provided secret key.
+	 * Creates a copy of a secret key.
```
sbernard31: same here ?

scandium-core/src/main/java/org/eclipse/californium/scandium/util/SecretUtil.java
sophokles73: No harm intended, it's just that `SecretKey` extends `Destroyable`, so IMHO it is obvious that the returned object implements `Destroyable`. Or am I missing something?

sbernard31: `SecretKey` extends `Destroyable` since java 8 only, so as we aims java7 maybe it is worth to be clear about that.
(but I have clearly not a strong opinion on this)

sophokles73: My bad, I didn't realize that and I keep forgetting that we are still Java 7 ...
I agree that we should document this explicitly. I will change it accordingly. Thanks for spotting :+1:

sophokles73: @sbernard31 @boaks why don't we move `DestroyableSecretKeySpec` into its own class file and return that instead of returning a `SecretKey`? Then everybody knows what he gets back from the factory methods ...

sbernard31: Maybe one argument could be that generally this is better for a factory to return interface instead of class. :thinking: 

My opinion : both are OK for me !

(Not directly linked but maybe in the javadoc we can add that the key destroy() SHOULD be called after key usage)

boaks: Even if it's not implemented for now, but using e.g. android HSM support, will return a `SecretKey`and not my internal, temporary work-around "DestroyableSecretKeySpec`implementation. So I would prefer to keep the types as they are.

sophokles73: @boaks @sbernard31 Then what about an interface `DestroyableSecretKey` which extends `SecretKey` and `Destroyable` and returning that instead? The current implementation can then return `DestroyableSecretKeySpec` (which should then implement `DestroyableSecretKey`) and later, when you also support Android HSM you wrap the object that you get from Android appropriately before returning it. WDYT?

boaks: Most are not longer using java 7. That "work-around" is mainly caused by the very delayed release. For me, it's not really worth to introduce something, which the most will wonder about, because it try to fix something in an deprecated java version. I only did it, to still comply to the java 7 promise.

sophokles73:
@sbernard31 @boaks I have added a note that the returned key implements `Destroyable` ...

boaks:
LGTM.

Should I add then something about the considerations for the "algorithm"? We shortly discussed that.

sophokles73:
@boaks not sure what you mean, I believe that I have added some wording regarding the *PSK* algorithm name. Or am I missing something?

boaks:
> PSK algorithm name. Or am I missing something?

Yes, that's what I'm refer to.

sbernard31:
(LGTM)

PR #1110, opened 2019-10-28
Setup benchmark clients with defined socket buffers.

boaks:
Increase possible possible number of clients per benchmark process.


sbernard31:
(no objection)

PR #1111, opened 2019-10-29
Benchmark startup

boaks:
Speedup benchmark startup, especially for DTLS.


sbernard31:
no objections !

Issue #1113, opened 2019-10-30
How to fix the Token ?


chejdj:
version: 2.0.0-M12
problem: The token is generated by `RandomTokenGenerator` , it is random.  Are there a way to provide the request Token by me??

boaks:
2.0.0-M12 is one year old and currently we have release the 2.0.0-M18.

With 2.0.0-M18 you just set the token in the request, complying to the "scope convention" of the `TokenGenerator`. 
Using the default `RandomTokenGenerator` the first two bits of the first byte are used to encode the scope: [scoped token](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/RandomTokenGenerator.java#L75)

For application provided tokens, the [`getScope`](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/RandomTokenGenerator.java#L82)  defines the resolution. So just use a shorter token and usually that should do it.

chejdj:
Thank you so much. But I think you should make a detail document.

PR #1114, opened 2019-10-30
Add javadoc for application-defined tokens.

boaks:
Issue #1113


californium-core/src/main/java/org/eclipse/californium/core/network/RandomTokenGenerator.java
```diff
@@ -78,6 +88,9 @@ public Token createToken(Scope scope) {
 		return Token.fromProvider(token);
 	}
 
+	/**
+	 * 
+	 */
```
sbernard31: empty javadoc ? :)

PR #1115, opened 2019-10-31
Tcp update to netty 4.1.42

boaks:
Update to netty 4.1.42.
Add checks to unit test in order to prevent exceptions.


sbernard31:
no objection !

PR #1119, opened 2019-11-02
OSCORE rederivation.

boaks:
Add ObjectSecurityContextLayer mostly on top of the other layers to send
the additional dummy request.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreStack.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreStack.java
@@ -59,8 +59,9 @@ public class OSCoreStack extends BaseCoapStack {
 			reliabilityLayer = new ReliabilityLayer(config);
 		}
 
-		Layer layers[] = new Layer[] { new ExchangeCleanupLayer(config), new ObserveLayer(config), new BlockwiseLayer(config),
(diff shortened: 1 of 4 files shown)


boaks:
@rikard-sics 

Please review this, if it matches your ideas/requirement.
And please, check the unit test, if it test the right artifacts.
Is it possible to check the "changed context" itself, e.g. comparing the secrets?

rikard-sics:
Thanks alot. The PR and the JUnit test looks good to me.

I was thinking about what more can be checked in the contexts. The master secret and master salt will be the same in all the contexts generated during the procedure. The key difference should be a new ID Context which will lead to generating different sender and recipient keys. But those keys are in a sense checked due to the fact that communication using OSCORE continues to work.

Perhaps further things could be checked by knowing the ID Contexts that are exchanged and the contextRederivationKey used on the server side. To manually generate expected values and compare them to what is autmatically generated in the new contexts.

I could add some further checks in the code if you'd like but the fact that communication is working should mean that matching sender/recipient keys are generated and things are working.

boaks:
OK.
I merge it, and you may add some checks afterwards, if you want.

rikard-sics:
> OK.
> I merge it, and you may add some checks afterwards, if you want.

Okay, sounds good to me.

PR #1121, opened 2019-11-04
Add byte-optimized implementations for reading and writing int and long values

boaks:

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/util/DatagramReader.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/util/DatagramReader.java
@@ -180,26 +180,35 @@ public final class DatagramReader {
 	 * @throws IllegalArgumentException if provided numBits exceeds available bytes
 	 */
 	public long readLong(final int numBits) {
-
+		if (numBits < 0 || numBits > Long.SIZE) {
+			throw new IllegalArgumentException("bits must be in range 0 ... 64!");
+		}
 		long bits = 0; // initialize all bits to zero
 
-		for (int i = numBits - 1; i >= 0; i--) {
-
-			// check whether new byte needs to be read
-			if (currentBitIndex < 0) {
-				readCurrentByte();
+		if (currentBitIndex < 0 && (numBits & 0x7) == 0) {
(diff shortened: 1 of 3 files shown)


element-connector/src/main/java/org/eclipse/californium/elements/util/DatagramReader.java
```diff
@@ -180,26 +180,34 @@ public void close() {
...
+			throw new IllegalArgumentException("bits must be in range 0 ... 64!");
+		}
 		long bits = 0; // initialize all bits to zero
 
-		for (int i = numBits - 1; i >= 0; i--) {
-
-			// check whether new byte needs to be read
-			if (currentBitIndex < 0) {
-				readCurrentByte();
+		if (currentBitIndex < 0 && (numBits & 0x7) == 0) {
```
sbernard31: `(numBits & 0x7) == 0`, I think the idea is to ensure that numbits is a multiple of 8.
For me using modulo is more intuitive.

(but probably subjective)

element-connector/src/main/java/org/eclipse/californium/elements/util/DatagramReader.java
```diff
@@ -180,26 +180,34 @@ public void close() {
...
+			for (int i = numBits; i > 0; i -= 8) {
```
sbernard31: Why not, the classical way ?
``` java
for (int i = 0; i < numBits; i+= 8) {
}
```

element-connector/src/main/java/org/eclipse/californium/elements/util/DatagramReader.java
```diff
@@ -180,26 +180,34 @@ public void close() {
...
+				bits <<= 8;
+				bits |= readByte();
```
sbernard31: Should we add a comment just to say  this if block is optimization ?

boaks:
> Did you observe any benefits of this ?

I hoped so, but the benchmark fails to show it ;-).

boaks:
`// byte-wise, no maverick bits left`

is that comment ok? other proposal?

sbernard31:
:+1:

PR #1124, opened 2019-11-04
Add log tag to exchangestore

boaks:

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -125,6 +125,7 @@ import org.eclipse.californium.elements.UDPConnector;
 import org.eclipse.californium.elements.util.ClockUtil;
 import org.eclipse.californium.elements.util.DaemonThreadFactory;
 import org.eclipse.californium.elements.util.ExecutorsUtil;
+import org.eclipse.californium.elements.util.StringUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
```diff
@@ -1330,6 +1342,17 @@ public Builder setCoapStackFactory(CoapStackFactory coapStackFactory) {
...
 		}
 
+		/**
+		 * Set logging tag.
+		 * 
+		 * @param tag logging tag. Defautls to connector's scheme.
+		 * @return this
+		 */
+		public Builder setLoggingTag(String tag) {
+			this.tag = tag;
```
sbernard31: I'm not sure but should we normalize tag here ?

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
```diff
@@ -126,6 +133,8 @@ public InMemoryMessageExchangeStore(final NetworkConfig config, TokenGenerator t
 		this.tokenGenerator = tokenProvider;
 		this.endpointIdentityResolver = endpointResolver;
 		this.config = config;
+		this.tag = StringUtil.normalizeLoggingTag(tag);
+		LOGGER.debug("{}using default TokenProvider {}", tag, RandomTokenGenerator.class.getName());
```
sbernard31: default token  provider is maybe not used here.

californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
boaks: Without there, `null`could be applied to use the default tag.

californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java
boaks: Ok.

PR #1125, opened 2019-11-04
Coap health logger

boaks:


sbernard31:
You need this when you run "bench" tests ? and for junit ?

boaks:
> You need this when you run "bench" tests ?

Mainly for that. But lets first work on the other stuff get merged.

boaks:
If I use that library, and have the complete usage under control, it's possible to externalize this.

From inside, it's not hat easy. For debugging I usually like "simple and direct implementations", but I will check, if I can reduce the impact, if not used.

boaks:
withdrawn for PR #1130

PR #1126, opened 2019-11-04
Dtls health logger

boaks:


boaks:
I applied the "alternative design".

PR #1127, opened 2019-11-04
Add ECDHE support to benchmark.

boaks:
Add benchmark for different handshake types.


sbernard31:
no objection :)

PR #1132, opened 2019-11-05
Introduce coap health interface.

boaks:
Add retransmission parameter to message observer onSent().

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Message.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Message.java
@@ -333,25 +333,32 @@ public abstract class Message {
 	/**
 	 * Sets the 16-bit message identification.
 	 *
-	 * Reset {@link #bytes} to force new serialization.
-	 *
 	 * Provides a fluent API to chain setters.
 	 *
 	 * @param mid the new mid
 	 * @return this Message
+	 * @throws IllegalArgumentException if mid is out of range {@link #NONE} to
+	 *             {@link #MAX_MID}
+	 * @throws IllegalStateException if message is already serialized
+	 *             ({@link #setBytes(byte[]) has been called before)
 	 */
 	public Message setMID(int mid) {
 		// NONE is allowed as a temporary placeholder
 		if (mid > MAX_MID || mid < NONE) {
 			throw new IllegalArgumentException("The MID must be an unsigned 16-bit number but was " + mid);
 		}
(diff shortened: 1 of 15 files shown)


californium-core/src/main/java/org/eclipse/californium/core/coap/Message.java
```diff
@@ -438,6 +439,7 @@ public Message setToken(byte[] tokenBytes) {
 	public Message setToken(Token token) {
 		this.token = token;
 		bytes = null;
+		sent = false;
```
sbernard31: Why setting Token or MID imply `sent = false` ?

boaks: Because it resets also the bytes ...
The first commit uses `sent` as indicator for retransmission instead of initial tansmission. 
According the transmission layer, a message with a new MID is a new message and so no retransmission.
According the request/response layer, an new token is also a new request/response and so no retransmission.
Do you feel, that this causes some issues?
We may check, why these values are changed at all after serialization.

boaks: Maybe it's better, not to support to alter the MID or token after serialization?
I did a first check, and that seems to work as well (with small changes).

sbernard31: > Do you feel, that this causes some issues?

I don't know. My guess was Message is initialized with `sent = false`, then `sent` is set to `true` after once the message is sent.

So I was surprised that we need to set it to false on `setToken` or `setMID`.

> Maybe it's better, not to support to alter the MID or token after serialization?

If we could avoid it, this is probably better.

boaks: Currently only the Matcher `sendResponse` was violating this.
I changed it, but I need to run the benchmark tests.
Maybe you check, if that setMID/Token after serialization is something leshan uses ...

boaks: My tests have been successful.
Leshan unit tests also.
I squash and merge :-)

sbernard31: setMID is never used in Leshan.
setToken only to create active cancel observe request. (done before we send the request)

sbernard31:
commit 8b21206  looks good to me too.

PR #1136, opened 2019-11-08
OSCORE: update cbor to 4.0.0 and numbers to 1.4.3.

boaks:

### DIFF
--- a/cf-oscore/pom.xml
+++ b/cf-oscore/pom.xml
@@ -13,10 +13,6 @@
 	<name>Cf-OSCORE</name>
 	<description>Californium (Cf) OSCORE, Object Security for Constrained RESTful Environments</description>
 
-	<properties>
-		<numbers.version>1.0.2</numbers.version>
-	</properties>
-
 	<dependencies>
 		<dependency>
 			<groupId>${project.groupId}</groupId>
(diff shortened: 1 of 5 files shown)


boaks:
@rikard-sics
@LudwigSeitz  

Please check, if it's OK for you to update the dependencies of cbor and number.

boaks:
The lifted cose classes only depends directly from cbor, numbers is only used indirect through cbor.
This direct dependency to numbers is removed in cose in the meantime. 
And it seems that it works also for oscore to depend only indirect from numbers.
If you don't have an objection or reason to stick to a direct dependency to numbers, I would like to remove it.

rikard-sics:
> The lifted cose classes only depends directly from cbor, numbers is only used indirect through cbor.
> This direct dependency to numbers is removed in cose in the meantime.
> And it seems that it works also for oscore to depend only indirect from numbers.
> If you don't have an objection or reason to stick to a direct dependency to numbers, I would like to remove it.

Yes, that makes sense to me. I do not foresee any issues with doing that.

boaks:
Thanks! I wait for the related CQ to be resolved, then I will merge it.

sophokles73:
@boaks CQs have been approved, you may proceed with checking in the changes ...

boaks:
Thanks!

PR #1138, opened 2019-11-11
Fix exchange leak in con responses

boaks:
Minor cleanup in unit tests

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ExchangeCleanupLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ExchangeCleanupLayer.java
@@ -18,6 +18,7 @@
 package org.eclipse.californium.core.network.stack;
 
 
+import org.eclipse.californium.core.coap.CoAP.Type;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.network.Exchange;
@@ -72,9 +73,12 @@ public class ExchangeCleanupLayer extends AbstractLayer {
 	 */
 	@Override
 	public void sendResponse(final Exchange exchange, final Response response) {
-
-		if (response.isConfirmable() && !response.isNotification()) {
-			response.addMessageObserver(new CleanupMessageObserver(exchange));
+		if (!response.isNotification()) {
+			Type type = response.getType();
+			if (type == null || type == Type.CON) {
+				// if type is set later, add the cleanup preventive
+				response.addMessageObserver(new CleanupMessageObserver(exchange));
+			}
 		}
 		super.sendResponse(exchange, response);
 	}
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/ErrorInjector.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/ErrorInjector.java
@@ -17,6 +17,7 @@ package org.eclipse.californium.core.test;
 
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.eclipse.californium.TestTools;
 import org.eclipse.californium.core.coap.MessageObserverAdapter;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
@@ -34,6 +35,10 @@ public class ErrorInjector extends MessageInterceptorAdapter {
 		errorOnEstablishedContext.set(true);
 	}
 
+	/**
+	 * Cause error, when onSent is reported. The message is actually sent, but
+	 * the reporting causes errors!
+	 */
 	public void setErrorOnSent() {
 		errorOnSent.set(true);
 	}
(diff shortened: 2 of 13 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/stack/ExchangeCleanupLayer.java
```diff
@@ -72,9 +73,12 @@ public void sendRequest(final Exchange exchange, final Request request) {
...
 	@Override
 	public void sendResponse(final Exchange exchange, final Response response) {
-
-		if (response.isConfirmable() && !response.isNotification()) {
-			response.addMessageObserver(new CleanupMessageObserver(exchange));
+		if (!response.isNotification()) {
+			Type type = response.getType();
+			if (type == null || type == Type.CON) {
+				// if type is set later, add the cleanup preventive
+				response.addMessageObserver(new CleanupMessageObserver(exchange));
```
sbernard31: is there drawback to add this (and so close exchange) even on none CON response ?

californium-core/src/test/java/org/eclipse/californium/core/test/ErrorInjector.java
```diff
@@ -44,11 +49,13 @@ public void setErrorOnReadyToSend() {
 
 	@Override
 	public void sendRequest(final Request request) {
+		TestTools.removeMessageObservers(request, ErrorInjectorMessageObserver.class);
```
sbernard31: I don't why we need to do this ? :thinking: 
(To avoid to add same injector twice ?)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ExchangeCleanupLayer.java
boaks: Beside of a small performance drawback [Exchange.complete - remote origin](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/Exchange.java#L830) I can't see any malfunction. The unit tests are still green. Or do I miss something?

californium-core/src/test/java/org/eclipse/californium/core/test/ErrorInjector.java
boaks: With the fix, the CON responses are sent multiple times. And that showed irritating logs.

1. transmission => 1 log line
2. transmission => 2 log lines more (from the first and the second observer).

without a implementation error, the message will have not been retransmitted, therefore that case doesn't occur in the unit test so far.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ExchangeCleanupLayer.java
sbernard31: I think it's OK too.
Eventually, if we want a behavior closest than before maybe we can check if message is CON in CleanupMessageObserver before to complete but this means more code and not sure there is any concrete benefits.

californium-core/src/test/java/org/eclipse/californium/core/test/ErrorInjector.java
sbernard31: OK. It could also be implemented as addIfNotAlreadyContained way ?

Maybe this is even something which could be add in the API as when you add observer you rarely want to add the same twice ? :thinking: 

(It's Ok if we keep this like this too)

boaks: In this use-case, a MessageObserver replaces not the same instance but a instance of the same implementation class. I don't see, that this is a "general criteria". Also , the most application message observers are added before the message is handed over to the CoapEndoint. Therefore the "adds" are not repeated on retransmission. My feeling is, the error injector and retransmission within the unit tests are very, very special cases. It costed me some more than a hour to find out, why there are "so many duplicated" log lines while the root cause of the error wasn't fixed and the messages have been retransmitted even if the sent has failed. For that "analysis time", I thought, that fixing it for analysing future errors, is a good idea.

sbernard31: I have no objection at all about removing irritating logs !  I think its a good point.

My point was more that a `addIfNotAlreadyContained` for message observer could make sense. I already experiment case where callback are called several time but this is not a so important feature and we release an RC so, again it's ok to let it like this. (that was just an idea :-))

> In this use-case, a MessageObserver replaces not the same instance but a instance of the same implementation class.

Of course, instead of create a new instance we should change the code to use only one instance of `ErrorInjectorMessageObserver` by `ErrorInjector` and so we will be in the case where we replace the same instance which seems a "more general" use case.

boaks: The instance of `LoggingMessageObserver` is currently accessing the request/response. To reuse the instance, that must be adapted into a field. For me, remove in the `TestTools` was the solution with the fewest changes.

Issue #1139, opened 2019-11-14
About Android Demo


MonsoonSa:
Base on the **cf-android** demo，I tried to build a client to post some Json data to server ,but it seems not work well，**json data can't be posted completely to the server**，there were allways 'Block 1'，but no 'Block 2'，like this:

`coapServer.py - parseCoapData - 369 - INFO - ====================================================================================================
coapServer.py - parseCoapData - 376 - INFO - version=1 message_type=1 token_len=8 code=2==POST mid=37512
coapServer.py - parseCoapData - 386 - DEBUG - ====>To Parse Options
coapServer.py - parseCoapData - 424 - DEBUG - ====>option_num:11  option_value:b'u'b'p'b'l'b'o'b'a'b'd'
coapServer.py - parseCoapData - 440 - INFO - options:Uri-Path==upload
coapServer.py - parseCoapData - 386 - DEBUG - ====>To Parse Options
coapServer.py - parseCoapData - 424 - DEBUG - ====>option_num:12  option_value:50
coapServer.py - parseCoapData - 440 - INFO - options:Content-Type==(b'2',)
coapServer.py - parseCoapData - 386 - DEBUG - ====>To Parse Options
coapServer.py - parseCoapData - 424 - DEBUG - ====>option_num:27  option_value:13
coapServer.py - parseCoapData - 427 - INFO - options:Block1 num:0-more:1-size:512
coapServer.py - parseCoapData - 386 - DEBUG - ====>To Parse Options
coapServer.py - parseCoapData - 424 - DEBUG - ====>option_num:60  option_value:6285
coapServer.py - parseCoapData - 440 - INFO - options:Size1==(b'\x18', b'\x8d')
coapServer.py - parseCoapData - 446 - DEBUG - ====>To Parse Payload
[...]

PR #1140, opened 2019-11-14
Update android demo.

boaks:
Update to EPL 2.0.
Use client.p12 for client certificate path.


boaks:
Fix for coap endpoint initialization errors reported in issue #1139

Issue #1134, opened 2019-11-07
Release 2.0.0 - renaming "...-tcp" into "...-netty-tcp"


boaks:
I would like to rename `element-connector-tcp` into `element-connector-netty-tcp` in order to use the `element-connector-tcp` for a future "simple client implementation using plain java sockets".

Any objections or opinions?

Just to be clear:
I personally don't plan such a implementation in the next time, nor do I plan to complete the TCP function according the RFC.

sbernard31:
`element-connector-tcp-netty` ?

boaks:
Would work as well.

sophokles73:
And the plain socket implementation could then be `element-connector-tcp-plain` ...

PR #1141, opened 2019-11-14
Rename module element-connector-tcp into element-connector-tcp-netty.

boaks:
Issue #1134


sbernard31:
:heavy_check_mark:

PR #1142, opened 2019-11-18
Make receiving methods overridable.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -585,6 +585,8 @@ public class DTLSConnector implements Connector, RecordLayer {
 	 * Re-starts the connector binding to the same IP address and port as
 	 * on the previous start.
 	 * 
+	 * Note: intended for unit tests only!
+	 * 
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -585,6 +585,8 @@ public final synchronized void start() throws IOException {
 	 * Re-starts the connector binding to the same IP address and port as
 	 * on the previous start.
 	 * 
+	 * Note: intended for unit tests only!
+	 *
```
sbernard31: Why is it for unit tests only ?

boaks: It's package scope and final. So my feeling is, it's only usable for unit test.

boaks: If you want to change it, feel free.
I would like to include this change in the 2.0.0RC2 tomorrow.
Therefore I merge now this PR.

sbernard31: Uups sorry to not reply about that.

You can merge this !

PR #1144, opened 2019-11-20
Get public key only once.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
@@ -261,17 +261,16 @@ public class ServerHandshaker extends Handshaker {
 	private void receivedClientCertificate(final CertificateMessage message) throws HandshakeException {
 
 		clientCertificate = message;
+		clientPublicKey = message.getPublicKey();
 		if (clientAuthenticationRequired && message.getCertificateChain() != null
-				&& message.getPublicKey() == null) {
+				&& clientPublicKey == null) {
 			LOGGER.debug("Client authentication failed: missing certificate!");
 			AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE,
 					session.getPeer());
 			throw new HandshakeException("Client Certificate required!", alert);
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -261,17 +261,16 @@ protected void doProcessMessage(HandshakeMessage message) throws HandshakeExcept
 	private void receivedClientCertificate(final CertificateMessage message) throws HandshakeException {
 
 		clientCertificate = message;
+		clientPublicKey = message.getPublicKey();
 		if (clientAuthenticationRequired && message.getCertificateChain() != null
-				&& message.getPublicKey() == null) {
+				&& clientPublicKey == null) {
```
sbernard31: I suppose this is a kind of optimisation ? 

If it is, maybe we should go to execute `message.getPublicKey()` only `if clientAuthenticationRequired && message.getCertificateChain() != null` is `true`

```java
clientPublicKey = null; // I'm not sure this must be initialized to null
if (clientAuthenticationRequired && message.getCertificateChain() != null){
    clientPublicKey = message.getPublicKey();
    if  (clientPublicKey == null) {
```

But looking at this it seems that verifyCertificate will call message.getPublicKey() too... :thinking: 

So, maybe I'm totally wrong and the purpose of this is not "an optimization" at all ? :sweat_smile:

sbernard31: Ok I looked at #1145 and this PR is clearly not an "optimization".

So the question is what is its purpose :sweat_smile:  ?

boaks: I implemented that before the other PR.
FMPOV, calling get multiple times expecting the same value is somehow smelly.

sbernard31: for a getter this does not disturb me a lot

boaks: Depends on expecting the same value or updated values.

PR #1146, opened 2019-11-21
Verify certificates also without trusted certificates.

boaks:
Add checks for extended key usage.
Fix issue #748

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/auth/X509CertPath.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/auth/X509CertPath.java
@@ -23,14 +23,11 @@ import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
-import javax.security.auth.x500.X500Principal;
-
 import org.eclipse.californium.elements.util.Bytes;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.eclipse.californium.elements.util.CertPathUtil;
 
 /**
  * A path of X.509 certificates asserting the identity of a peer.
@@ -41,7 +38,6 @@ import org.slf4j.LoggerFactory;
  */
 public class X509CertPath extends AbstractExtensiblePrincipal<X509CertPath> {
 
-	private static final Logger LOGGER = LoggerFactory.getLogger(X509CertPath.class.getCanonicalName());
 	private static final String TYPE_X509 = "X.509";
 	private static final String ENCODING = "PkiPath";
 	private final CertPath path;
@@ -110,19 +106,19 @@ public class X509CertPath extends AbstractExtensiblePrincipal<X509CertPath> {
 	 * 
 	 * @param certificateChain chain of certificates
 	 * @return created x509 certificate path
+	 * @throws NullPointerException if provided certificateChain is {@code null}
+	 * @throws IllegalArgumentException if certificateChain is empty, or a
+	 *             certificate is provided, which is no x509 certificate.
 	 */
 	public static X509CertPath fromCertificatesChain(Certificate... certificateChain) {
-		if (certificateChain != null && certificateChain.length == 0) {
+		if (certificateChain == null) {
+			throw new NullPointerException("Certificate chain must not be null!");
+		}
+		if (certificateChain.length == 0) {
 			throw new IllegalArgumentException("Certificate chain must not be empty!");
 		}
-		List<X509Certificate> chain = new ArrayList<>();
-		for (Certificate cert : certificateChain) {
-			if (!(cert instanceof X509Certificate)) {
-				throw new IllegalArgumentException("Given certificate is not X.509! " + cert);
-			}
-			chain.add((X509Certificate)cert);
-		}
-		CertPath certPath = generateCertPath(true, chain);
+		List<X509Certificate> chain = CertPathUtil.toX509CertificatesList(Arrays.asList(certificateChain));
+		CertPath certPath = CertPathUtil.generateCertPath(chain);
 		return new X509CertPath(certPath);
 	}
 
@@ -131,12 +127,17 @@ public class X509CertPath extends AbstractExtensiblePrincipal<X509CertPath> {
 	 * 
 	 * @param certificateChain chain of x509 certificates
 	 * @return created x509 certificate path
+	 * @throws NullPointerException if provided certificateChain is {@code null}
+	 * @throws IllegalArgumentException if certificateChain is empty
 	 */
-	public static X509CertPath fromCertificatesChain(List<X509Certificate>  certificateChain) {
-		if (certificateChain != null && certificateChain.isEmpty()) {
+	public static X509CertPath fromCertificatesChain(List<X509Certificate> certificateChain) {
+		if (certificateChain == null) {
+			throw new NullPointerException("Certificate chain must not be null!");
+		}
+		if (certificateChain.isEmpty()) {
 			throw new IllegalArgumentException("Certificate chain must not be empty!");
 		}
-		CertPath certPath = generateCertPath(true, certificateChain);
+		CertPath certPath = CertPathUtil.generateCertPath(certificateChain);
 		return new X509CertPath(certPath);
 	}
 
@@ -147,49 +148,17 @@ public class X509CertPath extends AbstractExtensiblePrincipal<X509CertPath> {
 	 *            path, {@code false} otherwise.
 	 * @param certificateChain list with chain of x509 certificates
 	 * @return created certificate path
+	 * @deprecated use {@link CertPathUtil}
 	 */
+	@Deprecated
 	public static CertPath generateCertPath(boolean includeRoot, List<X509Certificate> certificateChain) {
 		if (certificateChain == null) {
 			throw new NullPointerException("Certificate chain must not be null!");
 		}
-		List<X509Certificate> certificates = new ArrayList<>();
-		X500Principal issuer = null;
-		try {
-			int index = 0;
-			CertificateFactory factory = CertificateFactory.getInstance(TYPE_X509);
-			for (Certificate cert : certificateChain) {
-				if (!(cert instanceof X509Certificate)) {
-					throw new IllegalArgumentException("Given certificate is not X.509! " + cert);
-				}
-				X509Certificate xcert = (X509Certificate) cert;
-				LOGGER.debug("Current Subject DN: {}", xcert.getSubjectX500Principal().getName());
-				if (issuer != null && !issuer.equals(xcert.getSubjectX500Principal())) {
-					LOGGER.debug("Actual Issuer DN: {}", xcert.getSubjectX500Principal().getName());
-					throw new IllegalArgumentException("Given certificates do not form a chain");
-				}
-				++index;
-				if (xcert.getIssuerX500Principal().equals(xcert.getSubjectX500Principal())) {
-					// a self-signed root certificate
-					if (index != certificateChain.size()) {
-						throw new IllegalArgumentException(
-								"Given certificates do not form a chain, root is not the last!");
-					}
-					if (includeRoot || certificateChain.size() == 1) {
-						certificates.add(xcert);
-					}
-				} else {
-					// not a self-signed certificate
-					certificates.add(xcert);
-					issuer = xcert.getIssuerX500Principal();
-					LOGGER.debug("Expected Issuer DN: {}", issuer.getName());
-				}
-			}
-			return factory.generateCertPath(certificates);
-		} catch (CertificateException e) {
-			// should not happen because all Java 7 implementation MUST
-			// support
-			// X.509 certificates
-			throw new IllegalArgumentException("could not create X.509 certificate factory", e);
+		if (includeRoot) {
+			return CertPathUtil.generateCertPath(certificateChain);
+		} else {
+			return CertPathUtil.generateValidatableCertPath(certificateChain, null);
 		}
 	}
 
--- a/element-connector/src/main/java/org/eclipse/californium/elements/util/CertPathUtil.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/util/CertPathUtil.java
@@ -1,0 +1,450 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Bosch Software Innovations GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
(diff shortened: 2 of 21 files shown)


boaks:
This introduces to check the key usage of the certificates, when configured and when received from the other side.
The default `StaticCertificateVerifier` is extended to verify a certificate path even with an "empty trust store" (trust all) using the root-certificate of the path as trust. That is intended to ensure, that the other information in the certificate path is validated.

boaks:
I don't insist, that this is part of the 2.0.0 release.

boaks:
If it's not part of the 2.0.0-release, adjusting the javadoc of the "trust all" function should be considered.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1299,11 +1300,22 @@ protected final void expectChangeCipherSpecMessage() {
...
 		if (message.getCertificateChain() != null) {
+			List<? extends Certificate> certificates = message.getCertificateChain().getCertificates();
+			if (!certificates.isEmpty()) {
+				Certificate certificate = certificates.get(0);
+				if (certificate instanceof X509Certificate) {
+					if (!X509Util.isKeyUsage((X509Certificate) certificate, client)) {
+
+					}
+				}
+			}
```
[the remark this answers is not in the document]
sbernard31: would the idea to add `boolean client` parameter to  `CertificateVerifier`API ?

sbernard31:
> I don't insist, that this is part of the 2.0.0 release.

Generally between 2 RCs we only do bug fixes. I think this is a good rules but on the other hand  this features seems to provides new API meaning we will need to release a 2.1 version and we will need to keep the old one for backward compatibility. (or post-pone to the 3.0)

So, I have no strong opinion but I would rather go to integrate it.

(Do you have other issue like this in mind ? or AFAYK this is the last one)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/X509Util.java
```diff
@@ -0,0 +1,186 @@
...
+			List<? extends Certificate> list = certPath.getCertificates();
+			if (list.isEmpty()) {
+				throw new CertPathValidatorException("empty certificate chain!");
+			}
+			int last = list.size() - 1;
+			Certificate root = list.get(last);
+			// don't fill the original one.
+			trustAnchors = new HashSet<TrustAnchor>();
+			if (root instanceof X509Certificate) {
+				trustAnchors.add(new TrustAnchor((X509Certificate) root, null));
```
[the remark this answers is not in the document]
boaks: OK

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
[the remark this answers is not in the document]
boaks: A `HandshakeException` is missing.
I didn't want to change the `CertificateVerifierAPI`, therefore I implemented it in the handshaker.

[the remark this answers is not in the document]
sbernard31: Is it not more flexible to put this in the CertificateVerifierAPI ?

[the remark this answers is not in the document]
sbernard31: Or we keep it like you do and we see if more flexibility is needed later. 
If this is needed we will create a new Interface and in code we call either the old or new API depending of which interface is used by the user.

element-connector/src/main/java/org/eclipse/californium/elements/auth/X509CertPath.java
```diff
@@ -146,49 +146,95 @@ public static X509CertPath fromCertificatesChain(List<X509Certificate>  certific
 	 * @param includeRoot {@code true}, to include the root certificate in the
 	 *            path, {@code false} otherwise.
 	 * @param certificateChain list with chain of x509 certificates
-	 * @return created certificate path
+	 * @return generated certificate path
+	 * @throws NullPointerException if provided certificateChain is {@code null}
+	 * @throws IllegalArgumentException if provided certificateChain is empty
```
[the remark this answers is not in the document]
boaks: OK

boaks:
Not verifying the certificates of the path for "trust all" and not check the key usage may be considered as bug.

> (Do you have other issue like this in mind ? or AFAYK this is the last one)

Only the async credentials API, but that will only be implemented, if for some legal stuff the 2.0.0 release must be shifted to next year :-).

boaks:
I did some more tests and (openssl) interoperability tests. My results seems to be, that you may be able to configure the certificates well, but it's also too easy to break it. I wasn't able to fully mitigate the effects and I'm not sure, when I can spend the time to complete my work. Maybe today, maybe next week or alter. 
FMPOV, include it in 2.0.0 will therefore depend on the "final solution" and when this gets available.

boaks:
After reading a lot, I think, that the definition for certificate path validation does not forbid to truncate a certificate chain before passing it to the validation. The new implementation is now able to truncate the path on sending and receiving. Currently I implemented it configurable. Maybe we agree, that being able to use the new `AdvancedCertificateVerifier` obsoletes the other configuration flags.

boaks:
Just to mention, I merged the `CertPathGenerator` and the `X509Util` back into one class to support the (now @Deprecated) `X509CertPath.generateCertPath`.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -211,22 +215,33 @@
 	protected ReassemblingHandshakeMessage reassembledMessage;
 
 	/** The handshaker's private key. */
-	protected PrivateKey privateKey;
+	protected final PrivateKey privateKey;
 
 	/** The handshaker's public key. */
-	protected PublicKey publicKey;
+	protected final PublicKey publicKey;
```
sbernard31: If we want to keep api compatibility I think that we are not allowed to change this to final :thinking:

(This is where tooling could help)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1303,28 +1320,51 @@ protected final void expectChangeCipherSpecMessage() {
...
+
+			List<? extends Certificate> certificates = certPath.getCertificates();
+			if (certificates.isEmpty()) {
+				if (isClient) {
+					LOGGER.debug("Certificate validation failed: empty server certificate!");
+					AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.BAD_CERTIFICATE,
+							session.getPeer());
+					throw new HandshakeException("Empty server certificate!", alert);
+				}
+			}
```
sbernard31: This new empty check should not be moved in AdvancedCertificateVerifier ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
```diff
@@ -544,34 +541,6 @@ SignatureAndHashAlgorithm getSupportedSignatureAlgorithm(PublicKey key) {
...
 
-	/**
-	 * Truncates a given certificate chain at the first certificate encountered having
-	 * a subject listed in <em>certificateAuthorities</em>.
-	 * 
-	 * @param chain The original certificate chain.
-	 * @return A (potentially) truncated copy of the original chain.
-	 * @throws NullPointerException if the given chain is {@code null}.
-	 */
-	public List<X509Certificate> removeTrustedCertificates(List<X509Certificate> chain) {
```
sbernard31: [API break] should we just deprecated it ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
```diff
@@ -388,18 +374,16 @@ private boolean addCerticiateAuthorities(List<X500Principal> authorities) {
 	 *         maximum encoded length allowed for the certificate request message's
 	 *         certificate authorities vector (2^16 - 1 bytes).
 	 */
-	public boolean addCertificateAuthorities(X509Certificate[] trustedCas) {
```
sbernard31: [API break] should we also keep the old version ?

element-connector/src/main/java/org/eclipse/californium/elements/util/CertPathUtil.java
```diff
@@ -0,0 +1,447 @@
...
+ * by one of the listed CAs."
+ * </p>
+ * 
+ * <a href="https://tools.ietf.org/html/rfc5280#section-6">RFC5280, Section 6,
+ * Certification Path Validation</a>
+ * <p>
+ * "Valid paths begin with certificates issued by a trust anchor." ... "The
+ * procedure performed to obtain this sequence of certificates is outside the
+ * scope of this specification".
+ * </p>
```
sbernard31: :+1: !

element-connector/src/main/java/org/eclipse/californium/elements/util/CertPathUtil.java
```diff
@@ -0,0 +1,447 @@
...
+	 */
+	public static CertPath generateCertPath(List<X509Certificate> certificateChain) {
+		if (certificateChain == null) {
+			throw new NullPointerException("Certificate chain must not be null!");
+		}
+		return generateCertPath(certificateChain, certificateChain.size());
+	}
+
+	/**
+	 * Create certificate path from x509 certificates chain upto the provided
```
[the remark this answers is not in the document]
boaks: OK

element-connector/src/main/java/org/eclipse/californium/elements/util/CertPathUtil.java
```diff
@@ -0,0 +1,447 @@
...
+	/**
+	 * Create list of subject from certificates.
+	 * 
+	 * @param certificates list of certificates. Maybe {@code null}.
+	 * @return list of subjects of provided certificates. maybe empty, if
+	 *         provided list was empty or {@code null}.
+	 */
+	public static List<X500Principal> toSubjects(List<X509Certificate> certificates) {
+		List<X500Principal> subjects = new ArrayList<X500Principal>(certificates.size());
+		if (certificates != null) {
```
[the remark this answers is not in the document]
boaks: :-), good catch

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -803,6 +817,42 @@ public Boolean useHandshakeStateValidation() {
 		return useHandshakeStateValidation;
 	}
 
+	/**
+	 * Use key usage verification for x509.
+	 * 
+	 * @return {@code true}, enable check of key usage (x509 extension)
```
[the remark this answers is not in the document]

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -2057,6 +2154,15 @@ public DtlsConnectorConfig build() {
...
 			}
+			if (config.useTruncatedCertificatePathForClientsCertificateMessage == null) {
+				config.useTruncatedCertificatePathForClientsCertificateMessage = Boolean.TRUE;
+			}
+			if (config.useTruncatedCertificatePathForValidation == null) {
+				config.useTruncatedCertificatePathForValidation = Boolean.TRUE;
+			}
+			if (config.useKeyUsageVerification == null) {
+				config.useKeyUsageVerification = Boolean.TRUE;
+			}
```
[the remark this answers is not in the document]
boaks: Half true. Though the tests are more complete now (not only client handshaker on own certificate), the true enables now more checking. For me this is intended, and if someone doesn't want the key-usage check, it may be disabled at all.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: If all classes and all functions are considered to be API, then it will get hard to change something at all. 
"Protected" limit the scope to sub-classes and I don't see, that the API offers a possibility to inject a custom handshaker from outside. Or do I miss something?

> (This is where tooling could help 😅 )

If you consider everything to be the API, then maybe. Otherwise it will be the configuration of that tool, which maybe difficult. I saw your proposal on issue #1159, just create a PR to check, how it works.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: If your server certificate is empty, I would guess, your handshake will then fail anyway later. Otherwise scandium must explicitly support anonymous server.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
boaks: FMPOV, that's internal API and not considered to be not adapted.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
boaks: FMPOV, that's internal API and not considered to be not adapted.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sbernard31: Maybe we should clarify API question before.

boaks: :-).

You already started a issue. Maybe you provide the changes to use that plugins. 
Then we will see, when a API break happens. And hopefully, if we agree, we can exclude "case by case" some stuff from the API check of the plugin. 

Do you modify the `publicKey`? 
Do you have a use-case, which makes use of modifying the `publicKey`?

sbernard31: >  Maybe you provide the changes to use that plugins

I will try to find time to look at this.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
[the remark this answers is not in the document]
sbernard31: Like said in https://github.com/eclipse/californium/issues/1166#issuecomment-569504207.
Behavior break is not desirable for a 2.1.
But as I said in my email privately, I just notice this kind of issue but not insist on it anymore.

sbernard31:
No objection

PR #1148, opened 2019-11-25
Fix unknown signature- and hashalgorithms

boaks:
Fix NullPointerExceptions caused by unknown signature- and hashalgorithms during logging. Extending openssl interoperability tests uncovered this issue. 


boaks:
That other options will change more.

PR #1149, opened 2019-11-25
Add certificate verification to openssl test.

boaks:
Add openssl client auth.


boaks:
If openssl doesn't use a CAfile for trusts, openssl only sends to the certificate without the path. Scandium fails to validate that. FMPOV, that failure is the expected behavior.
Any opinions?

boaks:
Using "-cert_chain" will fix it.

boaks:
The tests demonstrates, that the verification fails on scandium, if there is no trusted certificate in the certificate-chain (expected), but also, if the trusted certificate is not at the end of the certificate-chain. Openssl seems not to truncate the certificate chain based on the received list of certificate authorities as scandium does.

FMPOV:
- merge this test
- update the certificates according the 1. commit in PR #1146 
- add javadoc about the "best practice" (use top-level certificates of chains as trusted certificates). 

PR #1146 will then be merged to 2.1.0

sbernard31:
> If openssl doesn't use a CAfile for trusts, openssl only sends to the certificate without the path. Scandium fails to validate that. FMPOV, that failure is the expected behavior.
> Any opinions?

I think it is the expected behavior too.

> The tests demonstrates, that the verification fails on scandium

Does it means when I launch interoperability tests, I should see some tests failing ?

> if there is no trusted certificate in the certificate-chain (expected), but also, if the trusted certificate is not at the end of the certificate-chain. 

About the last point (not at the end of the certificate-chain). It seems not expected to me ? Do you agree ?

boaks:
> Does it means when I launch interoperability tests, I should see some tests failing ?

No, I marked them with @Ignore :-)

> About the last point (not at the end of the certificate-chain). It seems not expected to me ? Do you agree ?

Consider to only validate the public keys, right, this failing verification is not expected. Considering [RFC5280 - Certification Path Validation](https://tools.ietf.org/html/rfc5280#section-6), which defines also a lot of additional different checks, then it seems to be important to validate a chain always from the begin, and not to start in the middle.

> In Section 6.1, the text describes basic path validation.  Valid
> paths begin with certificates issued by a trust anchor.  The
> algorithm requires the public key of the CA, the CA's name, and any
> constraints upon the set of paths that may be validated using this
> key. 

But for **now** (release 2.0.0), my feeling is, document that and then fix it later.

sbernard31:
> No, I marked them with @ignore :-)

Uups  I missed that :grin: 

> it seems to be important to validate a chain always from the begin, and not to start in the middle.

Read RFC5280 - Certification Path Validation, I think you're right :thinking:  The trust Anchor must be at the end.

I know that LWM2M specification points the [rfc6698](https://tools.ietf.org/html/rfc6698#section-2.1.1) about certificate usage and the only case where certchain/trust anchor is not used is **3-domain-issued certificate**.
There is also one case where you provide a CA certificate which is not at the end : **0-CA constraint** but this CA certificate is an additional constraint and you still need a trust anchor at the end.

> But for now (release 2.0.0), my feeling is, document that and then fix it later.

no objection about that :)

PR #1152, opened 2019-11-28
Check for emtpy server certificate.

boaks:
Fail handshake, if server doesn't provide a certificate.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
@@ -1304,6 +1304,12 @@ public abstract class Handshaker implements Destroyable {
 	 */
 	public void verifyCertificate(CertificateMessage message) throws HandshakeException {
 		if (message.getCertificateChain() != null) {
+			if (isClient && message.getCertificateChain().getCertificates().isEmpty()) {
+				LOGGER.debug("Certificate validation failed: empty server certificate!");
+				AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.BAD_CERTIFICATE,
+						session.getPeer());
+				throw new HandshakeException("Empty server certificate!", alert);
+			}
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1304,6 +1304,12 @@ protected final void expectChangeCipherSpecMessage() {
 	 */
 	public void verifyCertificate(CertificateMessage message) throws HandshakeException {
 		if (message.getCertificateChain() != null) {
+			if (isClient && message.getCertificateChain().getCertificates().isEmpty()) {
+				LOGGER.debug("Certificate validation failed: empty server certificate!");
+				AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.BAD_CERTIFICATE,
+						session.getPeer());
+				throw new HandshakeException("Empty server certificate!", alert);
+			}
```
sbernard31: Should we do that in `certificateVerifier` ?

boaks: The check fails only for the client-handshaker, the server handshaker must be able to process empty certificates.

sbernard31: I get this and I'm aware that with the current code `certificateVerifier` is not aware that it handles a client or a server certificate message.

But If we think this should be done in `certificateVerifier` we can also change the code to let certificateVerifier be aware of this.

boaks: Hm, my plan was to do this after 2.0.0 . Should that plan be adapted?

sbernard31: OK to integrate this change for the 2.0.0 without too much code modification !

And let's see later for integrated or not this check in `certificateVerifier`.

PR #1153, opened 2019-11-28
OSCORE: Update JUnit test for secure re-generation of contexts

rikard-sics:
This pull request updates the JUnit test for the functionality in Appendix B.2. of the OSCORE RFC which describes how to securely re-generate contexts.

It also makes a small update to the context re-derivation procedure to make sure the context rederivation key is kept when contexts are re-generated.

### DIFF
--- a/cf-oscore/src/test/java/org/eclipse/californium/oscore/ContextRederivationTest.java
+++ b/cf-oscore/src/test/java/org/eclipse/californium/oscore/ContextRederivationTest.java
@@ -140,6 +140,8 @@ public class ContextRederivationTest {
 		// Length of Context ID in context
 		int contextIdLen = currCtx.getIdContext().length;
 		assertEquals(3 * SEGMENT_LENGTH, contextIdLen);
+		// Check length of Context ID in the request
+		assertEquals(3 * SEGMENT_LENGTH, requestTestObserver.requestIdContext.length);
 
 		// Check R2 value derived by server using its key with received one
 		// The R2 value is composed of S2 || HMAC(K_HMAC, S2).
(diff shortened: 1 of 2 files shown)


cf-oscore/src/test/java/org/eclipse/californium/oscore/ContextRederivationTest.java
```diff
@@ -79,6 +85,7 @@
 	private final static byte[] rid = new byte[] { 0x01 };
 
 	private static int SEGMENT_LENGTH = ContextRederivation.SEGMENT_LENGTH;
+	private static byte[] requestIdContext;
```
boaks: would it be possible to move this into `RequestTestObserver`?

rikard-sics: Yes, I could do that. I made a new commit a09c1edbc4962b62c9d8be61e5da59f0c5e25de9
Does this fit with what you had in mind?

boaks: I meant the field `requestIdContext`only, but keep the test (assert...) in the main thread.

rikard-sics: Okay, I understand. Like I have now updated it in commit 807413a38e3c224c999d33bc314bc6508d1ae89d?

boaks:
Looks better.

I tried some experiment:

```
// Length of Context ID in context in the request
assertEquals(3 * SEGMENT_LENGTH, requestTestObserver.requestIdContext.length);
String id = resp.advanced().getSourceContext().get(OSCoreEndpointContextInfo.OSCORE_CONTEXT_ID);
assertNotNull(id);
assertEquals(6 * SEGMENT_LENGTH, id.length());
```

According that, the id in the sending context of the request has the same length as in the receiving context of the response. Is this random?

boaks:
I think, if that response's id is the same as the request's id, that may be changed later.

rikard-sics:
Thanks for merging it, sorry for the delayed response.

> According that, the id in the sending context of the request has the same length as in the receiving context of the response. Is this random?

Looking at the figure here:

The JUnit test should be doing request # 2 and response # 2 (as request # 1 and response # 1 happens elsewhere like in the ObjectSecurityContextLayer).

In request # 2 and response # 2 the ID Context in the server and client OSCORE contexts should be R2||R3 and thus the same size.

PR #1157, opened 2019-12-02
Add handshake probe.

boaks:
Intended to be used, if connectivity may be the cause for missing responses.
If the 1. handshake flight is not responded, the handshake is cancelled and the session still kept.
(Stop handshake after "deep sleep").

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -436,6 +436,8 @@ public class DTLSConnector implements Connector, RecordLayer {
 					Connection connection = handshaker.getConnection();
 					if (handshaker.isRemovingConnection()) {
 						connectionStore.remove(connection, false);
+					} else if (handshaker.isProbing()) {
+						LOGGER.debug("Handshake with [{}] failed within probe!", handshaker.getPeerAddress());
 					} else if (connection.getEstablishedSession() == handshaker.getSession()) {
 						// failure after established (last FINISH),
 						// but before completed (first data)
@@ -445,8 +447,7 @@ public class DTLSConnector implements Connector, RecordLayer {
 						LOGGER.warn("Handshake with [{}] failed, but has an established session!",
 								handshaker.getPeerAddress());
 					} else {
-						LOGGER.warn("Handshake with [{}] failed, connection preserved!",
-								handshaker.getPeerAddress());
+						LOGGER.warn("Handshake with [{}] failed, connection preserved!", handshaker.getPeerAddress());
 					}
 				}
 			};
@@ -1236,6 +1237,16 @@ public class DTLSConnector implements Connector, RecordLayer {
 			}
 
 			record.applySession(session);
+
+			if (handshaker != null && handshaker.isProbing()) {
+				// received record, probe successful
+				if (connection.hasEstablishedSession()) {
+					connectionStore.removeFromEstablishedSessions(connection.getEstablishedSession(), connection);
+				}
+				connection.resetSession();
+				handshaker.resetProbing();
+				LOGGER.debug("handshake probe successful {}", connection.getPeerAddress());
+			}
 
 			switch (record.getType()) {
 			case APPLICATION_DATA:
@@ -2091,10 +2102,18 @@ public class DTLSConnector implements Connector, RecordLayer {
 		LOGGER.debug("Sending application layer message to [{}]", message.getEndpointContext());
 
 		Handshaker handshaker = connection.getOngoingHandshake();
-		if (handshaker != null && handshaker.isExpired()) {
-			// handshake expired during Android / OS "deep sleep"
-			// on sending, abort, keep connection for new handshake
-			handshaker.handshakeAborted(new Exception("handshake already expired!"));
+		if (handshaker != null) {
+			if (handshaker.isExpired()) {
+				// handshake expired during Android / OS "deep sleep"
+				// on sending, abort, keep connection for new handshake
+				handshaker.handshakeAborted(new Exception("handshake already expired!"));
+			} else if (handshaker.isProbing()) {
+				if (checkOutboundEndpointContext(message, null)) {
+					message.onConnecting();
+					handshaker.addApplicationDataForDeferredProcessing(message);
+				}
+				return;
+			}
 		}
 
 		if (connection.isActive()) {
@@ -2172,8 +2191,9 @@ public class DTLSConnector implements Connector, RecordLayer {
 				return;
 			}
 		} else {
+			boolean probing = DtlsEndpointContext.HANDSHAKE_MODE_PROBE.equals(handshakeMode);
 			boolean full = DtlsEndpointContext.HANDSHAKE_MODE_FORCE_FULL.equals(handshakeMode);
-			boolean force = full || DtlsEndpointContext.HANDSHAKE_MODE_FORCE.equals(handshakeMode);
+			boolean force = probing || full || DtlsEndpointContext.HANDSHAKE_MODE_FORCE.equals(handshakeMode);
 			if (force || connection.isAutoResumptionRequired(getAutResumptionTimeout(message))) {
 				// create the session to resume from the previous one.
 				if (serverOnly) {
@@ -2185,17 +2205,28 @@ public class DTLSConnector implements Connector, RecordLayer {
 				}
 				message.onConnecting();
 				Handshaker previousHandshaker = connection.getOngoingHandshake();
-				SessionTicket ticket;
-				SessionId sessionId;
+				SessionTicket ticket = null;
+				SessionId sessionId = null;
 				if (session != null) {
 					sessionId = session.getSessionIdentifier();
 					ticket = session.getSessionTicket();
-					connectionStore.removeFromEstablishedSessions(session, connection);
+					if (!probing) {
+						connectionStore.removeFromEstablishedSessions(session, connection);
+					}
 				} else {
-					sessionId = connection.getSessionIdentity();
-					ticket = connection.getSessionTicket();
+					if (!full) {
+						sessionId = connection.getSessionIdentity();
+						ticket = connection.getSessionTicket();
+					}
+					probing = false;
 				}
-				connection.resetSession();
+				if (probing) {
+					// Only reset the resumption trigger, but keep the session for now
+					// the session will be reseted with the first received data
+					connection.setResumptionRequired(false);
+				} else {
+					connection.resetSession();
+				}
 				Handshaker newHandshaker;
 				if (full || sessionId.isEmpty()) {
 					// server may use a empty session id to indicate,
@@ -2210,7 +2241,7 @@ public class DTLSConnector implements Connector, RecordLayer {
 					SecretUtil.destroy(ticket);
 					resumableSession.setHostName(message.getEndpointContext().getVirtualHost());
 					newHandshaker = new ResumingClientHandshaker(resumableSession, this, connection, config,
-							maximumTransmissionUnit);
+							maximumTransmissionUnit, probing);
 				}
 				initializeHandshaker(newHandshaker);
 				if (previousHandshaker != null) {
@@ -2391,9 +2422,12 @@ public class DTLSConnector implements Connector, RecordLayer {
 
 	private void handleTimeout(DTLSFlight flight, Connection connection) {
 
-		if (!flight.isResponseCompleted() && !connection.hasEstablishedSession()) {
+		if (!flight.isResponseCompleted()) {
 			Handshaker handshaker = connection.getOngoingHandshake();
 			if (null != handshaker) {
+				if (!handshaker.isProbing() && connection.hasEstablishedSession()) {
+					return;
+				}
 				Exception cause = null;
 				String message = "";
 				if (!connection.isExecuting() || !running.get()) {
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
@@ -1300,6 +1300,31 @@ public abstract class Handshaker implements Destroyable {
 	}
 
 	/**
+	 * Test, if handshake was started in probing mode.
+	 * 
+	 * Usually a resuming client handshake removes the session from the
+	 * connection store with the start. Probing removes the session only with
+	 * the first data received back.
+	 * 
+	 * @return {@code true}, if handshake is in probing mode, {@code false},
+	 *         otherwise.
+	 * @see ResumingClientHandshaker
+	 */
+	public boolean isProbing() {
+		// intended to be overriden by the ResumingClientHandshaker
+		return false;
+	}
+
+	/**
+	 * Reset probing mode, when data is received during.
+	 * 
+	 * @see ResumingClientHandshaker
+	 */
+	public void resetProbing() {
+		// intended to be overriden by the ResumingClientHandshaker
+	}
+
+	/**
 	 * Test, if handshake is expired according nano realtime.
 	 * 
 	 * Used to mitigate deep sleep during handhsakes.
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
@@ -40,6 +40,7 @@ package org.eclipse.californium.scandium.dtls;
 import java.security.GeneralSecurityException;
 import java.security.MessageDigest;
 
+import org.eclipse.californium.elements.util.NoPublicAPI;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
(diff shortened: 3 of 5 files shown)


boaks:
Extension for 2.1.0

sbernard31:
Could explain the probe mode a bit more. This is not clear do me :confused:

sbernard31:
Thx for the clarification, I will review this now

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -299,7 +306,8 @@ protected Handshaker(boolean isClient, int initialMessageSeq, DTLSSession sessio
 		this.session.setMaxTransmissionUnit(maxTransmissionUnit);
 		this.applicationLevelInfoSupplier = config.getApplicationLevelInfoSupplier();
 		this.inboundMessageBuffer = new InboundMessageBuffer();
-
+		long timeout = config.getRetransmissionTimeout();
+		this.expiresNanosTimeout = TimeUnit.MILLISECONDS.toNanos(timeout << (config.getMaxRetransmissions() + 1));
```
sbernard31: Any chance to write this without `<<` ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -176,6 +177,11 @@
 	/** Realtime nanoseconds of last sending a flight */
 	private long flightSendNanos;
 
+	/** Expire realtime nanoseconds */
+	private long expiresNanos;
```
sbernard31: Any idea about a naming which express that `expiresNanos` is a timestamp and `expiresNanosTimeout` a timeout duration ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1228,6 +1237,39 @@ public final void handshakeFailed(Throwable cause) {
 		SecretUtil.destroy(this);
 	}
 
+	/**
+	 * Test, if handshake was started in probing mode.
+	 * 
+	 * Usually a client handshake removes the session from the connection store
+	 * with the start. Probing removes the session only with the first data
+	 * received back.
+	 *
```
sbernard31: Maybe adding the link to your https://github.com/eclipse/californium/pull/1157#issuecomment-561223737 here too

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -233,6 +239,7 @@
 	protected int statesIndex;
 	protected HandshakeState[] states;
 
+	protected boolean probe = false;
```
sbernard31: Why adding all the "probe stuff" in Handshaker, it seems to me that this is only used in `ResumingClientHandshaker` ?

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -2130,39 +2157,44 @@ private void sendMessage(final long nanos, final RawData message, final Connecti
...
+							connectionStore.removeFromEstablishedSessions(session, connection);
+						}
 					} else {
 						// then ticket != null
 						sessionId = connection.getSessionIdentity();
 					}
-					connection.resetSession();
-					Handshaker previous = connection.getOngoingHandshake();
+					if (probing) {
+						connection.setResumptionRequired(false);
```
sbernard31: This is the end of the week so maybe I'm a bit tired, but the code seems hard to read several line above we just `connection.setResumptionRequired(true)` if probing is true and now we set it to false because probing is true

And until now I believed that probing was only for abbreviated handshake (so resumption = true) :thinking:

boaks:
> The new isExpired state in handshaker in not so clear to me

That happens on android in "deep sleep". The timers are not cancelled, they are just paused. On wakeup, even after some minutes, they continue. That extension therefore holds them up to try to continue a stale handshake.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: Would adding a comment be enough?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: If you search for the usage, you should find, that it's mainly used in places, where a general `Handshaker` is available. To add dummy implementations in the base and override them in the `ResumingClientHandshaker` seems for me to make it more complex. But possible ...

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: Above, the `connection.setResumptionRequired(true)` was introduced to trigger the later
`if (connection.isAutoResumptionRequired(timeout))`. That was implemented to support message specific handshake behaviour.

```
if (probing) {
	connection.setResumptionRequired(false);
} else {
	connection.resetSession();
}
```

Here the `setResumptionRequired(false)` in case of probing replaces the `resetSession()`, which also includes that. The difference is to keep the session.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: Isn't "timeout" used for durations so far?
And I'm not sure, if we name the timestamps special, but I maybe wrong.

boaks:
I addressed some of the comments by a redesign of sendMessage.
I think, put "just the redesign in a own commit PR" without the expires and probe will simplify the this PR.
I will spend some time into that, but I'm not sure, when :-).

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sbernard31: About  the content : 
I guess you try to calculate the maximum time to wait before to consider the message as timed-out.
But I'm not sure to understand the math, e.g. for max retransmission = 3
`t + t*2 + t*2*2 + t*2*2*2 = t*(1+2+4+8) = t*15` **!=** `t*16 =  t*2^(3+1) = t <<(3+1)`
Did I miss something ? 
And by the way the DTLS specification [limit the waiting time to 60s](https://tools.ietf.org/html/rfc6347#section-4.2.4.1).
So maybe it should be more like :  `min(t,60)+min(t*2,60) + ... `

About the form : 
 If I'm correct about the content, I feel that  *2^a is nearer from the original idea than << a

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sbernard31: Yep I saw that and "dummy implementations" was what I have in mind.
because this avoid to jump from class to class to understand the idea when reviewing (and so understanding the code)
(but this is not a strong opinion)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sbernard31: I think you're right we do nothing special until now.
It's just I didn't get the meaning of this variable at first sight and I tell myself that maybe naming could help.
(not a strong opinion)

Maybe javadoc is enough ?

sbernard31:
> That happens on android in "deep sleep". The timers are not cancelled, they are just paused. On wakeup, even after some minutes, they continue. That extension therefore holds them up to try to continue a stale handshake.

I'm not sure I get it :thinking:.
This means that a java timer will not wake up the device from deep sleep ?

boaks:
> I'm not sure I get it thinking.
> This means that a java timer will not wake up the device from deep sleep ?

Yes, if android pushs the app in the background, android wakes it up on a conditional scheduled task (here, about every 5 minutes, when connectivity is available). In the sleep time (e.g. 4 minutes) `System.nanoTime()` is not running and so all timer and jobs  of the java scheduler are also not updated. 
I also work on the isolation of the expire into a own PR.
But after more intensive tests I found, that it will take some more time to work.
Currently I would appreciate to switch to the x509 certificate stuff (PR #1146), just because that seems to be clearer.

sbernard31:
Does it means you want I review #1146 ?

boaks:
> Does it means you want I review #1146 ?

Yes, since yesterday evening, I think, that will make more sense for now.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: About the 60s: OK, do it in a other small PR.

With that, the calculation will be adapted to a "loop", to address the 60s limit. 
By the way, the expire timeout should be some larger than the usual sum of the retransmit times to ensure, that usually a handshake timesout prior to expire.

boaks:
In the meantime I fixed my issues.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -2091,10 +2102,18 @@ private void sendMessage(final long nanos, final RawData message, final Connecti
...
-			// handshake expired during Android / OS "deep sleep"
-			// on sending, abort, keep connection for new handshake
-			handshaker.handshakeAborted(new Exception("handshake already expired!"));
+		if (handshaker != null) {
+			if (handshaker.isExpired()) {
+				// handshake expired during Android / OS "deep sleep"
+				// on sending, abort, keep connection for new handshake
+				handshaker.handshakeAborted(new Exception("handshake already expired!"));
+			} else if (handshaker.isProbing()) {
+				if (checkOutboundEndpointContext(message, null)) {
```
sbernard31: Not linked to this PR, but maybe a method with a clearer name and without `null` param could make this clearer.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -2185,17 +2205,28 @@ private void sendMessageWithSession(final RawData message, final Connection conn
...
+					if (!full) {
+						sessionId = connection.getSessionIdentity();
+						ticket = connection.getSessionTicket();
+					}
+					probing = false;
+				}
+				if (probing) {
+					// Only reset the resumption trigger, but keep the session for now
+					// the session will be reseted with the first received data
+					connection.setResumptionRequired(false);
```
sbernard31: I didn't get this part.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
```diff
@@ -83,11 +117,12 @@
 	 *            if session, recordLayer or config is <code>null</code>
 	 */
 	public ResumingClientHandshaker(DTLSSession session, RecordLayer recordLayer, Connection connection,
-			DtlsConnectorConfig config, int maxTransmissionUnit) {
+			DtlsConnectorConfig config, int maxTransmissionUnit, boolean probe) {
```
sbernard31: [API break] It would be easy to not break.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: Maybe you can give some more hints, what is hard to get?

The intention of probing is documented in the javadoc of `ResumingClientHandshaker`.

```
if (probing) {
	// Only reset the resumption trigger, but keep the session for now
	// the session will be reseted with the first received data
	connection.setResumptionRequired(false);
} else {
	connection.resetSession();
}
```

And in `processRecord` ("the session will be reseted with the first received data"):

```
if (handshaker != null && handshaker.isProbing()) {
	// received record, probe successful
	if (connection.hasEstablishedSession()) {
		connectionStore.removeFromEstablishedSessions(connection.getEstablishedSession(), connection);
	}
	connection.resetSession();
	handshaker.resetProbing();
	LOGGER.debug("handshake probe successful {}", connection.getPeerAddress());
}
```

sbernard31: I try to be clearer. 

I saw the documentation in `ResumingClientHandshaker` :+1: 

What I understand is that with probing mode, we try to do a handshake and if it failed we didn't remove the original/previous connection. (because maybe we lost connectivity but not connection)

What I didn't get is the line : `connection.setResumptionRequired(false);`
I knew there is a comment just above, but I didn't get why using probing means "no resumption required"

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
boaks: Sure, but I even don't want someone encourage to create nor implement a own `Handshaker`. 

For me this is one the strange points of revapi:
A public interface may be sued for two use-cases:
- as callback interface, where it is intended to provide own implementations. Such interfaces can't be modified without major release. To add functionality requires additional interfaces (e.g. `CloseSupportingConnectionStore`) and a merge of these with the next major version.
- as "functionality expose interface", where it is not intended to provide a own implementation. Here it's very easy to add function without breaking the API.

revapi doesn't reflect this nor provide guidance to deal with. At least I didn't found such guidance. Starting issues in revapi will help out, but as I already also said, for me this "public API definition" has, amoung other tasks, so I don't want to spend more of my time in it. Feel free, to spend you own time. Maybe you start to use revapi for leshan, even if the usage there is more to gain best practice with that tool (or @sophokles73 for hono). 

As I already wrote "sometimes", without prior preparing the sources according agreed patterns and rules, it's in my opinion not too helpful, to apply strictness.
The `Handshaker`got exposed by a request of a user (checkout for the introduction of `onInitializeHandshaker`). At that time, no one concerned about unknown API design patterns. I also already asked you, to spend your time into investigate for the `Handshaker`. That is unfortunately time consuming, but would help more than insist in "not changing stuff at all". With agreed patterns and rules, the user's request at that time would have been implemented using such "expose API interfaces" and would have cause more changes, than the actually chosen approach. Fell free to spend your time in provide a proposal for that.

Please don't spend your time in answers without express your opinion on the two interface use-cases and how to setup revapi accordingly.        
     
 
That pattern

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: Ah OK, no it requires one resumption, but this one is started in the lines afterwards.
Reset that mark, either by `connection.setResumptionRequired(false)` or 
`connection.resetSession()`, suppresses further resumptions without new trigger.
(see javadoc of `connection.resetSession()`).

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ResumingClientHandshaker.java
sbernard31: Uups I didn't expect that you take too much time to answer to this remark. As I said in our  private discussion I didn't fight for this, I just notice it. (But maybe you didn't read this and so you could be irritated by my [API break] remarks, sorry about that)

About the 2 interface uses cases, I agree about those 2 ways to use interface.
Even if maybe, for the `expose interface` way,  using a class (facade) to do the job is maybe better. So adding method is not a problem anymore.
With java8 default implementation is a solution too.
So I would say there is many way to handle this and available solution are not the same depending of the situation. That's why I think we can not agreed about patterns or rules before to really start to do the job.

My vision when revapi raise an issue : 
 - if there is a simple way to handle it without exception we go with it
 - if there is no solution or the solution is too ugly or complex, we eventually create an exception if this make sense.
In a general way, we could also discuss with revapi developer as he probably already think about that and could be interested by our feedback.

> Maybe you start to use revapi for leshan

I will as soon as I release the 1.0.0. But we plan to consider all the public/protected stuff as part of the API. I plan 1.0.x release but I don't know if there will be a 1.1.x. The 2.0.0 will probably target LWM2M 1.1 and so will come with API break.
So I'm not sure this would help.

> I also already asked you, to spend your time into investigate for the Handshaker

I already answer about that (still in the private mail) with the interface pattern way which work but taking into account your remark about extensibility,  create a class (facade) is maybe better.

> That is unfortunately time consuming, but would help more than insist in "not changing stuff at all".

Here this is a "straw man" of my position, anyway I suppose remarks is the result of irritability generate by the fact you didn't read my last email.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
sbernard31: Ok I finally get it.
We will initiate a handshake and so in case where resumption was required, we currently take it into account and so we go back in a "normal state"
Generally we go back to this normal state/"no resumption required" using resetSession()
But here we want to keep the session and so we just note that resumption is taking into account.

PR #1160, opened 2019-12-09
Simplify the exceptions for failing handshakes.

boaks:


boaks:
FMPOV, exception messages "may not be considered as part of the API". At least, I hope so :-):

PR #1161, opened 2019-12-09
Split send message into smaller functions.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -2044,7 +2044,6 @@ public class DTLSConnector implements Connector, RecordLayer {
 	 */
 	private void sendMessage(final long nanos, final RawData message, final Connection connection) throws HandshakeException {
 
-		InetSocketAddress peerAddress = message.getInetSocketAddress();
 		if (connection.getPeerAddress() == null) {
 			long delay = TimeUnit.NANOSECONDS.toMillis(ClockUtil.nanoRealtime() - nanos);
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -2056,120 +2055,124 @@ private void sendMessage(final long nanos, final RawData message, final Connecti
...
+	 * Starts handshake, if not already pending, and queue message.
+	 * 
+	 * @param message message to send after handshake completes
+	 * @param connection connectin to send message
+	 * @throws HandshakeException If exception occure starting the handshake
+	 */
+	private void sendMessageWithoutSession(final RawData message, final Connection connection)
+			throws HandshakeException {
 
+		if (!checkOutboundEndpointContext(message, null)) {
```
sbernard31: :warning:  A bit out of topic :warning:
I know you already explained me that a thousand time but each time I read this line I asked myself what it means :sweat_smile: 
(checking outbound context with null :thinking: )
Maybe we should add a method with a more explicit name without the  null parameter. :thinking:

boaks: Only in the DTLSConnector or in the EndpointContextMatcher?

sbernard31: At least in DTLSConnector.
As this will not bring to API break, it sounds pretty easy (expect maybe finding a good name :sweat_smile:) 

About EndpointContextMatcher, I don't know :thinking:  but maybe adding javadoc could be a good start. (I'm not sure)

PR #1165, opened 2019-12-12
Add realtime expiring handshakes.

boaks:
Support android/os deep sleep during handshake.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
@@ -75,6 +75,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 import javax.crypto.SecretKey;
@@ -180,6 +181,12 @@ public abstract class Handshaker implements Destroyable {
 	/** Realtime nanoseconds of last sending a flight */
 	private long flightSendNanos;
 
+	/** Realtime nanoseconds when handshakes get's expired. */
+	private long nanosExpireTime;
+
+	/** Timeout in nanoseconds to expire handshakes. */
+	private final long nanosExpireTimeout;
+
 	/** The current flight number. */
 	protected int flightNumber = 0;
 
@@ -251,6 +258,7 @@ public abstract class Handshaker implements Destroyable {
 
 	private boolean changeCipherSuiteMessageExpected = false;
 	private boolean sessionEstablished = false;
+	private boolean handshakeAborted = false;
 	private boolean handshakeFailed = false;
 	private Throwable cause;
 	private ApplicationLevelInfoSupplier applicationLevelInfoSupplier;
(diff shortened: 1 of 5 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1239,6 +1262,50 @@ public final void handshakeFailed(Throwable cause) {
...
+	 * If {@link #setFailureCause(Throwable)} was called before, only calls with
+	 * the same cause will notify the listeners. If
+	 * {@link #setFailureCause(Throwable)} wasn't called before, sets the
+	 * <em>cause</em> property to the given cause.
+	 * 
+	 * @param cause The reason for the abort.
+	 * @see #handshakeFailed(Throwable)
+	 * @see #isRemovingConnection()
+	 */
+	public final void handshakeAborted(Throwable cause) {
```
sbernard31: I do not understand the benefits of this `final` ?

boaks: The api-changes.json contains:

"justification": "keep API consistent with other similar functions" 

FMPOV, similar functions are:
[handshakeCompleted](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java#L1205) or [handshakeFailed(Throwable cause)](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java#L1224), both are already final.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -176,6 +177,12 @@
 	/** Realtime nanoseconds of last sending a flight */
 	private long flightSendNanos;
 
+	/** Realtime nanoseconds when handshakes get's expired. */
+	private long nonosExpireTime;
```
sbernard31: typo : nonos => nanos

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1239,6 +1262,50 @@ public final void handshakeFailed(Throwable cause) {
...
+	 */
+	public final void handshakeAborted(Throwable cause) {
+		this.handshakeAborted = true;
+		handshakeFailed(cause);
+	}
+
+	/**
+	 * Test, if handshake is expired according nano realtime.
+	 * 
+	 * Used to mitigate deep sleep during handhsakes.
```
sbernard31: Maybe ? (personally I didn't know deep sleep before those PRs)

`Used to mitigate **Android** deep sleep during handhsakes`

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sbernard31: Without API break in mind : 
I don't understand the benefits of all those `final` keyword. Most of the time I really don't like the use of `final` for methods or classes.

With API break in mind :
Even if I agree that `finalMethodAddedToNonFinalClass` is a questionable break, I would not break just for this, mainly because I find `final` keyword for method a bad idea most of the time.

Issue #1166, opened 2019-12-12
Voting for Versioning


boaks:
Hi Californians,

currently there is a discussion about versioning here #1159 .
Please participate in that discussion and add your comments.
And there are first changes, which are suspicious to break the API, e.g. [add final to - protected final PublicKey publicKey](https://github.com/eclipse/californium/pull/1146#discussion_r356595673) .

I wonder now, which approach would create the most benefit for the users.

Consider any public API breaking change as "major" version. Perhaps using a tool to verify that.

Consider "reachable" public API breaking change as "major" version. Though reachable maybe sometimes not clear, that may result in requests, to reestablish the old API.

My personal impression is, that a lot of the public API is just not used external. And considering major releases for changing them seems to be for me a wrong sign. 
2.0.0 toke now years, also because the API was changed a lot. But maintaining  the tools and actinium showed, that there are required changes, but not that much.  
I would therefore prefer to do it in a "soft" way, but if the majority wants something else, then this will be the way to go. 

(My impression is, that in that past, keeping that internal APIs as they where external, prevent them from being cleaned up.)

boaks:
Consider any public API breaking change as "major" version. Perhaps using a tool to verify that. 

If you think, that's the way you benefit most, then vote on this comment.

boaks:
Consider "reachable" public API breaking change as "major" version. Though reachable maybe sometimes not clear, that may result in requests, to reestablish the old API.

If you think, that's the way you benefit most, then vote on this comment.

boaks:
And, sure, additional proposals are welcome!

(You can't vote for this :-)  )

boaks:
From the current discussions in #1168 

[Asn1DerDecoder.OID](https://github.com/eclipse/californium/blob/master/element-connector/src/main/java/org/eclipse/californium/elements/util/Asn1DerDecoder.java#L123)

is reported as inconsistent API.

> [INFO] java.class.nonPublicPartOfAPI: class org.eclipse.californium.elements.util.Asn1DerDecoder.OidEntityDefinition: Class 'org.eclipse.californium.elements.util.Asn1DerDecoder.OidEntityDefinition' is indirectly included in the API (by the means of method return type for example) but the class is not accessible (neither public nor protected). (breaks semantic versioning)

There now two ways, to work on that:
1. [strict](https://github.com/eclipse/californium/pull/1168#issuecomment-565460106)

[...]

PR #1169, opened 2019-12-15
Revapi ignores 

boaks:
Fix inconsistent API scopes.
Add "api-changes.json" to ignore APIfixes.
Skip revapi for module cf-pubsub, not deployed for 2.0.0.

### DIFF
--- a/cf-pubsub/pom.xml
+++ b/cf-pubsub/pom.xml
@@ -42,6 +42,14 @@
 					</descriptorRefs>
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>org.revapi</groupId>
+				<artifactId>revapi-maven-plugin</artifactId>
+				<configuration>
+					<!-- not depolyed to repos -->
+					<skip>true</skip>
+				</configuration>
+			</plugin>
 		</plugins>
 	</build>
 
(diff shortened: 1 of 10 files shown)


sbernard31:
Please do what you want about that.

sbernard31:
Reading again my previous comment, and sounds a bit aggressive :worried: : 
I meant I have no strong opinion about that so if you think this is the right way, go for this.
(more details [here](https://github.com/eclipse/californium/issues/1166#issuecomment-566024894))

(just in case I open an issue about that at [revapi github](https://github.com/revapi/revapi/issues/created_by/sbernard31))

boaks:
So, start to go for PR #1168 and then this, and afterwards adjust the other pending PR's accordingly?

sbernard31:
Just to be sure I was talking about the "Non-Public Part of API" inconsistency.

What does it mean : "adjust the other pending PR's accordingly?"

boaks:
Either try to add changes to api-ignore or try to not change the API.
FMPOV, it's much more important to change it according a good future solution then to keep a (not used) API for "strict technical compatibility".

boaks:
Just to mention:
Agree on these two PRs, doesn't mean to agree on the other PRs, even if the get adapted. It's more like, these two are the minimal consensus, the rest maybe discussion.

sbernard31:
So I agree on this PR and #1168.

For the rest, I would like to know  your opinion about https://github.com/eclipse/californium/pull/1168#issuecomment-566027052 ?

Issue #1170, opened 2019-12-18
can not run multicast server demo in Mac OS


RaymondMcGuire:
I can't run the multicast demo in Mac, but in windows is ok..

the following is the log info:

[main] INFO org.eclipse.californium.core.network.config.NetworkConfig - loading properties from file /Users/raymondmg/Project/CoapSimulate/coap_server/Californium.properties
[main] INFO org.eclipse.californium.core.network.RandomTokenGenerator - using tokens of 8 bytes in length

[main] INFO org.eclipse.californium.core.network.CoapEndpoint - coap CoapEndpoint uses udp plain

[main] INFO org.eclipse.californium.core.network.stack.BlockwiseLayer - BlockwiseLayer uses MAX_MESSAGE_SIZE=1024, PREFERRED_BLOCK_SIZE=512, BLOCKWISE_STATUS_LIFETIME=300000, MAX_RESOURCE_BODY_SIZE=8192, BLOCKWISE_STRICT_BLOCK2_OPTION=false

[main] INFO org.eclipse.californium.core.network.stack.ReliabilityLayer - ReliabilityLayer uses ACK_TIMEOUT=2000, ACK_RANDOM_FACTOR=1.5, and ACK_TIMEOUT_SCALE=2.0 as default
[main] INFO org.eclipse.californium.core.CoapServer - Starting server
coap server: 70

[main] INFO org.eclipse.californium.core.network.CoapEndpoint - coap Stopping endpoint at coap://0.0.0.0:5683

[main] ERROR org.eclipse.californium.core.CoapServer - cannot start server endpoint [0.0.0.0/0.0.0.0:5683]

java.net.SocketException: Can't assign requested address
	at java.net.PlainDatagramSocketImpl.join(Native Method)
	at java.net.AbstractPlainDatagramSocketImpl.join(AbstractPlainDatagramSocketImpl.java:178)
	at java.net.MulticastSocket.joinGroup(MulticastSocket.java:323)
	at org.eclipse.californium.elements.UdpMulticastConnector.start(UdpMulticastConnector.java:89)
	at org.eclipse.californium.core.network.CoapEndpoint.start(CoapEndpoint.java:506)
	at org.eclipse.californium.core.CoapServer.start(CoapServer.java:252)
	at coap_server.CoapMultiCastServer.main(CoapMultiCastServer.java:30)
Exception in thread "main" java.lang.IllegalStateException: None of the server endpoints could be started
[...]

PR #1171, opened 2019-12-19
Add improved IPv6 multicast support.

boaks:
Add NetworkInterfacesUtil for any interface features.
Issue #1170

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/UdpMulticastConnector.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/UdpMulticastConnector.java
@@ -16,9 +16,17 @@
 package org.eclipse.californium.elements;
 
 import java.io.IOException;
+import java.net.Inet4Address;
+import java.net.Inet6Address;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.MulticastSocket;
+import java.net.SocketException;
+
+import org.eclipse.californium.elements.util.NetworkInterfacesUtil;
+import org.eclipse.californium.elements.util.StringUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * {@link Connector} which inherits functionality of {@link UDPConnector}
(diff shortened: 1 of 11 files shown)


element-connector/src/main/java/org/eclipse/californium/elements/UdpMulticastConnector.java
```diff
@@ -27,6 +35,7 @@
  * Protocol (CoAP) to the registered multicast group.
  */
 public class UdpMulticastConnector extends UDPConnector {
+	public static final Logger LOGGER = LoggerFactory.getLogger(UdpMulticastConnector.class.getName());
```
sbernard31: why not ? LoggerFactory.getLogger(UdpMulticastConnector.class);

element-connector/src/main/java/org/eclipse/californium/elements/util/NetworkInterfacesUtil.java
```diff
@@ -0,0 +1,141 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Utility class for NetworkInterfaces. Determine MTU, IPv4, IPv6 support.
+ */
+public class NetworkInterfacesUtil {
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(NetworkInterfacesUtil.class.getCanonicalName());
```
sbernard31: Why not ? LoggerFactory.getLogger(NetworkInterfacesUtil.class);

element-connector/src/main/java/org/eclipse/californium/elements/UdpMulticastConnector.java
boaks: Just copied the UDPConnector stuff.

element-connector/src/main/java/org/eclipse/californium/elements/util/NetworkInterfacesUtil.java
boaks: I checked it right now. it seems not to be used that frequently in californium.
I'm not sure, what the benefit will be, but we can adapt that then generally (except some special health logger, where this will not work).

PR #1174, opened 2019-12-30
Preserve ACK/RST for CON responses.

boaks:
Send-/Receive-timefilter drops only retransmitted responses.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
@@ -68,7 +68,7 @@ public class ReliabilityLayer extends AbstractLayer {
 	private final ReliabilityLayerParameters defaultReliabilityLayerParameters;
 
 	private final AtomicInteger counter = new AtomicInteger();
-	
+
 	/**
 	 * Constructs a new reliability layer. Changes to the configuration are
 	 * observed and automatically applied.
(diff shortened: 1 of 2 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -284,24 +284,55 @@ public void receiveResponse(final Exchange exchange, final Response response) {
...
+				if (response.getNanoTimestamp() < exchange.getSendNanoTimestamp()) {
+					// received response duplicate before ACK/RST
+					// or last request retransmission was sent
+					// => drop response
+					// Note: if the response is received the 1. time, no ACK nor RST was sent
+					// so far. Therefore the send timestamp is related to the request
+					// retransmission only. In that case always ACK/RST to try stopping future
+					// response retransmissions.
+					int count = counter.incrementAndGet();
+					LOGGER.info("{}: {} duplicate response {}, server sent ACK delayed, ignore response", count,
```
sbernard31: Is this LOGGER change from `debug` to `info` is intended ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -284,24 +284,55 @@ public void receiveResponse(final Exchange exchange, final Response response) {
...
+							exchange, response);
+					return;
+				}
+				// resend last ack or rst, don't update, request state may have changed!
+				if (response.isRejected()) {
+					ack = false;
+					LOGGER.debug("{} reject duplicate CON response, request canceled.", exchange);
```
sbernard31: Is the "request canceled" in the log intended ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
boaks: I changed it back.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
boaks: The "canceled request" is the cause for the "rejected response".

PR #1175, opened 2019-12-31
Add ACK to health statistic.

boaks:
Expose coap health handler in builder.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -130,6 +130,7 @@ import org.eclipse.californium.elements.UDPConnector;
 import org.eclipse.californium.elements.util.ClockUtil;
 import org.eclipse.californium.elements.util.DaemonThreadFactory;
 import org.eclipse.californium.elements.util.ExecutorsUtil;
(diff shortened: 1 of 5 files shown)


boaks:
The Builder uses the interface only.
The `CoapEndpoint` then creates a default instance using the logging implementation, if health logging is configured and no other interface implementation is provided.
For me this seems to provide all benefits: you may place any implementation or disable it at all.

sbernard31:
>  For me this seems to provide all benefits: you may place any implementation or disable it at all.

This is not, the problem I raised here.

My point if that it already exist a way to monitor called `MessageExchangeInterceptor`(and maybe also `messageObserver`)

When you introduce `coapHealthHandler`, I said OK if you need this now but for the future I think we should rather improve our listeners to make it usable for this kind of use case instead of create a new way.

I reiterate my concern about multiplying this kind of interface to do almost the same job.

boaks:
`MessageExchangeInterceptor`, do you mean `MessageInterceptor`?

If so, yes, it looks similar, and I guess try to extend that will have advantages.
For my concerns, that requires an additional interface for failed and successful sends, and for receiving duplicates and dropped messages. I will give it a try.

boaks:
See PR #1179 for alternative implementation.

boaks:
See #1179 or #1180

PR #1178, opened 2020-01-03
Use "classic" way to initialize logger.

sbernard31:
This PR was triggered by https://github.com/eclipse/californium/pull/1171#discussion_r362503747.

Reading the [SLF4J manual](http://www.slf4j.org/manual.html) this sounds to be the "normal" way.

AFAIK, the only benefits is that this is more concise.

(I do this quickly with a find/replace, then review the diff, so if this is not accepted this is not a big deal :-) )


boaks:
LGTM.
(I merge it, to be bale to update the other pending PRs)

PR #1179, opened 2020-01-07
Message interceptor based health

boaks:
Add `HealthStatisticLogger` and `MessagePostInterceptor` and use it in benchmark and extended-plugtest-server. Deprecate `CoapEndpointHealth` and `CoapEndpointHealthLogger`.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/Message.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/Message.java
@@ -960,7 +960,9 @@ public abstract class Message {
 
 	/**
 	 * Checks if this message is a duplicate.
-	 *
+	 * 
+	 * Since 2.1 this also reflects, if the message is resent.
+	 * 
 	 * @return true, if is a duplicate
 	 */
 	public boolean isDuplicate() {
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java
@@ -150,8 +150,9 @@ import org.slf4j.LoggerFactory;
  * layer to layer. An {@link Exchange} represents the known state about the
  * exchange between a request and one or more corresponding responses. The
  * matcher remembers outgoing messages and matches incoming responses, acks and
- * rsts to them. MessageInterceptors receive every incoming and outgoing
- * message. By default, only one interceptor is used to log messages.
+ * rsts to them. MessageInterceptors registered with
+ * {@link #addInterceptor(MessageInterceptor)} receive every incoming and
+ * outgoing message. By default, only one interceptor is used to log messages.
  * 
  * <pre>
  * +-----------------------+
@@ -191,7 +192,7 @@ import org.slf4j.LoggerFactory;
  * The endpoint and its layers use an {@link ScheduledExecutorService} to
  * execute tasks, e.g., when a request arrives.
  */
-public class CoapEndpoint implements Endpoint {
+public class CoapEndpoint implements Endpoint, MessagePostProcessInterceptors {
 
 	/** the logger. */
 	private static final Logger LOGGER = LoggerFactory.getLogger(CoapEndpoint.class);
@@ -233,7 +234,13 @@ public class CoapEndpoint implements Endpoint {
 
 	/** tag for logging */
 	private final String tag;
-	
+
+	/**
+	 * @deprecated use {@link MessageInterceptor} registered with
+	 *             {@link #addPostProcessInterceptor(MessageInterceptor)}
+	 *             instead.
+	 */
+	@Deprecated
 	private final CoapEndpointHealth health;
 
 	/** The executor to run tasks for this endpoint and its layers */
@@ -251,6 +258,9 @@ public class CoapEndpoint implements Endpoint {
 	/** The list of interceptors */
 	private List<MessageInterceptor> interceptors = new CopyOnWriteArrayList<>();
 
+	/** The list of post process interceptors */
+	private List<MessageInterceptor> postProcessInterceptors = new CopyOnWriteArrayList<>();
+
 	/** The list of Notification listener (use for CoAP observer relations) */
 	private List<NotificationListener> notificationListeners = new CopyOnWriteArrayList<>();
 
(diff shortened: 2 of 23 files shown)


boaks:
Intended to replace PR #1175

sbernard31:
I feel this PR go in the right way and we can merge it like this.

But for methods which exist in both interface, I would like to understand why `MessageInterceptor` and MessagePostInterceptor` is not called in a same way.

Is it to not break behavior ? If we didn't want to keep this behavior compatibility, did you merge both ?

If yes I understand the code is like this and so we will expect that we remove the old one in a next major release and maybe we deprecate `MessageInterceptor`.
If not, I should rather go to add some kind of RequestSent() method and call both event for `MessagePostInterceptor` (which I called MessageInterceptor2)

boaks:
The `MessageInterceptor`s are called "between" the coap-stack and the connector. throwing a exception there or cancel the message stops the further processing. I'm not sure, how frequently that is used, but it's part of the public reachable (callback) API and openHAB requires to have callbacks before the coap-stack starts to process the message. So, this **CAN NOT BE REMOVED WITHOUT HARMING OTHERS**. 

The `MessagePostInterceptor`s are called "after" the coap-stack and the connector. My intention is, that for health it's interesting, if messages are re-transmitted or re-received. And for health it's also interesting, if sending the message has failed on the connector. The re-???-ratio may indicate an "overload level". The error-ration may indicate environment issues. 

Both share their callback functions, but are called at different stages.
Yes, strictly, a `MessagePostInterceptor` is no `MessageInterceptor`, but they share their methods. Alternatively, the methods can be moved to a common base  interface. Or remove the `MessageInterceptor` extension from the `MessagePostInterceptor` and copy the methods (maybe also change the name from send to sent and from receive to received). That would also double the add/remove functions, the list is already doubled.

> I reiterate my concern about multiplying this kind of interface to do almost the same job.

The question is, how that "almost" is then implemented.

Anyway, I feel better with that approach than with the `CoapEndpointHealth`.
Even if the `MessageInterceptor` extension gets removed from the `MessagePostInterceptor` and so some duplication is created.

boaks:
> If not, I should rather go to add some kind of RequestSent() method and call both event for MessagePostInterceptor (which I called MessageInterceptor2)

FMPOV, that seems more to justify the "extends" than it reflects my use-case. If both interfaces are not related and the methods for the "Post" are renamed, a implementing class can implement both interfaces and add to both lists. That would be more flexible.

sbernard31:
> it's part of the public reachable (callback) API and openHAB requires to have callbacks before the coap-stack starts to process the message

OK, I understand (:fearful:  no need to use upper case + bold)

So My understanding is that we need both event before and after.

> Anyway, I feel better with that approach than with the CoapEndpointHealth.

Me too :)

> The question is, how that "almost" is then implemented.

I will look at this deeper.

sbernard31:
I tried to create a quick&dirty PR about my `MessageInterceptor2` idea.

If you think this is **not** the right way, then I think we should to clearly separate `MessageInterceptor` and `MessagePostIntercepor`.
But it seems to me this second solution move us away from the original idea : 

> I reiterate my concern about multiplying this kind of interface to do almost the same job.

boaks:
There is one proposal left:
introduce a common base interface (e.g. MessageListener with onRequest(Request)) and derive the `MessageInterListener` and `MessagePostListener`.
And deprecate the `MessageInterceptor` (sure, not removed before a 3.0 but may be even longer depending on openHABs possibilities to do that slightly update).

sbernard31:
Yep, this is another solution but what would be the benefits of this compare to `MessageIntercepor2` which allow to keep compatibility without deprecation ? :thinking: 

(bonus question, I see the meaning of `Post` but not of `inter`)

boaks:
`MessageIntermediateListener`

sbernard31:
I understood that inter means intermediate but why intermediate ?

boaks:
Because it's called between the connector and the coap-stack.

It may also be possible, just to have two lists with add/remove and use the new MessageListener without an additional marker interface. The would move the "special" into the documentation of the add/remove.
Bascially it's not the API of the `MessageInterceptor`, it's more the documentation for what is is exactly used. 

May be with just add the new add/remove and change the documentation this PR will also do it.

sbernard31:
> It may also be possible, just to have two lists with add/remove and use the new MessageListener without an additional marker interface. The would move the "special" into the documentation of the add/remove.

I think this does not really make sense as sendError/failedToSent does not make sense why it should rather in one than the other.

boaks:
> It may also be possible, just to have two lists with add/remove and use the new MessageListener without an additional marker interface. 

I was wrong, I ment without "two additional", sure the "post" with the error will still be required.

boaks:
Newest proposal/personal conclusion:

The interface `MessageInterceptor` should just be specified more relaxed from the processing stage.
The register methods instead should define the processing stage and if cancel is supported.

With that, it's not too much change (at least in my opinion).

sbernard31:
About the newest proposal : 

1. How do you handle the backward compatibility ? 
2. I suspect that we will want to add some new events which are not particularly linked to a processing stage (like sentError) and so I don't know if this is really more elegant.
3. This makes difficult if I want to be notified at each stage for same object because I will just get event several time without knowing the stage where they come from. (a solution could be to add an enum as parameter of the callback, but I'm not sure this is better than have several methods)
4. About support of cancel, I didn't get what you means.

Currently, I still feel that the #1180 solution is better as it isn't affected by issue above.

boaks:
> How do you handle the backward compatibility ?

This is either simply naturally by that approach (all already available MessageInterecptors using the already available register will keep their functionality, though the combination of both defines now the same behaviour as the interface did before on it's own), or, I don't get what you see as critical?

> I suspect that we will want to add some new events which are not particularly linked to a processing stage (like sentError) and so I don't know if this is really more elegant.

The sentError is now just detected by `getSendError()`.

> This makes difficult if I want to be notified at each stage for same object because I will just get event several time without knowing the stage where they come from.

Yes, but none of the current use-case uses such a combinatino. Instead each is only interested in a single event. And it will be easy to just use two interceptor instances, to distinguish the stage, if this is not reflected in the message fields. So this approach reflects directly the use-cases.

> About support of cancel, I didn't get what you means.

That was part of the previous javadoc of the interceptor. You simply can't cancel a message in order to prevent it from being sent, after the message was already sent.

boaks:
> This sounds obvious and I can not see the issue here. A way to make this even clearer is just to specify where you can use cancel in each method javadoc of MessageInterceptor

I'm wondering, why 3x javadocs at MessageInterceptor is clearer the one javadoc at the register.

boaks:
> I already proposed a use case which this could be useful 

To measure times you need to match the messages. I would guess, that this will also extend the complexity somehow, so that it will not stay that "very simple".

boaks:
Generally it seems, that now there are two proposals:

Use a extended interceptor interfaces with more explicit events (copying similar methods).
Use the available interceptor with more register points (copying similar add/remove/get).

boaks:
> (Here we can see that just talking about the event is more complicate. I must write a sentence instead of the just call the even like requestSent)

I just forgot to generalize the javadoc.

boaks:
The approach of this PR is also used for other stuff within Californium, e.g. the "Layer" of the "CoapStack". Anyway, if the other approach (new special event methods) should be chosen, please obey the fixes of the last commits here, otherwise the "post processed" events may not be called consequently.

sbernard31:
> I'm wondering, why javadoc at MessageInterceptor is clearer then javadoc at the register.

This is not what I said.
My understanding was that you consider there is a problem with javadoc about `MessageInterceptor` and cancel feature with `MessageInterceptor2` solution. Personally I didn't see the problem but I proposed to move the javadoc from class to methods to disambiguate that if needed.
(To answer to your "wondering", It seems to me when playing with listener, you mainly read the listener javadoc instead of the `addListener`javadoc. That's why maybe javadoc on listener could be consider as clearer but that was not my point here :-))

> Anyway, if the other approach (new special event methods) should be chosen, please obey the fixes of the last commits here,

I will try to do it if this is chosen.

> Generally it seems, that now there are two proposals:
> Use a extended interceptor interfaces with more explicit events (~copying similar methods~ adding new events).
> Use the available interceptor with more register points (copying similar add/remove/get).

It seems we didn't succeed to agree about some details :sweat_smile:.
Anyway, I think we agree that both solution are an improvement to Cf monitoring feature and are a better way to implement health stuff.
So do not hesitate to chose the one which seems better to you.

boaks:
> So do not hesitate to chose the one which seems better to you.

Thanks!

PR #1180, opened 2020-01-09
add Message interceptor2

sbernard31:
This is an alternative PR to #1179 
This is just a quick&dirty PR to illustrate this idea : https://github.com/eclipse/californium/pull/1179#issuecomment-572081188

If we agree on the idea, I can clean this or I let you do it if you prefer.


californium-core/src/main/java/org/eclipse/californium/core/network/interceptors/HealthStatisticLogger.java
```diff
@@ -183,7 +183,7 @@ public void dump() {
 	}
 
 	@Override
-	public void sendRequest(Request request) {
+	public void requestSent(Request request) {
 		if (request.isSent()) {
```
sbernard31: Writting this PR, I asked myself why `isSent` is called which make me feel that I maybe missed something :thinking:

boaks:
```
public void onSent() {
	for (MessageInterceptor interceptor : interceptors) {
		if (interceptor instanceof MessageInterceptor2) {
			MessageInterceptor2 postInterceptor = (MessageInterceptor2) interceptor;
			if (message instanceof Request) {
				postInterceptor.requestSent((Request) message);
			} else if (message instanceof Response) {
				postInterceptor.receiveResponse((Response) message);
			} else if (message instanceof EmptyMessage) {
				postInterceptor.emptyMessageSent((EmptyMessage) message);
			}
		}
	}
	message.setSent(true);
}
```
and

```
public void requestSent(Request request) {
	if (request.isSent()) {
		resentRequests.increment();
	} else {
		sentRequests.increment();
	}
}
```

The idea is simply: 
On transmission, the handler is called before "setSent(true)", so `sent` is `false`. 
On retransmission, `sent` is already `true` (set by the first transmission). 
Therefore `sent` could be used to distinguish between "retransmission" and "transmission". Alternatively, the `Message.duplicate` would be possible to use, if it gets also set on sending, currently it's only used for receiving.

boaks:
For me this looks more as work-around then as "natural extension".
I would prefer to have either a new common base interface or to similar interfaces without a common base.

sbernard31:
Could you explain why this make you think this is a work-around ? I didn't get it.

boaks:
Mainly, because I don't see (at least for no) common use-case for both.

Without the special current `MessageInterceptor` I don't think, that this approach would be chosen. Therefore it looks for me as workaround, that something currently (too) special, gets reused more general without really adapting the stuff to the "new" awareness.

sbernard31:
For me it makes sense that a message interceptor have several event for same message (before/after, we can even imagine more).

> Mainly, because I don't see (at least for no) common use-case for both.

Maybe both could be used for example to monitor time to handle request or response.
Currently we use `MessageInterceptor` for metrics and I would not be disturb at all if the interface propose me all those events and  I only choose the one I need.

> Without the special current MessageInterceptor I don't think, that this approach would be chosen. 

I don't know why ? but anyway the fact that `MessageInterceptor` is here and we want to keep it for backward compatibility should be taking into account to take the decision.

> Therefore it looks for me as workaround, that something currently (too) special, gets reused more general without really adapting the stuff to the "new" awareness.

I didn't get what you consider "too special" and what is the "new awareness".

sbernard31:
#1179 way was preferred

Issue #1186, opened 2020-01-16
Some strange behavior about Handshake mode NONE (#1104)


sbernard31:
We finally integrate leshan-1.0.0-M13(based on Cf-2.0.0) in production . \o/

In Leshan, we add the feature where each requests sent to a UQ device use the handshake mode NONE. (#1104)
For now, almost all our devices use Q mode (as server initiated in still experimental), but we : 
 - still see in our metrics a lot of server initiated handshake.
 - still see a lot of "Hanshake flight 1 timeout".

Where we expected a decrease of this.

So I add an `MessageInterceptor` to log if EndpointContext is well set before to enter in the coap stack.
And all sounds OK excepts maybe for transparent blockwise where destination endpoint context parameter of parent request sounds maybe not transmit to "children" block requests. (I'm not sure)

Anyway I'm pretty sure this "blockwise bug" is not the main reason, as requests which failed on "Handshake flight 1 timeout" rarely used block wise transfer.

I looked at the californium code and for now, I can not explained this behavior :thinking:

boaks:
[BLOCK1 - Request](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java#L120)
[BLOCK2 - Request](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java#L980)

Seems the endpoint-context is tranferred.

sbernard31:
For block2 : `block.setDestinationContext(response.getSourceContext())`, the  parent request destination context sounds not to be used. So parameters like HANDSHAKE_MODE seems lost :thinking:. Did I miss something ?

boaks:
Maybe a `MessageObserver`, which checks `onSendError(Throwable error)` for the error-message ("Hanshake flight .* timeout"), and reports more details about the affected `Message`?

(Just to mention, from time to time, I used to cleaned up the error-messages, so just ensure, you grep for the right one.)

boaks:
> For block2 : block.setDestinationContext(response.getSourceContext()), the parent request destination context sounds not to be used. So parameters like HANDSHAKE_MODE seems lost thinking. Did I miss something ?

Hm,  the context of the response is used, to ensure, that the crypto stuff get's added.
Maybe change that to a variant, which checks, if the original Request's context already contains the information and use that instead, would help here.

boaks:
[...]

PR #1188, opened 2020-01-16
Fix handshake mode for blockwise.

boaks:

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseStatus.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseStatus.java
@@ -35,6 +35,9 @@ import org.eclipse.californium.core.coap.BlockOption;
 import org.eclipse.californium.core.coap.Message;
 import org.eclipse.californium.core.coap.OptionSet;
 import org.eclipse.californium.core.network.Exchange;
+import org.eclipse.californium.elements.DtlsEndpointContext;
+import org.eclipse.californium.elements.EndpointContext;
+import org.eclipse.californium.elements.MapBasedEndpointContext;
 
 /**
  * A tracker for the status of a blockwise transfer of a request or response body.
@@ -50,6 +53,7 @@ public abstract class BlockwiseStatus {
 	protected boolean randomAccess;
 	protected final ByteBuffer buf;
 	protected Exchange exchange;
+	protected EndpointContext followUpEndpointContext;
 
 	private ScheduledFuture<?> cleanUpTask;
 	private Message first;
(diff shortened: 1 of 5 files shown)


sbernard31:
I have a quick look at this PR and I think it should work better.

This code is very specific to HANDSHAKE_MODE and this create a kind of "dependency" between HANDSHAKE_MODE and transparent block wise.

To avoid that I proposed to add a kind of type to endpoint context attributes which define if attribute should be used once or for each "children" requests (https://github.com/eclipse/californium/issues/1186#issuecomment-575246399). This concept could works for transparent blockwise but also maybe for CoAP retransmission.  
Maybe using a prefix or suffix like KEY_PREFIX_NONE_CRITICAL.
But maybe this is a bit overkill as for now we don't have so much attributes :thinking:.

Talking about  CoAP retransmission, I guess we maybe face the kind of issue ? (not sure at all)

boaks:
>But maybe this is a bit overkill

No, that may/will be the plan for a 3.0.

Changing the value of final String may cause much more stress.
Changing the key value of the handshake-mode using new prefixes, which makes that approach more generic breaks the API logically in a way, which gets hard to detect.
I'm not sure, if this fix is required for a 2.0.1, therefore I tried to change it in a "save way".

sbernard31:
> I'm not sure, if this fix is required for a 2.0.1, therefore I tried to change it in a "save way".

:+1:

boaks:
;-(

requires more work ... notifies from other addresses will require to use not the endpoint context of the original request, though that's destination is not longer valid.

 ;-).

boaks:
After some investigation and tests, I consider to use the response's endpoint context and preserve the handshake-mode "none", is the most proper solution.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseStatus.java
```diff
@@ -231,6 +235,35 @@ final synchronized int getBlockCount() {
...
+			// considering notifies with address changes,
+			// use the response's endpoint-context to compensate that
+			EndpointContext context = exchange.getRequest().getDestinationContext();
+			String mode = context.get(DtlsEndpointContext.KEY_HANDSHAKE_MODE);
+			if (mode != null && mode.equals(DtlsEndpointContext.HANDSHAKE_MODE_NONE)) {
+				// restore handshake-mode "none"
+				blockContext = MapBasedEndpointContext.addEntries(blockContext, DtlsEndpointContext.KEY_HANDSHAKE_MODE,
+						DtlsEndpointContext.HANDSHAKE_MODE_NONE);
+			}
+			followUpEndpointContext = blockContext;
```
sbernard31: clearly a details but why not ? 

```java
if (mode != null && mode.equals(DtlsEndpointContext.HANDSHAKE_MODE_NONE)) {
	// restore handshake-mode "none"
	followUpEndpointContext = MapBasedEndpointContext.addEntries(blockContext, DtlsEndpointContext.KEY_HANDSHAKE_MODE, DtlsEndpointContext.HANDSHAKE_MODE_NONE);
} else {
	followUpEndpointContext = blockContext;
}
```

boaks: If you prefer ...

sbernard31: Don't do it if you don't think, it's clearer.

boaks: Done ... over the time I get used to the other style, mainly to initialize "final" stuff. But I have no real preference.

PR #1196, opened 2020-01-20
Expose simplified http-server.

boaks:
Update apache http libraries.

### DIFF
--- a/californium-proxy/pom.xml
+++ b/californium-proxy/pom.xml
@@ -17,14 +17,12 @@
 	<description>Cross-proxy module</description>
 
 	<properties>
-		<httpasyncclient.version>4.1.2</httpasyncclient.version>
+		<httpasyncclient.version>4.1.4</httpasyncclient.version>
 		<httpasyncclient.version.spec>
 			version="[${versionmask;==;${httpasyncclient.version}},${versionmask;+;${httpasyncclient.version}})"
 		</httpasyncclient.version.spec>
-		<httpcore.version>4.4.5</httpcore.version>
-		<httpcore.version.spec>version="[${versionmask;==;${httpcore.version}},${versionmask;+;${httpcore.version}})"</httpcore.version.spec>
-		<httpclient.version>4.5.2</httpclient.version>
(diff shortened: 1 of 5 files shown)


boaks:
The intention to expose this simplified http-server is to use it later for a simple demo in cf-proxy.

californium-proxy/pom.xml
```diff
@@ -17,14 +17,12 @@
...
-		<httpasyncclient.version>4.1.2</httpasyncclient.version>
+		<httpasyncclient.version>4.1.4</httpasyncclient.version>
 		<httpasyncclient.version.spec>
 			version="[${versionmask;==;${httpasyncclient.version}},${versionmask;+;${httpasyncclient.version}})"
 		</httpasyncclient.version.spec>
-		<httpcore.version>4.4.5</httpcore.version>
-		<httpcore.version.spec>version="[${versionmask;==;${httpcore.version}},${versionmask;+;${httpcore.version}})"</httpcore.version.spec>
-		<httpclient.version>4.5.2</httpclient.version>
-		<httpclient.version.spec>version="[${versionmask;==;${httpclient.version}},${versionmask;+;${httpclient.version}})"</httpclient.version.spec>
+		<httpcorenio.version>4.4.12</httpcorenio.version>
```
sophokles73: does this mean that httpcore and httpclient are no longer used/required?

boaks: AFAIK, they used indirect via dependencies from the left two.
I chose the httpasynclient as leading dependency and updated the httpcore-nio to a even newer one than httpasynclient depends on.

So currently these two updated versions includes:
httpasyncclient.4.1.4
httpclient4.5.6
httpcore-nio.4.4.12
httpcore-4.4.10

A current check shows, that there is now also a new httpcore (4.4.13) and httpclient (4.5.11), but I would like to add now first my other updates/changes in order to be able again to do some stress- and longterm-tests.

sophokles73: @boaks Just make sure that the transitive deps still have the same major.minor version. Otherwise, we'll need a new CQ for the dependencies ...

PR #1200, opened 2020-01-22
Introduce dtls_retransmission_timeout.

boaks:

### DIFF
--- a/demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/ClientInitializer.java
+++ b/demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/ClientInitializer.java
@@ -181,6 +181,7 @@ public class ClientInitializer {
 		int staleTimeout = config.getInt(NetworkConfig.Keys.MAX_PEER_INACTIVITY_PERIOD);
 		int senderThreads = config.getInt(NetworkConfig.Keys.NETWORK_STAGE_SENDER_THREAD_COUNT);
 		int receiverThreads = config.getInt(NetworkConfig.Keys.NETWORK_STAGE_RECEIVER_THREAD_COUNT);
+		int retransmissionTimeout = config.getInt(Keys.ACK_TIMEOUT);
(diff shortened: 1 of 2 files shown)


boaks:
Sure, that `NetworkConfig` is somehow strange. I assume a better solution will take reasonable time, but currently I don't see, that I can spend that time into changing it.
The reason for me to add this value was, that I got aware, that scandium uses still 1s as default. Therefore I choose to add that time to the `NetworkConfig`.

sbernard31:
If I well understand your needs is about plugtests and sample.
You need it to be set at 2s for those components ? or you need to be able to configure it ?

boaks:
> You need it to be set at 2s for those components ? 

Depending on the environment, that is required to be adapted. Using Cat-NB makes it hard to find good timeouts. I don't think that compile/deploy for each test is the right approach.

> or you need to be able to configure it ?

That was my intention of this.

sbernard31:
:thinking:  and if you are using the key in plugtests & sample but you don't add it too `NetworkConfig` ?

boaks:
> and if you are using the key in plugtests & sample but you don't add it too 

As `BenchmarkClient.KEY_BENCHMARK_CLIENT_THREADS`?

If so, I would rename them then to "EXTKEY_"  and use that pattern also for the key value (e.g. "")

private static final String EXTKEY_BENCHMARK_CLIENT_THREADS = "EXT_BENCHMARK_CLIENT_THREADS";

sbernard31:
As we don't really use it in `californium-core`, that sounds better to me. :-)

boaks:
Trying it, it ends up in 3 copies of the same key ;-).

sbernard31:
I have pretty much the same kind of issue with Leshan. Our demos contain some duplicated code :/.
I prefer to have duplicate code in demo rather than having useless code in the core library.
I often ask myself if I should create a common project for demo but never did it ...

Anyway, you know my preferences and there is no perfect solution so I let you choose where to put the :hankey:. :grin:

boaks:
I removed the new key and "miss"use the ACK_TIMEOUT. For the most experiments, that timeout is rather related to the transmission and not the calculations on the nodes. That will fail for ecc on too constraint nodes but for now, I think it's the best way to do it.

PR #1201, opened 2020-01-22
Support custom socket-impl for dtls connector.

boaks:
Schedule retransmission after sendFlightOverNetwork.


boaks:
Supports to inject a custom DatagramSocketImpl. Makes implementations for at-command modems or bluetooth transmission easier.

PR #1202, opened 2020-01-23
OSCORE: Support for (inner) Block-Wise functionality

rikard-sics:
This pull request adds support for using OSCORE with Block-Wise requests and responses. This is specifically for the "inner Block-Wise" mode of OSCORE where the Block CoAP options are encrypted.

See [https://tools.ietf.org/html/rfc8613#section-4.1.3.4.1](https://tools.ietf.org/html/rfc8613#section-4.1.3.4.1)

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
@@ -112,7 +112,7 @@ public class ObjectSecurityLayer extends AbstractLayer {
 	}
 
 	@Override
-	public void sendRequest(Exchange exchange, Request request) {
+	public void sendRequest(final Exchange exchange, final Request request) {
 		Request req = request;
 		if (shouldProtectRequest(request)) {
 			try {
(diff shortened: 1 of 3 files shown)


cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
```diff
@@ -152,6 +152,16 @@ public void sendRequest(Exchange exchange, Request request) {
 					@Override
 					public void onReadyToSend() {
 						Token token = preparedRequest.getToken();
+
+						// For Block-Wise requests update the Token in the
+						// original request so the request/response matching
+						// works. (OSCORE creates a new request so the
+						// MessageObserver attached in the BlockwiseLayer will
+						// not do this.)
+						if (request.getOptions().hasBlock1()) {
```
boaks: The blockwise message observers are just try to read that from the "block-request" and transfer it to the "outer-request". But the "block-request", is replaced by the "prepared-request" and therefore don't contain that token any longer. To fix that, I introduced right now (master) a `Message.add(index, messageObserver)` and if that MessageObserver is added at the head, it works. It transfers the token from the "prepared-request" to the "block-request" before the other observer then forwards this token again.

```
preparedRequest.addMessageObserver(0, new MessageObserverAdapter() {

	@Override
	public void onReadyToSend() {
		Token token = preparedRequest.getToken();

		// add at had of message observers to update
		// the token of the original request first,
		// before calling other message observers!
		if (request.getToken() == null) {
			request.setToken(token);
		}

		ctxDb.addContext(token, finalCtx);
		ctxDb.addSeqByToken(token, seqByToken);
	}
});
```

rikard-sics: Thanks alot. I will update the code to use this method.

rikard-sics: I have now updated the code to use this method instead in commit b470ba265e78cf61c898039623e0fa4586e48614

boaks:
Thanks! 
LGTM
(Please, squash the commits.)

rikard-sics:
> Thanks!
> LGTM
> (Please, squash the commits.)

Thanks!

The commits have now been squashed.

PR #1204, opened 2020-01-27
Use scheme specific endpoint pool.

boaks:
Enables to use encryption for outgoing requests.
Pass network config to endpoint pool.


boaks:
This only removes API, which is introduced by "incremental improving the proxy".
No 2.0 API is removed, but a lot of stuff is deprecated, because it doesn't comply to RFC7252 in the "final version".

PR #1208, opened 2020-01-31
Revert proxy to 2.0.0.

boaks:
Adapt logging to org.slf4j.


boaks:
The PR with the new proxy in new modules will come next week.

PR #1214, opened 2020-02-06
Fix requests extra options

boaks:
Ensure backwards compatibility for request setting uri-path or uri-query
before URI.

Plugtest failed after cleanup Request. This enable to set the uri-query
before the URI again. Will be changed for 3.0.0, which requires then to
set the URI before additional options.


boaks:
Maybe this behavior of keeping the uri-query-option is intended to use the CoapClient advanced(Request)  function in combination with the CoapClient URI. But that still seem to be more surprising than intended to me.

PR #1219, opened 2020-02-12
Fix reregistration leak.

boaks:
Prevent the cleanup- and observation-message-observer from being
transferred to the reregistration.
(Deprecated healthchecks must also be excluded from being transferred.)

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
@@ -39,6 +39,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.eclipse.californium.core.coap.InternalMessageObserver;
 import org.eclipse.californium.core.coap.CoAP;
 import org.eclipse.californium.core.coap.MessageObserver;
 import org.eclipse.californium.core.coap.Request;
@@ -158,9 +159,14 @@ public class CoapObserveRelation {
 			refresh.setOptions(request.getOptions());
 
 			// use same message observers
-			for (MessageObserver mo : request.getMessageObservers()) {
-				request.removeMessageObserver(mo);
-				refresh.addMessageObserver(mo);
+			for (MessageObserver observer : request.getMessageObservers()) {
+				if (observer instanceof InternalMessageObserver) {
+					if (((InternalMessageObserver) observer).isInternal()) {
+						continue;
+					}
+				}
+				request.removeMessageObserver(observer);
+				refresh.addMessageObserver(observer);
 			}
 
 			this.request = refresh;
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/InternalMessageObserver.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/InternalMessageObserver.java
@@ -1,0 +1,33 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Bosch.IO GmbH and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v20.html
+ * and the Eclipse Distribution License is available at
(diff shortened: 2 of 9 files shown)


sbernard31:
Tell me if I'm wrong but we are principally interested in keeping the `ObserveMessageObserverImpl` created in `CoapObserveRelation.observe() ` ? 

If I'm right, I would prefer we only transfer this observer from register request to reregister one.

My  concern with the current approach is about creating an interface which is only needed for `CoapClient`/`CoapObserverRelation` but we need to ensure to use it correctly anywhere else ?

I mean If we need an interface maybe its better to create a specific marker interface which should be used if you want  your `messageObserver` to be transferred from register to re-register request ? (This avoid a kind of indirect dependency of `CoapObserverRelation` on the rest of the stack)

boaks:
> Tell me if I'm wrong but we are principally interested in keeping the ObserveMessageObserverImpl created in CoapObserveRelation.observe() ?
> 
> If I'm right, I would prefer we only transfer this observer from register request to reregister one.

That depends more on the interpretation of the "message observer" contract in combination with other topic, as observe/notify. At least some unit test may depend on more "message observers". So my impression: all message observers added outside the californium core are intended to be kept.

> My concern with the current approach is about creating an interface which is only needed for CoapClient/CoapObserverRelation but we need to ensure to use it correctly anywhere else ?

Yes, as soon as someone got aware of such problems somewhere else. I did a "fast check" for the core and didn't found any critical stuff. But maybe the future shows something else. Anyway, this improves also future fixes.

> creating an interface

That's only done to not break the API. For the next major version it will be merge back to the `MessageObserver`. Therefore I mark all such interfaces with `@PublicAPIExtension(type = ???.class)`. 

> I mean If we need an interface maybe its better to create a specific marker interface which should be used if you want your messageObserver to be transferred from register to re-register request ?
 
That would force all, which interprets the message observer contract in combination with observe/notify to adapt their code. It would compile, but breaks the expectations.

sbernard31:
> That would force all, which interprets the message observer contract in combination with observe/notify to adapt their code. It would compile, but breaks the expectations.

Correct only if you consider that followed all message observer to re-register is the expected behavior which is not clear at all.  The API say that you put a messageObserver on a request and not on a ObserveRelation. So this could be consider as an unexpected behavior.

I still feel that this indirect dependency is not a so good design, that's why I would prefer to isolate the use case.  
But anyway feel free to integrate this if you don't think the same. :)

boaks:
> I still feel that this indirect dependency is not a so good design

I just don't get, why you consider that as indirect dependency?

`MessageInterceptor` are added for application or core-internal usage. Therefore, they need to be moved sometimes, and sometimes not. 
(Sometimes ago, when I started to fix this leak, I used the term AutoMessageInterceptor because they are added automatically by the core. I tried also a version, where such AutoMessageInterceptor are removing them self. But that remove caused some more pain and so I decided to just not move them).

sbernard31:
> I just don't get, why you consider that as indirect dependency?

I try to explain better.

A `MessageObserver` is something to observe a message.
But in your case you want to transfer some observer from one message(register) to another (re-register). This is specific to the concept of re-register of a coapclient which is currently isolated.
To do that you introduce a new concept (via a new interface or later merge in the MessageObserver) which impact all the CoAP layer stack. (This is what I call indirect dependency, because all the stack should be coded in a way that this re-register feature work : that sounds not so good to me)

My position is if we can implement this without this indirect dependency, I think we should. (of course if there is no other solution, we add the indirect dependency)

Here my feeling is that you don't want to observe a `Message` but an `ObserveRelation`, so I would add a new (class/interface or whatever like an `ObserveRelationObserver`) which introduces/clarifies this concept and which will be isolated to the re-register use case.

boaks:
OK. My impression was, that the observers are copied/moved more around, so for me this wasn't related only to the reregistration. I remember also some discussion about these observers related to the blockwise stuff. But yes, for single suppress the leak in reregistration  an other solutions may also work.

sbernard31:
> OK. My impression was, that the observers are copied/moved more around, so for me this wasn't related only to the reregistration. 

But the new interface seems to be introduce specifically for this usage. (Or at least this is what say the [javadoc](https://github.com/eclipse/californium/pull/1219/files#diff-639eda4f0eacdc8c8b3b45625dd8c4a5R21)). 

> I remember also some discussion about these observers related to the blockwise stuff.

Yes and I think that current behavior/API about `MessageObserver` and transparent blockwise is not clean too. (But this means more time and maybe this will not be so easy)

In this specific use case (re-registration), it seems to me that adding a new `ObserveRelationObserver` should be easily doable (maybe I'm wrong) and should make the code clearer.

PR #1221, opened 2020-02-17
Fix reregistration

boaks:
Next try :-).

Implement the request-pending monitor using a `MessageObserver` instead of the order-filtered `onResponse`. Reset the notification-orderer before sending the reregistration-request (resolving a rare race condition). Update the observe option value of the reregistration-response, otherwise that may be filtered out by the notification-orderer.
Add separate unit test for proactive cancel.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapResource.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapResource.java
@@ -63,6 +63,7 @@ import org.eclipse.californium.core.server.resources.CoapExchange;
 import org.eclipse.californium.core.server.resources.Resource;
 import org.eclipse.californium.core.server.resources.ResourceAttributes;
 import org.eclipse.californium.core.server.resources.ResourceObserver;
+import org.eclipse.californium.core.server.resources.ResourceObserverAdapter;
 
 /**
  * CoapResource is a basic implementation of a resource. Extend this class to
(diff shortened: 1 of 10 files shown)


californium-core/src/main/java/org/eclipse/californium/core/CoapResource.java
```diff
@@ -341,20 +342,35 @@ public void checkObserveRelation(Exchange exchange, Response response) {
...
 			return; // because request did not try to establish a relation
 		}
 		if (CoAP.ResponseCode.isSuccess(response.getCode())) {
-			response.getOptions().setObserve(notificationOrderer.getCurrent());
-			
+
 			if (!relation.isEstablished()) {
 				relation.setEstablished();
+				// work-around for missing return value of addObserveRelation
+				// maybe replaced by next major version
```
sbernard31: Do you mean you add this observer just in case someone override `addObserveRelation`  (or I don't get your point) ?

boaks:
> CoapObserveRelation.cancel is executed just before BaseMatcher.registerObserver

For the proactive cancel, I hope, [onResponse - without observe option](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/BaseMatcher.java#L321) removes it, when the cancel request gets it's response without a observe option.

For a reactive response, maybe ... 
(From my side, I currently don't have too much time to spend into it :-) ).

californium-core/src/main/java/org/eclipse/californium/core/CoapResource.java
boaks:
```
public void addObserveRelation(ObserveRelation relation);
```

It's part of a public interface, and FMPOV, part of the "public API", though implementing a specific `CoapResource` is very basic for californium's usage. I don't see, that something prevents someone from overriding that method (a final would do so, but it's not used). so, yes, adding a boolean return breaks in my opinion the API and therefore I added the work-around with the comment.

sbernard31:
> For the proactive cancel, I hope, onResponse - without observe option removes it, when the cancel request gets it's response without a observe option.

You probably right.

> For a reactive response, maybe ...
(From my side, I currently don't have too much time to spend into it :-) ).

I understand, not a big deal.

boaks:
> Maybe a solution would be to mark the request as "re-register" (maybe using Request.UserContext or something else) and for re-register add in the store only in a previous observation with the same token/path already exists.

My feeling is, that ensure the reregistration request gets canceled, would remove the observation. But I spend already too much time into the reactive-cancel, so I will not go for that for the next weeks. The proactive case was one of the "failure hotspots" on my "continued testing system", therefore I want to have it fixed.

PR #1224, opened 2020-02-20
Add command and control support to hono client.

boaks:


boaks:
I would like to merge this at the end of this week, Thursday. 
It only affects the hono demo client.

sbernard31:
no objection

PR #1226, opened 2020-02-25
Add DtlsHandshakeTimeoutException in scandium

### LINKED ISSUE #1223: Add a specific exception for HandshakeTimeout ?
In Leshan, when a [request timeout for any reason](https://github.com/eclipse/leshan/wiki/Request-Timeout) we raise a TimeoutException.
We decide recently to consider a Handshake flight timeout as Leshan request timeout.

But currently there no easy way to know handshake timeout because we raise an `Exception`.
So the code looks like something like this : 
```java
    if (error instanceof Exception && error.getMessage().startsWith("Handshake flight")) {
```
Should we raise a specific exception ? 

(Ideally I would like to be able to distinguish a CoAP timeout from a Blockwise timeout but I guess this would be a really more complex task , this is maybe relative to https://github.com/eclipse/californium/issues/818 ?)

sbernard31:
It aims to fix #1223.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -183,7 +183,9 @@ import org.eclipse.californium.scandium.dtls.ConnectionIdGenerator;
 import org.eclipse.californium.scandium.dtls.ContentType;
 import org.eclipse.californium.scandium.dtls.DTLSFlight;
 import org.eclipse.californium.scandium.dtls.DTLSSession;
+import org.eclipse.californium.scandium.dtls.DtlsException;
 import org.eclipse.californium.scandium.dtls.DtlsHandshakeException;
+import org.eclipse.californium.scandium.dtls.DtlsHandshakeTimeoutException;
 import org.eclipse.californium.scandium.dtls.HandshakeException;
 import org.eclipse.californium.scandium.dtls.HandshakeMessage;
 import org.eclipse.californium.scandium.dtls.Handshaker;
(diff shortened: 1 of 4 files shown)


sbernard31:
I don't know if this : 
```java
// limit of retransmissions not reached
// but handshake expired during Android / OS "deep sleep"
message = " Stopped by expired realtime!";
```
should be consider as a timeout.

sbernard31:
Not a big deal, but I didn't get the difference between : 

```java
else if (tries > max) {
	LOGGER.debug("Flight for [{}] has reached timeout, discarding ...", flight.getPeerAddress());
	message = " Stopped by timeout!";
	timeout = true;
} 
```
and
```java
else {
	LOGGER.debug(
			"Flight for [{}] has reached maximum no. [{}] of retransmissions, discarding ...",
			flight.getPeerAddress(), max);
	message = " Stopped by timeout after " + max + " retransmissions!";
	timeout = true;
}
```

sbernard31:
Not linked directly to this PR but I failed to launch single scandium test, neither in eclipse (tests failed) or with maven `mvn -Dtest=DTLSConnectorTest test`  :thinking:

boaks:
> -Dtest=DTLSConnectorTest -DfailIfNoTests=false

on the "parent" works "on my machine".

sbernard31:
It works on my machine too at parent level.

boaks:
> It works on my machine too at parent level.

So, you complain about executing it at module level "scandium"?

boaks:
>  I didn't get the difference between

The first is used for "isEarlyStopRetransmission".

boaks:
> should be consider as a timeout.

I would so.

sbernard31:
> The first is used for "isEarlyStopRetransmission".

I get it now.

> So, you complain about executing it at module level "scandium"?

"Complain" that was not my intention :sweat_smile:. I just share an issue I faced (Mainly to know if this is a shared/intended/known behavior)

> I would so.

Done.

boaks:
LGTM

PR #1227, opened 2020-02-25
Move attributes from Builder to DtlsConnectorConfig.

### LINKED ISSUE #1225: `recommendedCipherSuiteOnly` is not hold by `DtlsConnectorConfig`.
About `DtlsConnectorConfig`, `clientOnly` and `recommendedCipherSuitesOnly` attributes belong to the builder instead of the `DtlsConnectorConfig`.

The way it is used in Leshan, this information is lost. (see [CaliforniumEndpointsManager](https://github.com/eclipse/leshan/blob/leshan-1.0.0-M13/leshan-client-cf/src/main/java/org/eclipse/leshan/client/californium/CaliforniumEndpointsManager.java#L86))

Could we move those in `DtlsConnectorConfig` ?

sbernard31:
It aims to fix #1225.

### DIFF
--- a/californium-proxy2/.gitignore
+++ b/californium-proxy2/.gitignore
@@ -1,2 +1,2 @@
-/Proxy.properties
+/Proxy2.properties
 /ProxyMapping.properties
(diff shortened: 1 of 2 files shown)


californium-proxy2/.gitignore
```diff
@@ -1,2 +1,2 @@
-/Proxy.properties
+/Proxy2.properties
```
boaks: Thanks!

boaks:
LGTM

PR #1238, opened 2020-03-09
Introduce message offload to reduce heap consumption for deduplication.

boaks:
Add configuration switch for offloading incoming requests when sending
the response.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapObserveRelation.java
@@ -155,6 +155,7 @@ public class CoapObserveRelation {
 				.getLong(NetworkConfig.Keys.NOTIFICATION_REREGISTRATION_BACKOFF);
 		this.scheduler = executor;
 		this.request.addMessageObserver(pendingRequestObserver);
+		this.request.setProtectFromOffload();
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


sbernard31:
Interesting idea :)

Did you have some number about memory gain about that ? 
(From my side each time, I looked at memory consumption the main hot stop was dtls connection store)

boaks:
> Did you have some number about memory gain about that ?

That depends on the situation. Though only incoming requests, which are stored in the deduplicator are offloaded, it requires some load to see that (benchmark.sh, but comment out the proxy test). The benchmark uses 40 bytes messages, so the memory statistic in visualvm shows for that case some improvements, but not too large ones (it' more the frequency of the gc).

Additionally the benchmark client also offloads responses, but that only effective for separate responses, which are not too much used in the benchmarks.

The intention of this PR is, to reduce the heap used by the deduplicator. Mainly in order to make that more robust.

boaks:
Just to add:
It doesn't use less heap, it just release not longer used heap earlier.

sbernard31:
> It doesn't use less heap, it just release not longer used heap earlier.

Yep, I get it.

Should I review this more deeply ? (or do you plan to modify it a bit more ?)

boaks:
Let's go for your stuff first - I will need some more time and currently I'm too busy with other jobs.

boaks:
On the client side it's up to the application implementation, to use the offloading.
I was mainly interested on the server side.

PR #1239, opened 2020-03-10
Enhance memory consumption

sbernard31:
After reading #1238, I looked at some memory dump and find ways to reduce a bit memory consumption at scandium side.
I measured value with visualVM. I did 10000 handshake on ExampleDTLSServer.java, then I perform a GC.

**Do not keep unnecessary reference on handshaker : e26ecd3**
This allow the JVM to garbage collect the `Handshaker` just after the handshake is complete instead of waiting the 60s. (after the 60s there is no more benefits)
(before : 77MB, after : 43MB, gain 34MB for 10000 connections)
I don't see any drawback here.

**Do not store the whole ClientHello message : 1450641**
This allow the JVM to garbage collect the `ClientHello`  with the handshaker and only keep necessary value. (after the 60s there is no more benefits)
(before : 43MB, after : 33MB : gain 10MB for 10000 connections)
Maybe a little drawback is that after the 60s we use some more byte to store the seqNumber but I think this is worthy.

**Use only 1 string object for identity and name if possible : d86d474**
This is a very little optimization without any drawback.
(before : 33MB, after : 32MB : gain 1MB for 10000 connections)

**Use array instead of HashMap to store sequence numbers : 6385e37**
Looking at heap dump I see that seq numbers uses a HashMap with an internal cache array with a length of 15. As we do not allow renegotiation I think that we can replace it by a simple array[2].
(before: 32MB, after : 28 MB : gain 4MB for 10000 connections)
If instead of using a simple array we use an hashmap iniatialised to 2. (and so with an array of 4 internally) the gain is ~1MB for 10000 connections)

We could gain ~1MB more if `SessionID.text` attribute was removed but this means it must be regenerate each time we send a message. (see DTLSSession.getConnectionWriteContext()`)

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/auth/PreSharedKeyIdentity.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/auth/PreSharedKeyIdentity.java
@@ -73,8 +73,8 @@ public final class PreSharedKeyIdentity extends AbstractExtensiblePrincipal<PreS
 			throw new NullPointerException("Identity must not be null");
 		} else {
 			scopedIdentity = sni;
-			StringBuilder b = new StringBuilder();
 			if (sni) {
+				StringBuilder b = new StringBuilder();
 				if (virtualHost == null) {
 					this.virtualHost = null;
 				} else if (StringUtil.isValidHostName(virtualHost)) {
@@ -84,16 +84,16 @@ public final class PreSharedKeyIdentity extends AbstractExtensiblePrincipal<PreS
 					throw new IllegalArgumentException("virtual host is not a valid hostname");
 				}
 				b.append(":");
+				b.append(identity);
+				this.name = b.toString();
 			} else {
 				if (virtualHost != null) {
 					throw new IllegalArgumentException("virtual host is not supported, if sni is disabled");
 				}
 				this.virtualHost = null;
+				this.name = identity;
 			}
 			this.identity = identity;
-
-			b.append(identity);
-			this.name = b.toString();
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


boaks:
e26ecd3 - looks good

1450641 - "Note: used outside of serial-execution!" - so may require some kind of protection. I'm not sure about that.

d86d474  - looks good. Cosmetics: maybe the ifs could be merged into one

6385e37 - looks good

boaks:
1450641 - also the old solution may have surprises (I forgot to load the field into a local variable before check and using it). I would go for a class { Random, MessageSeq} and such a local variable to get gain a more atomic behaviour.

sbernard31:
1450641 - I will give a try tomorrow.

boaks:
I analyzed it again:
The fields are accessed three times. Twice within `processNewClientHello` protected by a synchronized. And once from a timer started in `handshakeCompleted`.
So, I would adapt the 

```
public boolean isStartedByClientHello(ClientHello clientHello) {
	if (clientHello == null) {
		throw new NullPointerException("client hello must not be null!");
	}
        Random startingClientHelloRandom = this.startingClientHelloRandom;
	if (startingClientHelloRandom != null) {
		if (startingClientHelloRandom.equals(clientHello.getRandom())) {
			if (startingClientHelloMessageSeq >= clientHello.getMessageSeq()) {
				return true;
			}
		}
	}
	return false;
}
```
and

```
public void startByClientHello(ClientHello clientHello) {
	if (clientHello == null) {
		startingClientHelloRandom = null;
	} else {
		startingClientHelloMessageSeq = clientHello.getMessageSeq();
		startingClientHelloRandom = clientHello.getRandom();
	}
}
```

that should do it. Reading the states will then either see the last starting client hello or the `null`.

sbernard31:
About ClientHello I changed it as you proposed.
I changed the code here too : https://github.com/eclipse/californium/pull/1239/commits/f631a0e56f0a2fc9b931fdf4b3637d5a679136bf#diff-f7d385d8c88c68885e577bef2d4c909cL569
(I do not set the startingClientHelloMessageSeq to 0 like you proposed for startByClientHello())

About PreSharedKeyIdentity, I rewrite it a bit trying to remove the "double if"

boaks:
LGTM

PR #1243, opened 2020-03-16
Add unix health logger based on netstat.

boaks:
Read /proc/net/snmp to access udp diagnose statistics from OS.

### DIFF
--- a/cf-unix-health/pom.xml
+++ b/cf-unix-health/pom.xml
@@ -1,0 +1,95 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<project
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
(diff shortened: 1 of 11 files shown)


boaks:
> (By the way for command line : my first impression of https://picocli.info/ was really good)

Maybe I missed something, but, I wouldn't name a 360K library "pico" :-).

sbernard31:
I agree :D

Last time I take a look it was more arround 250K. This is still very fat :D.

But for demo or server I think it's OK

boaks:
> But for demo or server I think it's OK

That's right.

boaks:
What's about just use "commons-cli" as leshan?

sbernard31:
It's another possibility.

I tested both and my personal preference goes to **picocli** if you don't care so much about library size:
- easier to use,
- could also be used with jline to create some interactive command,
- support color.

If I would had time, I would have migrated to it for the 1.0.0. I will maybe do it later. (https://github.com/eclipse/leshan/issues/569)

PR #1246, opened 2020-03-17
Adjust handshake mode on retransmission.

boaks:

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/EndpointContextUtil.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/EndpointContextUtil.java
@@ -61,4 +61,32 @@ public class EndpointContextUtil {
 		}
 		return matchAll;
 	}
+
+	/**
+	 * Get follow-up endpoint context.
+	 * 
+	 * Using {@link DtlsEndpointContext#KEY_HANDSHAKE_MODE} requires to adjust
+	 * the reuse of the endpoint context in case of blockwise transfers or
+	 * retransmissions.
+	 * 
+	 * @param messageContext messages's endpoint context.
+	 * @param connectionContext the connection's endpoint context. Either
+	 *            reported with
+	 *            {@link RawData#onContextEstablished(EndpointContext)} or
+	 *            contained in a received message.
+	 * @return endpoint context to be used for follow-up messages.
+	 */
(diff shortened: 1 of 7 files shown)


boaks:
@sbernard31 

check, if you want the fix in 2.2.0. For me it will be also OK, to include that after the 2.2.0.

sbernard31:
Just to know, what could be the reason to not integrate it ? you consider you don't test it enough ?

boaks:
> you consider you don't test it enough ?

Yes, because of the short time. And it now changes the `Request`'s `EndpointContext` itself. I assume, this will be not too often effective, because the HandshakeModes are rarely used.

sbernard31:
Just to be sure this fix is only needed when default handshake mode is AUTO ?

I mean in Leshan we set a default handshake mode to NONE and set manually on the request  `HANDSHAKE_MODE_AUTO` when needed ?

boaks:
It keeps NONE and remove the others. If leshan uses only NONE, we don't need it for 2.2.0.

sbernard31:
:+1:

element-connector/src/main/java/org/eclipse/californium/elements/EndpointContextUtil.java
```diff
@@ -61,4 +61,33 @@ public static boolean match(String name, Set<String> keys, EndpointContext conte
...
+	 * 
+	 * @param messageContext messages's endpoint context.
+	 * @param connectionContext the connection's endpoint context. Either
+	 *            reported with
+	 *            {@link RawData#onContextEstablished(EndpointContext)} or
+	 *            contained in a received message.
+	 * @return endpoint context to be used for follow-up messages.
+	 * @since 2.3
+	 */
+	public static EndpointContext getFollowUpEndpointContext(EndpointContext messageContext,
```
sbernard31, line 80: clearly a detail but I would change the name to make clear this is for block. (even if this is already mention in the javadoc)

boaks: It's also for "retransmissions of CONs". 
So I'm not sure, which name you have in mind?

sbernard31: correct forget it !

PR #1248, opened 2020-03-18
OSCORE: Added case when context re-generation is initiated by the server

rikard-sics:
This pull request adds the case for the context re-generation functionality where it is the server that loses the mutable parts of the OSCORE context and thus takes initiative to start the context re-derivation procedure.

That is the client will not explicitly send Request 1 as described in the context re-derivation procedure below. Rather the client will sends a normal request and since it is the server that has lost the mutable parts of the context it is the server that will then continue with the re-derivation procedure.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
@@ -50,7 +50,7 @@ public class ContextRederivation {
 	 *
 	 */
 	public static enum PHASE {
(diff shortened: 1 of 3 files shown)


boaks:
Please add the "revapi ignore".

```
	<build>
		<plugins>
			<plugin>
				<artifactId>maven-assembly-plugin</artifactId>
				<configuration>
					<descriptorRefs>
						<descriptorRef>enhanced-jar-with-licenses</descriptorRef>
					</descriptorRefs>
				</configuration>
			</plugin>
			<plugin>
				<groupId>org.revapi</groupId>
				<artifactId>revapi-maven-plugin</artifactId>
				<configuration>
					<!-- not depolyed to repos -->
					<skip>true</skip>
				</configuration>
			</plugin>
		</plugins>
	</build>
```

boaks:
LGTM
I will add the ignore by myself.

rikard-sics:
> LGTM
> I will add the ignore by myself.

Thanks for the merge!

Issue #1245, opened 2020-03-17
NPE in serverHandshaker.


sbernard31:
If  `CertificateVerifier.getAcceptedIssuers()` returns null this generate an NPE.

```
unexpected error occurred:
java.lang.NullPointerException: null
	at java.util.Objects.requireNonNull(Objects.java:203)
	at java.util.Arrays$ArrayList.<init>(Arrays.java:3813)
	at java.util.Arrays.asList(Arrays.java:3800)
	at org.eclipse.californium.scandium.dtls.ServerHandshaker.createCertificateRequest(ServerHandshaker.java:532)
	at org.eclipse.californium.scandium.dtls.ServerHandshaker.receivedClientHello(ServerHandshaker.java:400)
	at org.eclipse.californium.scandium.dtls.ServerHandshaker.doProcessMessage(ServerHandshaker.java:200)
	at org.eclipse.californium.scandium.dtls.Handshaker.processMessage(Handshaker.java:638)
	at org.eclipse.californium.scandium.DTLSConnector.startNewHandshake(DTLSConnector.java:1842)
	at org.eclipse.californium.scandium.DTLSConnector.processClientHello(DTLSConnector.java:1737)
	at org.eclipse.californium.scandium.DTLSConnector.access$1500(DTLSConnector.java:217)
	at org.eclipse.californium.scandium.DTLSConnector$12.run(DTLSConnector.java:1675)
	at org.eclipse.californium.elements.util.SerialExecutor$1.run(SerialExecutor.java:276)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

The javadoc says : 

```java
/**
 * Return an array of certificate authority certificates which are trusted
 * for authenticating peers.
 * 
 * @return the trusted CA certificates (possibly <code>null</code>)
 */
X509Certificate[] getAcceptedIssuers();
```

So there is bug either in javadoc or in the code ?

My bet is that allowing null make sense as it seems optional to me to provide AcceptedIssuers. (I should double check this)

If there is a bug in the code we should fix it for the 2.2. (#1244)

sbernard31:
Reading [rfc5246§7.4.4](https://tools.ietf.org/html/rfc5246#section-7.4.4), I would say this is an issue in the code.
> certificate_authorities
[...]

PR #1249, opened 2020-03-18
Adapt javadoc of CertificateVerifier.

boaks:
Suppport only "empty array" for issuers.
null is considered to be valid return value.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/x509/CertificateVerifier.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/x509/CertificateVerifier.java
@@ -42,7 +42,11 @@ public interface CertificateVerifier {
 	 * Return an array of certificate authority certificates which are trusted
 	 * for authenticating peers.
 	 * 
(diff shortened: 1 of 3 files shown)


boaks:
See issue #1245

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/x509/CertificateVerifier.java
```diff
@@ -42,7 +42,11 @@
 	 * Return an array of certificate authority certificates which are trusted
 	 * for authenticating peers.
 	 * 
-	 * @return the trusted CA certificates (possibly <code>null</code>)
+	 * The javadoc of previous versions (2.1.0 and before) permits to use
+	 * {@code null}. This causes an failure, please adapt to use an empty array.
```
sbernard31:
```suggestion
	 * {@code null}. This causes a failure, please adapt to use an empty array.
```

boaks: :-) thanks!

PR #1252, opened 2020-03-19
OSCORE: Support for (outer) Block-Wise functionality

rikard-sics:
This pull request adds further support for using OSCORE with Block-Wise requests and responses. It implements the "outer Block-Wise" mode of OSCORE where the Block CoAP options are not encrypted.

A proxy could split a large OSCORE protected request before forwarding it to the server (or split a response before forwarding it to the client). In such case the block options will be unencrypted and external. That is, the individual block messages are not protected with OSCORE, as is the case with inner Block-Wise, but rather the whole message (before being split) is protected with OSCORE. Thus the message must be re-assembled by the block-wise layer before it is unprotected by OSCORE.
See https://tools.ietf.org/html/rfc8613#section-4.1.3.4.2

### DIFF
--- a/cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreOuterBlockwiseTest.java
+++ b/cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreOuterBlockwiseTest.java
@@ -1,0 +1,565 @@
+/*******************************************************************************
+ * Copyright (c) 2020 RISE SICS and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * 
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v20.html
+ * and the Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.html.
+ *
+ * This test class is based on org.eclipse.californium.integration.test.SecureBlockwiseTest
+ * 
+ * Contributors: 
+ *    Rikard Höglund (RISE SICS) - testing OSCORE outer Block-Wise messages
+ ******************************************************************************/
+package org.eclipse.californium.oscore;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
+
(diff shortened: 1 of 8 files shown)


rikard-sics:
Currently I am using functionality from californium-proxy2. If it is better I could switch to using californium-proxy instead.

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreOuterBlockwiseTest.java
```diff
@@ -0,0 +1,567 @@
...
+	/**
+	 * Create simple non-OSCORE proxy.
+	 * 
+	 * @param proxyRequestBlockwise the proxy sends requests with block-wise
+	 * @param proxyResponseBlockwiseEnabled the proxy responds with block-wise
+	 */
+	private void createSimpleProxy(final boolean proxyRequestBlockwise, final boolean proxyResponseBlockwiseEnabled) {
+
+		final Coap2CoapTranslator coapTranslator = new Coap2CoapTranslator();
+		final CoapUriTranslator uriTranslator = new CoapUriTranslator();
```
boaks: A `Coap2CoapTranslator` extends a `CoapUriTranslator`. Therefore you may replace the separate `uriTranslator` and just use the `coapTranslator` instead.

boaks:
> Currently I am using functionality from californium-proxy2. If it is better I could switch to using californium-proxy instead.

It's OK to stay with the new proxy2.

cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreOuterBlockwiseTest.java
rikard-sics: I see, I did not notice that. I have now updated the code accordingly in commit 4886d668163d16f0597577e74fe94c7c8a1535fe

rikard-sics:
> It's OK to stay with the new proxy2.

Alright, sounds good. I will stay with proxy2 then.

rikard-sics:
I found an issue not caught by the JUnit tests and fixed it in commit 861dabb1c7744a71bc20f667437b4d588aeaa18c

boaks:
LGTM

@sbernard31 

Does leshan 1.0 use OSCORE? If not, is it OK from your point to merge this for 2.2.0?

boaks:
@rikard-sics 

Please squash. I merge it either for the 2.2.0 of after that release.

rikard-sics:
> Please squash. I merge it either for the 2.2.0 of after that release.

Thanks, sounds good. I have now squashed the commits.

PR #1253, opened 2020-03-19
Signature algorithms extension support

ecimpsmain:
In our project we meet an issue where we had to provide the supported signature algorithms by the client. Otherwise the DTLS communication was not working. I have updated the library to add possibility to provide the supported signature algorithms.

PR #1256, opened 2020-03-25
Add sig hash extension for server-side

boaks:
Add SignatureAlgorithmsExtension for server-side.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -42,6 +42,7 @@ import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 import org.eclipse.californium.elements.DtlsEndpointContext;
@@ -213,7 +214,10 @@ public final class DtlsConnectorConfig {
 	/** the supported cipher suites in order of preference */
 	private List<CipherSuite> supportedCipherSuites;
 	
-	/** the supported signature and hash algorithms */
+	/**
+	 * the supported signature and hash algorithms.
+	 * @since 2.3
+	 */
 	private List<SignatureAndHashAlgorithm> supportedSignatureAlgorithms;
 
 	/** the trust store for RPKs **/
@@ -560,12 +564,17 @@ public final class DtlsConnectorConfig {
 	public List<CipherSuite> getSupportedCipherSuites() {
 		return supportedCipherSuites;
 	}
-	
+
 	/**
-	 * Gets the signature algorithms the connector should advertise in a DTLS
-	 * handshake.
+	 * Gets the supported signature and hash algorithms the connector should
+	 * advertise in a DTLS handshake.
 	 * 
-	 * @return the supported signature algorithms (ordered by preference)
+	 * @return the supported signature and hash algorithms (ordered by
+	 *         preference). If empty, the client does not advertise it's
+	 *         supported signature and hash algorithms, and the server assumes
+	 *         the {@link SignatureAndHashAlgorithm#DEFAULT} as list of
+	 *         supported signature and hash algorithms
+	 * @since 2.3
 	 */
 	public List<SignatureAndHashAlgorithm> getSupportedSignatureAlgorithms() {
 		return supportedSignatureAlgorithms;
@@ -1467,16 +1476,40 @@ public final class DtlsConnectorConfig {
 			}
 			return setSupportedCipherSuites(suites);
 		}
-		
+
 		/**
 		 * Sets the signature algorithms supported by the connector.
 		 * <p>
-		 * The connector will use these signature algorithms (in exactly the same
-		 * order) during the DTLS handshake.
+		 * The connector will use these signature algorithms (in exactly the
+		 * same order) during the DTLS handshake.
 		 * 
-		 * @param supportedSignatureAlgorithms the supported signature algorithms in the order of
-		 *            preference
+		 * @param supportedSignatureAlgorithms the supported signature
+		 *            algorithms in the order of preference. No arguments, if no
+		 *            specific extension is to be used for a client, and the
+		 *            server uses {@link SignatureAndHashAlgorithm#DEFAULT}.
 		 * @return this builder for command chaining
+		 * @since 2.3
+		 */
+		public Builder setSupportedSignatureAlgorithms(SignatureAndHashAlgorithm... supportedSignatureAlgorithms) {
+			List<SignatureAndHashAlgorithm> list = null;
+			if (supportedSignatureAlgorithms != null) {
+				list = Arrays.asList(supportedSignatureAlgorithms);
+			}
+			return setSupportedSignatureAlgorithms(list);
+		}
+
+		/**
+		 * Sets the signature algorithms supported by the connector.
+		 * <p>
+		 * The connector will use these signature algorithms (in exactly the
+		 * same order) during the DTLS handshake.
+		 * 
+		 * @param supportedSignatureAlgorithms the list of supported signature
+		 *            algorithms in the order of preference. Empty, if no
+		 *            specific extension is to be used for a client, and the
+		 *            server uses {@link SignatureAndHashAlgorithm#DEFAULT}.
+		 * @return this builder for command chaining
+		 * @since 2.3
 		 */
 		public Builder setSupportedSignatureAlgorithms(List<SignatureAndHashAlgorithm> supportedSignatureAlgorithms) {
 			config.supportedSignatureAlgorithms = supportedSignatureAlgorithms;
@@ -2416,6 +2449,9 @@ public final class DtlsConnectorConfig {
 				throw new IllegalStateException("PSK store set, but no PSK cipher suite!");
 			}
 
+			if (config.supportedSignatureAlgorithms == null) {
+				config.supportedSignatureAlgorithms = Collections.<SignatureAndHashAlgorithm>emptyList();
+			}
 			if (!certifacte) {
 				if (config.privateKey != null || config.publicKey != null) {
 					throw new IllegalStateException("Identity set, but no certificate based cipher suite!");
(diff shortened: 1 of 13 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -521,13 +528,17 @@ private void createServerKeyExchange(final ClientHello clientHello, final DTLSFl
 
 	private boolean createCertificateRequest(final ClientHello clientHello, final DTLSFlight flight) throws HandshakeException {
 
-		if ((clientAuthenticationWanted || clientAuthenticationRequired) && signatureAndHashAlgorithm != null) {
+		if ((clientAuthenticationWanted || clientAuthenticationRequired) && negotiatedSignatureAndHashAlgorithms != null
+				&& !negotiatedSignatureAndHashAlgorithms.isEmpty()) {
```
sbernard31: `negotiatedSignatureAndHashAlgorithms != null && !negotiatedSignatureAndHashAlgorithms.isEmpty()`

**I probably didn't get the point here**, but this sounds a bit strange that if `negotiatedSignatureAndHashAlgorithms` or `negotiatedSignatureAndHashAlgorithms` are `null` we just don't create a certificate request.

This sounds more critical to me.

boaks:
> Anyway reading the specification, that default signature_algorithm seems to be {sha1,ecdsa}
> And It seems to me that scandium use {sha256, ecdsa} as default, so I probably missed something thinking

The x509 certificates must match that sig-hash-algorithm and it seems, that the java keytool generates the certificates with {sha256,ecdsa}. Therefore the old implementation used that as fixed list in the ServerHandshaker

```
// TODO SHA256withECDSA is default but should be configurable
signatureAndHashAlgorithm = new SignatureAndHashAlgorithm(SignatureAndHashAlgorithm.HashAlgorithm.SHA256, SignatureAndHashAlgorithm.SignatureAlgorithm.ECDSA);
```

There was also a discussion in the IETF tls-mailing list about deprecating md5/sha1 for certificate signing.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1467,7 +1467,25 @@ public Builder setSupportedCipherSuites(String... cipherSuites) {
...
+		/**
+		 * Sets the signature algorithms supported by the connector.
+		 * <p>
+		 * The connector will use these signature algorithms (in exactly the same
+		 * order) during the DTLS handshake.
+		 * 
+		 * @param supportedSignatureAlgorithms the supported signature algorithms in the order of
+		 *            preference
+		 * @return this builder for command chaining
+		 * @throws NullPointerException if the given array is <code>null</code>
```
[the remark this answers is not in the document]
boaks: yes.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
```diff
@@ -341,6 +339,18 @@ public void addSignatureAlgorithm(SignatureAndHashAlgorithm signatureAndHashAlgo
...
 	}
 
+	/**
+	 * Appends a list of signature and hash algorithms to the end of the list of supported algorithms.
+	 * <p>
+	 * The algorithm's position in list indicates <em>least preference</em> to the
+	 * recipient (the DTLS client) of the message.
+	 * 
+	 * @param signatureAndHashAlgorithms The algorithms to add.
+	 */
```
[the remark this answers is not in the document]
boaks: I will add the tags.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks: If there are no common sig_hash_algo, the client will not  be able to send successfully it's certificate, though [isSignedWithSupportedAlgorithm](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java#L499) will fail.

boaks: Before this approach, it is "assumed/required", that both sides support the {sha256,ecdsa}. And though all californium certificates use that, all work well. But it comes also with some inflexibility. The implementation was mainly triggered by PR #1253 , though I prefer to support functionality on both sides.

sbernard31: I'm totally agree to integrate the support of this extention :+1: 

On this point : 
 -  either `negotiatedSignatureAndHashAlgorithms` can not be null or empty because check was done before (and so this check is useless)
 - or if it is null or empty and null or empty means the is no  "compatible" sign/hash then we must stop the handshake instead of just "not sending the CertificateRequest"

Maybe this is the current behavior but when you only read this method, it doesn't look like.

boaks:
Tags "since 2.3" are added.

sbernard31:
Ok I'm lost so I warn you I will ask a lot of questions :sweat_smile:

>The x509 certificates must match that sig-hash-algorithm

So when client initiated a handshake, it says :"hey I support all those sign_hash".
1.This make sense only for RPK/x509, right?
2 .Then server check if one sign_hash match. If there is "no match" => handshake failure ?
3. What means match exactly, reading the code it seems it should match `supportedSignatureAndHashAlgorithms` of the connector but reading you I finally understand this must match the certificate(or pubkey?) sign_hash ?

boaks:
> 1.This make sense only for RPK/x509, right?

Yes.

> 2 .Then server check if one sign_hash match. If there is "no match" => handshake failure ?

The server checks for common sign_hashs and checks its certificate-path, if that matches. Generally, also a PSK suite may chosen, so the mismatch will cause "no common cipher suites".

> 3. What means match exactly, reading the code it seems it should match supportedSignatureAndHashAlgorithms of the connector but reading you I finally understand this must match the certificate(or pubkey?) sign_hash ?
 
It must match both, the supported sign_hash and the key or cert-path.

[7.4.1.4.1.  Signature Algorithms](https://tools.ietf.org/html/rfc5246#section-7.4.1.4.1)

"The semantics of this extension are somewhat complicated because ..."

In my opinion the RFC tries to add this new functionality an keep backwards compatibility. That makes it somehow illogical.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks: In case of a configuration for both PSK and EC cipher suites, the negotiated list maybe empty. Though the `createCertificateRequest` is always called, also for PSK, it checks itself, if the request should be included or not.

sbernard31: just to be clear for PSK this does make sense to send a certificateRequest ?

boaks: Though PSK requires the same secret on both sides and with that, no additional data is needed, PSK doesn't use a certificate request.

boaks: Just additionally: 
The signing (+ hash) itself is used for the certificate chain and the "server key exchange" and the (client) "certificate verify" message. It's send in the "certificate request" to ensure, that the clients certificate and "certificate verify" also matches the server's limitations.

sbernard31: > Though the createCertificateRequest is always called

So it seems to me not so intuitive, I think we should not call this method for PSK.

Now, imagine that PSK is excluded. What happens if `negotiatedSignatureAndHashAlgorithms` is null or empty. It seems to me that this should stop the handshake and here this looks like it just imply that we will not send the CertificateRequest.

sbernard31:
> It must match both, the supported sign_hash and the key or cert-path

But "sign_hash of key/certpath" is a subset of "supported sign_hash" so checking againts "key/certpah" should be enough ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks:
```
private void negotiateCipherSuite(final ClientHello clientHello, final HelloExtensions serverHelloExtensions) throws HandshakeException {
	...
	// if none of the client's proposed cipher suites matches throw exception
	AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.HANDSHAKE_FAILURE, session.getPeer());
	throw new HandshakeException("Client proposed unsupported cipher suites only", alert);
}
```

If there is no matching cipher suite, the handshake is aborted here.

```
if (cipherSuite.requiresServerCertificateMessage()) {
	SignatureAndHashAlgorithm algo = getNegotiatedSignatureAndHashAlgorithm(signatures);
	if (algo != null) {
		negotiatedSignatureAndHashAlgorithms = signatures;
		session.setSignatureAndHashAlgorithm(algo);
	}
}
```

That check ensures, that `negotiatedSignatureAndHashAlgorithms` indicates, if ECDSA cipher suite is selected and therefore the signature and hash algorithms are negotiated.

For me you seems to have different idea's. I would prefer to add more docu to the current implementation.

boaks:
> But "sign_hash of key/certpath" is a subset of "supported sign_hash" so checking againts "key/certpah" should be enough ?

There are two things to do:
Check, if a certificate chain could be used because of the already contained signs.
Negotiate, which signature and hashs algorithm is to be used for signing using the public key during the handshake. So I'm not sure, why "checking againts "key/certpah" should be enough ?".

sbernard31:
I read some documentation and your comment and I think I better understand now.

A client send the "signature and hashs" extension to let know which algorithm it can execute.
Those  algorithms will be used to :
 - check certificate signature which will be send  in [Server Certificate](https://tools.ietf.org/html/rfc5246#section-7.4.2) (algo usage is fixed, it depends on stored certificates)
 - sign/check [Server Key Exchange](https://tools.ietf.org/html/rfc5246#section-7.4.3)  (algo usage is dynamic, it depends of server/client capabilities)

As far as I know scandium does not support multiple certificate selection. (I mean there is only 1 certchain/public key by endpoint)

So when the server get the CLIENT_HELLO, at cipher suite negotiation step,  for "certificate based cipher", It should ensure that : 
- it exists 1 or several algorithm supported by both (server and client)
- the client support the algorithm used in by server CertChain.

Correct ?

boaks:
> Correct ?

That's my understanding, even if the current implementation is not exactly like that. I opened a issue #1259 to discus the solution but I would prefer to have the two pending PRs merged before we polish the ciphersuite selection.

sbernard31:
> That's my understanding, even if the current implementation is not exactly like that

Good to hear.

>  I would prefer to have the two pending PRs merged before we polish the ciphersuite selection.

let's talk about it  #1259.

boaks:
So, merge this, discus the other and then discus your points in issue #1259 ?

sbernard31:
yes

PR #1258, opened 2020-03-26
Cleanup ecdhe

boaks:
Introduce redesigned `ECDHECryptography`as `XECDHECryptography`.

Update implementation based on RFC 8422.
Simpliy and cleanup ECDH related handshake messages.
Add XDH (X25519/X448) support for ECHD (java 11 only!).
Add configuration for supported groups (curves).
Add sigAls and curves for openssl interoparability test.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -60,6 +60,7 @@ import org.eclipse.californium.scandium.dtls.SessionCache;
 import org.eclipse.californium.scandium.dtls.SignatureAndHashAlgorithm;
 import org.eclipse.californium.scandium.dtls.cipher.CipherSuite;
 import org.eclipse.californium.scandium.dtls.cipher.CipherSuite.KeyExchangeAlgorithm;
+import org.eclipse.californium.scandium.dtls.cipher.XECDHECryptography.SupportedGroup;
 import org.eclipse.californium.scandium.dtls.pskstore.PskStore;
 import org.eclipse.californium.scandium.dtls.rpkstore.TrustAllRpks;
 import org.eclipse.californium.scandium.dtls.rpkstore.TrustedRpkStore;
@@ -228,10 +229,18 @@ public final class DtlsConnectorConfig {
 	private List<CipherSuite> supportedCipherSuites;
 	
 	/**
-	 * the supported signature and hash algorithms.
+	 * the supported signature and hash algorithms in order of preference.
+	 * 
 	 * @since 2.3
 	 */
 	private List<SignatureAndHashAlgorithm> supportedSignatureAlgorithms;
+
+	/** 
+	 * the supported groups (curves) in order of preference.
+	 * 
+	 * @since 2.3
+	 */
+	private List<SupportedGroup> supportedGroups;
 
 	/** the trust store for RPKs **/
 	private TrustedRpkStore trustedRPKs;
@@ -366,6 +375,8 @@ public final class DtlsConnectorConfig {
 	private Boolean clientOnly;
 	
 	private Boolean recommendedCipherSuitesOnly;
+
+	private Boolean recommendedSupportedGroupsOnly;
 
 	private DtlsConnectorConfig() {
 		// empty
@@ -582,8 +593,11 @@ public final class DtlsConnectorConfig {
 	}
 
 	/**
-	 * Gets the cipher suites the connector should advertise in a DTLS
-	 * handshake.
+	 * Gets the supported cipher suites.
+	 * 
+	 * On the client side the connector advertise these cipher suites in a DTLS
+	 * handshake. On the server side the connector limits the acceptable cipher
+	 * suites to this list.
 	 * 
 	 * @return the supported cipher suites (ordered by preference)
 	 */
@@ -604,6 +618,24 @@ public final class DtlsConnectorConfig {
 	 */
 	public List<SignatureAndHashAlgorithm> getSupportedSignatureAlgorithms() {
 		return supportedSignatureAlgorithms;
+	}
+
+	/**
+	 * Gets the supported groups (curves).
+	 * 
+	 * On the client side the connector advertise these supported groups
+	 * (curves) in a DTLS handshake. On the server side the connector limits the
+	 * acceptable supported groups (curves) to this list. According
+	 * <a href= "https://tools.ietf.org/html/rfc8422#page-11">RFC 8422, 5.1.
+	 * Client Hello Extensions, Actions of the receiver</a> This affects both,
+	 * curves for ECDH and the certificates for ECDSA.
+	 * 
+	 * @return the supported groups (curves, ordered by preference)
+	 * 
+	 * @since 2.3
+	 */
+	public List<SupportedGroup> getSupportedGroups() {
+		return supportedGroups;
 	}
 
 	/**
@@ -980,6 +1012,16 @@ public final class DtlsConnectorConfig {
 	}
 
 	/**
+	 * @return <code>true</code> if only recommended supported groups (curves) are used.
+	 * @see Builder#setRecommendedSupportedGroupsOnly(boolean)
+	 * 
+	 * @since 2.3
+	 */
+	public Boolean isRecommendedSupportedGroupsOnly() {
+		return recommendedSupportedGroupsOnly;
+	}
+
+	/**
 	 * @return a copy of this configuration
 	 */
 	@Override
@@ -1008,6 +1050,7 @@ public final class DtlsConnectorConfig {
 		cloned.certChain = certChain;
 		cloned.supportedCipherSuites = supportedCipherSuites;
 		cloned.supportedSignatureAlgorithms = supportedSignatureAlgorithms;
+		cloned.supportedGroups = supportedGroups;
 		cloned.trustedRPKs = trustedRPKs;
 		cloned.outboundMessageBufferSize = outboundMessageBufferSize;
 		cloned.maxDeferredProcessedOutgoingApplicationDataMessages = maxDeferredProcessedOutgoingApplicationDataMessages;
@@ -1037,6 +1080,7 @@ public final class DtlsConnectorConfig {
 		cloned.healthHandler = healthHandler;
 		cloned.clientOnly = clientOnly;
 		cloned.recommendedCipherSuitesOnly = recommendedCipherSuitesOnly;
+		cloned.recommendedSupportedGroupsOnly = recommendedSupportedGroupsOnly;
 		return cloned;
 	}
 
@@ -1134,6 +1178,24 @@ public final class DtlsConnectorConfig {
 			config.recommendedCipherSuitesOnly = recommendedCipherSuitesOnly;
 			if (recommendedCipherSuitesOnly && config.supportedCipherSuites != null) {
 				verifyRecommendedCipherSuitesOnly(config.supportedCipherSuites);
+			}
+			return this;
+		}
+
+		/**
+		 * Set usage of recommended supported groups (curves).
+		 * 
+		 * @param recommendedSupportedGroupsOnly {@code true} allow only
+		 *            recommended supported groups, {@code false}, also allow not
+		 *            recommended supported groups. Default value is {@code true}
+		 * @return this builder for command chaining
+		 * 
+		 * @since 2.3
+		 */
+		public Builder setRecommendedSupportedGroupsOnly(boolean recommendedSupportedGroupsOnly) {
+			config.recommendedSupportedGroupsOnly = recommendedSupportedGroupsOnly;
+			if (recommendedSupportedGroupsOnly && config.supportedGroups != null) {
+				verifyRecommendedSupportedGroupsOnly(config.supportedGroups);
 			}
 			return this;
 		}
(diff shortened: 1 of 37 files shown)


boaks:
Analyzing `ECDHECryptography` my impression was, that it uses encapsulation of internally required fields and exposing of external artifacts not proper.
Therefore I implemented a new one `XECDHECryptography` which makes it possible to sort out a lot of complexity from the previous handshake messages.
Adding the first x25519 support is more a experiment, if implementing on java8 and using java11 crypto extensions will work. It looks promising.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SupportedEllipticCurvesExtension.java
```diff
@@ -80,26 +72,29 @@ protected void addExtensionData(DatagramWriter writer) {
...
 
-		List<Integer> groupIds = new ArrayList<Integer>();
+		List<SupportedGroup> groups = new ArrayList<>();
 		int listLength = extensionDataReader.read(LIST_LENGTH_BITS);
 		DatagramReader rangeReader = extensionDataReader.createRangeReader(listLength);
 		while (rangeReader.bytesAvailable()) {
 			int id = rangeReader.read(CURVE_BITS);
-			groupIds.add(id);
+			SupportedGroup group = SupportedGroup.fromId(id);
+			if (group != null && group.isUsable()) {
```
sbernard31: This is extension deserialization ? 

I didn't expect that we check that group is usable here ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -782,41 +820,50 @@ private void addServerHelloExtensions(final CipherSuite negotiatedCipherSuite, f
...
 	 * 
-	 * @param clientHello
-	 *            the peer's <em>CLIENT_HELLO</em> message containing its
-	 *            preferred elliptic curves
-	 * @return the selected curve or {@code null} if server and client have no curves in common
+	 * @param clientHello the peer's <em>CLIENT_HELLO</em> message containing
+	 *            its preferred elliptic curves
+	 * @return a list of common supported curves. Maybe empty, if server and
+	 *         client have no curves in common
+	 *
```
sbernard31: it seems to me that javadoc is not correct ? 
We don't search the EC but all the supported group handled by both side ordered by client preferences ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -717,16 +739,17 @@ private void negotiateCipherSuite(final ClientHello clientHello, final HelloExte
 
 		CertificateType supportedServerCertType = getSupportedServerCertificateType(clientHello);
```
sbernard31: A bit out of topic, but why certtype are not a list ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -717,16 +739,17 @@ private void negotiateCipherSuite(final ClientHello clientHello, final HelloExte
...
 
 		for (CipherSuite cipherSuite : clientHello.getCipherSuites()) {
 			// NEVER negotiate NULL cipher suite
 			if (cipherSuite != CipherSuite.TLS_NULL_WITH_NULL_NULL && supportedCipherSuites.contains(cipherSuite)) {
-				if (isEligible(cipherSuite, supportedServerCertType, supportedClientCertType, group, signatures)) {
+				if (isEligible(cipherSuite, supportedServerCertType, supportedClientCertType, groups, format, signatures)) {
 					negotiatedServerCertificateType = supportedServerCertType;
 					negotiatedClientCertificateType = supportedClientCertType;
-					negotiatedSupportedGroup = group;
+					negotiatedSupportedGroups = groups;
```
sbernard31: why negociateGroups is a list ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/XECDHECryptography.java
```diff
@@ -0,0 +1,751 @@
...
+		keyAgreement.doPhase(peerPublicKey, true);
+		byte[] secret = keyAgreement.generateSecret();
+		SecretKey secretKey = SecretUtil.create(secret, "TlsPremasterSecret");
+		Bytes.clear(secret);
+		return secretKey;
+	}
+
+	@Override
+	public void destroy() {
+		// SecretUtil.destroy(privateKey);
```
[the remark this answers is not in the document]
boaks: Removed, it throws an exception, if called ;-(.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/XECDHECryptography.java
```diff
@@ -0,0 +1,751 @@
...
+		X448(30, 56, XDH_KEYPAIR_GENERATOR_ALGORITHM, true);
+
+		private final int id;
+		private final String algorithmName;
+		private final int keySizeInBytes;
+		private final boolean usable;
+		private final boolean recommended;
+
+		/**
+		 * Create not usable group.
```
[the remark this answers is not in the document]
boaks: Good catch. Must be corrected.
Copy & paste error from the previous implementation. There it was assumed, that the parameter must be known. They could be read from a generated key(pair).

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/XECDHECryptography.java
```diff
@@ -0,0 +1,751 @@
...
+			} catch (GeneralSecurityException e) {
+				LOGGER.trace("Group [{}] is not supported by JRE! {}", name(), e.getMessage());
+			}
+			this.keySizeInBytes = keySize;
+			this.usable = curve != null;
+			EC_CURVE_MAP_BY_ID.put(code, this);
+		}
+
+		/**
+		 * Create not usable group.
```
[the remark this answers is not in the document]
boaks: see above

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -2479,9 +2657,22 @@ public DtlsConnectorConfig build() {
 					throw new IllegalStateException("certificate has no proper key usage!");
 				}
 			}
+			if (config.publicKey != null) {
+				SupportedGroup group = SupportedGroup.fromPublicKey(config.publicKey);
+				if (!config.supportedGroups.contains(group)) {
```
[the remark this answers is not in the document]
boaks: We can go for adding them.
With recommended only using a certchain causes an exception, if none recommended curves are used?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks: Previous, because in the server_hello it is only possible to send one cert type, no list. Without sign_hash_algo and supported curves, this could be negotiated here. With them, it will get a little complicated. Therefore I postponed it to issue #1259

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks: If list or single value depends also on the extensions.
For some the client sends a list, and the server the selected item.
For others the client sends a list, and the server sends the selected sublist.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SupportedEllipticCurvesExtension.java
boaks: The ids, where scandium doesn't even know the "supported group name", are also skipped here. So for me it didn't make too much difference.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks: Yes, but in order to use them for ECDHE (and sometimes also to check the public key or certificate chain, depending on other parameters).

boaks:
> Too bad that java API does not allow to avoid reflection stuff disappointed

The stuff could be encapsulated in XDHUtil, which uses the reflection default implementation, and enables java 11 code to be set as none-reflection handler.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/XECDHECryptography.java
```diff
@@ -0,0 +1,751 @@
...
+		brainpoolP384r1(27, false),
+		brainpoolP512r1(28, false),
+		ffdhe2048(256, false),
+		ffdhe3072(257, false),
+		ffdhe4096(258, false),
+		ffdhe6144(259, false),
+		ffdhe8192(260, false),
+		arbitrary_explicit_prime_curves(65281, false),
+		arbitrary_explicit_char2_curves(65282, false),
+
```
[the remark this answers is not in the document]
boaks: removed

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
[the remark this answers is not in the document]
boaks: added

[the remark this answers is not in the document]
sbernard31: I would say : we always try to add it if config.groups is null.
But if is recommanded is true we check that this is a recommended one.
I will look at the code to see.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SupportedEllipticCurvesExtension.java
sbernard31: The idea behind this comment : 
Here we deserialize, so we should see what is  sent by the peer.
Later we negotiate and so remove the not usable.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
sbernard31: My remarks was for both certtype. (ok to discuss about that later)

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
sbernard31: currently the code only use "the selected item" (1 group)
So this list is just for the future ? (or I missed something?)

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -2633,5 +2829,57 @@ private void verifySignatureAndHashAlgorithms() {
...
+				addSupportedGroups(defaultGroups, config.publicKey);
+			}
+			return defaultGroups;
+		}
+
+		private void addSupportedGroups(List<SupportedGroup> defaultGroups, PublicKey publicKey) {
+			if (publicKey != null) {
+				SupportedGroup group = SupportedGroup.fromPublicKey(publicKey);
+				if (group != null && group.isUsable() && !defaultGroups.contains(group)) {
+					if (!config.recommendedSupportedGroupsOnly || !group.isRecommended()) {
```
sbernard31:
```suggestion
					if (!config.recommendedSupportedGroupsOnly || group.isRecommended()) {
```

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SupportedEllipticCurvesExtension.java
boaks: For ids, unknown by scandium, this will not work. 
The read list is not really used to visualize, and if, with the comment in the first line, it would require to change the enum implementation, otherwise the unknown gets lost. So, my feeling was, it's not worth to spend too much in things, which are anyway not longer recommended.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks: Exactly, currently **one** is selected, because the server hello extensions are limited to **one** and this implementation is mainly from the time before adding the signing stuff and evaluating the curves!
With #1259 I would change that to:
common parameter: lists
negotiated parameter: lists and selected items (depending on the specs and usage).

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: Good catch!

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SupportedEllipticCurvesExtension.java
sbernard31: > For ids, unknown by scandium, this will not work.

I understand.

In term of separation of concern, I still thing separating deserialization & negotiation would be better. But it's not a big deal.

sbernard31:
LGTM

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SupportedEllipticCurvesExtension.java
boaks: I changed to include the unused/unsupported, but not the unknown. Otherwise a second list just with the ids would be required. If that causes trouble in the future, we can add such a raw-list. For dtls details I'm more used to use wireshark than the scandium logs. That convinces more, than the own logs :-).

PR #1261, opened 2020-03-29
Cleanup imports

boaks:
Remove obsolete imports.


boaks:
Though the changes don't affect the function and the diff is very easy to be verified, I merge this PR without review ahead. Any comments will still be welcome afterwards.

PR #1262, opened 2020-03-30
Bugfix for retransmission of a separate NON response

boaks:
Bugfix for retransmission of a separate NON message as response for a duplicated observe request.

Responses with observe option are kept in the exchange store even if
their type is NON. So their retransmission requires to increment the
failed transmissons counter.

Add unit test as well.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
@@ -219,8 +219,10 @@ public class ReliabilityLayer extends AbstractLayer {
 					if (request.isConfirmable()) {
 						// resend ACK,
 						// comply to RFC 7252, 4.2, cross-layer behavior
-						EmptyMessage ack = EmptyMessage.newACK(request);
-						sendEmptyMessage(exchange, ack);
+						if (request.acknowledge()) {
+							EmptyMessage ack = EmptyMessage.newACK(request);
+							sendEmptyMessage(exchange, ack);
+						}
 					}
 					if (type == Type.CON) {
 						// retransmission cycle
@@ -235,6 +237,11 @@ public class ReliabilityLayer extends AbstractLayer {
 							sendResponse(exchange, currentResponse);
 						}
 						return;
+					} else if (currentResponse.isNotification()) {
+						// notifications are kept in the exchange store, so
+						// prepare retransmission counter for retransmission
+						int failedCount = exchange.getFailedTransmissionCount() + 1;
+						exchange.setFailedTransmissionCount(failedCount);
 					}
 				}
 				LOGGER.debug("{} respond with the current response to the duplicate request", exchange);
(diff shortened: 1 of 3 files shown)


californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -235,6 +237,11 @@ public void receiveRequest(final Exchange exchange, final Request request) {
 							sendResponse(exchange, currentResponse);
 						}
 						return;
+					} else if (currentResponse.isNotification()) {
+						// notifications are kept in the exchange store, so
+						// prepare retransmission counter for retransmission
+						int failedCount = exchange.getFailedTransmissionCount() + 1;
+						exchange.setFailedTransmissionCount(failedCount);
```
sbernard31: I don't get the point here. 
`FailedTransmissountCount` is a best effort counter which tries to count how many time we failed to send a message for this exchange ?

And as we are able to count this even for NON notification we try to do it ? 
I try to read the test to understand the use case but still don't get it :grimacing:

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
```diff
@@ -219,8 +219,10 @@ public void receiveRequest(final Exchange exchange, final Request request) {
 					if (request.isConfirmable()) {
 						// resend ACK,
 						// comply to RFC 7252, 4.2, cross-layer behavior
-						EmptyMessage ack = EmptyMessage.newACK(request);
-						sendEmptyMessage(exchange, ack);
+						if (request.acknowledge()) {
+							EmptyMessage ack = EmptyMessage.newACK(request);
+							sendEmptyMessage(exchange, ack);
+						}
```
sbernard31: reading this, it seems to me that we always send an empty ACK even if we replied with a piggyback, but I suppose I miss read the code ?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
boaks: The exchange store keeps only messages, when a answer from the other peer is considered. this is mainly CON and for notifies also NON to react on RST to cancel the observation.

[Check NON with MID collision](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/InMemoryMessageExchangeStore.java#L259)

That check takes the `FailedTransmissountCount` to check, if this is a failure or a retransmission. Therefore I increment this now also for non notifies retransmission.

I added two test, one with piggy-backed response, which would work without this fix. And one with separate response, which is affected by this bug.
If the request gets retransmitted, the duplicate will cause first the ACK (as on the first receive of that request), and then the NON notify , which was already send and so was already stores in the exchange store. With this fix, this is now handled as retransmission an so the check is passed.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
boaks: Piggy-backed response have type ACK, and will bypass this.
If the response was NON, usually a ACK was sent, so this just retransmit the ACK as well.

sbernard31: I get it.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/ReliabilityLayer.java
sbernard31: a bit clearer to me, thx

sbernard31:
No objection.

PR #1263, opened 2020-03-30
Prepare for reregistration even if response is not newer.

boaks:
A reregister request may get a response, which is not changed since the
last.
Reverts reregistration fix of PR #1221.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/observe/ObserveNotificationOrderer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/observe/ObserveNotificationOrderer.java
@@ -83,10 +83,11 @@ public class ObserveNotificationOrderer {
 	}
 
 	/**
-	 * Returns true if the specified notification is newer than the current one.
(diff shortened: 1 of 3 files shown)


californium-core/src/main/java/org/eclipse/californium/core/observe/ObserveNotificationOrderer.java
```diff
@@ -83,14 +83,27 @@ public int getCurrent() {
...
+	}
+
+	/**
+	 * Check, if the provided value of the observer option is newer than the
+	 * current one.
+	 * 
+	 * @param observe the value of the observe option. Maybe {@code null}, if
+	 *            the observation is canceled
+	 * @return true if the notification is new, or canceled
+	 * @since 2.3
```
sbernard31: 2.2.1 or 2.3 ?

(new API so 2.3 ? or maybe this part of the code in considering as internal)

sbernard31:
LGTM

PR #1269, opened 2020-03-31
Move anmial-sniffer to profile.

boaks:
Skip revapi using properties.
Initially I wanted to skip the animal sniffer automatically, when java 11 is selected as target. That's not achieved. But anyway this PR demonstrates to use the skips configured in the properties instead of define a specialized plugin with skipped execution.


sbernard31:
LGTM

PR #1270, opened 2020-03-31
Add new cipher suite and parameter selection.

boaks:
Used during handshake on the server side.
Select algorithm maybe exchanged via dtls configuration.


boaks:
If PR #1258 is merged, the diff should be much smaller.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
```diff
@@ -509,9 +509,15 @@ public SignatureAndHashAlgorithm getSignatureAndHashAlgorithm(PublicKey key) {
 	 *         compatible with any of the supported certificate types or any of the supported signature algorithms.
 	 */
 	public SignatureAndHashAlgorithm getSignatureAndHashAlgorithm(List<X509Certificate> chain) {
```
sbernard31, line 511: This method sounds never used to me ? 

Should it be a list returned ? I mean for a certChain it could be several algorithm ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -558,23 +548,23 @@ private void createServerKeyExchange(final ClientHello clientHello, final DTLSFl
...
 	}
 
-	private boolean createCertificateRequest(final ClientHello clientHello, final DTLSFlight flight) throws HandshakeException {
-
-		if ((clientAuthenticationWanted || clientAuthenticationRequired) && negotiatedSignatureAndHashAlgorithms != null) {
-
+	private boolean createCertificateRequest(final ClientHello clientHello, final DTLSFlight flight)
+			throws HandshakeException {
+		if ((clientAuthenticationWanted || clientAuthenticationRequired)
+				&& session.getCipherSuite().requiresServerCertificateMessage()) {
```
sbernard31, line 554: :+1:

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -2856,7 +2900,10 @@ private void addSupportedGroups(List<SupportedGroup> defaultGroups, PublicKey pu
 		private void verifySupportedGroups(List<SupportedGroup> list) {
 			if (config.certChain != null) {
 				for (X509Certificate certificate : config.certChain) {
-					verifySupportedGroups(list, certificate.getPublicKey());
+					PublicKey publicKey = certificate.getPublicKey();
+					if (SupportedGroup.isEcPublicKey(publicKey)) {
+						verifySupportedGroups(list, publicKey);
+					}
```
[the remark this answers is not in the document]
boaks: The `config.publicKey` is always the head of `config.certChain`. So that would be a double check.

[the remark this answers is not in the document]
boaks: > (like for verifySignatureAndHashAlgorithms)

public key => `SignatureAndHashAlgorithm.getSupportedSignatureAlgorithm`
chain => `SignatureAndHashAlgorithm.isSignedWithSupportedAlgorithms`

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/CertificateRequest.java
boaks: Typo in `ClientHandshaker.createCertificateMessage`. Now it's used again.

boaks:
I changed the parameter and selector slightly to enable customized implementations, which don't request client certificates, if client auth is only wanted (not required) and no common parameter exists. But as default I would keep the current implementation, which doesn't select that.

sbernard31:
It seems you skip this comment : 
> I just ask myself if this is a good Idea to mix the class which contents parameter to select and the result (parameter selected)

but not a big deal :)

I think it's ok now.
Maybe for big PR like this it would be easy if you do not push force on the PR, so I can just review the modified part. then at the end we squash useless commit ?

boaks:
> I just ask myself if this is a good Idea to mix the class which contents parameter to select and the result (parameter selected)

I tried a "selected parameter extends parameter", but for me, that was just too much copying.
Otherwise a lot of copies will be required. And the the "selected parameter class" will be

boaks:
> Maybe for big PR like this it would be easy if you do not push force on the PR, so I can just review the modified part. then at the end we squash useless commit ?

I will try to do so.

PR #1271, opened 2020-04-03
Add hono benchmark.

boaks:
Add hono-style request and PSK-storefile.
Add base64 encoding/decoding support to StringUtils for values without padding.
Enabled via env variable.


boaks:
I would like to merge this at the end of this week, Thursday. 
It only affects the benchmark client.

sbernard31:
no objection

PR #1274, opened 2020-04-07
Add logging messages and fix closing uncomplete handshakes

boaks:
Add logging for complete or failing handshakes.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
@@ -40,6 +40,7 @@ package org.eclipse.californium.scandium.dtls;
 
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.util.ConcurrentModificationException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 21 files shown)


boaks:
In the meantime one cause of the NPE could be reproduced.
It is the processing of a close notify, when the handshake has a established session, but hasn't completed (sent FINISH, wait for appl. data). The left the handshake in an inconsistent state, which caused the NPE.

Unit tests are currently missing, I hope I can add them the next days.

boaks:
FMPOV, a 2.2.2 should only contain 11a4071 and db8439d

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Connection.java
```diff
@@ -569,13 +571,35 @@ public void sessionEstablished(Handshaker handshaker, DTLSSession session) throw
...
+			}
 			if (ongoingHandshake.compareAndSet(handshaker, null)) {
 				LOGGER.debug("Handshake with [{}] has been completed", handshaker.getPeerAddress());
 			}
 		}
 
 		@Override
 		public void handshakeFailed(Handshaker handshaker, Throwable error) {
+			SerialExecutor executor = serialExecutor;
+			if (executor != null && !executor.isShutdown()) {
```
sbernard31, line 593: should we make the check only available is LOGGER_OWNER is ErrorEnabled ? (same for handshake complete)

boaks: I can add the check, but I would recommend to have it enabled.

boaks:
> It's not clear to me about the other commits should I review it too ?

Depends on the time you're able to spend. The other stuff is mainly moving code and add a unit test.

sbernard31:
I looked at the reamining commit quickly. No objections :)

PR #1275, opened 2020-04-13
Fix startup error on none-secure communication.

boaks:
Adapt content types for benchmark.
Force handshakes after errors.


sbernard31:
No objections.

PR #1279, opened 2020-04-14
Keep closed session and mark the closed receive window.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -176,7 +176,6 @@ import org.eclipse.californium.scandium.dtls.ApplicationMessage;
 import org.eclipse.californium.scandium.dtls.AvailableConnections;
 import org.eclipse.californium.scandium.dtls.ClientHandshaker;
 import org.eclipse.californium.scandium.dtls.ClientHello;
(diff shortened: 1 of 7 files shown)


boaks:
A "cheap cheat", not CLOSE, just mark the session with the record_sequence_number. Drop messages after that number, process messages before that number.

sbernard31:
I looked at this and I understand the idea. (I think it should work)
But I have some concern about some part of the code where connection could be retrieved and not active. (like maybe where we raise `EndpointUnconnectedException` like `DTLSConnector.send(RawData)`)

boaks:
For incoming, the records should be stopped by "session.isRecordProcessable" (small fix for session == null is on the way).
For outgoing, the difference maybe, that with this, a resumption is tried, with the other a full handshake. I think, the resumption is not too bad. Or do you see something, I oversee?

sbernard31:
I read the code quickly again and you're right it should be OK.

(Just a detail but first time I read the code, "closing a session" sounds strange to me because as I understand the rfc a session can not be closed, this is the connection which is closed. I think the issue here is that our Connection / Session object does not really match the connection/session concept of the RFC)

boaks:
> Just a detail 

you're right :-).

boaks:
I added test, redesigned it a little, and deprecated the obsolete `CloseSupportingConnectionStore`.

PR #1285, opened 2020-04-20
Add interoperability test for RSA certificate in chain.

boaks:
Redesign utility APIs.
Openssl s_server seems to just use the provided certificate and doesn't check the signature and hash algorithms accordingly. 


boaks:
I start also to include libcoap in the tests. Therefore I will move the test into a separate module.

boaks:
In order to have a "moving only commit", I would like to merge this today, late afternoon.

PR #1286, opened 2020-04-20
Do not warn for Handshake without APP_DATA of evicted connection.

sbernard31:
(For the fix in 2.3.x branch, I think we need to create a dedicated Exception for eviction)

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -458,8 +458,14 @@ public class DTLSConnector implements Connector, RecordLayer {
 						} else {
 							// failure after established (last FINISH),
 							// but before completed (first data)
-							LOGGER.warn("Handshake with [{}] failed after session was established!",
-									handshaker.getPeerAddress(), error);
+							// TODO a dedicated Exception should be created.
+							if (error instanceof RuntimeException && "Evicted!".equals(error.getMessage())) {
+								LOGGER.debug("Handshake with [{}] never get APPLICATION_DATA",
+										handshaker.getPeerAddress(), error);
+							} else {
+								LOGGER.warn("Handshake with [{}] failed after session was established!",
+										handshaker.getPeerAddress(), error);
+							}
 						}
 					} else if (connection.hasEstablishedSession()) {
 						LOGGER.warn("Handshake with [{}] failed, but has an established session!",


boaks:
I would prefer, just to change warn into debug.

sbernard31:
My guess,  if handshake failed because another exception than "eviction", it's a real warn.

boaks:
:-) we will see, what the next cause will be.

LGTM

sbernard31:
Are you OK to backport this in 2.3 with a new dedicated exception (e.g ConnectioEvictedException extends DtlsException) ?

boaks:
> Are you OK to backport this in 2.3 with a new dedicated exception (e.g ConnectioEvictedException extends DtlsException) ?

Hm, a `DtlsException` seesm to have a `peer address`, that's not really true for all "evicted connections". Some may have been already removed from an active address.

sbernard31:
> that's not really true for all "evicted connections"

Correct but in a meantime a connection could have a null peer address so we could imagine that a  dtls exception could have too.

Anyway, If you prefer  `ConnectionEvictedException extends RuntimeException`, it's OK for me too.

boaks:
For me, the session listener tries to keep track of the connections in the store.
The if branches are mainly used as alternative to

```
if (handshaker.isRemovingConnection()) {
   connectionStore.remove(connection, false);
```

I would prefer to check, if the connection is still in the store, and so for the eviction case, this would then not longer write any message.

sbernard31:
I'm not sure to see the benefits of this. :thinking: 
And this would notl add an access to a synchronized method of InMemoryConnectionStore just for a log if we can avoid this.

Using this PR check or your solution, I still think it's better to raise a dedicated Exception instead of a Runtime.

boaks:
> I'm not sure to see the benefits of this. thinking

The reason for logging there is the "house keeping" of the connections.
If that is in doubt, the reason for the failure maybe a good hint.
If the house-keeping is not in doubt, there is FMPOV no reason, to write a log.
 
> And this would notl add an access to a synchronized method of InMemoryConnectionStore just for a log if we can avoid this.

I don't think, that this happens too often.

boaks:
> Using this PR check or your solution, I still think it's better to raise a dedicated Exception instead of a Runtime.

If the information is required for something, I would go for it. 
But I guess, this is mostly not needed (because the handshakes will be completed), and the few left, the reason seems to get less and lesser interesting. That's at least my impression.
Anyway, if you feel more comfortable, go for a dedicated Exception and adjust to log-level using that.

sbernard31:
> If the house-keeping is not in doubt, there is FMPOV no reason, to write a log.

The current log help me to detect incorrect behavior at client side.
As said before, I think this could be valuable to use the DEBUG level to help to detect this in production. 

> If the information is required for something, I would go for it.

We have a sessionListener to monitor Handshake (metrics for success, failure ...) with an `EvictionException` this would be easier to filter Handshake Failure because of this.

boaks:
> We have a sessionListener to monitor Handshake (metrics for success, failure ...) with an EvictionException this would be easier to filter Handshake Failure because of this.

OK, so go for it :-).

PR #1290, opened 2020-04-23
Do not warn for Handshake without APP_DATA of evicted connection.

sbernard31:


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ConnectionEvictedException.java
```diff
@@ -0,0 +1,31 @@
...
+ * Raised when a connection is evicted from
+ * {@link ResumptionSupportingConnectionStore}
+ * 
+ * @since 2.3
+ */
+public class ConnectionEvictedException extends DtlsException {
+
+	private static final long serialVersionUID = 1L;
+
+	public ConnectionEvictedException(String message, InetSocketAddress peer) {
```
boaks, line 28: The peer maybe `null`. I'm not sure, if that would happen also before. At least, I would prefer to add now javadoc to DtlsException including the version, when `null` is introduced.

sbernard31: I will add the javadoc. But my guess is a `null` peer address could maybe already happen. So not sure it makes sense to add the `since version`. By the way nothing in the javadoc says that peer address can not be null too.

What do you think ? Should I add the `since version` ?

(This is not part of this PR, but my guess is that the correct API could be to have `Connection`(or something like this) instead of peer address.)

sbernard31: (I added the javadoc without since for now)

boaks: > By the way nothing in the javadoc says that peer address can not be null too.

There is also no indication, that `null` maybe returned.

> What do you think ? Should I add the since version ?

That makes it clear, that at least from that version on, `null` maybe returned.
If I would get "new" NullPointerExceptions, I would appreciate, to read, that this is cased by a change with the 2.3.

boaks: At least, if you add a `null`check (NullPointerException) and run `mvn clean test`, the tests are not failing.
In the most scenarios, I was able to check (without wasting too much time for this), the exceptions results from receiving messages, in which case the current address is available.

If `null`is already returned, existing code would be aware of that.

The comment just helps in cases, where until now, no `null`was returned to already existing code. Just to mention: I didn't found, that californium calls the getter at all.

sbernard31: I will add the since version warn because I don't really want to take more time to talk about the color of the bike.

sbernard31: Done

boaks:
LGTM

(Do you prefer `<code> ... </code>` over `{@code ...}`?)

sbernard31:
No preference, if there is a dedicated tag (@code) for javadoc maybe we should use it ?

boaks:
I usually use that, therefore I asked, because your comment use the other.

sbernard31:
OK I change this.

boaks:
LGTM

sbernard31:
So I fixed the whole class javadoc : 3651936

Issue #1291, opened 2020-04-23
Allow integration with HSM


JaimieW:
I'm using Californium to write a CoAP and DTLS server implementation.
My problem is that the DTLS PSK Key is stored in a Hardware Security Module.
This means that the key material cannot be extracted and used to create the pre master secret.
It also means that the master secret derivation has to be performed on the HSM itself.

I have a PR containing the changes I needed to make, which are fairly minimal.

boaks:
The extension in general is great.

I would prefer a implementation, which uses a new interface passed in via `DtlsConnectorConfig`. Let me know, if you could consider to change your PR accordingly, or if we should go for now with your implementation.

JaimieW:
I'm looking into changing the code to use an interface passed in via the config.
I should have a commit ready later today.


________________________________

boaks:
Sounds great!

boaks:
It's merged, so can we close this issue?

PR #1292, opened 2020-04-23
Allow integration with Hardare Security Modules

JaimieW:
Added methods which can be overridden to allow integration with Hardware Security Modules
I added issue #1291 for this

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
@@ -885,11 +885,22 @@ public abstract class Handshaker implements Destroyable {
 	 */
 	private SecretKey generateMasterSecret(SecretKey premasterSecret) {
 		byte[] randomSeed = Bytes.concatenate(clientRandom, serverRandom);
-		byte[] secret = PseudoRandomFunction.doPRF(session.getCipherSuite().getThreadLocalPseudoRandomFunctionMac(),
-				premasterSecret, Label.MASTER_SECRET_LABEL, randomSeed);
+		byte[] secret = deriveMasterSecret(premasterSecret, randomSeed);
 		SecretKey masterSecret = SecretUtil.create(secret, "MAC");
 		Bytes.clear(secret);
 		return masterSecret;
+	}
+	
+	/**
+	 * Derives the master secret from the premaster secret and the random seed.
+	 * 
+	 * @param premasterSecret the Pre Master Secret Key
+	 * @param randomSeed the random seed bytes
+	 * @return the derived master secret bytes
+	 */
+	protected byte[] deriveMasterSecret(SecretKey premasterSecret, byte[] randomSeed) {
+		return PseudoRandomFunction.doPRF(session.getCipherSuite().getThreadLocalPseudoRandomFunctionMac(),
+				premasterSecret, Label.MASTER_SECRET_LABEL, randomSeed);
 	}
 
 	protected final void setCurrentReadState() {
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
@@ -142,11 +142,18 @@ public abstract class Handshaker implements Destroyable {
 	private boolean destroyed;
 
 	protected final DTLSSession session;
+
+	/** 
+	 * The logic in charge of deriving the Master Secret 
+	 */
+	protected final MasterSecretDeriver masterSecretDeriver;
+	
 	/**
 	 * The logic in charge of verifying the chain of certificates asserting this
 	 * handshaker's identity
 	 */
 	protected final CertificateVerifier certificateVerifier;
+	
 
 	/** The trusted raw public keys */
 	protected final TrustedRpkStore rpkStore;
@@ -316,6 +323,7 @@ public abstract class Handshaker implements Destroyable {
 		this.publicKey = config.getPublicKey();
 		this.certificateChain = config.getCertificateChain();
 		this.certificateVerifier = config.getCertificateVerifier();
+		this.masterSecretDeriver = config.getMasterSecretDeriver();
 		this.rpkStore = config.getRpkTrustStore();
 		this.pskStore = config.getPskStore();
 		this.session.setMaxTransmissionUnit(maxTransmissionUnit);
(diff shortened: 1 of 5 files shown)


boaks:
Thanks for your contribution!
Please read [CONTRIBUTING](https://github.com/eclipse/californium/blob/master/CONTRIBUTING.md)

Especially, the ECA is required. Please use the same e-mail for registering there as you use for your commits. Finally the commit needs a "sign", which is just a text as:

```
```

JaimieW:
Hi,
I have read the Contributing page.
I have signed the ECA.
Do I just add the text to the commit message?

Thanks,
Jaimie
________________________________

boaks:
yes ... then force push ... and it should go green ...

boaks:
Please take care, that you use the same e-mail for the commits, sign and the ECA.

boaks:
> I have signed the ECA.

At least the "check tool" reports, that you didn't sign a ECA with the e-mail in the commits.

JaimieW:
Sorry, I'm new to open source contributing.

Is there a particular format for
Signed-off-by: Jaimie Whiteside <jaimiewhiteside@hotmail.com>

My github and eclipse emails are the same.
I'm not sure what I'm doing wrong.

boaks:
Your commits shows an other e-mail as "author" and "commiter" then the signing and the ECA.

"jaimiewhiteside@hotmail.com" has a valid ECA, that's great.

Now create commits with that as author.

`git commit --amend --no-edit --reset-author`

would do that, if the e-mail config would be the hotmail and not the business mail.

JaimieW:
Got it now updated now, thank you.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -885,12 +885,23 @@ protected void calculateKeys(SecretKey masterSecret) {
...
 	}
+	
+	/**
+	 * Derives the master secret from the premaster secret and the random seed.
+	 * 
+	 * @param premasterSecret the Pre Master Secret Key
+	 * @param randomSeed the random seed bytes
+	 * @return the derived master secret bytes
+	 */
+	protected byte[] deriveMasterSecret(SecretKey premasterSecret, byte[] randomSeed) {
```
boaks, line 901: Did you have a look at the [SecretUtil](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/util/SecretUtil.java)? Maybe you can adapt the return type to a `SecretKey`?

boaks:
Generally, scandium uses some "callback" interfaces, e.g. [x509 - CertificateVerifier](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/x509/CertificateVerifier.java) to make such functions configurable. To permit to use a own `Handshaker` subclass will lock more functionality as "public API" and would make it harder to change scandium for maintenance.
 
If you would have a look on such an approach, I would appreciate it.
Otherwise, we can merge it for now. But, though the async PSK API is on my list, I may change that again before release a 2.3.

JaimieW:
Added a commit to use an interface to derive the master secret in the Handshaker.
If it is not configured in the DtlsConnectorConfig, then the default behaviour is to use the existing PseudoRandomFunction static method to derive the master secret.

boaks:
The interface looks very good! I prefer that way.

Currently I'm wondering, if your HSM stores only one psk?
Or do you encode the identity in the returned key of `PskStore.getKey()`?
What happens, if a `premasterSecret` of a ECDSA handshake is provided for the call?

JaimieW:
My implementation only supports PSK, this is because I have to be compliant with an internal specification.
Therefore I haven't tried an ECDSA handshake.

I use the HSM to store the PSK for each client that will connect to the server.
The client's unique identifer is transmitted in the psk id. 
My custom PskStore uses the identifierto lookup and return the corresponding PSK key reference from the HSM.
My custom deriver class is a wrapper, which gets the HSM to derive the master secret using the PSK reference.

boaks:
OK, now I understand better how it works.
Assuming the interface would be a 

```
interface HsmPskStore {
SecretKey getMasterKey(ServerNames serverName, PskPublicInformation identity, Mac hmac, byte[] randomSeed);
SecretKey getMasterKey(PskPublicInformation identity, Mac hmac, byte[] randomSeed);
}
```

and implementing that as e.g.

```
class XyzHsmPskStore implements PskStore, HsmPskStore {
...
}
```

Do you think, that will also possible? It would require some more internal changes in scandium, so I would merge your PR "as it is", and adapt the implementation in the next two or three weeks.

JaimieW:
I'm not sure I fully understand your intention with the HsmPskStore interface.
Would the call to getMasterKey be made from receivedClientKeyExchange?
Is it a replacement for the sequence:
```
SecretKey premaster = pskUtil.generatePremasterSecretFromPSK(null);
...
generateKeys(premasterSecret)
```


There may be a problem in that the HSM client may not conform to the JCE api, so there may not be a valid Mac object.

boaks:
> Would the call to getMasterKey be made from receivedClientKeyExchange?
> Is it a replacement for the sequence:
> 

Yes, after some additional adaption.

The `MAC` is passed in empty. You may just use the "getAlgorithm" to check, which MAC to use on the HSM (or the cipher suite doesn't match your HSM). There is no need to call update or doFinal on it, just use the algorithm to calculate the master secret.

boaks:
Just to write it again:

For me, this PR is "LGTM".
I just ask the questions in order to ensure, that such a solution works also for you. It will be my task to change scandium accordingly.

JaimieW:
Yes, fine by me.On 24 Apr 2020 14:51, Achim Kraus <notifications@github.com> wrote:
Just to write it again:
For me, this PR is "LGTM".
I just ask the questions in order to ensure, that such a solution works also for you. It will be my task to change scandium accordingly.

boaks:
LGTM

sbernard31:
I had a quick look : no objection :)

boaks:
@sbernard31 

Just to mention it:
I started to work on the async PSK store at the weekend.
My preference is currently, to offer a interface, which returns the master secret asynchronous, not only the secret key. If someone only wants to fetch the secret key asynchronous, then it will be very easy to use that to provide the master secret and use then the asynchronous handler. My current design is "stateless", I use the CID as ID to match the master secret for the handshakes.
Generally, the idea already works, but need more work. Hope I can finish my exploration this week or next weekend to make a first PR to get more feedback.

sbernard31:
@boaks  :+1:

Issue #1259, opened 2020-03-27
new ciphersuite selection


boaks:
With PR #1256 and #1258 new possibilities are enabled, but also the risk, that a handshake is aborted or encryption function are chosen, which may be considered weaker (e.g. RPK instead of x509, or PSK instead of RPK).

Information to select the cipher suite:
- the supported ciphersuites of both sides (pre-selected by the available credential types)
- the supported signing and hash algorithms of both sides (scandium actively uses only EC-based, but may passively use also other to verify the certificate chain, if the JRE supports that)
- the supported curves of both sides 
- the certificate-type (x509 or RPK) of both sides 

That affects mostly the server-side selection.
Though both PRs are changing al lot of common code, I would therefore prefer to postpone the implementation (and discussion) of this selection details after merging the two PRs.

Any opinions?

sbernard31:
> I would therefore prefer to postpone the implementation (and discussion) of this selection details after merging the two PRs.

I'm lost the only new selection stuff is about sign and hash algorithms ? why not discuss this for the PR which introduce it ? :thinking:

boaks:
Because, if you also look on the other PR, the changes for the "select ciphersuite" are affecting mostly the same source-lines as the other PR and so requires always a manual merges after adapting it (again and again :-) ).

sbernard31:
This would be clearer to do 

So I don't know so much how to handle this, so I will list my comments/concerns here.
For  #1256 : 
1. the cipher selection
2. some part of the code could be clearer. (especially the create certificate call)
3. the default value for sign_hash. 

I will look at #1258 next week. (+2,692 −1,004 :cold_sweat:)

boaks:
> I will look at #1258 next week. (+2,692 −1,004 cold_sweat)

Greta! [Start with XECDHECryptography](https://github.com/eclipse/californium/pull/1258#issuecomment-604372009) and a lot of the other changes will be much easier too understand.

boaks:
> 1. the cipher selection

Therefore I opened this issue to signal, that this must be discussed.

> 2. some part of the code could be clearer. (especially the createCertificateRequest)

As always :-). And, also as always, that depends much on personal preferences. We will see ...

> 3. the default value for sign_hash.

With the openssl issue of today, I would go to pick additional sign_hash from the provided certificate chain as default. Scandium seems to comply with the definitions of [RFC 7252, 9.1.3.2](https://tools.ietf.org/html/rfc7252#section-9.1.3.2)

[...]

PR #1293, opened 2020-04-23
Add libcoap interoperability tests.

boaks:
The CQ for libcoap is pending.
The tests shows some details in the signature and hash algorithms processing:
They maybe used with wireshark to demonstrate the behavior of other implementations.
I will document my results in #1259 . 


boaks:
The CQ is processed.

sbernard31:
no objection

PR #1300, opened 2020-04-30
OSCORE: Block-wise MAX_UNFRAGMENTED_SIZE parameter

rikard-sics:
When using outer block-wise with OSCORE a proxy can maliciously inject block fragments. To protect against this a message with size exceeding this parameter should never be sent without inner block-wise. Likewise when receiving a message using outer block-wise it should be discarded if the cumulated size exceeds this parameter.
See https://tools.ietf.org/html/rfc8613#section-4.1.3.4.2

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
@@ -29,6 +29,8 @@ import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
 
 import org.eclipse.californium.core.coap.CoAP.Code;
+import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.network.config.NetworkConfig.Keys;
 
 import com.upokecenter.cbor.CBORObject;
 
@@ -111,7 +113,18 @@ public class OSCoreCtx {
 	 * See https://tools.ietf.org/html/rfc8613#appendix-B.2
 	 */
 	private boolean contextRederivationEnabled;
-	
+
+	/**
+	 * When using outer block-wise with OSCORE a proxy can maliciously inject
+	 * block fragments. To protect against this a message with size exceeding
+	 * this parameter should never be sent without inner block-wise. Likewise
+	 * when receiving a message using outer block-wise it should be discarded if
+	 * the cumulated size exceeds this parameter.
+	 * 
+	 * See https://tools.ietf.org/html/rfc8613#section-4.1.3.4.2
+	 */
+	private int maxUnfragmentedSize;
+
 	/**
 	 * URI this Context is associated with if any.
 	 *
@@ -258,6 +271,9 @@ public class OSCoreCtx {
 
 		overrideContextId = null;
 		contextRederivationPhase = ContextRederivation.PHASE.INACTIVE;
+
+		// Set default value of MAX_UNFRAGMENTED_SIZE
+		maxUnfragmentedSize = NetworkConfig.getStandard().getInt(Keys.MAX_RESOURCE_BODY_SIZE);
 
 		//Set digest value depending on HKDF
 		String digest = null;
@@ -549,6 +565,26 @@ public class OSCoreCtx {
 	 */
 	public void setContextRederivationEnabled(boolean contextRederivationEnabled) {
 		this.contextRederivationEnabled = contextRederivationEnabled;
+	}
+
+	/**
+	 * Gets the current value of the MAX_UNFRAGMENTED_SIZE parameter. It is used
+	 * to prevent malicious behaviour by a proxy when using block-wise.
+	 * 
+	 * @return the current value of MAX_UNFRAGMENTED_SIZE
+	 */
+	public int getMaxUnfragmentedSize() {
+		return maxUnfragmentedSize;
+	}
+
+	/**
+	 * Sets the current value of the MAX_UNFRAGMENTED_SIZE parameter. It is used
+	 * to prevent malicious behaviour by a proxy when using block-wise.
+	 * 
+	 * @param maxUnfragmentedSize the desired value of MAX_UNFRAGMENTED_SIZE
+	 */
+	public void setMaxUnfragmentedSize(int maxUnfragmentedSize) {
+		this.maxUnfragmentedSize = maxUnfragmentedSize;
 	}
 
 	/**
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
@@ -21,6 +21,7 @@ package org.eclipse.californium.oscore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Arrays;
 import org.eclipse.californium.core.coap.EmptyMessage;
 import org.eclipse.californium.core.coap.Message;
 import org.eclipse.californium.core.coap.MessageObserverAdapter;
@@ -29,6 +30,7 @@ import org.eclipse.californium.core.coap.OptionSet;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
 import org.eclipse.californium.core.coap.Token;
+import org.eclipse.californium.core.coap.CoAP.ResponseCode;
 import org.eclipse.californium.core.network.Exchange;
 import org.eclipse.californium.core.network.stack.AbstractLayer;
 import org.eclipse.californium.elements.util.Bytes;
@@ -160,6 +162,10 @@ public class ObjectSecurityLayer extends AbstractLayer {
 				final Request preparedRequest = prepareSend(ctxDb, request);
 				final OSCoreCtx finalCtx = ctxDb.getContext(uri);
 
+				if (outgoingExceedsMaxUnfragSize(preparedRequest, outerBlockwise, ctx.getMaxUnfragmentedSize())) {
+					throw new IllegalStateException("outgoing request is exceeding the MAX_UNFRAGMENTED_SIZE!");
+				}
+
 				preparedRequest.addMessageObserver(0, new MessageObserverAdapter() {
 
 					@Override
@@ -215,8 +221,16 @@ public class ObjectSecurityLayer extends AbstractLayer {
 			try {
 				OSCoreCtx ctx = ctxDb.getContext(exchange.getCryptographicContextID());
 				addPartialIV = ctx.getResponsesIncludePartialIV() || exchange.getRequest().getOptions().hasObserve();
-				
-				response = prepareSend(ctxDb, response, ctx, addPartialIV, outerBlockwise);
+
+				Response preparedResponse = prepareSend(ctxDb, response, ctx, addPartialIV, outerBlockwise);
+
+				if (outgoingExceedsMaxUnfragSize(preparedResponse, outerBlockwise, ctx.getMaxUnfragmentedSize())) {
+					super.sendResponse(exchange,
+							Response.createResponse(exchange.getCurrentRequest(), ResponseCode.INTERNAL_SERVER_ERROR));
+					throw new IllegalStateException("outgoing response is exceeding the MAX_UNFRAGMENTED_SIZE!");
+				}
+
+				response = preparedResponse;
 				exchange.setResponse(response);
 			} catch (OSException e) {
 				LOGGER.error("Error sending response: " + e.getMessage());
@@ -238,6 +252,12 @@ public class ObjectSecurityLayer extends AbstractLayer {
 			// For OSCORE-protected requests with the outer block1-option let
 			// them pass through to be re-assembled by the block-wise layer
 			if (request.getOptions().hasBlock1()) {
+
+				if (request.getMaxResourceBodySize() == 0) {
+					int maxPayloadSize = getIncomingMaxUnfragSize(request, ctxDb);
+					request.setMaxResourceBodySize(maxPayloadSize);
+				}
+
 				super.receiveRequest(exchange, request);
 				return;
 			}
@@ -270,6 +290,7 @@ public class ObjectSecurityLayer extends AbstractLayer {
 			LOGGER.error("No request tied to this response");
 			return;
 		}
+
 		try {
 			//Printing of status information.
 			//Warns when expecting OSCORE response but unprotected response is received
@@ -282,6 +303,12 @@ public class ObjectSecurityLayer extends AbstractLayer {
 			// For OSCORE-protected response with the outer block2-option let
 			// them pass through to be re-assembled by the block-wise layer
 			if (response.getOptions().hasBlock2()) {
+
+				if (response.getMaxResourceBodySize() == 0) {
+					int maxPayloadSize = getIncomingMaxUnfragSize(response, ctxDb);
+					response.setMaxResourceBodySize(maxPayloadSize);
+				}
+
 				super.receiveResponse(exchange, response);
 				return;
 			}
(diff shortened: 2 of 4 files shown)


rikard-sics:
As we discussed over mail I chose the option to expose some of the information in the block-wise layer so it is accessible in the Object Security Layer. If you have any feedback regarding this I could change the way this is done accordingly.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
```diff
@@ -349,4 +382,105 @@ private static boolean shouldProtectRequest(Request request) {
...
+	 * @param ctxDb the context database used
+	 * 
+	 * @return if the cumulative payload exceeds the MAX_UNFRAGMENTED_SIZE
+	 */
+	private boolean incomingExceedsMaxUnfragSize(Message message, Exchange exchange, OSCoreCtxDB ctxDb) {
+
+		int cumulativeSize;
+		OSCoreCtx ctx = null;
+		if (message instanceof Request) {
+			ctx = ctxDb.getContext(getRid(message.getOptions().getOscore()));
```
boaks, line 426: The rid is used to lookup the ctx, which contains the max size?

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
```diff
@@ -349,4 +382,105 @@ private static boolean shouldProtectRequest(Request request) {
...
+			cumulativeSize = blockwiseTransferInfo.getReceivedBodySize(exchange, (Request) message);
+			cumulativeSize += message.getPayloadSize(); // Count current payload
+		} else {
+			ctx = ctxDb.getContextByToken(message.getToken());
```
boaks, line 430: And fro responses the Token is used to lookup the ctx, which contains the max size?

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
```diff
@@ -349,4 +382,105 @@ private static boolean shouldProtectRequest(Request request) {
...
+			cumulativeSize += message.getPayloadSize(); // Count current payload
+		}
+
+		// Do not consider it exceeded if no context is found, it will be
+		// handled elsewhere
+		if (ctx == null) {
+			return false;
+		}
+
+		if (cumulativeSize > ctx.getMaxUnfragmentedSize()) {
```
boaks, line 441: The max size seems to be defined by the rid or token.

On the client side, the uri ctx is transferred to the token context and so taken in account for responses.
I'm not sure, if adding such a "max response size" to a request would not be easier and also protects other transfers, not only oscore.

But the server side, I'm not sure, how it works.
Is that "rid" request specific (it's calculated from the received oscore option)?
Does the server need to be prepared for that exact request?
Or is that "rid" more resource specific?

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
rikard-sics: Exactly, I need to parse the Recipient ID from the incoming request to retrieve the associated context. And the context contains the max size limit.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
rikard-sics: Yes, since I know which context the corresponding request was sent using, I can look up that same context using the Token for the response.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
boaks: I would understand, that a specific resource may have a limit.
But using that rid seems for me to be the wrong item to associate that size.
I think, if we add the "size limitation" to requests (for the expected response) and resources (for incoming requests), then this would be clearer to use. The it will be even a general feature, and not a oscore.
WDYT?

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
rikard-sics: > The max size seems to be defined by the rid or token.

The max size is defined in the OSCORE context used for a particular response or request. And the context used (for incoming messages) is retrieved by using either the Recipient ID (RID) from the OSCORE option of an incoming request, or the Token for an incoming response.

A user or application developer can set specific size limits per OSCORE context (perhaps one server needs a lower limit so the context shared with that server can be configured to have a lower limit).

> But the server side, I'm not sure, how it works.
> Is that "rid" request specific (it's calculated from the received oscore option)?
> Does the server need to be prepared for that exact request?
> Or is that "rid" more resource specific?

Yes indeed, the Recipient ID (RID) is request specific. It is contained in the OSCORE option and parsed from there. By doing that the server can identity the OSCORE context associated with the incoming request and retrieve the particular size limit from that context.

The server does not need to be prepared for that particular request except for having an OSCORE context associated to the RID in the OSCORE option of the request. The RID is not resource-specific.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
rikard-sics: > I would understand, that a specific resource may have a limit.
> But using that rid seems for me to be the wrong item to associate that size.

My idea was to implement this as a general limit for a whole OSCORE context, and not as a resource-specific limitation. So for a specific OSCORE context there is a MAX_UNFRAGMENTED_SIZE parameter. My understanding from the OSCORE RFC is that this is more a general security measure to prevent malicious proxies from injecting messages than a limit that takes into account what a particular resource contains.

Also this limitation is only when using outer block-wise with OSCORE. So for instance if the limit is 1000 bytes and I get a block-wise response with 5 * 250 bytes using outer block-wise it should be rejected. However if I get the same 5 * 250 bytes response with inner block-wise it should be accepted.

Since for inner block-wise each block response is individually OSCORE-protected and can be verified as it arrives, for outer block-wise I must wait for reassembly before verifying them)

boaks: For a general limit, wouldn't the already available [MAX_RESOURCE_BODY_SIZE](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java#L212) work?

That is applied, when the payload is assembled, so that may even work without any changes.

rikard-sics: > For a general limit, wouldn't the already available [MAX_RESOURCE_BODY_SIZE](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java#L212) work?
> 
> That is applied, when the payload is assembled, so that may even work without any changes.

So that would act as a limit for the size of how big a re-assembled block-wise message can be (whether incoming request or response)?

One potential issue is that it would not be possible to have different limits for OSCORE inner and outer block-wise. Since for inner block-wise each block message is individually protected a malicious proxy could never inject messages, so the size limit for inner block-wise could be more generous than for outer block-wise.

Let me have a look in the code for how MAX_RESOURCE_BODY_SIZE is used and get back to you.

rikard-sics: I have now checked the code more, and we also had a discussion regarding this during our regular meeting about this work.

It seems that MAX_RESOURCE_BODY_SIZE serves a similar purpose. However if we use it we must restrict the maximum size of all block-wise transfers. That is regardless if the individual block messages are protected with OSCORE using inner block-wise, or if the messages are protected with OSCORE before they are fragmented by a proxy (outer block-wise).

We feel there is a use case where a smaller limit can be set for outer block-wise where a malicious proxy can inject messages, compared to the case for inner block-wise where each individual block message is protected with OSCORE. In the inner block-wise case, since each individual message is verified, an application may want to set a much more generous limit.

In the way I have implemented things now it is possible for an application to set the MAX_UNFRAGMENTED_SIZE to a small limit thus being strict when it comes to outer block-wise going through untrusted proxies, while at the same time having a more generous general limit (using the MAX_RESOURCE_BODY) for inner block-wise.

boaks: Yes, you may use different sizes.
But, how do a client know, that it will be a inner or outer transfer?
And how will the server know, that it is a inner or outer transfer?

boaks: Let me try to reduce misunderstandings:

[InnerBlockwise](https://github.com/eclipse/californium/pull/1300/files#diff-30dc1a5cd61b70377a692213961fa02dR152)

[OuterBlockwise](https://github.com/eclipse/californium/pull/1300/files#diff-0cc71273f0cb77efb79646f03b3a3b38R413)

I don't see the difference on the client. Is there one?
If not, how should a client setup the right size, if there is no difference in inner and outer on the client side?

rikard-sics: > I don't see the difference on the client. Is there one?
> If not, how should a client setup the right size, if there is no difference in inner and outer on the client side?

The difference is in the BlockWiseLayer and ObjectSecurityLayer. Since the MAX_RESOURCE_BODY_SIZE is a limit for all block-wise transfers but MAX_UNFRAGMENTED_SIZE only limits outer block-wise transfers.

So let's say a client (communicating using OSCORE) sets MAX_RESOURCE_BODY_SIZE to 2000 and MAX_UNFRAGMENTED_SIZE to 200. That means that when it gets a response from a server using inner block-wise the MAX_UNFRAGMENTED_SIZE limit will not apply and the limit for inner block-wise will thus be 2000. However if it gets a response from a proxy, where the proxy split the response into blocks (outer block-wise) then MAX_UNFRAGMENTED_SIZE will apply and the limit will be 200.

Likewise a server can set these parameters to for instance say that it will allow incoming requests up to 10000 bytes cumulative payload size (using MAX_RESOURCE_BODY_SIZE) when inner OSCORE block-wise is used, but only 1000 bytes when using OSCORE outer block-wise (using MAX_UNFRAGMENTED_SIZE).

With inner block-wise there is no fear of a proxy injecting messages so a much large payload size may be acceptable. Then you check each individual block message with OSCORE so you know a proxy cannot try to inject blocks to waste your memory. With outer block-wise you don't have that guarantee and may thus not want to allow it to use as much memory. So a client and server can use these 2 parameters to set overall limits for inner and outer block-wise.

boaks: Please first try to sort out the client side.
A client may use different sizes, but if the client doesn't know, if it will be inner or outer, how should the client chose the right size?

boaks: > MAX_UNFRAGMENTED_SIZE only limits outer block-wise transfers.

[OSCoreInnerBlockwiseTest.testOscoreExceedMaxUnfragmentedSize](https://github.com/eclipse/californium/pull/1300/files#diff-30dc1a5cd61b70377a692213961fa02dR216)

Why is that limit then checked in the test for inner blockwise?

boaks: Maybe I start to understand:

If the transmitted message has a block1/block2 option (outer), then you check that against the new limit.

if the message has a block1/block2 option after the "juggle" (inner), then you leave it to blockwise max resource.

Is that right?

rikard-sics: > Maybe I start to understand:
> 
> If the transmitted message has a block1/block2 option (outer), then you check that against the new limit.
> 
> if the message has a block1/block2 option after the "juggle" (inner), then you leave it to blockwise max resource.
> 
> Is that right?

Yes exactly. For incoming messages the MAX_UNFRAGMENTED_SIZE is only checked for block1/block2 outer options (in the original incoming CoAP message). But if the message uses block1/block2 options after it has been unprotected/juggled (inner block-wise) only the MAX_RESOURCE_BODY_SIZE limit is applied.

That is because the _incomingExceedsMaxUnfragSize_ method will only be called for incoming messages when outer block-wise is used. I did not have it as an explicit parameter in the method, but it is only applied in the case of outer block-wise:
https://github.com/eclipse/californium/blob/f375eb886b2e7749e96fc7466a2e92ac9e9d83a6/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java#L257
https://github.com/eclipse/californium/blob/f375eb886b2e7749e96fc7466a2e92ac9e9d83a6/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java#L310
(If the message has a block1/block2 option before it is unprotected/juggled.)

As you say for incoming messages the MAX_RESOURCE_BODY_SIZE would apply in the case of inner block-wise while MAX_UNFRAGMENTED_SIZE does not.

> [OSCoreInnerBlockwiseTest.testOscoreExceedMaxUnfragmentedSize](https://github.com/eclipse/californium/pull/1300/files#diff-30dc1a5cd61b70377a692213961fa02dR216)
> 
> Why is that limit then checked in the test for inner blockwise?

The MAX_UNFRAGMENTED_SIZE should mainly function as a check on incoming messages. However it is also stated in the OSCORE RFC that a client/server should transmit messages where the payload size is over the MAX_UNFRAGMENTED_SIZE using inner block-wise. So this JUnit test is for checking the _outgoingExceedsMaxUnfragSize_ method to ensure that a client cannot send a message exceeding the MAX_UNFRAGMENTED_SIZE unless it is using inner block-wise. So for outgoing messages with inner block-wise the check versus MAX_UNFRAGMENTED_SIZE will not apply.
https://github.com/eclipse/californium/blob/f375eb886b2e7749e96fc7466a2e92ac9e9d83a6/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java#L401

https://tools.ietf.org/html/rfc8613#section-4.1.3.4.2
_Messages exceeding this size SHOULD be fragmented by the sending
endpoint using Inner Block options (Section 4.1.3.4.1)._


I put it in the JUnit test file for inner block-wise as the test is for making sure the client is using inner block-wise when it should. But perhaps I should split the MAX_UNFRAGMENTED_SIZE related tests into its own JUnit test class?

boaks:
I created a "wip_max_payload_size" with an alternative implementation.
It adds a max-payload-size to each message, so that it could be also used without oscore. In my opinion this simplifies also the implementation with oscore (see the usage in the new branch).
Maybe you can check, if it works for you. Then the new `BlockwiseTransferInfo` coudl be removed and I prepare a PR to introduce this max-payload-size into the master.
(Unfortunately, I have to postpone that to next week, sorry.)

rikard-sics:
> I created a "wip_max_payload_size" with an alternative implementation.
> It adds a max-payload-size to each message, so that it could be also used without oscore. In my opinion this simplifies also the implementation with oscore (see the usage in the new branch).
> Maybe you can check, if it works for you. Then the new `BlockwiseTransferInfo` coudl be removed and I prepare a PR to introduce this max-payload-size into the master.

Thanks! I will check it and come back to you.

> (Unfortunately, I have to postpone that to next week, sorry.)

No worries, thanks for all the assistance.

rikard-sics:
> I created a "wip_max_payload_size" with an alternative implementation.
> It adds a max-payload-size to each message, so that it could be also used without oscore. In my opinion this simplifies also the implementation with oscore (see the usage in the new branch).
> Maybe you can check, if it works for you. Then the new `BlockwiseTransferInfo` coudl be removed and I prepare a PR to introduce this max-payload-size into the master.
> (Unfortunately, I have to postpone that to next week, sorry.)

Thanks again for the help regarding this.

I have checked the code and functionality, and it looks like just what we are after. It functions like what I had in mind for my code. Please tell me if I can assist in any way for making the final PR.

boaks:
OK, I cherrypick an polish my changes.

boaks:
I pushed the change to PR #1326 .

Please prepare this PR to use the changes there.
I plan to merge #1326 on Monday, and this PR then on Tuesday, if its ready.

rikard-sics:
> I pushed the change to PR #1326 .
> 
> Please prepare this PR to use the changes there.
> I plan to merge #1326 on Monday, and this PR then on Tuesday, if its ready.

Okay, sounds great. I will proceed with updating this PR accordingly.

rikard-sics:
> Okay, sounds great. I will proceed with updating this PR accordingly.

Alright, this PR should now be updated according to the code from PR https://github.com/eclipse/californium/pull/1326 you prepared.

If everything looks good I can squash and rebase. (I guess I should wait until the PR you made has been merged first.)

boaks:
> (I guess I should wait until the PR you made has been merged first.)

It's already merged. Don't hesitate to rebase, squash and test.
On success forced push your branch.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
```diff
@@ -111,7 +113,18 @@
...
+	/**
+	 * When using outer block-wise with OSCORE a proxy can maliciously inject
+	 * block fragments. To protect against this a message with size exceeding
+	 * this parameter should never be sent without inner block-wise. Likewise
+	 * when receiving a message using outer block-wise it should be discarded if
+	 * the cumulated size exceeds this parameter.
+	 * 
+	 * See https://tools.ietf.org/html/rfc8613#section-4.1.3.4.2
+	 */
+	private int MAX_UNFRAGMENTED_SIZE;
```
boaks, line 126: Only static final uses UPPER_CASE. Please use CaMeL, `maxUnfragmentedSize`.

rikard-sics: Thanks, I will change it.

rikard-sics:
> t's already merged. Don't hesitate to rebase, squash and test.
> On success forced push your branch.

Sounds good, I will proceed with that later today.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
rikard-sics: This has now been fixed.

rikard-sics:
> It's already merged. Don't hesitate to rebase, squash and test.
> On success forced push your branch.

I have now rebased it, squashed the commits and tested with "mvn test".

boaks:
Thanks!

PR #1302, opened 2020-05-05
Add map of algorithms to thread local crypto functions.

boaks:
Use that for signature and hash algorithms.


sbernard31:
no objection

PR #1306, opened 2020-05-05
OSCORE: Fix getRTT method when using OSCORE

rikard-sics:
Previously the getRTT method did not work on responses when using OSCORE. This has now been fixed by assigning the RTT when creating the new response.


boaks:
LGTM

Issue #1307, opened 2020-05-06
Prevent to use 2 certificate with same subject truststore ?


sbernard31:
It seems there is nothing to prevent to put 2 different certificate with the same ~issuer~ subject in the scandium trust store.

Should we add a check about that ?

boaks:
I'm not sure, if you mean "same subject"?
For issuers I'm not sure, why that should be checked.

sbernard31:
Yep I would mean same subject.

boaks:
I'm not sure about the rules, which applies to x509.

[Multiple Certificates with same DN](https://security.stackexchange.com/questions/174640/multiple-certificates-with-same-dn)

seems to permit that.

sbernard31:
Let's explain my story :)

I was doing some tests of Leshan (using demo). I created several root certificates that I used as truststore at server side.  
And at client side, I was doing some test with several certificate signed by those root certificates.

I faced a situation where handshake failed. After a while I finally discovered this was because I have several root certificates with exactly the same subject and  [search method](https://github.com/eclipse/californium/blob/2.2.2/element-connector/src/main/java/org/eclipse/californium/elements/util/CertPathUtil.java#L344) doesn't  pick the right one.

So my first idea was to avoid to have several certificate with same subject to avoid this behavior.

boaks:
OK, maybe we add a short javadoc to `setTrustStore`, that explains that rule and mention `setCertificateVerifier`,  if that rule should be bypassed?
(by the way, I think, the check order needs to be fixed as well. And:
```
public Builder setTrustStore(Certificate[] trustedCerts) {
	if (trustedCerts == null) {
		throw new NullPointerException("Trust store must not be null");
	} else if (config.certificateVerifier != null) {
		throw new IllegalStateException("Trust store must not be used after certificate verifier is set!");
	} else if (trustedCerts.length == 0) {
		config.trustStore = new X509Certificate[0];
	} else {
		config.trustStore = SslContextUtil.asX509Certificates(trustedCerts);
		checkTrustStore();
	}
	return this;
}
```
)

sbernard31:
> OK, maybe we add a short javadoc to setTrustStore, that explains that rule and mention setCertificateVerifier, if that rule should be bypassed?

We can do that (simple solution) or we check against all certificates (less simple one)

What do you mean by check order ? 

`checkTrustStore()` before `config.trustStore = SslContextUtil.asX509Certificates(trustedCerts);` ?

[...]

PR #1308, opened 2020-05-06
Ensure there is not several certificates with same subject in truststore

sbernard31:
(link to #1307)

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -42,7 +42,11 @@ import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+
+import javax.security.auth.x500.X500Principal;
 
 import org.eclipse.californium.elements.DtlsEndpointContext;
 import org.eclipse.californium.elements.util.CertPathUtil;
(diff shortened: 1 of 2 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -2487,5 +2488,17 @@ private void determineCipherSuitesFromConfig() {
 
 			config.supportedCipherSuites = ciphers;
 		}
+
+		private void checkTrustStore() {
+			X509Certificate[] store = config.trustStore;
+			for (int i = 0; i < store.length; i++) {
```
boaks, line 2494: What's about using `CertPathUtil.toSubjects` and `List.contains`?

boaks:
As I wrote in [my comment](https://github.com/eclipse/californium/issues/1307#issuecomment-624718489), I'm not sure about the rules, which applies to x509.

If you think, that is worth to be checked, go for it.
Maybe you replace the nested for-loops by one loop and a `List.contains`.

```
while (principal = list.remove(0)) {
 if (list.contains(principal)) {
 }
}
```

sbernard31:
Let's think and discuss a bit more about that in #1307

sbernard31:
(I suspect this will need more discussion than expected so maybe we can let this PR aside for the 2.2.3)

boaks:
You may add it, I don't object that.
I would prefer to add the javadoc I recommend in the issue comment.

boaks:
LGTM.
(squash before merge?)

boaks:
> (squash before merge?)

Uups, I mixed up the number of files and commits.

sbernard31:
(integrated in master too 3e5bb4d)

PR #1314, opened 2020-05-12
Terminate connection on close, if no established session is available.

boaks:


boaks:
This is required for `Connection.isClosed()`, which fails without session.
For 2.2.x I'm not sure, if this changes anything.

PR #1316, opened 2020-05-13
Async API variant

sbernard31:
Just to discuss about #1304 . Not intended to be merged.


scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
```diff
@@ -1021,8 +1017,8 @@ protected PskSecretResult requestPskSecretResult(PskPublicInformation pskIdentit
 		String hmacAlgorithm = session.getCipherSuite().getPseudoRandomFunctionMacName();
 		pskRequestPending = true;
 		this.otherSecret = SecretUtil.create(otherSecret);
-		return advancedPskStore.requestPskSecretResult(connection.getConnectionId(), serverNames, pskIdentity,
-				hmacAlgorithm, otherSecret, generateRandomSeed());
+		return advancedPskStore.requestMasterSecret(connection.getConnectionId(), serverNames, pskIdentity,
+				hmacAlgorithm, otherSecret, generateRandomSeed(), resultHandler);
```
boaks, line 1021: Now the mechanism has the `resultHandler` as state.
My intention of the API was to allow a stateless implementation (at least no new state).
It would just send away information and any result back could be processed without additional state.
The "other_secret" belongs more to the state, the handshaker needs anyway.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/AbstractAsyncPskStore.java
```diff
@@ -0,0 +1,64 @@
...
+			final String hmacAlgorithm, SecretKey otherSecret,  byte[] seed, final PskSecretResultHandler callback) {
+		
+		final byte[] localSeed = Arrays.copyOf(seed, seed.length);
+		final SecretKey localOtherSecret = SecretUtil.create(otherSecret);
+		requestPSKSecret(serverName, identity, new PskSecretResultHandler() {
+			@Override
+			public void apply(PskSecretResult secretResult) {
+				// no psk for this identity.
+				if (secretResult == null || secretResult.getSecret() == null) {
+					callback.apply(null);
```
boaks, line 40: a callback calls a callback with `null`. 
And I can't see, where the cid is returned to map that result.

boaks:
This approach may match your expectations, but for me, it doesn't make it simpler.
I will check it out to see, how it works.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/AbstractAsyncPskStore.java
sbernard31: I write it quickly, I suppose I did a mistake here we probably need to return a PskSecretResult with null as key.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
sbernard31: I didn't take this stateless requirement into account.
Without it, I think it's pretty common to use callback for async API.

Do you have an exemple of code or a usecase where this stateless requirement is needed ? I have difficulty to see the benenfit right now.

boaks:
Let me try to explain my approach again:

Collect all information, which is required to generate the master secret:

cid
server names
psk indentity
hmac algorithm
other secret
seed

That may be processed or send as message to a remote store.

The result contains:

cid
psk identity (optionally normalized, for none utf-8)
secret (optional, including algorithm, master secret or psk secret key)

That could be processed without other additional information by the dtls connector/handshaker.

The callback is a general one, no handshake specific.
The communication with a remote store will not require request/response in the protocol of that communication, both messages contain all information required for processing.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java
boaks: Why should such a specific callback be used, if the general one works as well?
There is just no reason to always create a new callback instead of using the single one for the connector.

boaks: By the way: 
I thought, we have been discussing, if the master secret should have always been returned. I don't see, that this change focus on the difference of returning the master secret. Or do I miss something?

sbernard31: 1. To avoid the setter in an Interface which is generally not so elegant.
2. We could imagine to not look in the connectionStore when processing result by creating a dedicate callback (but this is not compatible with your stateless requirement)
3. this allow to have a consistent API with [AbstractSyncPskStore](https://github.com/eclipse/californium/blob/async_psk_store/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/AbstractSyncPskStore.java#L51)

sbernard31:
I understand your approach, It's just hard to me to find an example of a such kind of protocol in this case :thinking:

boaks:
I don't get it.
If I use the async flow here in your proposal, it is the same information passed around.
The only difference is the callback, you create for every handshake and pass it in the call. You may even use always the same callback instance and pass that to all handshakers, then the difference is, that you pass the callback for each request call, I pass it once in the initialization.

boaks:
Maybe, as additional hint:
The task to solve for the async PSK is to pass the async processing of the result to the serialized execution. Using a java callback alone will not work. I introduced the callback not for the async processing of the result, I introduced it to decouple the psk store from the dtls connector, otherwise both will depend on each other.

sbernard31:
(could we either continue the discussion in the main thread or in the review comment above :grin:)

Ok to be clearer : 
If stateless implementation is needed, I have no idea better idea than your
If stateless is not required, I think my solution is a bit more intuitive.

That's why I try to understand why/if this is required ?

sbernard31:
> You may even use always the same callback instance and pass that to all handshaker

This is what I do [here](https://github.com/eclipse/californium/blob/async_psk_store/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Handshaker.java#L1015)
But I can not do it  [here](https://github.com/eclipse/californium/blob/async_psk_store/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/pskstore/AbstractAsyncPskStore.java#L35)

And as I didn't consider the stateless requirement that was not an issue.

sbernard31:
I leave you for today. It's time for cooking :man_cook:

boaks:
I would like to end this discussion.
It will be easy for you, to provide a "abstract class" in leshan, which injects the callback into each call as you want. For me, this idea does not bring any benefit.

sbernard31:
I think a bit more about this, this night but visibly for nothing.

I still don't understand why we need a stateless API.
I think about the current backend we are using at sierra for pskstore  : 
- redis
- SQL
- csv file
- HSM with an http API
- backend with REST API
I can not see any of them where this stateless API could help but I would be happy to change my mind.

And if stateless is not needed, I feel the API I propose is more intuitive :
- method names are more clear
- method does not return 2 completely different kind of value
- if you want to implement a sync store it's more straightforward
- if you want to implement a async store it's more straightforward.
- remove the setter (which should generally be avoid in service interface)
- (probably not a big deal but with your solution a pskStore can not be shared by several connector)

Anyway, it seems your choice is already done ...

boaks:
> Anyway, it seems your choice is already done ...

Yes, decades ago :-).

sbernard31:
> Yes, decades ago :-).

...

boaks:
My decision was even before I started to work with Californium.
For me, the "callback hell" is something, I don't like.

sbernard31:
"Your decision", Yes Sir !

sbernard31:
By the way, as I understand "callback hell", this is not about using callback, this is just about how bad you could write your code based on callback API.

Issue #1041, opened 2019-08-20
Release cf-nat ?


sbernard31:
I'm currently using `NatUtil` for writing some tests.
AFAIK, `NatUtil` is not released so I copy/paste it.

Maybe it could make sense to release it ?
I don't see any issue to do that. Maybe the only one could be the `logback.xml` in `/cf-nat/src/main/resources` which could make trouble if it is integrated in a real application. (because if there are several `logback.xml` we don't know which one will be used...)

boaks:
> Maybe it could make sense to release it ?

If wanted, we can add it to the release job.

sbernard31:
Ok.

For logback.xml issue a solution could be to rename `logback.xml` in something like `logback-config.xml` and add this static block : 

```java
    static {
        // Define a default logback.configurationFile
        String property = System.getProperty("logback.configurationFile");
        if (property == null) {
            System.setProperty("logback.configurationFile", "logback-config.xml");
        }
    }
```
in `Nat` class.

But If you prefer we can start by just releasing cf-nat like this and see if we really face logback.xml classpath issue.

boaks:
Sorry, I forget to answer.
Please add that stuff to the Nat.

sbernard31:
I integrated the #1041 in master and modify jobs :
- [californium-deploy-maven-central](https://ci.eclipse.org/californium/view/Release/job/californium-deploy-maven-central/)
- [californium-deploy-eclipse](https://ci.eclipse.org/californium/view/Release/job/californium-deploy-eclipse/)

by adding this to the maven step : 
> --projects
californium-osgi,californium-proxy,cf-oscore,demo-apps/cf-plugtest-server,**demo-apps/cf-nat**
-am
deploy

I hope I didn't miss something.

@boaks, if it's OK for you we can close this issue.

boaks:
The jobs looks good for me.
We will see next week, if it works :-),

sbernard31:
After so many time I finally try to integrate the release version instead of copy/paste one and I get this error : 
```
Failed to collect dependencies at org.eclipse.californium:cf-nat:jar:2.2.3:
Failed to read artifact descriptor for org.eclipse.californium:cf-nat:jar:2.2.3: 
Could not find artifact org.eclipse.californium:demo-apps:pom:2.2.3 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
```

I didn't investigate so much but my guess is `demo-apps` need to be release too ?

boaks:
[...]

PR #1318, opened 2020-05-14
Add cf-utils folder.

boaks:
Introduce bom module.
Move cf-nat and cf-unix-health into cf-utils.
Move californium-integration-tests and
californium-interoperability-tests into tests.


boaks:
This PR seems to require a "parent bootstrap".
I think, it's required to introduce a "bom" module as well.

boaks:
@sophokles73 

It would be great, if you can check if the introduced "bom" module complies to common patterns. I tried to implement it according the bom in hono.

cf-utils/cf-nat/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+	xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.eclipse.californium</groupId>
+		<artifactId>cf-bom</artifactId>
+		<version>2.3.0-SNAPSHOT</version>
+		<relativePath>../../bom</relativePath>
+	</parent>
```
sophokles73, line 13: Is there a particular reason why you do not use *cf-utils* as the parent?

boaks: My intention is to use the folder just as collection. If a pom is added to the "cf-utils" folder, that must be published as well in order to be able to us the cf-nat (see issue #1041 ).  Though the modules may vary much more than others "cf-nat" (independent) and (new to come) "cf-cli" (basic command line interface with picocli, highly dependent on californium), the pom on cf-utils will not contain too much information.

Issue #1321, opened 2020-05-20
ObservationUtil.shallowClone() exception


gaomingok:
https://github.com/eclipse/californium/blob/97a4d50e7e3cbdc11837cea963830fc862ba6a46/californium-core/src/main/java/org/eclipse/californium/core/observe/ObservationUtil.java#L49
**clonedRequest.setPayload(request.getPayload());**
If request.isUnintendedPayload ()==true && request.getCode()==GET, this line will throw an exception, causing an observe exception.

boaks:
Please check, if PR #1322 would fix this bug.

GET with payload is considered to be not compliant. Maybe you consider to use FETCH.
[FETCH, PATCH, IPATCH](https://tools.ietf.org/html/rfc8132)

boaks:
PR #1322 is merged. 
If that fixes this issue, please close it.

PR #1322, opened 2020-05-20
Fix issue 1321.

boaks:
Clone "unintended payload" before the payload itself.

Add newFetch, newPatch, and new IPatch to Request.


boaks:
Fixes #1321

PR #1324, opened 2020-05-20
Fix session resumption using RPK

boaks:
Limit supported cipher suites to selected cipher suite.
Reconstruct certificate type from principal.


sbernard31:
Not really specific to this bug, but do you plan to backport bug fixes in 2.2.x branch when this is important bug or easy to backport one ?

I ask this because, I think about trying to stay on : 
 - californium 2.2.x for Leshan 1.0.x.
 - californium 2.x for Leshan 1.x

(I didn't find any clear recommendation about semantic versioning and dependency upgrade :confused: )

boaks:
If it is easy, yes.
If it is important, depends on the time to invest.

For this bug I even don't know, if this was introduced by the "cipher suite selector" or not. For the next days I guess I will be too busy to backport it,

Issue #1301, opened 2020-05-02
OSCORE Pre-shared Key 


eliot4321:
Hi,

I can see that in the OSCORE  example you are assuming that the keys are pre-shared "Following the OSCORE RFC".  However, the first request sent from an OSCORE client within the same session it's always taking too long than other following requests(for example 500 ms first request, 
 20 to 30  ms the following ones). Could you please explain the reason behind this behavior? 

Thanks in advance

boaks:
What is your network setup?
Intranet?
Internet?  
Could you check the times using "plain coap", no DTLS and no OSCORE?

eliot4321:
Two nodes one machine running. the oscore server and one machine running the oscore client, both they in the same private WLAN network provided as a hotspot by the server machine,  to have a clean network.

The first request within CoAP is normal as the others.

eliot4321:
Hi,
A clarification:  Should the first OScore request behave as the consecutive requests?

Thanks

onovo1:
Hi,

I got similar results. The first OSCORE request really takes very long time. I have tested it in my own Intranet. Is there any configuration setup to improve this performance?

boaks:
@rikard-sics 

Do you have any guess about that?

rikard-sics:
Hello.

Thank you both for reporting this issue. I need to investigate further to understand why this is happening and if something is taking long in the code for the first request. There is nothing in inherent in OSCORE itself (like a handshake) that should slow down the first request.

May I ask if you also experience these issues when having the OSCORE-stack enabled with _OSCoreCoapStackFactory.useAsDefault(db)_ but then sending the messages as CoAP-only messages (without adding the OSCORE option)?

I have tried running some tests of my own and I do see a slowdown for the first request (around 100 ms compared to 2 ms for the following messages). For CoAP-only I see around 30 ms for the first request versus 1 ms for the rest.

(I also realize now that the getRTT() method does not work on responses when using OSCORE. I will fix this in a future PR.)

eliot4321:
Hi,

I tried after disabling the OSCORE option and I'm still getting latency on the first request, even I have other comments that the following requests took 15 ms in average  on running in localhost that's is quite high value.

rikard-sics:
Thanks for the quick follow-up.

> I tried after disabling the OSCORE option and I'm still getting latency on the first request.
I have another question: the following requests took 15 ms "quite high value" on average while running in localhost?

[...]

PR #1332, opened 2020-05-26
OSCORE: Initialize cipher on context creation

rikard-sics:
Initializes the cipher object during creation of the OSCORE context by calling CCMBlockCipher.encrypt (which in turn initializes the cipher). Doing this reduces the latency for the first request since it would otherwise happen upon encryption/decryption of the first request.

As discussed in https://github.com/eclipse/californium/issues/1301


boaks:
LGTM

PR #1336, opened 2020-05-28
OSCORE: Use ID Context to disambiguate when retrieving contexts

rikard-sics:
When retrieving an OSCORE context for an incoming request both the Recipient ID (RID/KID) and ID Context should be used. However the ID Context may not be included in a request (to save overhead). In such case the correct context can still be found if there is only 1 stored under that RID. If there are multiple contexts stored under that RID an error message will be returned and the client can retransmit the request now including an ID Context.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
@@ -26,13 +26,15 @@ import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import org.eclipse.californium.core.coap.CoAP.ResponseCode;
 import org.eclipse.californium.core.coap.Token;
+import org.eclipse.californium.elements.util.Bytes;
 
 /**
  * 
@@ -46,7 +48,9 @@ public class HashMapCtxDB implements OSCoreCtxDB {
 	 */
 	private static final Logger LOGGER = LoggerFactory.getLogger(HashMapCtxDB.class);
 
-	private HashMap<ByteId, OSCoreCtx> ridMap;
+	// The outer HashMap has RID as key and the inner ID Context
+	private HashMap<ByteId, HashMap<ByteId, OSCoreCtx>> contextMap;
+
 	private HashMap<Token, OSCoreCtx> tokenMap;
 	private HashMap<String, OSCoreCtx> uriMap;
 	private HashMap<Token, Integer> seqMap;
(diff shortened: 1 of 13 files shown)


cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
```diff
@@ -59,20 +64,62 @@
...
-			return ridMap.get(new ByteId(rid));
-		} else {
+	public synchronized OSCoreCtx getContext(byte[] rid, byte[] IDContext) throws CoapOSException {
+		// Do not allow a null RID
+		if (rid == null) {
 			LOGGER.error(ErrorDescriptions.BYTE_ARRAY_NULL);
 			throw new NullPointerException(ErrorDescriptions.BYTE_ARRAY_NULL);
 		}
+
+		HashMap<ByteId, OSCoreCtx> matchingRidMap = contextMap.get(new ByteId(rid));
```
boaks, line 86: Just as exercise for good coding practice: count the `new ByteId(byte[])` and check, if this could be reduced by at least reuse that ByteId within the method. It may get even less used, if you store the rid as ByteId and not as byte array.

cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
```diff
@@ -59,20 +64,62 @@
...
+			}
+
+		} else {
+			// If retrieving using both RID and ID Context
+			return matchingRidMap.get(new ByteId(IDContext));
+		}
+	}
+
+	/**
+	 * Retrieve context using only RID when it is certain it is unique.
```
boaks, line 111: I would prefer a exception, if that "certain" is broken.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OptionJuggle.java
```diff
@@ -328,4 +329,73 @@ public static OptionSet merge(OptionSet eOptions, OptionSet uOptions) {
...
 
+	/**
+	 * Retrieve RID value from an OSCORE option.
+	 * 
+	 * @param oscoreOption the OSCORE option
+	 * @return the RID value
+	 */
+	static byte[] getRid(byte[] oscoreOption) {
+		if (oscoreOption.length == 0) {
+			oscoreOption = new byte[] { 0x00 };
```
boaks, line 340: For me this case will always return `null`, or?

cf-oscore/src/main/java/org/eclipse/californium/oscore/OptionJuggle.java
```diff
@@ -328,4 +329,73 @@ public static OptionSet merge(OptionSet eOptions, OptionSet uOptions) {
...
+
+	/**
+	 * Retrieve ID Context value from an OSCORE option.
+	 * 
+	 * @param oscoreOption the OSCORE option
+	 * @return the ID Context value
+	 */
+	static byte[] getIDContext(byte[] oscoreOption) {
+		if (oscoreOption.length == 0) {
+			oscoreOption = new byte[] { 0x00 };
```
boaks, line 377: For me this case will always return `null`, or?

boaks:
Beside the repeats of `new ByteId(byte[])` and the processing of empty oscore options in  `getRid` and `getIDContext` it's LGTM.
If you able to spend some time in reducing the `new ByteId(byte[])` I would prefer that. Otherwise we can merge it with the news and you may address that in a later PR.

cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
rikard-sics: Yes, better to check and throw an exception. I have now added this in commit 53c1507badbae05cbd12830f8e750701eab8f3bb

cf-oscore/src/main/java/org/eclipse/californium/oscore/OptionJuggle.java
rikard-sics: Yes, you mean that I could return null straight in that if, so having?
```
static byte[] getIDContext(byte[] oscoreOption) {
		if (oscoreOption.length == 0) {
			return null;
}
```
(And same for getRID).

rikard-sics: I updated to return null immediately in commit 9a654a3f67270da32a728120f62cfc0e5919d242

cf-oscore/src/main/java/org/eclipse/californium/oscore/OptionJuggle.java
rikard-sics: Changed in commit 9a654a3f67270da32a728120f62cfc0e5919d242

rikard-sics:
> Beside the repeats of `new ByteId(byte[])` and the processing of empty oscore options in `getRid` and `getIDContext` it's LGTM.
> If you able to spend some time in reducing the `new ByteId(byte[])` I would prefer that. Otherwise we can merge it with the news and you may address that in a later PR.

Sounds great. Thanks for the feedback. I will try to work now to reduce the usage of `new ByteId`.

cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
rikard-sics: True, it was used excessively. Thanks for the feedback. I have now made a commit trying to reduce how much it is used, see commit 19f0fc1ebdae42eaff3b70dbedd60866730c2ab7

As for storing the rid differently, perhaps that could be a thing for future improvements?

boaks:
Please rebase and squash.

rikard-sics:
> Please rebase and squash.

Okay, I have now rebased and squashed.

PR #1338, opened 2020-06-11
OSCORE: Add support for more AES CCM algorithms

rikard-sics:
Added support for using the following algorithms with OSCORE:
AES-CCM-64-64-128
AES-CCM-16-128-128
AES-CCM-64-128-128

Previously only AES-CCM-16-64-128 was supported.

The actual encryption/decryption uses the CCMBlockCipher.encrypt and CCMBlockCipher.decrypt methods from Scandium.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/cose/EncryptCommon.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/cose/EncryptCommon.java
@@ -54,7 +54,8 @@ import org.eclipse.californium.scandium.dtls.cipher.CCMBlockCipher;
  */
 public abstract class EncryptCommon extends Message {
 
-	private final int AES_CCM_16_IV_LENGTH = 13;
+	private final static int AES_CCM_16_IV_LENGTH = 13;
+	private final static int AES_CCM_64_IV_LENGTH = 7;
 	protected String context;
 	protected byte[] rgbEncrypt;
 
@@ -65,14 +66,11 @@ public abstract class EncryptCommon extends Message {
 		if (rgbEncrypt == null)
 			throw new CoseException("No Encrypted Content Specified");
 
-		switch (alg) {
-		case AES_CCM_16_64_128:
-			AES_CCM_Decrypt(alg, rgbKey);
-			break;
-
-		default:
+		if (!isSupportedAesCcm(alg)) {
 			throw new CoseException("Unsupported Algorithm Specified");
 		}
+
+		AES_CCM_Decrypt(alg, rgbKey);
 
 		return rgbContent;
 	}
@@ -84,25 +82,13 @@ public abstract class EncryptCommon extends Message {
 		if (rgbContent == null)
 			throw new CoseException("No Content Specified");
 
-		switch (alg) {
-		case AES_CCM_16_64_128:
-			AES_CCM_Encrypt(alg, rgbKey);
-			break;
-
-		default:
+		if (!isSupportedAesCcm(alg)) {
 			throw new CoseException("Unsupported Algorithm Specified");
 		}
+
+		AES_CCM_Encrypt(alg, rgbKey);
 	}
 
-	private int getAES_CCM_IVSize(AlgorithmID alg) throws CoseException {
-		switch (alg) {
-		case AES_CCM_16_64_128:
-			return AES_CCM_16_IV_LENGTH;
-		default:
-			throw new CoseException("Unsupported Algorithm Specified");
-		}
-	}
-	
 	//Method taken from EncryptCommon in COSE. This will provide the full AAD / Encrypt0-structure.
     private byte[] getAADBytes() {
         CBORObject obj = CBORObject.NewArray();
@@ -127,7 +113,7 @@ public abstract class EncryptCommon extends Message {
 		}
 
 		// obtain and validate IV
-		final int ivLen = getAES_CCM_IVSize(alg);
+		final int ivLen = ivLength(alg);
 		CBORObject iv = findAttribute(HeaderKeys.IV);
 		if (iv == null) {
 			throw new CoseException("Missing IV during decryption");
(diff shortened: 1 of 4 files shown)


cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
```diff
@@ -834,7 +842,11 @@ public synchronized void rollBack() {
 	private static int ivLength(AlgorithmID alg) {
```
boaks, line 842: I would prefer, if this is moved to "EncryptCommon" (use the finals there) and made public.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
```diff
@@ -689,14 +689,22 @@ public void setSeqMax(int seqMax) {
...
+			switch (common_alg) {
+			case AES_CCM_16_64_128:
+			case AES_CCM_64_64_128:
+			case AES_CCM_16_128_128:
+			case AES_CCM_64_128_128:
 				iv_length = ivLength(common_alg);
-				id_length = 7;
-				key_length = common_alg.getKeySize() / 8; // 16;
-			} else {
+				id_length = iv_length - 6; // RFC section 5.2
```
boaks, line 699: If the iv-length is 7, the resulting id-length is 1. Is that intended?

rikard-sics: Yes, that is correct. That is why OSCORE only supports AEAD algorithms with nonces 7 bytes or more.

Having a maximum length of the sender/recipient ID be 1 byte only is small, but there is also the ID Context that can be used to identify a specific context (if more than 256 are needed).

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
rikard-sics: Yes, that makes sense. This has now been done in commit 8c0bee8df9e1eb0fb862725a36d47c2ed6f4965a

cf-oscore/src/main/java/org/eclipse/californium/cose/EncryptCommon.java
```diff
@@ -97,7 +104,12 @@ void encryptWithKey(byte[] rgbKey) throws CoseException, IllegalStateException {
 	private int getAES_CCM_IVSize(AlgorithmID alg) throws CoseException {
```
boaks, line 104: Now the difference to the new `public static int ivLength(AlgorithmID alg)` seems to be very small. Maybe you redesign it a little to remove this one?

rikard-sics: True, they are very similar now. I will redesign things to only keep 1.

Just to clarify, you mean removing getAES_CCM_IVSize and keeping ivLength, or the opposite?

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
```diff
@@ -689,14 +690,22 @@ public void setSeqMax(int seqMax) {
...
-				id_length = 7;
-				key_length = common_alg.getKeySize() / 8; // 16;
-			} else {
+
+			switch (common_alg) {
+			case AES_CCM_16_64_128:
+			case AES_CCM_64_64_128:
+			case AES_CCM_16_128_128:
+			case AES_CCM_64_128_128:
+				iv_length = EncryptCommon.ivLength(common_alg);
```
boaks, line 699: `ivLength` would return a `-1` for not supported algorithms (or maybe a exception after redesign). Checking for the returned value (or exception) would obsolete the repeated "switch(algo)".

cf-oscore/src/main/java/org/eclipse/californium/cose/EncryptCommon.java
```diff
@@ -218,4 +232,20 @@ private void AES_CCM_Encrypt(AlgorithmID alg, byte[] rgbKey) throws CoseExceptio
...
+		case AES_CCM_16_64_128:
+		case AES_CCM_16_128_128:
+			return AES_CCM_16_IV_LENGTH;
+		case AES_CCM_64_64_128:
+		case AES_CCM_64_128_128:
+			return AES_CCM_64_IV_LENGTH;
+		default:
+			return -1;
+		}
+	}
```
boaks, line 250: Maybe adding a  `boolean isSupportedAesCcm(AlgorithmID alg)` obsoletes some general `switch(alg)` as above in line 69 or 91. Or use return value of this `ivLength()` instead of the `switch(alg)`.

cf-oscore/src/main/java/org/eclipse/californium/cose/EncryptCommon.java
boaks: I guess, the `public static` one is the one to keep. if you keep the -1 or the exception depends on the calling contexts. Just check, what will fit better for you.

rikard-sics:
I made a new commit now baed0fb9426cbc57dd303a290de9e474f8cb762c with some updates from our discussion.

Changes done:

> Removed the duplicate method for getting the IV length
> Added a method for checking if an algorithm is supported
> Removed some switch statements

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
boaks: It's OK, if you want to keep this switch. I would replace it with `iv_length > 0`.

boaks:
You may replace the switch (or not :-). If you ready, please rebase and squash.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCtx.java
rikard-sics: I will update this one also. I was too focused on the EncryptCommon and missed this one.

rikard-sics:
> You may replace the switch (or not :-). If you ready, please rebase and squash.

I updated removed that switch statement and have now rebased and squashed.

PR #1340, opened 2020-06-15
Adapt oscore bytes processing.

boaks:


boaks:
@rikard-sics 

Would it be possible for you to check, if these changes are working for you?

rikard-sics:
> Would it be possible for you to check, if these changes are working for you?

Yes indeed. Let me have a look and reply shortly.

rikard-sics:
> Would it be possible for you to check, if these changes are working for you?

Yes, the changes look good for me. Thanks for the update.

Btw. I was thinking about removing the client boolean at some point as it is only used in a few JUnit tests.

Issue #1323, opened 2020-05-20
2.3.0-RC2 - Released


boaks:
I plan to release the 2.3.0-RC1 end of next week, 28. May.

Though it comes with many changes, I consider a RC1 to get more feedback about the changes.

Changes in DTLS:
- supports hello_extension for signature and hash algorithms.
  The default changes from "ecdsa_sha256" to "ecdsa_sha256,rsa_sha256".
  It's also possible to configure a different list of supported signature and hash
  algorithms   
- supports x25519 curve for ECDHE (running java 11)
  The default curves changes from "secp256r1,secp384r1,secp521r1" to
  "secp256r1,x25519,x448,secp384r1".  It's also possible to configure a
  different list of supported curves.
- Asynchronous PSK store API, supports returning master secret or secret key
- CipherSuiteSelector for more control on selecting cipher suite and parameters
  on the server side.

Changes in CoAP:
- add (experimental) peer's based deduplicator to limit maximum number of messages
  kept for deduplication. 
- improve multicast support and add broadcast support. 
- add picocli command-line-interfaces to plugtest clients and servers.

Additionally, some bugs have been fixed, please refer to the reflog.

boaks:
2.3.0-RC1 is available at the [Eclipse Repo](https://repo.eclipse.org/content/repositories/californium-releases/org/eclipse/californium/) and on [Maven Central](https://search.maven.org/search?q=g:org.eclipse.californium%20a:parent%20v:2.3.0-RC1).

Feedback is very welcome and helps, to release the 2.3.0 on 16. June.

rikard-sics:
One piece of feedback from me. I see the following line printed when running my CoAP applications for testing:
https://github.com/eclipse/californium/blob/8fa8051aafae279d54116165d2c87af712e1092a/element-connector/src/main/java/org/eclipse/californium/elements/util/NetworkInterfacesUtil.java#L153

Is it supposed to be printed with println or the LOGGER?

boaks:
Thanks! I fix it.

sbernard31:
- `getPskStore()` is now deprecated and `setPskStore` is not.
- `setPskStore` does not set advancedPskStore attribute.

As a user I can use `setPskStore` or `setAdvancedPskStore`
But when I handle an `incompleteConfig` , If I want to check if a kind of pskStore is set I need to check both using `getAdvancedPskStore()` and the deprecated `getPskStore()`
[...]

PR #1343, opened 2020-06-16
Reduce cache synchronization

janssk1:
The InMemoryCache synchronization is taking too much locks. Even secondary cache lookups are were done in synchronized blocks. 
This commit moves all secondary cache read/updates out of synchronized blocks and also removes some other unnecessary synchronization.

### DIFF
--- a/scandium-core/api-changes.json
+++ b/scandium-core/api-changes.json
@@ -168,6 +168,11 @@
 					"old": "class org.eclipse.californium.scandium.dtls.PskUtil",
 					"justification": "not part of the public API - obsolete",
 					"oldArchive": "org.eclipse.californium:scandium:jar:2.1.0"
+				},{
+					"code": "java.method.addedToInterface",
+					"new": "method org.eclipse.californium.scandium.dtls.Connection org.eclipse.californium.scandium.dtls.ResumptionSupportingConnectionStore::get(java.net.InetSocketAddress, org.eclipse.californium.scandium.dtls.ResumptionSupportingConnectionStore.ConnectionMapper)",
+					"oldArchive": "org.eclipse.californium:scandium:jar:2.1.0",
+					"justification": "Performance improvement. Cache is responsible now for all synchronization"
 				}
 			]
 		}
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -1063,36 +1063,43 @@ public class DTLSConnector implements Connector, RecordLayer {
 	 *         exhausted or if the connection is not available and the provided
 	 *         parameter create is {@code false}.
 	 */
-	private final Connection getConnection(InetSocketAddress peerAddress, ConnectionId cid, boolean create) {
-		ExecutorService executor = getExecutorService();
-		synchronized (connectionStore) {
+	private final Connection getConnection(final InetSocketAddress peerAddress, final ConnectionId cid, final boolean create) {
+		final ExecutorService executor = getExecutorService();
 			Connection connection;
 			if (cid != null) {
 				connection = connectionStore.get(cid);
 			} else {
-				connection = connectionStore.get(peerAddress);
-				if (connection == null && create) {
-					LOGGER.trace("create new connection for {}", peerAddress);
-					Connection newConnection = new Connection(peerAddress, new SerialExecutor(executor));
-					if (running.get()) {
-						// only add, if connector is running!
-						if (!connectionStore.put(newConnection)) {
-							return null;
+				connection = connectionStore.get(peerAddress, new ResumptionSupportingConnectionStore.ConnectionMapper() {
+
+					@Override
+					public Connection remapConnection(Connection existingConnection) {
+						if (existingConnection == null) {
+							if (create) {
+								LOGGER.trace("create new connection for {}", peerAddress);
+								return new Connection(peerAddress, new SerialExecutor(executor));
+							} else {
+								return null;
+							}
+						} else {
+							return existingConnection;
 						}
 					}
-					return newConnection;
-				}
+
+				});
 			}
 			if (connection == null) {
 				LOGGER.trace("no connection available for {},{}", peerAddress, cid);
-			} else if (!connection.isExecuting() && running.get()) {
-				LOGGER.trace("revive connection for {},{}", peerAddress, cid);
-				connection.setExecutor(new SerialExecutor(executor));
 			} else {
-				LOGGER.trace("connection available for {},{}", peerAddress, cid);
+				synchronized (connection) {
+					if (!connection.isExecuting() && running.get()) {
+						LOGGER.trace("revive connection for {},{}", peerAddress, cid);
+						connection.setExecutor(new SerialExecutor(executor));
+					} else {
+						LOGGER.trace("connection available for {},{}", peerAddress, cid);
+					}
+				}
 			}
 			return connection;
-		}
 	}
 
 	/**
@@ -1694,7 +1701,7 @@ public class DTLSConnector implements Connector, RecordLayer {
 	 * @param record record of CLIENT_HELLO message
 	 */
 	private void processNewClientHello(final Record record) {
-		InetSocketAddress peerAddress = record.getPeerAddress();
+		final InetSocketAddress peerAddress = record.getPeerAddress();
 		if (LOGGER.isTraceEnabled()) {
 			StringBuilder msg = new StringBuilder("Processing new CLIENT_HELLO from peer [")
 					.append(peerAddress).append("]").append(":").append(StringUtil.lineSeparator()).append(record);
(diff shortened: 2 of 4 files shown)


boaks:
Thanks for your contribution.
Did you read [CONTRIBUTING](https://github.com/eclipse/californium/blob/master/CONTRIBUTING.md)?
If you want to contribute, you need a [ECA](http://www.eclipse.org/legal/ECA.php). That permits other to use your work. Please use the e-mail of that ECA to sign the commit.

janssk1:
Did read contributing. I'm trying to get all the legal stuff right.  I figured i might just a well get some early feedback. If the changes don't make sense, no need for me to continue the painful procedure..

boaks:
Sure. I currently build it and I will run some benchmarks to see the benefit.

boaks:
Did you measure a performance gain?

usually I use the cf-extplugtest-server running with

```
java -Xmx10g -XX:+UseG1GC -Xverify:none -jar cf-extplugtest-server-${CF_VERSION}.jar --no-ipv6 --benchmark
```

And the cf-extplugtest-client running the "benchmark.sh" (you may disable the plain and tcp execution by 

```
USE_TCP=0
USE_UDP=1
USE_PLAIN=0
USE_SECURE=1
USE_HTTP=0
USE_OBSERVE=1
```

in that script.

Result of master:

> [2020-06-16 18:06:35]	Linux, 4.15.0-106-generic, amd64
> [2020-06-16 18:06:35]	-XX:+UseG1GC -Xmx6g -Xverify:none 
> [2020-06-16 18:06:35]	coaps://nuc5005:5784/benchmark?rlen=40 --clients 2000 --requests 5000 --no-stop 
> [2020-06-16 18:06:35]	uptime: 302213 ms, 16 processors
> [2020-06-16 18:06:35]	cpu-time: 1377160 ms (per-processor: 86072 ms, load: 28%)
> [2020-06-16 18:06:35]	gc: 1680 ms, 203 calls
> [2020-06-16 18:06:35]	average load: 5,59
> [2020-06-16 18:06:36]	10000000 requests sent, 10000000 expected
> [2020-06-16 18:06:36]	10000000 requests in 302213 ms, 33089 reqs/s
> [2020-06-16 18:06:36]	195602 retransmissions (1,96%)
> [2020-06-16 18:06:36]	2 transmission errors (0,00%)

Result of this PR:

> [2020-06-16 18:32:33]	Linux, 4.15.0-106-generic, amd64
> [2020-06-16 18:32:33]	-XX:+UseG1GC -Xmx6g -Xverify:none 
> [2020-06-16 18:32:33]	coaps://nuc5005:5784/benchmark?rlen=40 --clients 2000 --requests 5000 --no-stop 
> [2020-06-16 18:32:33]	uptime: 305848 ms, 16 processors
> [2020-06-16 18:32:33]	cpu-time: 1387002 ms (per-processor: 86687 ms, load: 28%)
> [2020-06-16 18:32:33]	gc: 1847 ms, 238 calls
> [2020-06-16 18:32:33]	average load: 4,02
> [2020-06-16 18:32:34]	10000000 requests sent, 10000000 expected
> [2020-06-16 18:32:34]	10000000 requests in 305848 ms, 32696 reqs/s
> [2020-06-16 18:32:34]	198117 retransmissions (1,98%)

That is extremely close and the diff is within in the diffs from run to run even with the same version.

So:
Did you measure the gain? Could you describe the results and how your test is constructed?

janssk1:
We did extensive performance tests using our server. The synchronization issue manifest itself when using secondary storage for DTLS sessions. When lots of devices hit the server at the same time, one thread updating a DTLS session toward the secondary storage blocks all other threads trying to read anything from the (primary) cache. 
I think you will have to add secondary session cache to your perf tests to reproduce the problem.

boaks:
So you use a `SessionCache`?
Is that a local one? Or a remote one?

boaks:
I would assume/consider:

```
SessionCache {
	void put(DTLSSession session);
	SessionTicket get(SessionId id);
	void remove(SessionId id);
}
```

that the `get` is the critical one. That is used in `Connection find(final SessionId id)`and to reduce the scope of  the synchronization there should be easier. 

So, do you have different experience?
Could you give me some estimations, how long your put-get-remove calls take?

janssk1:
Sessioncache indeed. Our implementation is remote (redis). AFAIK, the point of the session cache is to be remote. Removing syncrhonization during get will help. But i also don't see a reason to block the complete cache when updating one element. 
Duration of the calls depends on various factors. But if you use  a few milliseconds you should be able to easily see the difference between between this patched and the unpatched version.

boaks:
> But i also don't see a reason to block the complete cache when updating one element.

I plan to release 2.3.0-RC2 today, and the 2.3.0 on next Wednesday, 24. June (see issue #1323 ). Unfortunately, If have only one workday left (this Wednesday), until next Wednesday I'm out of the office. That makes it hard to react on your findings for 2.3.0.

Some parts this PR is changing (the `DTLSConnector synchronized (connectionStore)` by introducing the `ConnectionMapper`) have been trouble-hot-spots in the past. So I would prefer not to change them for 2.3.0. For a 2.4.0 that `ConnectionMapper` may be an idea.

Generally the `SessionCache` seems for me to be not completely "specified". I don't know, if it is assumed to be "consistent" or "eventually consistent". The "removes" in case of a missing session id in the cache seems to point to the first, but intended "remote implementations" would then slow down too much. So, FMPOV, there is even specification work left, which then may change the implementations. That will be "time consuming".

I currently don't use this `SessionCache` and currently I prefer to improve the connection id usage in a cluster. Therefore my plan to invest time is more to make the "connection id cluster" working, than a clustered session resumption. With that it depends on others (including you :-) ), how much time you want to spend into the `SessionCache`.

I prepared a PR #1344 with less changes. Assuming "eventually consistent", the `put` and `remove` could be executed asynchronous and so the implementations may just start that operations. With that, I don't see, that for now it is important to adapt the `put` as well.

I will also prepare an issue for those who are interested to work on the `SessionCache`. We will see, how that is evolving.

janssk1:
I have also noticed that ConnectionId's don't support clustering yet. So are you saying that going forward scandium will focus on fixing/extending the connectionId handling, and phase out the SessionCache ? With a perfect connectionId cache, is there still a need for sessioncache ? In other words, is the connectionId cache also going to cache session details ?

sbernard31:
> Therefore my plan to invest time is more to make the "connection id cluster" working

Could you describe what doesn't work for now ?
And what would mean exactly  "connection id cluster" working ?
Are you talking about a share/persistent connection store ?

boaks:
> I have also noticed that ConnectionId's don't support clustering yet. 

Did you read [DTLS-1.2-connection-ID-based-load-balancer](https://github.com/eclipse/californium/wiki/DTLS-1.2-connection-ID-based-load-balancer)? 

> So are you saying that going forward scandium will focus on fixing/extending the connectionId handling, and phase out the SessionCache ? 

"going forward scandium" depends on more people than just me. If there is interest in the `SessionCache` and others are spending their time into it, my part of the work will not be that large. At least, I don't plan a " phase out the SessionCache". But I also  don't plan to spend too much of my time into it.

> With a perfect connectionId cache, is there still a need for sessioncache ?

My current idea of using the connection id is not to "share the connection state" in a cache, it's to forward the message to the right node.

boaks:
> AFAIK, the point of the session cache is to be remote.

Just to mention.
The session cache may also be used to persist the session for restarts. The doesn't require remote components. And there are also implementations possible, which share the session ahead and don't load them on demand.

boaks:
@janssk1 

OK to postpone this to 2.4.0?

Or would you prefer the alternative PR #1344 with less changes be included in 2.3.0?

janssk1:
I think its fine to postpone. You are bringing up some good points. The (current) cost of remote DTLS session caching probably does not justify its benefit (prevent a few full handshakes). We were thinking of either fixing the synchronization issues or removing it all together. We will go for the former for now.

boaks:
Let me add my view:

The penalty from "double sync" is not that large (compared to the crypto-calculations done afterwards).
But, if a pool-thread is blocked for longer to enter the sync-monitor, that results in penalties.

From my experience, it is important to keep the connection store consistent.
But it may be possible to access the session cache out side that consistency.
Though some may rely on that "sync" behavior, I would adapt my PR to make it configurable, if the session cache is processed within that sync or outside.
That may be a first step. If the session cache calls blocks the calling thread, that affects also the performance, and would require to use "much more threads", than cpu-cores (as before the  async PSK of the 2.3), or it requires to add a "AsynSessionCache". But that would require more time, than I currently want to spend in.

So, I would prefer, if you generally try to check, if the other PR comes also with performance benefits for you. If so, I would go to make it configurable.

boaks:
@janssk1 

Are you still interested? What do you think about my proposal in the last comment above?

boaks:
Just for your information:

PR #1550 comes now with a cleanup of this `SessionCache` API. It's based on the discussions in #1345. I still didn't spent the required in it, so it's still not that well tested. Anyway, I assume it's better than before. Let me know, if your interested in testing it.

boaks:
You may have a look at PR #1618 

With that, it should be possible to implement strategies to resume a session within a cluster (e.g. "lazy loading/loading on request").

Issue #1298, opened 2020-04-27
Have more information about the foreign peer in logs.


sbernard31:
It would great to have more information about the foreign peer in logs (Initially needed for scandium but maybe it makes sense on californium too)

In Scandium, the idea could to use `connection.toString` each time it's possible (instead of just peer address), but for now `connection.toString` does not show the peer identity (e.g. PSK id or Public key)

It could be added to the current `toString` OR/AND  a new interface could be created to allow user to print information they need.

Personally, my guess is `peer address` + `conId` + `sessionId` + `peerIdentity` should cover most of the use case, so maybe the interface is not really needed.

(This issue was created from the discussion in : https://github.com/eclipse/californium/issues/1284)

boaks:
For me it seems to be worth to check, if [MDC](http://www.slf4j.org/manual.html#mdc), e.g. introduce a "PEER" and load it with the proper information, would solve this evene a little more general.

So it would require some "volunteers" to investigate on that. I assume, I will not be able to spend my time into that for the next week (may be months).

AndreasMager:
We have some trouble with debugging scandium connections too.
Adding a `session id` to the log lines would enhance debugging of dtls connections.

I have played with your MDC approach in the org.eclipse.californium.scandium.dtls.Handshaker. 
MDC is a static resource for the current running thread. It must be filled and cleaned on each thread change.
If we add more Key/Value pairs for peer, conId & peerIdenty to MDC it becomes even more complicated.

All code sections with importance should be changed to this:
`
MDCCloseable closeMe = fillMDC();
try {
  processMessage(record) 
} finally {
 closeMe.close();
}`

It is not an easy task to figure the right spots for these changes.

Refactoring all log lines and add a "connection.toString" seems to be easier approach.

boaks:
Thanks for your feedback.

My idea was to provide the peer's information in "callbacks" for the `SerialExecutor` of the `Connection` (means: if the serial executer starts a job in `scheduleNextJob()`, there will be a callback to set the MDC, and a callback to clean it. Or something similar).

Beside of that, threads are only rarely assigned to peers:
Incoming: protected void processDatagram(DatagramPacket packet)
Outgoing: public final void send(final RawData message)

FMPOV, that should be not that big issue.

> We have some trouble with debugging scandium connections too.

Any trouble to report? Any found bugs?

AndreasMager:
I will take a deeper dive in the code and give you feedback. 

> Any trouble to report? Any found bugs?

[...]

PR #1346, opened 2020-06-23
WIP: Add mdc logging fields 

AndreasMager:
See #1298 for more information.


boaks:
Thanks for your contribution.

I would prefer, if the setup of the MDC is handled in a callback.
That keeps the `SerialExecutor` more decoupled from the logging internals.
Could you check, if such an approach would work for you as well?

AndreasMager:
Sure will test your branch and close this one.

PR #1357, opened 2020-07-01
Handle ENTITY TOO LARGE for transparent blockwise transfert

### LINKED ISSUE #395: Handle ENTITY TOO LARGE  for transparent blockwise transfert
Base on the [RFC 7959](https://tools.ietf.org/html/rfc7959#section-2.9.3) 

> a 4.13 response to a request with a Block1 Option (control
   usage, see Section 2.3) where the response carries a smaller SZX in
   its Block1 Option is a hint to try that smaller SZX.

sbernard31:
This is a WIP PR aiming to fix #395.

Triggered by [mailing list discussion](https://www.eclipse.org/lists/cf-dev/msg01762.html).

Currently, I only support that a block1 request with too large block size is negotiated into block1 request with smaller block size.

But we should also be able to handle a normal request (without block1) with too large payload which should be turned out in block1 request.

The [specification says](https://tools.ietf.org/html/rfc7959#section-2.9.3) : 
> It also allows  the server to return a 4.13 response to a request that does not
   employ Block1 as a hint for the client to try sending Block1.

But this is not clear how :thinking:  ?
1.  should the coap server add a block1 option 
2. or should we use the size1 option ?

solution 1.  sounds only used to answer to request with block1 option.
solution 2. sounds to so adapted as size1 is more about the whole payload size and this forces client to guess the block size to use.. 

Any thought about that ?

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -791,7 +791,7 @@ public class BlockwiseLayer extends AbstractLayer {
 			newSize = currentSize;
 			newSzx = status.getCurrentSzx();
 		}
-		int nextNum = status.getCurrentNum() + currentSize / newSize;
+		int nextNum = (status.getCurrentNum() + 1) * status.getCurrentSize() / newSize;
 		LOGGER.debug("sending next Block1 num={}", nextNum);
 		Request nextBlock = null;
 		try {
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -76,6 +76,7 @@ import org.eclipse.californium.core.coap.MessageObserverAdapter;
 import org.eclipse.californium.core.coap.OptionSet;
 import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.coap.Response;
+import org.eclipse.californium.core.coap.Token;
 import org.eclipse.californium.core.network.Exchange;
 import org.eclipse.californium.core.network.config.NetworkConfig;
 import org.eclipse.californium.core.network.config.NetworkConfigDefaults;
@@ -153,6 +154,10 @@ public class BlockwiseLayer extends AbstractLayer {
 	 * matches the example in the draft.
 	 */
 
+	// Minimal block size : 2^4 bytes
+	// (see https://tools.ietf.org/html/rfc7959#section-2.2)
+	private static final int MINIMAL_BLOCK_SIZE = 16;
+
 	private static final Logger LOGGER = LoggerFactory.getLogger(BlockwiseLayer.class);
 	private static final Logger HEALTH_LOGGER = LoggerFactory.getLogger(LOGGER.getName() + ".health");
 	private final LeastRecentlyUsedCache<KeyUri, Block1BlockwiseStatus> block1Transfers;
@@ -167,6 +172,8 @@ public class BlockwiseLayer extends AbstractLayer {
 	private int maxResourceBodySize;
 	private boolean strictBlock2Option;
 	private int healthStatusInterval;
+	/* @since 2.4 */
+	private boolean enableAutoFailoverOn413;
 
 	/**
 	 * Creates a new blockwise layer for a configuration.
@@ -221,6 +228,9 @@ public class BlockwiseLayer extends AbstractLayer {
 		strictBlock2Option = config.getBoolean(NetworkConfig.Keys.BLOCKWISE_STRICT_BLOCK2_OPTION, NetworkConfigDefaults.DEFAULT_BLOCKWISE_STRICT_BLOCK2_OPTION);
 
 		healthStatusInterval = config.getInt(NetworkConfig.Keys.HEALTH_STATUS_INTERVAL, 60); // seconds
+
+		enableAutoFailoverOn413 = config.getBoolean(NetworkConfig.Keys.BLOCKWISE_ENTITY_TOO_LARGE_AUTO_FAILOVER,
+				NetworkConfigDefaults.DEFAULT_BLOCKWISE_ENTITY_TOO_LARGE_AUTO_FAILOVER);
 
 		LOGGER.info(
 				"BlockwiseLayer uses MAX_MESSAGE_SIZE={}, PREFERRED_BLOCK_SIZE={}, BLOCKWISE_STATUS_LIFETIME={}, MAX_RESOURCE_BODY_SIZE={}, BLOCKWISE_STRICT_BLOCK2_OPTION={}",
@@ -314,7 +324,7 @@ public class BlockwiseLayer extends AbstractLayer {
 				
 				if (requiresBlockwise(request)) {
 					// This must be a large POST or PUT request
-					requestToSend = startBlockwiseUpload(exchange, request);
+					requestToSend = startBlockwiseUpload(exchange, request, preferredBlockSize);
 				}
 			}
 		}
@@ -323,7 +333,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		lower().sendRequest(exchange, requestToSend);
 	}
 
-	private Request startBlockwiseUpload(final Exchange exchange, final Request request) {
+	private Request startBlockwiseUpload(final Exchange exchange, final Request request, int blocksize) {
 
 		final KeyUri key = getKey(exchange, request);
 
@@ -336,10 +346,14 @@ public class BlockwiseLayer extends AbstractLayer {
 				status.cancelRequest();
 				clearBlock1Status(key, status);
 			}
-			status = getOutboundBlock1Status(key, exchange, request);
+			status = getOutboundBlock1Status(key, exchange, request, blocksize);
 
 			final Request block = status.getNextRequestBlock();
 			block.setDestinationContext(request.getDestinationContext());
+			Token token = request.getToken();
+			if (token != null) {
+				block.setToken(token);
+			}
 			block.addMessageObserver(new MessageObserverAdapter() {
 
 				@Override
@@ -630,6 +644,10 @@ public class BlockwiseLayer extends AbstractLayer {
 				case REQUEST_ENTITY_INCOMPLETE: // 4.08
 					// we seem to have uploaded blocks not in expected order
 				case REQUEST_ENTITY_TOO_LARGE: // 4.13
+					if (handleEntityTooLarge(exchange, response)) {
+						return;
+					}
+
 					// server is not able to process the payload we included
 					KeyUri key = getKey(exchange, exchange.getCurrentRequest());
 					Block1BlockwiseStatus status = getBlock1Status(key);
@@ -694,6 +712,84 @@ public class BlockwiseLayer extends AbstractLayer {
 			exchange.setResponse(response);
 			upper().receiveResponse(exchange, response);
 		}
+	}
+
+	/**
+	 * Handle 4.13 Entity Too Large error.
+	 * 
+	 * @param exchange current exchange
+	 * @param response the Entity Too Larger response.
+	 * @return {@code true} if the response is handled by auto failover
+	 */
+	private boolean handleEntityTooLarge(Exchange exchange, Response response) {
+		if (enableAutoFailoverOn413) {
+			if (response.getOptions().hasBlock1()) {
+
+				BlockOption block1 = response.getOptions().getBlock1();
+				final KeyUri key = getKey(exchange, exchange.getRequest());
+
+				synchronized (block1Transfers) {
+					Block1BlockwiseStatus status = getBlock1Status(key);
+					if (status == null) {
+						// We sent a request without using block1 and
+						// server give us hint it want it with block1
+						Request request = exchange.getRequest();
+						if (!exchange.getRequest().isCanceled() && block1.getNum() == 0
+								&& block1.getSize() < request.getPayloadSize()) {
+							// Start block1 transfer
+							Request blockRequest = startBlockwiseUpload(exchange, request,
+									Math.min(block1.getSize(), preferredBlockSize));
+							exchange.setCurrentRequest(blockRequest);
+							lower().sendRequest(exchange, blockRequest);
+							return true;
+						}
+					} else if (!status.hasMatchingToken(response)) {
+						// a concurrent block1 transfer has been started in
+						// the meantime which has "overwritten" the status
+						// object with the new (concurrent) request to we simply
+						// discard the response
+						LOGGER.debug("discarding obsolete block1 response: {}", response);
+						return true;
+					} else if (exchange.getRequest().isCanceled()) {
+						clearBlock1Status(key, status);
+						return true;
+					} else {
+						// we handle only Entity Too Large
+						// at begin of the transfer and
+						// if blocksize requested is smaller
+						if (status.getCurrentNum() == 0 && block1.getSize() < status.getCurrentSize()) {
+							// re-send first block with smaller size
+							sendBlock(exchange, response, key, status, 0, block1.getSzx());
+							return true;
+						}
+					}
+				}
+			} else if (!exchange.getRequest().isCanceled()
+					&& getBlock1Status(getKey(exchange, exchange.getRequest())) == null) {
+				// We sent a request without using block1 and
+				// server give us hint it want it with block1
+				Request request = exchange.getRequest();
+				// Try to guess the a block size to use
+				Integer maxSize = null;
+				if (response.getOptions().hasSize1() && response.getOptions().getSize1() >= MINIMAL_BLOCK_SIZE
+						&& response.getOptions().getSize1() < request.getPayloadSize()) {
+					maxSize = response.getOptions().getSize1();
+				} else if (request.getPayloadSize() > MINIMAL_BLOCK_SIZE) {
+					maxSize = request.getPayloadSize() - 1;
+				}
+
+				// Start blockwise if we guess a correct size
+				if (maxSize != null) {
+					int blocksize = Integer.highestOneBit(maxSize);
+					Request requestToSend = startBlockwiseUpload(exchange, request,
+							Math.min(blocksize, preferredBlockSize));
+					exchange.setCurrentRequest(requestToSend);
+					lower().sendRequest(exchange, requestToSend);
+					return true;
+				}
+			}
+		}
+		return false;
 	}
 
 	/**
@@ -778,7 +874,8 @@ public class BlockwiseLayer extends AbstractLayer {
 		}
 	}
 
-	private void sendNextBlock(final Exchange exchange, final Response response, final KeyUri key, final Block1BlockwiseStatus status) {
+	private void sendNextBlock(final Exchange exchange, final Response response, final KeyUri key,
+			final Block1BlockwiseStatus status) {
 
 		BlockOption block1 = response.getOptions().getBlock1();
 		int currentSize = status.getCurrentSize();
@@ -792,10 +889,15 @@ public class BlockwiseLayer extends AbstractLayer {
 			newSzx = status.getCurrentSzx();
 		}
 		int nextNum = (status.getCurrentNum() + 1) * status.getCurrentSize() / newSize;
-		LOGGER.debug("sending next Block1 num={}", nextNum);
+		sendBlock(exchange, response, key, status, nextNum, newSzx);
+	}
+
+	private void sendBlock(final Exchange exchange, final Response response, final KeyUri key,
+			final Block1BlockwiseStatus status, int num, int szx) {
 		Request nextBlock = null;
+		LOGGER.trace("sending Block1 num={}", num);
 		try {
-			nextBlock = status.getNextRequestBlock(nextNum, newSzx);
+			nextBlock = status.getNextRequestBlock(num, szx);
 			// we use the same token to ease traceability
 			nextBlock.setToken(response.getToken());
 			nextBlock.setDestinationContext(status.getFollowUpEndpointContext(response.getSourceContext()));
@@ -1057,12 +1159,12 @@ public class BlockwiseLayer extends AbstractLayer {
 		}
 	}
 
-	private Block1BlockwiseStatus getOutboundBlock1Status(final KeyUri key, final Exchange exchange, final Request request) {
+	private Block1BlockwiseStatus getOutboundBlock1Status(final KeyUri key, final Exchange exchange, final Request request, int blocksize) {
 
 		synchronized (block1Transfers) {
 			Block1BlockwiseStatus status = block1Transfers.get(key);
 			if (status == null) {
-				status = Block1BlockwiseStatus.forOutboundRequest(exchange, request, preferredBlockSize);
+				status = Block1BlockwiseStatus.forOutboundRequest(exchange, request, blocksize);
 				block1Transfers.put(key, status);
 				enableStatus = true;
 				LOGGER.debug("created tracker for outbound block1 transfer {}, transfers in progress: {}", status,
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
@@ -347,6 +347,40 @@ public class BlockwiseClientSideTest {
 	}
 
 	/**
+	 * Block1 late negotiation in the middle of the transfer.
+	 */
+	@Test
+	public void testLateNegotiationInTheMiddle() throws Exception {
+		System.out.println("Blockwise PUT with late negotiation in the middle");
+		reqtPayload = generateRandomPayload(290);
+		String path = "test";
+
+		Request request = createRequest(PUT, path, server);
+		request.setPayload(reqtPayload);
+		client.sendRequest(request);
+
+		server.expectRequest(CON, PUT, path).storeBoth("A").block1(0, true, 128).size1(reqtPayload.length())
+				.payload(reqtPayload.substring(0, 128)).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("A").block1(0, true, 128).go();
+
+		server.expectRequest(CON, PUT, path).storeBoth("B").block1(1, true, 128).payload(reqtPayload.substring(128, 256)).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("B").block1(1, true, 32).go(); // late negotiation
+
+		server.expectRequest(CON, PUT, path).storeBoth("C").block1(8, true, 32).payload(reqtPayload.substring(256, 288)).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("C").block1(8, true, 32).go();
+
+		server.expectRequest(CON, PUT, path).storeBoth("D").block1(9, false, 32)
+				.payload(reqtPayload.substring(288, 290)).go();
+		server.sendResponse(ACK, CHANGED).loadBoth("D").block1(9, false, 32).go();
+
+		Response response = request.waitForResponse(RESPONSE_TIMEOUT_IN_MS);
+		assertThat(response, is(notNullValue()));
+		assertThat(response.getPayloadSize(), is(0));
+		assertThat(response.getCode(), is(CHANGED));
+		assertThat(response.getToken(), is(request.getToken()));
+	}
+
+	/**
 	 * <pre>
 	 * CLIENT                                                     SERVER
      * |                                                          |
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
@@ -513,6 +513,284 @@ public class BlockwiseClientSideTest {
 	}
 
 	/**
+	 * Verifies that a block1 transfer with a 4.13 code at beginning with a smaller size1 option is retried with a smaller blocksize.
+	 * 
+	 * @throws Exception if the test fails.
+	 */
+	@Test
+	public void testBlockwisePUTWithBegining413AndSmallerSZX() throws Exception {
+
+		System.out.println("Block1 with  REQUEST_ENTITY_TOO_LARGE negotiation");
+		reqtPayload = generateRandomPayload(128 + 10);
+		String path = "test";
+
+		Request request = createRequest(PUT, path, server);
+		request.setPayload(reqtPayload);
+		client.sendRequest(request);
+
+		server.expectRequest(CON, PUT, path).storeBoth("A").block1(0, true, 128).size1(reqtPayload.length()).payload(reqtPayload, 0, 128).go();
+		server.sendResponse(ACK, REQUEST_ENTITY_TOO_LARGE).loadBoth("A").block1(0, false, 64).go();
+		server.expectRequest(CON, PUT, path).storeBoth("B").block1(0, true, 64).size1(reqtPayload.length()).payload(reqtPayload, 0, 64).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("B").block1(0, true, 64).go();
+		server.expectRequest(CON, PUT, path).storeBoth("C").block1(1, true, 64).payload(reqtPayload, 64, 128).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("C").block1(1, true, 64).go();
+		server.expectRequest(CON, PUT, path).storeBoth("D").block1(2, false, 64).payload(reqtPayload, 128, 138).go();
+		server.sendResponse(ACK, CHANGED).loadBoth("D").block1(2, false, 64).go();
+
+		Response response = request.waitForResponse(RESPONSE_TIMEOUT_IN_MS);
+		assertThat(response, is(notNullValue()));
+		assertThat(response.getPayloadSize(), is(0));
+		assertThat(response.getCode(), is(CHANGED));
+		assertThat(response.getToken(), is(request.getToken()));
+	}
+
+	/**
+	 * Verifies that none block request will turn in block transfer if 4.13 code
+	 * is received at beginning with a smaller size1 option.
+	 * 
+	 * @throws Exception if the test fails.
+	 */
+	@Test
+	public void testPUTWithBeginning413WithSize1TurnInBlockPUT() throws Exception {
+
+		System.out.println("REQUEST_ENTITY_TOO_LARGE with smaller size1 turn PUT in blockwise transfer");
+		reqtPayload = generateRandomPayload(128);
+		String path = "test";
+
+		Request request = createRequest(PUT, path, server);
+		request.setPayload(reqtPayload);
+		client.sendRequest(request);
+
+		server.expectRequest(CON, PUT, path).storeBoth("A").payload(reqtPayload).go();
+		server.sendResponse(ACK, REQUEST_ENTITY_TOO_LARGE).loadBoth("A").size1(90).go();
+		server.expectRequest(CON, PUT, path).storeBoth("B").block1(0, true, 64).size1(reqtPayload.length()).payload(reqtPayload, 0, 64).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("B").block1(0, true, 64).go();
+		server.expectRequest(CON, PUT, path).storeBoth("C").block1(1, false, 64).payload(reqtPayload, 64, 128).go();
+		server.sendResponse(ACK, CHANGED).loadBoth("C").block1(1, false, 64).go();
+
+		Response response = request.waitForResponse(RESPONSE_TIMEOUT_IN_MS);
+		assertThat(response, is(notNullValue()));
+		assertThat(response.getPayloadSize(), is(0));
+		assertThat(response.getCode(), is(CHANGED));
+		assertThat(response.getToken(), is(request.getToken()));
+	}
+
+	/**
+	 * Verifies that none block request will turn in block transfer if 4.13 code
+	 * is received at beginning with out option.
+	 * 
+	 * @throws Exception if the test fails.
+	 */
+	@Test
+	public void testPUTWithBeginning413WithoutOptionTurnInBlockPUT() throws Exception {
+
+		System.out.println("REQUEST_ENTITY_TOO_LARGE without option turn PUT in blockwise transfer");
+		reqtPayload = generateRandomPayload(128);
+		String path = "test";
+
+		Request request = createRequest(PUT, path, server);
+		request.setPayload(reqtPayload);
+		client.sendRequest(request);
+
+		server.expectRequest(CON, PUT, path).storeBoth("A").payload(reqtPayload).go();
+		server.sendResponse(ACK, REQUEST_ENTITY_TOO_LARGE).loadBoth("A").go();
+		server.expectRequest(CON, PUT, path).storeBoth("B").block1(0, true, 64).size1(reqtPayload.length()).payload(reqtPayload, 0, 64).go();
+		server.sendResponse(ACK, CONTINUE).loadBoth("B").block1(0, true, 64).go();
+		server.expectRequest(CON, PUT, path).storeBoth("C").block1(1, false, 64).payload(reqtPayload, 64, 128).go();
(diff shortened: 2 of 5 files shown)


boaks:
Questions about interpretations of a RFC are best asked in the IETF core mailing list.

I'm currently not too deep in RFC 7959. As far as I remember:

If the resource is too large, 4.13 may indicate that just by including a size1. That may be detected at the start of the transfer (if the client includes a size1) or somewhere in the middle.
In that case, I don't see, that the client is able to fix it. 

If the request is too large, 4.13 may be used to negotiate a smaller block1 size. But In my memory, such blocksize negotiation are usually done at the begin of the transfer and not "in between", as this PR might do. 

One left general question to me is, if the server will be able to response at all, if the request is too large.
Assuming using DTLS and a truncated message, this message will get filtered out by the MAC.
Assuming the message is received, I'm not sure, how many libraries will be able to process the just the header to get the MID and token, but not the rest. Let's see how many use-cases will really benefit from this.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -630,6 +630,12 @@ public void receiveResponse(final Exchange exchange, final Response response) {
 				case REQUEST_ENTITY_INCOMPLETE: // 4.08
 					// we seem to have uploaded blocks not in expected order
 				case REQUEST_ENTITY_TOO_LARGE: // 4.13
+					if (response.getOptions().hasBlock1() && exchange.getCurrentRequest().getOptions().hasBlock1()
+							&& response.getOptions().getBlock1().getSzx() < exchange.getCurrentRequest().getOptions()
+									.getBlock1().getSzx()) {
+						handleBlock1Response(exchange, response);
+						return;
+					}
```
boaks, line 638: I'm not sure, if a transfer should change the blocksize in the middle of the transfer. I would assume, this is only done at the begin.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -731,10 +737,14 @@ private void handleBlock1Response(final Exchange exchange, final Response respon
 
 				clearBlock1Status(key, status);
 
+			} else if (response.getCode() == ResponseCode.REQUEST_ENTITY_TOO_LARGE
+					&& block1.getSize() < status.getCurrentSize()) {
+
+				SendBlockAgain(exchange, response, key, status);
```
boaks, line 743:
```
block1.getSize() < status.getCurrentSize()
```

If the blocksize is changed only at the begin, I guess, these checks will have the same result as the above test.
If the blocksize should be changed within the transfer, I'm not sure, when this test should fail and why.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -792,10 +803,22 @@ private void sendNextBlock(final Exchange exchange, final Response response, fin
 			newSzx = status.getCurrentSzx();
 		}
 		int nextNum = status.getCurrentNum() + currentSize / newSize;
-		LOGGER.debug("sending next Block1 num={}", nextNum);
+		sendBlock(exchange, response, key, status, nextNum, newSzx);
+	}
+
+	private void SendBlockAgain(final Exchange exchange, final Response response, final KeyUri key,
```
boaks, line 809: Until now, I hope the most methods start with a lower case, `sendBlockAgain`.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -792,10 +803,22 @@ private void sendNextBlock(final Exchange exchange, final Response response, fin
...
+			final Block1BlockwiseStatus status) {
+		BlockOption block1 = response.getOptions().getBlock1();
+		int nextNum = status.getCurrentNum() * status.getCurrentSize() / block1.getSize();
+		sendBlock(exchange, response, key, status, nextNum, block1.getSzx());
```
boaks, line 813: As already mentioned, I don't think, that changing the blocksize in the middle makes sense.
For me this is just a new transfer starting at 0.
(For block2 it is assumed, that the first too large block is processed, but the client want to indicate (late), that using a smaller blocksize is preferred. But that's different from 4.13).

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: I understand your point but some part of the RFC makes me thinks that this could maybe be allowed. As this was not really more costly to handle it  I decided to go for it.

If you insist I can only support it for first answer.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: This is just a typo, I will fix it.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: > (For block2 it is assumed, that the first too large block is processed, but the client want to indicate (late), that using a smaller blocksize is preferred. But that's different from 4.13).

I don't really get your point about block2, this also exist for block1 : 
```
CLIENT                                                     SERVER
     |                                                          |
     | CON [MID=1234], GET, /status                     ------> |
     |                                                          |
     | <------   ACK [MID=1234], 2.05 Content, 2:0/1/128        |
     |                                                          |
     | CON [MID=1235], GET, /status, 2:2/0/64           ------> |
     |                                                          |
     | <------   ACK [MID=1235], 2.05 Content, 2:2/1/64         |
     |                                                          |
     | CON [MID=1236], GET, /status, 2:3/0/64           ------> |
     |                                                          |
     | <------   ACK [MID=1236], 2.05 Content, 2:3/1/64         |
     |                                                          |
     | CON [MID=1237], GET, /status, 2:4/0/64           ------> |
     |                                                          |
     | <------   ACK [MID=1237], 2.05 Content, 2:4/1/64         |
     |                                                          |
     | CON [MID=1238], GET, /status, 2:5/0/64           ------> |
     |                                                          |
     | <------   ACK [MID=1238], 2.05 Content, 2:5/0/64         |

              Figure 4: Block-Wise GET with Late Negotiation
```

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: [block1 size negotiation](https://tools.ietf.org/html/rfc7959#page-24)

I'm not sure, where changing the blocksize in the middle should be mentioned in RFC 7959.
Could you provide a link to that?

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: Yes, the point is

> I'm currently not too deep in RFC 7959. As far as I remember:

boaks: Just to mention: that changes the preferred blocksize at the begin.
That includes, that the server could process the large request, but prefers a smaller blocksize.
That uses 2.05 or 2.31, not 4.13.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: I keep it as I consider that this method should be responsible to check this too in case making it more safe to be reused and making it more readable (as the test is explicitly a translation of the RFC sentence)

But the code will maybe change again if we support to turn out a "normal" request (not block1) into a block1 one.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: The figure 4 is for block2, or?

boaks:
So, my questions are:
Are there really cases, where a server sends 4.13 to "negotiate a smaller blocksize".
Not, that this is more about fixing "broken servers".

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: My idea was to call sendBlockAgain already above.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: oups sorry : bad copy/paste : 

```
  CLIENT                                                     SERVER
     |                                                          |
     | CON [MID=1234], PUT, /options, 1:0/1/128    ------>      |
     |                                                          |
     | <------   ACK [MID=1234], 2.31 Continue, 1:0/1/32        |
     |                                                          |
     | CON [MID=1235], PUT, /options, 1:4/1/32     ------>      |
     |                                                          |
     | <------   ACK [MID=1235], 2.31 Continue, 1:4/1/32        |
     |                                                          |
     | CON [MID=1236], PUT, /options, 1:5/1/32     ------>      |
     |                                                          |
     | <------   ACK [MID=1235], 2.31 Continue, 1:5/1/32        |
     |                                                          |
     | CON [MID=1237], PUT, /options, 1:6/0/32     ------>      |
     |                                                          |
     | <------   ACK [MID=1236], 2.04 Changed, 1:6/0/32         |

          Figure 9: Simple Atomic Block-Wise PUT with Negotiation
```

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: Not clearly mention, I just see some sentences make me think that maybe this is allowed:  

> The present specification allows the server to return this response
   code at any time during a Block1 transfer to indicate that it does
   not currently have the resources to store blocks for a transfer that
   it would intend to implement in an atomic fashion.

(but reading it again, this is maybe more about stopping the transfer than negotiating the block size)

>    Finally, a 4.13 response to a request with a Block1 Option (control
   usage, see Section 2.3) where the response carries a smaller SZX in
   its Block1 Option is a hint to try that smaller SZX.

(not really explicit say this is only for first transfer) 

> Finally, the SZX block size given in a control Block1 Option
         indicates the largest block size preferred by the server for
         transfers toward the resource that is the same or smaller than
         the one used in the initial exchange; the client SHOULD use
         this block size or a smaller one in all further requests in the
         transfer sequence, even if that means changing the block size
         (and possibly scaling the block number accordingly) from now
         on.

(this makes maybe thing that this is only on first exchange)

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
sbernard31: It could be. I will taking this into account but first I would like to add the _"turn out a "normal" request (not block1) into a block1 one"_ use case.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: A question in the IETF core mailing list may bring clarification.

sbernard31:
> Questions about interpretations of a RFC are best asked in the IETF core mailing list.

I will ask all my concerns at core mainling list.

> Are there really cases, where a server sends 4.13 to "negotiate a smaller blocksize".

It seems there is hint which indicates there are : 
- the spec cover this case.
- several users ask for this.

I can also imagine a case where the buffer of the server has a limited static allocated size for block transfer.
This could also be just incompatible configuration issue, this negotiation makes this "incompatible configuration" interoperable.
So I don't know if you consider this as "broken servers" but It looks like pretty much real use case.

> As already mentioned, I don't think, that changing the blocksize in the middle makes sense.

I pretty much agree with this but
- rfc makes me a bit confuse (let's see if we get clarification on core mailing list)
-  this remark concerns also late negotiation (that uses 2.04, 2.05 or 2.31) and currently californium seems to support it. (block2 seems to work and block1 don't just because of a[ bad num block calculation](https://github.com/eclipse/californium/blob/2.3.0/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java#L794)) 
- support it could be as complex as not support it.

So either we support it, either we should define which behavior we want if a "broken server" try to do this kind of late negotiation.

boaks:
We will see, if those who have reported their issues, validate, that it works with this PR.

boaks:
> this remark concerns also late negotiation (that uses 2.04, 2.05 or 2.31) and currently californium seems to support it. (block2 seems to work and block1 don't just because of a bad num block calculation)

I'm not sure, if that works for block2 (2.05) or block1 (2.04 and 2.31).
My intention to point to this for 4.13 is, that in difference to the others, this one is used for more cases, including the cases to fail the transfer. So maybe using the first request, it will be easier to distinguish to retry with a smaller blocksize or to fail.

boaks:
sendNextBlock:

```
int nextNum = status.getCurrentNum() + currentSize / newSize;
```

SendBlockAgain:

```
int nextNum = status.getCurrentNum() * status.getCurrentSize() / block1.getSize();
```

My feeling is, your new stuff may work for current num != 0, but the old one will break there.

So with the answer from the list, it may be required, to fix the old stuff as well.

californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
```diff
@@ -464,6 +464,125 @@ public void testPUTFailsWith413IfBodyExceedsMaxBodySize() throws Exception {
...
+	 * Verifies that a block1 transfer with a 4.13 code at begining with a smaller szx option is retried with a smaller blocksize.
+	 * 
+	 * @throws Exception if the test fails.
+	 */
+	@Test
+	public void testBlockwisePUTWithBegining413AndSmallerSZX() throws Exception {
+
+		System.out.println("Blockwise PUT fails for excessive body size");
+		reqtPayload = generateRandomPayload(128 + 10);
+		respPayload = generateRandomPayload(30);
```
boaks, line 477: No big thing, but I would prefer not to use payload in the response for PUT.

boaks:
[CORE mailing list](https://mailarchive.ietf.org/arch/msg/core/BfOlcS_BpNXbksfjzz6iGpEa_SQ/)

About the question using block 1 in the response:

[RFC7959, Section 2.5, page 15](https://tools.ietf.org/html/rfc7959#page-15)

"(Note that a 4.13 response to a request that does not employ Block1 is
a hint for the client to try sending Block1, and a 4.13 response with a
smaller SZX in its Block1 Option than requested is a hint to try a
smaller SZX.)"

boaks:
Looks like the IETF core group is too busy.

I still feel (but sure can't proof), that only at the begin of a transfer (1. exchange), an adaption of the blocksize is intended. I would prefer to have the PR adapted. If you strongly want to keep a solution for changing it "in between", it's your work and I will not object it.

sbernard31:
> Looks like the IETF core group is too busy.

I was too :sweat_smile: 

So several remaining question : 
1. Do we really need this feature ? 
2. If yes, should we only begin transfer on 1 exchange ? 
  2.1 if yes, should we change the other negotiation behavior to only negotiate on first exchange ?
3. Should we also be able to handle a normal request (without block1) with too large payload which should be turned out in block1 request ? 
  3.1 if yes, how should it behave.

My current understanding : 
 1. You seem to not be so convinced about this (https://github.com/eclipse/californium/pull/1357#issuecomment-652837647)
Personally, I have doubt too but when user comes with this kind of remarks we have nothing to propose :confused: 
 2. I have no strong opinion, but I would prefer to have a consistent behavior between all block size negotiation mechanism so maybe the best solution is to just accept on first exchange for all negotiation ?
 
For 3.1, let's imagine some scenario : 
- client send a request without block
- server return 4.13 
- client will try to use block transfer  but it needs to calculate the block size : 
  - if there is no option : blocksize =  the "bigger 2^n" which is smaller than previous payload size ?
  - if there is a size1 : blocksize = the "bigger 2^n" which is smaller or equal  than size1  ?
  - if a block option with szx is used : blocksize = szx
- Now server could : 
   returns 2.31 => we continue the block transfer
   returns 4.13 with block option with smaller szx =>  we retry block mode using this szx.
   returns 4.13 with block option with greater or equal szx =>  we stop here, request failed.
   returns 4.13 without block option => we stop here, request failed.

tuve:
I am not entirely sure but when I implemented this for wakaama I think did as you suggest except in the case where the server returns 4.13 without block option, then I did retry with 2^n-1 until a lower limit of n.

boaks:
From my view, it's very simple:
I don't want to spend my time into blockwise for now. I'm too busy with other stuff.
The PR looks go to me, I would just prefer to change the blocksize only at the start.

sbernard31:
Thx @tuve for the feedback.

- I changed the code to limit ENTITY TOO LARGE negotiation only on first block.
- I fixed "normal" block1 size negotiation in the middle. (because this is easier than rejected it correctly)
- I added test for block2 size negotiation in the middle.
- I tried to implement : turning "normal" request into block transfer if 4.13 is returned.

But in the end I have a failing test about OSCORE :  `OSCoreOuterBlockwiseTest.testOuterBlockwiseExceedMaxUnfragmentedSizeProxyServerBW()`
(deactivated in last commit)
```
:04:09.883 WARN [UdpMatcher]: error receiving response ACK-4.13   MID=30467, Token=C4EB4CD55D9D6E7E, OptionSet={"Size1":64}, "body too large, can proc".. 40 bytes for Exchange[L1, complete] (o.e.c.c.n.UdpMatcher$4.run:442)
org.eclipse.californium.core.network.ExchangeCompleteException: Exchange[L1, complete] is already complete! CON-POST   MID=   -1, Token=null, OptionSet={"Object-Security":0x0902, "Proxy-Uri":"coap://127.0.0.1:54924"}, 
8c 46 3f 6a d1 78 0e 0a 1e 29 9e 76 5f 24 6a 50 23 50 2a e7 28 66 82 97 17 d6 74 49 af 5b c9 0b
b9 4a 99 3d 17 e6 93 fd 66 83 08 ee 97 07 17 eb 17 3b c9 95 6b 2e ab 7b 2b 49 26 c0 b6 22 54 ad
07 d6 69 dd 5e a4 05 5e e7 c3 56 7b 82 da 91 a4 af 81 42 a2 6c 8b dd 0f 53 af 
	at org.eclipse.californium.core.network.Exchange.assertIncomplete(Exchange.java:1125)
	at org.eclipse.californium.core.network.InMemoryMessageExchangeStore.registerWithMessageId(InMemoryMessageExchangeStore.java:236)
	at org.eclipse.californium.core.network.InMemoryMessageExchangeStore.registerOutboundRequest(InMemoryMessageExchangeStore.java:330)
	at org.eclipse.californium.core.network.UdpMatcher.sendRequest(UdpMatcher.java:132)
	at org.eclipse.californium.core.network.CoapEndpoint$OutboxImpl.sendRequest(CoapEndpoint.java:968)
	at org.eclipse.californium.core.network.stack.BaseCoapStack$StackBottomAdapter.sendRequest(BaseCoapStack.java:231)
	at org.eclipse.californium.core.network.stack.AbstractLayer.sendRequest(AbstractLayer.java:69)
	at org.eclipse.californium.oscore.ObjectSecurityLayer.sendRequest(ObjectSecurityLayer.java:197)
	at org.eclipse.californium.core.network.stack.ReliabilityLayer.sendRequest(ReliabilityLayer.java:107)
	at org.eclipse.californium.core.network.stack.BlockwiseLayer.receiveResponse(BlockwiseLayer.java:665)
	at org.eclipse.californium.core.network.stack.ReliabilityLayer.receiveResponse(ReliabilityLayer.java:339)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveResponse(AbstractLayer.java:89)
	at org.eclipse.californium.oscore.ObjectSecurityLayer.receiveResponse(ObjectSecurityLayer.java:357)
	at org.eclipse.californium.core.network.stack.AbstractLayer.receiveResponse(AbstractLayer.java:89)
	at org.eclipse.californium.core.network.stack.BaseCoapStack.receiveResponse(BaseCoapStack.java:143)
	at org.eclipse.californium.core.network.CoapEndpoint$1.receiveResponse(CoapEndpoint.java:309)
	at org.eclipse.californium.core.network.UdpMatcher$4.run(UdpMatcher.java:435)
	at org.eclipse.californium.elements.util.SerialExecutor$1.run(SerialExecutor.java:276)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:473)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:178)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:292)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1152)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:622)
```
I currently didn't really understand what happens. (@rikard-sics maybe you can give me some hint about that?)

boaks:
Maybe, if the processing of the 4.13 is configurable, I guess the test would be passed.
Would a flag in the request be possible?

rikard-sics:
> I currently don't really understand what happens. (@rikard-sics maybe you can give me some hint about that?)

Yes, let me look into this and the code for this PR to understand what is happening.

rikard-sics:
> Maybe, if the processing of the 4.13 is configurable, I guess the test would be passed.
> Would a flag in the request be possible?

Yes, possibly if there was a flag OSCORE could choose to not use this new functionality when its doing outer block-wise transfers.

sbernard31:
OK I think I get it. 

Here : https://github.com/eclipse/californium/pull/1357/files#diff-20c9d7fed566d6ad9272e84175d7b9c2R646

I want to ensure if my request was not a block1 request.
- If this is a block1 request, I stop all and deliver the error response to upper layer.
- if this is not a block1, I start a new blockwise transfer.

To do that I check if : !exchange.getCurrentRequest().getOptions().hasBlock1()
But in this case, the request was turned into oscore one and block1 option is not here.
`CON-PUT    MID=   -1, Token=4C258B4A23AC724C, OptionSet={"Object-Security":0x, "Content-Format":"text/plain", "Accept":"text/plain", "Proxy-Uri":"coap://127.0.0.1:48672/resource"}, "982573870057647435266673".. 128 bytes`
**==> is it the expected behavior ?**

I can eventually change my check by something like : `getBlock1Status(getKey(exchange, exchange.getCurrentRequest())) == null` (It seems this works)

sbernard31:
> Yes, possibly if there was a flag OSCORE could choose to not use this new functionality when its doing outer block-wise transfers.

But why OSCORE should be incompatible with this feature ?

rikard-sics:
> OK I think I get it.
> 
> Here : https://github.com/eclipse/californium/pull/1357/files#diff-20c9d7fed566d6ad9272e84175d7b9c2R646
> 
> I want to ensure that my request was not a block1 request.
> 
> * If this is a block1 request, I stop all and deliver the error response to upper layer.
> * if this is not a block1, I start a new blockwise transfer.
> 
> To do that I check if : !exchange.getCurrentRequest().getOptions().hasBlock1()
> But in this case, the request was turned into oscore one and block1 option is not here.
> `CON-PUT MID= -1, Token=4C258B4A23AC724C, OptionSet={"Object-Security":0x, "Content-Format":"text/plain", "Accept":"text/plain", "Proxy-Uri":"coap://127.0.0.1:48672/resource"}, "982573870057647435266673".. 128 bytes`
> **==> is it the expected behavior ?**
> 
> I can eventually change my check by something like : `getBlock1Status(getKey(exchange, exchange.getCurrentRequest())) == null` (It seems this works)

If you are referring to the test testOuterBlockwiseExceedMaxUnfragmentedSizeProxyServerBW then yes since the outgoing request from the client is not using block-wise, rather the proxy->server request will be using (outer) block-wise.

rikard-sics:
> But why OSCORE should be incompatible with this feature ?

As long as it aborts transfers using (outer) block-wise when they exceed the MAX_UNFRAGMENTED_SIZE it should be fine.

boaks:
My idea was a "enableAutoFailoverOn413" flag. That would make it easier, to switch back, if some new issues came up with this feature. That relaxes the timelines on solving them. With such a flag, it would be OK to first mention that in combination with OSOCRE there are issues left, and the remove that comment with solving them. Even for other future trouble, it may make sense, to be able to disable the auto-failover.

rikard-sics:
Let me expand a bit on what I wrote earlier. First, I think it would be good if OSCORE does supports this functionality (while still being able to enforce MAX_UNFRAGMENTED_SIZE ).

**Issue with the test for outer block-wise**
The issue you had with the test _testOuterBlockwiseExceedMaxUnfragmentedSizeProxyServerBW_ should only occur when block-wise is being performed by a proxy. That is a proxy splits an OSCORE message into blocks meaning that the block-option is an external "outer" option. In such case OSCORE has a restriction defined called MAX_UNFRAGMENTED_SIZE as I mentioned earlier. Any incoming block transfer that is using outer block options should be aborted if the cumulative payload size exceeds this value. For incoming messages this restriction is only applied when outer block-wise is used.

**Inner block-wise seems fine**
However for "inner" block-wise where for instance an OSCORE client sends a block-wise request I don't believe there should be an issue as this limit is not enforced for incoming messages. (Rather outgoing messages are checked to not exceed this in their individual payloads. If they do the application should send them using block-wise instead.) Also note that in the case of inner block-wise without any proxies involved there will be no block-option on the protected request after it has been processed with OSCORE as many options are encrypted and part of the ciphertext.

**Possible solutions**
So as far as I understand this issue is unique for when OSCORE message are being sent via a proxy that applies block-wise before sending it on to the receiver. Thinking more about it having a flag on the request the client application can set may not help since it is the proxy that splits the message using block-wise and would be the one "negotiating" with the server. Although I am not sure about this. The aim would also be to support OSCORE unaware proxies. 

Some possible ideas to solve it. If you believe a flag on the request could suppress this negotiation I can make sure that this flag is set to disable this functionality in the ObjectSecurityLayer or ObjectSecurityContext layer when an OSCORE client is communicating via a proxy (since then outer block-wise may be used by the proxy). Another idea I had is to change the response code from 4.13 to something else like bad request in the case that a transfer using OSCORE outer block-wise is aborted by the server due to the payload size exceeding MAX_UNFRAGMENTED_SIZE.

sbernard31:
Regardless the "OSCORE issue" I think it could make sense to add a way to deactivate this feature. 
@boaks , could you be more specific about what you have in mind about a "flag", is it something at NetworkConfig level ?

sbernard31:
About this test failure, I will try to explain a bit more my understanding : 

Without this new feature the behavior should be : 
```
client -> proxy : POST 
proxy  -> server: POST with outer block
server -> proxy : ENTITY TOO LARGE
proxy  -> client: ENTITY TOO LARGE
```

With this feature this should be something like this : 
```
client -> proxy : POST 
proxy  -> server: POST with outer block
server -> proxy : ENTITY TOO LARGE
proxy  -> client: ENTITY TOO LARGE
client -> proxy : POST with inner block // client try to use block transfer
proxy  -> server: POST with outer block
server -> proxy : ENTITY TOO LARGE 
server -> proxy : ENTITY TOO LARGE // block transfer doesn't works too, so we stop here.
```
(maybe a bit silly, but if client is not aware that proxy use blockwise this could maybe make sense :thinking:)


So why the test failed ?
This is not really about the use case.
As explained in my comments https://github.com/eclipse/californium/pull/1357#issuecomment-660213931, the code doesn't work because the `exchange.currentRequest()` does not contains the block1 option.  
Now I better understand why ? this is because the exchange.currentRequest() is not the transparent block1 request as I expected but the encrypted OSCORE one (with inner block1)

I didn't expect this as this sounds new  to me because before like said the javadoc : 
```java
	/**
	 * The current block of the request that is being processed. This is a
	 * single block in case of a blockwise transfer or the same as
	 * {@link #request} in case of a normal transfer.
	 */
	// Matching needs to know for what we expect a response
	private volatile Request currentRequest;
```
So If we want to keep it like, we should maybe update the javadoc and I hope there is no other place in the code where we expect to have the single block request when we call `exchange.getCurrentRequest()` because with OSCORE this could generate bugs.

In a perfect world, I suppose we should have a design where OSCORE are transparent for BlockLayer and so encrypted OSCORE request should not be visible by it. (Easier said than done :sweat_smile:)

Anyway, instead of using `currentRequest.hasBlock1()` I will use something like : `getBlock1Status(getKey(exchange, exchange.getCurrentRequest())) == null` which is another way to know if there is an ongoing block1 transfer for this exchange.

rikard-sics:
> So why the test failed ?
> This is not really about the use case.
> As explained in my comments [#1357 (comment)](https://github.com/eclipse/californium/pull/1357#issuecomment-660213931), the code doesn't work because the `exchange.currentRequest()` does not contains the block1 option.
> Now I better understand why ? this is because the exchange.currentRequest() is not the transparent block1 request as I expected but the encrypted OSCORE one (with inner block1)
> 
I am not sure I understand. So for testOuterBlockwiseExceedMaxUnfragmentedSizeProxyServerBW 
 on the client the exchange.currentRequest() does not have the block1 option? But here the client is not sending a block-wise request.

boaks:
@sbernard31 I hadn't something in mind, but I think, NetworkConfig will do it, at least to start with.

sbernard31:
> So for testOuterBlockwiseExceedMaxUnfragmentedSizeProxyServerBW
on the client the exchange.currentRequest() does not have the block1 option? But here the client is not sending a block-wise request. Or you mean for the new block-wise transfer you are attempting to start (I guess so)?

I mean _"the new block-wise transfer I'm attempting to start"_.

> I can certainly try to change things related to OSCORE block-wise to make it more transparent if you have any suggestions.

Unfortunately, I haven't concrete Idea. 
I just see that we have some layers which change/replace requests/responses on the fly. (like BlockLayer or ObjectSecurityLayer) and we didn't really have a generic way to address this.
We had this getRequest()/getCurrentRequest() on `Exchange` to handle this. But from the beginning I'm a bit disturb that transparent blockwise concept impacts the "generic" `Exchange`. (One of my  old dream was to remove this "currentRequest", I never succeed)

Maybe, changing layer method like this : 
```
void receiveResponse(Exchange exchange, Response response);
```
by
```
void receiveResponse(Exchange exchange, Request request, Response response);
```
This way each layer which replaces request from top to bottom layer, could provide (re-replace) the original request when receiving the response (top to bottom).
We probably also need a kind of parent/children request concept to be able to forward request state. (e.g. when you cancel a parent request, child request must be cancelled too and vice&versa) 

But this mean API breaks,  meaning californium 3.x. 
I probably missed a lot of issues, this will probably impact the code very much.
This maybe only works on paper (or rather in my head :sweat_smile:)

sbernard31:
I added the network config flag. I think this is in a reviewable state now.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
boaks: That has now grown to a larger couple o f lines. Please consider to move that into a own method.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -716,10 +764,21 @@ private void handleBlock1Response(final Exchange exchange, final Response respon
...
+								Math.min(block1.getSize(), preferredBlockSize));
+						exchange.setCurrentRequest(blockRequest);
+						lower().sendRequest(exchange, blockRequest);
+					} else {
+						deliverResponseToUpperLayer(exchange, response);
+					}
+				} else {
+					// request has not been sent blockwise
+					LOGGER.debug("discarding unexpected block1 response: {}", response);
+				}
```
boaks, line 781: I would prefer, if the failover already creates the Block1BlockwiseStatus, so that the null case could be kept simple here.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -731,10 +790,16 @@ private void handleBlock1Response(final Exchange exchange, final Response respon
 
 				clearBlock1Status(key, status);
 
+			} else if (enableAutoFailoverOn413 && response.getCode() == ResponseCode.REQUEST_ENTITY_TOO_LARGE) {
+				if (status.getCurrentNum() == 0 && block1.getSize() < status.getCurrentSize()) {
+					sendBlock(exchange, response, key, status, 0, block1.getSzx());
+				}else {
+					clearBlock1Status(key, status);
+					deliverResponseToUpperLayer(exchange,response);
+				}
```
boaks, line 799: I would prefer, if the failover is already handled so that this null case could be kept simple here.

boaks:
I would prefer a clean and slim solution, which just reacts on 4.13 and block1 option in the first response.

The  approach of this PR seems to try out, if it is somehow possible to transfer the request anyway. If the intention is to solve the communication with "buggy/dirty" clients, we will see, which work-around will be required next :-). Anyway, I would prefer to create the new block1status in one place, and not spread over in the code. That would in my opinion keep the required changes in a clearer scope.

sbernard31:
> I would prefer a clean and slim solution

What does it mean  ? :expressionless: 

> The approach of this PR seems to try out, if it is somehow possible to transfer the request anyway.

I do not understand what you mean but it seems that the behavior is not what you expect.
I spent some time to explain the behavior I plan to code here : https://github.com/eclipse/californium/pull/1357#issuecomment-659386050

You answer : _"I don't want to spend my time into blockwise for now. I'm too busy with other stuff."_

Now I implemented what I described and it's not OK. So complicated :tired_face:

sbernard31:
(By the way I only support 4.13 for the first block as you requested)

sbernard31:
> If the intention is to solve the communication with "buggy/dirty" clients

I didn't implement this or at least not deliberately.

boaks:
> What does it mean ?

Sorry, I stated that from the very begin: I prefer a solution, which only applies the failover for the 1. exchange.
Anyway, if you want to stick to this solution, which seems to prepare to apply it in between and also for exchanges without block 1 option, go for it. I just don't care.

BUT: 
Please, don't spread the creations of block1state "all over the code". If that is isolated in one method , then it is much easier to ensure, that the rest is unchanged.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -630,6 +643,37 @@ public void receiveResponse(final Exchange exchange, final Response response) {
 				case REQUEST_ENTITY_INCOMPLETE: // 4.08
 					// we seem to have uploaded blocks not in expected order
 				case REQUEST_ENTITY_TOO_LARGE: // 4.13
+					if (enableAutoFailoverOn413) {
+						if (response.getOptions().hasBlock1()) {
+							// may be a block size negotiation
```
boaks, line 648: That's the case, I would expect to be limited to the 1. exchange.

californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
```diff
@@ -630,6 +643,37 @@ public void receiveResponse(final Exchange exchange, final Response response) {
...
+							handleBlock1Response(exchange, response);
+							return;
+						} else if (!exchange.getRequest().isCanceled()
+								&& getBlock1Status(getKey(exchange, exchange.getCurrentRequest())) == null) {
```
boaks, line 652: That's the casesf, which makes me feel: let's try, how we the request could be transmitted anyway. If you want it, do it.

sbernard31:
> BUT:
Please, don't spread the creations of block1state "all over the code". If that is isolated in one method , then it is much easier to ensure, that the rest is unchanged.

(please avoid UPPERCASE)
I would try to do to this.  Probably means code duplication. (I didn't plan to spend so much time on this feature ...)


For the rest, I still don't understand which use case I currently cover that I should not.

boaks:
Sorry, you make me really frustrated!

I try to give you my opinion, that it's better to isolate the creation of the failover block1 state.Over and Over. But you seem to read all other comments and oversee that. Then I want to put your attention on to it, and you feel ...

> (By the way I only support 4.13 for the first block as you requested)

I can't see that in the code, maybe because I don't understand, where and under which condition you create the failover block 1 states. There are a couple of places, therefore I prefer one place.


```
testLateNegotiationInTheMiddle
```

was also make me belief, that is not only the first exchange.

sbernard31:
> Sorry, you make me really frustrated!

Frustration is present at both side :grin: 

> testLateNegotiationInTheMiddle

This is not related to Entity Too Large.
This just the classic one negotiation. As I said (https://github.com/eclipse/californium/pull/1357#issuecomment-660186563) : 
> I fixed "normal" block1 size negotiation in the middle. (because this is easier than rejected it correctly)
> I added test for block2 size negotiation in the middle.

I think this is a better behavior than answering block number with does makes sense (https://github.com/eclipse/californium/pull/1357#issuecomment-653029652).
And if want to reject it I don't know how to do that ? answer with a RST message ?  It sounds more work to me.

boaks:
I would response with "4.02 BAD OPTION". If you feel, the changes are a good solution for blocksize negotiation in the middle, keep it. As I wrote, I don't want to spent too much time into it. And I even spent much more than I want, just to achieve, that, the failover block1 state is create in one place.

So, do you plan to adapt the code to create the failover block1 state only in one place?

> Probably means code duplication.

Why? I can't see that. There is on place for 4.13, either report the error, or create a block1 state, if required.

sbernard31:
I can not respond with a "4.02 Bad Option" to "2.31 Continue" response.

> So, do you plan to adapt the code to create the failover block1 state only in one place?

Already said I will try.

> Probably means code duplication.

I think about case in `handleBlock1Response` like `if (!status.hasMatchingToken(response))` which I need to handle too.

boaks:
> I can not respond with a "4.02 Bad Option" to "2.31 Continue" response.

Sure, but the client can easily stop such an transfer. Just don't continue to send or request more blocks.

sbernard31:
Last try for me. :tired_face:

I tried to keep the code isolated (as much as possible) in one method.
This feature can be disable if needed.
I accept 4.13 negotiation only on first block.
Behavior is explained at the end of this https://github.com/eclipse/californium/pull/1357#issuecomment-659386050.

First two commit is about middle negotiation for old classic one. I just fix something which looks like a bug for block1.
If you don't want to support this middle negotiation anymore, feel free to adapt this in another PR.

boaks:
> Last try for me.

And you complain about upper case ...

LGTM.

If the core-list provides more details, it may (but not must) be changed again.

sbernard31:
>  And you complain about upper case ...

I just try to ask politely about avoiding unnecessary uppercase.
I don't get the problem with my : "Last try for me". I just express that I will not be able to spend more time on this feature.
If you explain the problem with this sentence maybe I could be better next time.

boaks:
"last try for me" includes "take it or leave it", doesn't it?

sbernard31:
Rather like "Unfortunately I can not spend more time on this".
So to you can integrate it like this 
or If this is still not good enough, you can adapt it now or later or never (if this feature does not deserve times).

Still don't get what is the problem with this. Is it about the form or the content ?

sbernard31:
We get the answer on the core mailing list.

For your 1. question this sounds pretty clear to me.
For your 2. one this is less clear.

My understanding of Carsten's answer is 

1. "Nothing prevent to do change size in the middle of blocksize". We all agree this sounds unusual/strange behavior.
    I agree with Carsten to say : "it's not so hard to support".
    => So I would keep it for classic (not 4.13) size negotiation. For 4.13, I would go for support it to be consistent.

2. Regarding the end of my https://github.com/eclipse/californium/pull/1357#issuecomment-659386050 :
   - the size1 case sounds to be out of spec. => I understand we should remove it.
   - the block option with szx sounds the preferred way. => we should keep it.
   - about the 4.13 with no option => this is not clear to me so I try to get clearer answer.

PR #1416, opened 2020-09-17
Add preselected cipher suites to dtls configuration.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -313,6 +313,21 @@ public final class DtlsConnectorConfig {
 	 */
 	private CipherSuiteSelector cipherSuiteSelector;
 
+	/**
+	 * Preselected cipher suites.
+	 * 
+	 * If no supported cipher suites are provided, consider only this subset of
+	 * {@link CipherSuite} to be automatically selected as supported cipher
+	 * suites depending on other setting (e.g. if settings allow only PSK, only
+	 * PSK compatible cipher suite from this list will be selected).
+	 * 
+	 * Not used, if supported cipher suites are provided.
+	 * 
+	 * @see #supportedCipherSuites
+	 * @since 2.5
+	 */
+	private List<CipherSuite> preselectedCipherSuites;
+
 	/** the supported cipher suites in order of preference */
 	private List<CipherSuite> supportedCipherSuites;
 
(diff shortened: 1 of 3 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -3413,7 +3512,16 @@ private void determineCipherSuitesFromConfig() {
...
-
+			if (config.preselectedCipherSuites != null) {
+				// remove not preselected cipher suites
+				Iterator<CipherSuite> iterator = ciphers.iterator();
+				while (iterator.hasNext()) {
+					CipherSuite cipherSuite = iterator.next();
+					if (!config.preselectedCipherSuites.contains(cipherSuite)) {
+						iterator.remove();
+					}
+				}
```
sbernard31, line 3523: It seems to me that order will be defined by default cipher suite order.
Should we rather use the `preselectedCipherSuites` cipher suite order ?

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -313,6 +314,20 @@
 	 */
 	private CipherSuiteSelector cipherSuiteSelector;
 
+	/**
+	 * Preselected cipher suites.
+	 * 
+	 * If no supported cipher suites are provided, consider only this subset of
+	 * {@link CipherSuite} to be selected as supported cipher suites for this
+	 * peer.
```
sbernard31, line 322:
```suggestion
	 * If no supported cipher suites are provided, consider only this subset of
	 * {@link CipherSuite} to be automatically selected as supported cipher suites depending 
	 * on other setting (e.g. if settings allow only PSK, only PSK compatible cipher suite from this list will be selected). 
```

Maybe this deserve a bit more explanation.
I propose something but feel free to adapt or just keep it like this. :slightly_smiling_face:

sbernard31 (commented):
Tested and it works for me.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
boaks: The order only complies to sub-sections in cipher suites. But the order of theses sections is:

- ECDSA
- ECDHE_PSK
- PSK

Anyway, sure we can use the order of the preselected cipher suites.

I adapt it.

boaks:
I adapted the order and the javadocs.

sbernard31 (approved):
LGTM.

Thx @boaks for this new feature :pray:

PR #1422, opened 2020-09-29
Fix nionat reorderer

boaks:


boaks:
@sbernard31 

I deprecated the `NatUtil` in favor of the new `NioNatUtil`. Starting the `NioNatUtil` some time ago, I also tried to use a common base for both. That's not impossible, but will take more time to do. So, for me I decided, that the new one will be the maintained one. If you find the time, check, if `NioNatUtil` will do it for you as well. If not, I will remove the deprecation, but I can't promise to really maintain `NatUtil` as well.

sbernard31:
I suppose `NioNatUtil` has pretty much the same feature set than `NatUtil`.
We are using `NatUtil` for tests but I guess we will be able to switch to `NioNatUtil` easily. (so probably no need to maintain both)

boaks:
Great! The main difference is, using `java.nio` it's not too easy to implement something as [DirectDatagramSocketImpl](https://github.com/eclipse/californium/blob/master/element-connector/src/test/java/org/eclipse/californium/elements/util/DirectDatagramSocketImpl.java) used for unit-tests. For californium's integration-tests I run it now for 3 weeks and with this fix, the unit-tests are stable as before.

boaks:
If you get aware, that you need to old one, just ping.

PR #1442, opened 2020-11-09
Certificate update

boaks:
Update certificates.
Add support for multiple CA certificates sharing the same DN.


sbernard31:
Not directly linked to this issue but could you explain(or point me to explanation) the truncated stuff about `CertPathUtil.validateCertificatePath()`, I'm not sure to get it :thinking: ?

sbernard31:
(back in time with my boring question on review :sweat_smile:)

boaks:
The PKIX CertPathValidator starts always with the last certificate in the chain. So trusting only intermediate CAs requires to truncate the chain. (It's optional, therefore the parameter) 

In order to make that truncation visible, that truncated path is returned. 
According to your issue #1443, it's also possible to truncate 1 after the CA or in append the CA.
My intention was to indicate the "verified part" of the chain.

boaks:
> (back in time with my boring question on review sweat_smile)

Your welcome, but don't insist in make PR #1441 in smaller commits ;-).

boaks:
There are unit tests, `CertPathUtilTest`. 
See `testServerCertificateTruncatingValidationWithIntermediateTrust`. Change the truncate true into false, the test fails.

sbernard31:
:pray: Thx, I will look this in a deeper way tomorrow

boaks:
I added the proposed change for issue #1443 .

sbernard31:
> Add support for multiple CA certificates sharing the same DN.

I didn't see any test about this and I ask myself if we are sure that PKIX java validation don't support it ? (on my side I didn't check that)

boaks:
> I didn't see any test about this

I added a second CA (caalt) with the same DN but different key-pair to the truststore.

The test `testServerCertificateValidationWithTrust` uses this trust store with the two cf-ca certificates.
You may check the test failing, using the new certificates, but the old `CertPathUtil`.

> I ask myself if we are sure that PKIX java validation don't support it ?

I didn't check that either. But from the experience in that past, the PKIX seems to be too strict.

boaks:
> I ask myself if we are sure that PKIX java validation don't support it ?

I did a short check:
- this till requires to truncate the path for intermediate CAs before passing it to the `CertPathValidator`.
- the scenario to trust the nodes-certificate itself doesn't work ( `testServerCertificateTruncatingValidationWithSelfTrust`).

boaks:
I updated the PR with the recent help of libcoap (-C ca is intended for a "common CA", for both client and server).

boaks:
```
		if (trustedCertificates.length == 0) {
			if (trust == null) {
				// prepare to fail :-)
				trust = trustedCertificates[0];
			}
			trustAnchors.add(new TrustAnchor(trust, null));
		} else {
			for (X509Certificate trusts : trustedCertificates) {
				trustAnchors.add(new TrustAnchor(trusts, null));
			}
		}
```

Replace that snippet in `CertPathUtil` to check, how far that works.

sbernard31:
> But I would like to release 2.5.0 on Thursday. So, if possible, can we focus on issue #1443 and PR #1442 ?

I'm currently on it  but tomorrow is public holiday in France and this takes me more time than expected. :confused:

sbernard31:
It seems there is a behavior changes in `validateCertificatePath` because when I ran Leshan SecurityTest with https://github.com/eclipse/leshan/pull/923 :  
 - with cf-2.4.1 , tests are green :heavy_check_mark: 
 - with this PR,  I get some failure. :red_circle: 

(error) registered_device_with_x509cert_to_server_with_x509cert_selfsigned_certificate_usage_ca_selfsigned_server_cert_given
(failure)registered_device_with_x509cert_to_server_with_x509cert_server_certificate_usage_ca_server_cert_wo_chain_given
(failure)registered_device_with_x509cert_to_server_with_x509cert_rootca_certificate_usage_taa
(error) registered_device_with_x509cert_to_server_with_x509cert_selfsigned_certificate_usage_taa_selfsigned_server_cert_given
(failure) registered_device_with_x509cert_to_server_with_x509cert_server_certificate_usage_taa_server_cert_wo_chain_given

(I don't know if there is a bug before and it is now fixed OR if this is a kind of regression)

For our use case, I think we don't need to truncate (what we do currently in https://github.com/eclipse/leshan/pull/923), if I change the code to not truncate. All tests pass (except registered_device_with_x509cert_to_server_with_x509cert_rootca_certificate_usage_taa_intca_given which I should not pass...) so maybe the regression is in the "truncated" block.

boaks:
> this PR 

This californium PR? Or the leshan PR 923?

sbernard31:
This californium PR.

boaks:
so leshan 923 with this?

boaks:
I test it. It changes the returned chain (now including the trusted ca). maybe the test fails on that change.

sbernard31:
Now in https://github.com/eclipse/leshan/pull/923, if I remove  the usage of `BaseCertificateVerifier.expandCertPath()` : 
```
// - must do PKIX validation with trustStore
CertPath certPath = expandCertPath(messageChain, trustedCertificates);
try {
    CertPathUtil.validateCertificatePath(true/false, certPath, trustedCertificates);
} 
```
and use the new `validateCertificatePathWithIssuer()` instead . 
```
// - must do PKIX validation with trustStore
CertPath certPath;
try {
    certPath = CertPathUtil.validateCertificatePathWithIssuer(true/false, messageChain, trustedCertificates);
} 
```

I get same result than https://github.com/eclipse/californium/pull/1442#issuecomment-724834133 (depending I truncate or not)

boaks:
Sorry,  you mix me up.

Sure the leshan PR assumes a lot, and in some cases even to fail, while californium truncates and succeeds.
That's the story:
- either truncate for intermediate CA trust and succeed
- or fail with java's PKIX.

So, if the failure is wanted, use false, if success is wanted use true.
 
I'm not sure, if that's the case. I currently on the way to check it.

sbernard31:
> Sorry, you mix me up.

I'm sorry about that, I would  have liked to give you more precise information but as my day is almost finished and I can work on this tomorrow, I just give you my current understanding. 

I agree we must know if we want to truncate or not at Leshan side. (I probably confuse you talking about this ...)

Anyway my point is tests passed with 2.4.1 + truncate but it doesn't pass with #1442 which make me think there is maybe a bug.
And second hint without truncate (which should be more restrictive) test pass with 2.4.1 and #1442.

sbernard31:
> Is this possible to run the test from the IDE?

just right click on `SecurityTest` run as JUnit.

sbernard31:
Currently, I guess the problem is around : 
```java
if (truncated || add) {
	if (add) {
		if (!truncated) {
			chain.add(trust);
		}
		verifyCertPath = generateCertPath(chain, last + 1);
	}
	return verifyCertPath;
} else {
	return certPath
```
We add the certificate in case we should not.
and so get this kind of error : 
```
java.lang.IllegalArgumentException: Given certificates do not form a chain, root is not the last!
	at org.eclipse.californium.elements.util.CertPathUtil.generateCertPath(CertPathUtil.java:231)
	at org.eclipse.californium.elements.util.CertPathUtil.validateCertificatePathWithIssuer(CertPathUtil.java:468)
	at org.eclipse.leshan.client.californium.CaConstraintCertificateVerifier.verifyCertificate(CaConstraintCertificateVerifier.java:93)
```

boaks:
You already use a keystore, which is not longer java 7 compatible! I adjusted my IDE to use java 8, and it works.

sbernard31:
> You already use a keystore, which is not longer java 7 compatible! I adjusted my IDE to use java 8, and it works.

This surprise me because our jenkind build againt java7 but I will double check this. (:pray thx)

boaks:
`registered_device_with_x509cert_to_server_with_x509cert_rootca_certificate_usage_taa`

For me this seems to check, if trusting the server's node certificate works. 
With Californium 2.5 that's fixed (2.4.a will fail).

I'm not sure, why that tests expects to fail:

> Expected outcome:
>  - Client denied the connection (not pkix chainable, client trust store not in use, server not sending root CA and TAA is not root CA)
> 

For me this seems to be mixed up in that leshan PR.

It takes a broken, old 2.4.1 Californium implementation, and constructs tests based on the wrong behavior.
I would recommend to go over that tests and decide again, what the outcome should be.

boaks:
https://github.com/eclipse/californium/commit/18a6de67bdc94bb78baa899064a83f34acf27a26

That's the californium fix for that leshan unit test. 

Just make sure, if you really want to fail trusting the server's node certificate itself.
FMPOV, this was a bug in 2.4.1. Fixed with the above commit, not related to this PR.

I will check the other failures ...

element-connector/src/main/java/org/eclipse/californium/elements/util/CertPathUtil.java
```diff
@@ -345,19 +345,20 @@ public static CertPath validateCertificatePath(boolean truncateCertificatePath,
 		} else if (truncateCertificatePath) {
 			mode = "anchor";
 			X509Certificate trust = null;
-			for (int index = 0; index < size; ++index) {
+			for (int index = 1; index < size; ++index) {
```
sbernard31, line 389: is this index changes expected ?

boaks: Yes, the idea before was to search for a CA which signed the cert at index 0.
Now it searches for the CA certificate itself => index +1

boaks:
> This surprises me because our jenkins builds against java7 but I will double check this. (pray thx)

Where are these certificates stored?

And by the way, did you check my comment on leshan's toolchain usage? But I guess, it's late, so shift that to Friday.

boaks:
I found the certificates and they are PKCS12, new, not java 7 compatible!

So, check https://github.com/eclipse/leshan/pull/855#issuecomment-653731302

sbernard31:
> And by the way, did you check my comment on leshan's toolchain usage?

I add this to my todo list.

element-connector/src/main/java/org/eclipse/californium/elements/util/CertPathUtil.java
sbernard31: I'm not sure but what about the case where peer certificate is selfsigned.

sbernard31:
> I would recommend to go over that tests and decide again, what the outcome should be.

Test was not written by me and I didn't double check it. I should probably do that. 

I must to stop for today, back on Thursday... sorry to abandon you. :disappointed:

boaks:
The other leshan unit test failure is really a bug in this PR, which fails for "self-signed". I fix it and add a test for that.

boaks:
> I must stop for today, back on Thursday... sorry to abandon you. disappointed

Enjoy the public holiday.

element-connector/src/main/java/org/eclipse/californium/elements/util/CertPathUtil.java
boaks: That case fails, but not because of that 1.
The self-signed is processed below that in

```java
trust = searchIssuer(root, trustedCertificates);
```

but set 

```java
add = true;
```

then returns a path with the self-singed twice in.

boaks:
```
[ERROR]   SecurityTest.registered_device_with_x509cert_to_server_with_x509cert_rootca_certificate_usage_taa:1176 No registration expected
[ERROR]   SecurityTest.registered_device_with_x509cert_to_server_with_x509cert_server_certificate_usage_ca_server_cert_wo_chain_given:935 No registration expected
[ERROR]   SecurityTest.registered_device_with_x509cert_to_server_with_x509cert_server_certificate_usage_taa_server_cert_wo_chain_given:1380 No registration expected
```

With the latest fixes, only the leshan tests for trusting the node's certificate directly now successfully executes a handshake. FMPOV, that's the expected behavior.

boaks:
@sbernard31 

I cleaned up my work.

Some more unit tests are added and a description of the used certificates for each tests.

With that, the unit tests of the leshan PR passes (except the test with trusting the server-cert itself, which is possible in Californium and I don't see, why that should fail as assumed by that PR.)

I would like to merge this tomorrow morning and the release the 2.5.0.

I know, that you will not be able to review this again that fast.
If some of the behavior is considered to be a bug, a 2.5.1 will solve it :-).

sbernard31:
> I know, that you will not be able to review this again that fast.
> If some of the behavior is considered to be a bug, a 2.5.1 will solve it :-)

There is no urgency for me. So if you prefer to integrate this like it and eventually fix it later, it's fine to me.
If you prefer to not integrate this PR at all and postpone it for next release, it's fine to me too.

element-connector/src/test/java/org/eclipse/californium/elements/util/CertPathUtilTest.java
```diff
@@ -169,117 +206,327 @@ public void testServerCertificateValidationWithoutTrust() throws Exception {
...
+		X509Certificate[] path = new X509Certificate[] { server };
+		X509Certificate[] trusts = new X509Certificate[] { ca, caalt };
+		X509Certificate[] verfied = new X509Certificate[] { server, ca };
+		CertPath certPath = CertPathUtil.generateCertPath(Arrays.asList(path));
+		CertPath verifiedPath = CertPathUtil.validateCertificatePathWithIssuer(false, certPath, trusts);
+		TestCertificatesTools.assertEquals(verfied, verifiedPath.getCertificates());
+
+		X509Certificate[] trusts2 = new X509Certificate[] { caalt, ca };
+		verifiedPath = CertPathUtil.validateCertificatePathWithIssuer(false, certPath, trusts2);
+		TestCertificatesTools.assertEquals(verfied, verifiedPath.getCertificates());
```
sbernard31, line 410: Should it be : 
```java
CertPath verifiedPath = CertPathUtil.validateCertificatePathWithIssuer(true, certPath, trusts); 
```
?

boaks: The test-names are not that "self explaining" .... 

Truncating: the send certificate path is truncated
Truncated: the path may be truncated on validation.

I will add the missing test-case.

boaks: I pushed the additional test case to master.

Issue #1451, opened 2020-11-18
Feedback about  release 2.5.0


sbernard31:
I put on my user cap by integrating Cf-2.5.0 in Leshan, some feedback above : 

### Deprecation PskStore/AdvancedPskStore :
1. frustrating to remove deprecated stuff and get other deprecated warning stuff instead : setResultHandler(~PskSecretResultHandler~ resultHandler)

2. API is probably more flexible but for user who only cares about providing key for a given identity it's really less straightforward. (too many  useless parameters in this case, you need to implement method you don't care about async, you need to ask yourself if you support ecdhe)
   For single psk/key pair, `AdvancedSinglePskStore` helps a lot.

### Deprecation RpkTrustStore/NewAdvancedCertificateVerifier
API is probably more flexible but for user who only cares about just saying if they trust or not a given RPK in a sync way this is  really less straightforward. (you need to handle X509 and RPK at same type, you need to care about CertificateType,  you need to implement method you don't care about async)

For static configuration, the `StaticNewAdvancedCertificateVerifier.Builder` helps a lot but it should maybe be more visible in javadoc where old stuff is deprecated. Currently only `NewAdvancedCertificateVerifier` interface is referenced.

### Deprecation CertificateVerifier&trust/NewAdvancedCertificateVerifier
More or less same remarks as above.

### DtlsConnectorConfig
1. I think there is maybe a backward compatibility bug. Before you was able to set a RpkTruststore + a CertificateVerifier. I suspect that now you can not do that.

2. I think that getTrustStore, getCertificateVerifier & getRpkTrustStore should not be deprecated. Pretty much the same issue than https://github.com/eclipse/californium/issues/1323#issuecomment-641184338 but it seems it can not be fixed in a same way.

### Global Feeling
My current feeling is that API bring probably more possibility at cost of complex API.
For some static usecase there is some API which hide this complexity. :+1: 
For some other classic sync usecase, the API is now clearly less user-friendly than before.

boaks:
> frustrating to remove deprecated stuff and get other deprecated warning stuff instead : setResultHandler(PskSecretResultHandler resultHandler)

Sure, therefore I plan to switch to 3.0. It's simply hard to change an API "again". The `PskSecretResultHandler` will be removed and fully replaced by the generic `HandshakeResultHandler`.
 
The issue of supporting both "in memory trusts stores" and "dynamic verifier callback" got somehow wired up.
[...]

PR #1458, opened 2020-11-24
Fix regression on DtlsConnectorConfig + remove deprecation.

sbernard31:
Aims to fix 2. and 3. from https://github.com/eclipse/californium/issues/1451#issuecomment-730231995

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -924,13 +924,12 @@ public final class DtlsConnectorConfig {
 	 * 
 	 * Only valid, if {@link Builder#setTrustStore(Certificate[])} is used.
 	 * 
+	 * Note: the upcoming next major version 3.0 will remove this method.
+	 * 
 	 * @return the root certificates. If empty (length of zero), all
 	 *         certificates are trusted. If {@code null}, the trust may be
 	 *         implemented by a {@link CertificateVerifier}.
-	 * @deprecated will be replaced by {@link #getAdvancedCertificateVerifier()}
-	 *             in a future version.
 	 */
-	@Deprecated
 	public X509Certificate[] getTrustStore() {
 		if (trustStore == null) {
 			return null;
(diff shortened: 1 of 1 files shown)


boaks:
Just as remark:

Removing

```
config.advancedCertificateVerifier = BridgeCertificateVerifier.builder()
					.setCertificateVerifier(verifier).build();
```

also fails the checks in the unit-tests like in ConnectionHelper:

```
if (incompleteConfig.getAdvancedCertificateVerifier() == null) {
	Builder verifierBuilder = BridgeCertificateVerifier.builder();
```

FMPOV, that isn't a that big problem, though I anyway will focus on the 3.0.
Hopefully that is not required to test larger bugfixes in the 2.x .
 

=>

LGTM

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -2656,10 +2656,11 @@ public Builder setCertificateVerifier(CertificateVerifier verifier) {
 				throw new NullPointerException("CertificateVerifier must not be null");
 			} else if (config.trustStore != null) {
 				throw new IllegalStateException("CertificateVerifier must not be used after trust store is set!");
+			} else if (config.advancedCertificateVerifier != null) {
+				throw new IllegalStateException(
+						"CertificateVerifier must not be used after new certificate verifier is set!");
 			}
 			config.certificateVerifier = verifier;
-			config.advancedCertificateVerifier = BridgeCertificateVerifier.builder()
-					.setCertificateVerifier(verifier).build();
```
sbernard31, line 2662: You are talking about removing this line ?

> also fails the checks in the unit-tests like in ConnectionHelper:

I'm not sure to get the issue. I run mvn clean install and I don't see any unit-tests failing ?
Did I missed something ?

boaks: Yes.
And yes, the unit test are not failing. Otherwise I would not have given LGTM.
And for now, no test is using the trust store nor the rpk trust store.

If some test are accidentally modified or added in the future using the setter for them (even if they are deprecated), then the result may be surprising.

sbernard31: I can add support of  `config.certificateVerifier != null` if you want.
But this means allowing to use `setCertificateVerifier` with `null` which is not allowed currently  like it was not allowed for RPKStore and trustStore. (I see you add it recently when you add new advanced certificate verifier).

boaks: No, it was just a remark, if unexpected, larger bugfixes gets required.

PR #1459, opened 2020-11-24
Add explicit default version for hello-verify-request.

boaks:
Switch back to 1.2.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
@@ -57,6 +57,7 @@ import org.eclipse.californium.scandium.dtls.CertificateMessage;
 import org.eclipse.californium.scandium.dtls.CertificateRequest;
 import org.eclipse.californium.scandium.dtls.CertificateType;
 import org.eclipse.californium.scandium.dtls.ConnectionIdGenerator;
+import org.eclipse.californium.scandium.dtls.HelloVerifyRequest;
 import org.eclipse.californium.scandium.dtls.ProtocolVersion;
 import org.eclipse.californium.scandium.dtls.RecordLayer;
 import org.eclipse.californium.scandium.dtls.SessionCache;
@@ -218,7 +219,7 @@ public final class DtlsConnectorConfig {
 	private Boolean enableMultiHandshakeMessageRecords;
 	/**
 	 * Protocol version to use for sending a hello verify request. Default
-	 * {@link ProtocolVersion#VERSION_DTLS_1_0}.
+	 * {@code null} to reply the clients version.
 	 * 
 	 * @since 2.5
 	 */
@@ -570,16 +571,25 @@ public final class DtlsConnectorConfig {
 	/**
 	 * Get protocol version for hello verify requests to send.
 	 * 
-	 * Before version 2.5.0, the protocol version 1.2 was used by Californium to
-	 * send hello verify requests. According
+	 * Before version 2.5.0, Californium used fixed the protocol version DTLS
+	 * 1.2 to send the HelloVerifyRequest. According
 	 * <a href="https://tools.ietf.org/html/rfc6347#section-4.2.1">RFC 6347,
-	 * 4.2.1. Denial-of-Service Countermeasures</a>, that hello verify request
-	 * should be sent using protocol version 1.0. That's now the default. In
-	 * order to provide backwards compatibility, it's also possible to configure
-	 * to use protocol version 1.2.
+	 * 4.2.1. Denial-of-Service Countermeasures</a>, that HelloVerifyRequest
+	 * SHOULD be sent using protocol version DTLS 1.0. But that found to be
+	 * ambiguous, because it's also requested that "The server MUST use the same
+	 * version number in the HelloVerifyRequest that it would use when sending a
+	 * ServerHello." With that, Californium from 2.6.0 on will, by default,
+	 * reply the version the client sent in the HelloVerifyRequest, and will
+	 * postpone the version negotiation until the client has verified it's
+	 * endpoint ownership. If that client version is below DTLS 1.0, a DTLS 1.0
+	 * will be used. If a different behavior is wanted, you may use the related
+	 * setter to provide a fixed version for the HelloVerifyRequest. In order to
+	 * provide backwards compatibility to version before 2.5.0 , configure to
+	 * use protocol version DTLS 1.2.
 	 * 
-	 * @return protocol version. Default
-	 *         {@link ProtocolVersion#VERSION_DTLS_1_0}.
+	 * @return fixed protocol version, or {@code null}, to reply the clients
+	 *         version. Default is {@code null}.
+	 * @see HelloVerifyRequest
 	 * @since 2.5
 	 */
 	public ProtocolVersion getProtocolVersionForHelloVerifyRequests() {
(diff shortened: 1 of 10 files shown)


boaks:
See [RFC 6347 - Section 4.2.1 - used version in a HelloVerifyRequest](https://mailarchive.ietf.org/arch/msg/tls/rQ3El3ROKTN0rpzhRpJCaKOrUyU/).

boaks:
Just on interest:

[tinydtls](https://github.com/eclipse/tinydtls/issues/46)

[mbedtls](https://github.com/ARMmbed/mbedtls/issues/3914)

boaks:
Just in addition:

wolfssl 4.5 uses version 1.2, at least, if it's only compiled to support dtls 1.2

boaks:
@sbernard31 

After some more analyzing, my personal conclusion/assumption is, that the intention of that definition is mainly to postpone the version-negotiation in order not prevent from wrong (version-)alerts caused by spoofing. See [RFC 5246 e.1](https://tools.ietf.org/html/rfc5246#appendix-E.1") for version alerts.

With that, the comment in mbedtls make sense, just reply the clients version and later do the negotiation baes on E.1 and reply with an alert to an verified endpoint.

I adapted Californium now to reply the client's version as default (configured version {@code null}) and offer also a configurable fixed version.

Ok, more test and verification for resumption behavior is pending, but FMPOV, that's the best way to go.

boaks:
I tried to verify the behavior of mbedtls about that. In combination with libcoap, it always alerts direct on a client hello with 1.0. I adapted the libcoap setup, but that seems to have no effect. 
Maybe it's required to setup interoperability tests direct with mbedtls and wolfssl, but not for 2.6.

sbernard31:
> that the intention of that definition is mainly to postpone the version-negotiation

That's my understanding too.

> With that, the comment in mbedtls make sense, just reply the clients version and later do the negotiation based on E.1 and reply with an alert to an verified endpoint.
I adapted Californium now to reply the client's version as default (configured version null) and offer also a configurable fixed version.

That's totally make sense :+1:  (based on fact that we want to postpone version-negotiation)

But I'm not sure to understand why we need to postpone version-negotiation :thinking:  (note that we don't specially need to understand this, this is more by curiosity)

> in order to prevent from wrong (version-)alerts caused by spoofing

I don't understand this "attack".
Spoofing attack could be an issue at server side it handling the client message is a costly task, right ? 
It seems to me that HELLO_VERIFY_REQUEST is there to stop the spoofing attack.
In case of spoofing attack, client which does not send the first CLIENT_HELLO will probably ignore the HELLO_VERIFY_REQUEST no matter the protocol_version used.
Or I missed something ? 

> I tried to verify the behavior of mbedtls about that. In combination with libcoap, it always alerts direct on a client hello with 1.0. I adapted the libcoap setup, but that seems to have no effect.

It looks like a bug in mbedtls ?  But even with this behavior I can not see the spoofing attack as attacked client will receive an ALERT instead of HELLO_VERIFY_REQUEST and should ignore it too. :thinking:

sbernard31:
(tell me when this is in reviewable state)

boaks:
> I don't understand this "attack".

I'm not sure, if that is an attack.
A client hello with a "wrong" version and a spoofed endpoint may cause an alert send by the server to that "spoofed endpoint". To prevent sending alerts to a none verified endpoint, I adapted the implementation according. I hop, I manage to add some unit tests in the afternoon.

boaks:
Ready for review.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
```diff
@@ -330,7 +330,13 @@ protected void receivedHelloVerifyRequest(HelloVerifyRequest message) throws Han
 	 */
 	protected void receivedServerHello(ServerHello message) throws HandshakeException {
 		// store the negotiated values
+
 		usedProtocol = message.getServerVersion();
+		if (usedProtocol.compareTo(ProtocolVersion.VERSION_DTLS_1_2) != 0) {
+			AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.PROTOCOL_VERSION, session.getPeer());
+			throw new HandshakeException("The client only supports DTLS v1.2", alert);
```
sbernard31, line 337: maybe rather something like : "The client does not support %usedProtocol but only DTLS v1.2" ?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
```diff
@@ -690,17 +690,20 @@ public void startHandshake() throws HandshakeException {
...
 		if (clientVersion.compareTo(ProtocolVersion.VERSION_DTLS_1_2) >= 0) {
 			return ProtocolVersion.VERSION_DTLS_1_2;
 		} else {
-			AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.PROTOCOL_VERSION, session.getPeer());
+			if (clientVersion.compareTo(ProtocolVersion.VERSION_DTLS_1_0) < 0) {
+				clientVersion = ProtocolVersion.VERSION_DTLS_1_0;
+			}
+			AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.PROTOCOL_VERSION, clientVersion,
+					session.getPeer());
 			throw new HandshakeException("The server only supports DTLS v1.2", alert);
```
sbernard31, line 707: same remarks for server

sbernard31 (approved):
LGTM

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -570,16 +571,25 @@ public Boolean useMultiHandshakeMessageRecords() {
...
 	 * <a href="https://tools.ietf.org/html/rfc6347#section-4.2.1">RFC 6347,
-	 * 4.2.1. Denial-of-Service Countermeasures</a>, that hello verify request
-	 * should be sent using protocol version 1.0. That's now the default. In
-	 * order to provide backwards compatibility, it's also possible to configure
-	 * to use protocol version 1.2.
-	 * 
-	 * @return protocol version. Default
-	 *         {@link ProtocolVersion#VERSION_DTLS_1_0}.
+	 * 4.2.1. Denial-of-Service Countermeasures</a>, that HelloVerifyRequest
+	 * SHOUL be sent using protocol version DTLS 1.0. But that found to be
```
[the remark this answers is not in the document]
boaks: Done.

scandium-core/src/main/java/org/eclipse/californium/scandium/config/DtlsConnectorConfig.java
```diff
@@ -1698,16 +1708,26 @@ public Builder setEnableMultiHandshakeMessageRecords(boolean enable) {
...
+		 * DTLS 1.2 to send the HelloVerifyRequest. According
 		 * <a href="https://tools.ietf.org/html/rfc6347#section-4.2.1">RFC 6347,
-		 * 4.2.1. Denial-of-Service Countermeasures</a>, that hello verify
-		 * request should be sent using protocol version 1.0. That's now the
-		 * default. In order to provide backwards compatibility, it's also
-		 * possible to configure to use protocol version 1.2.
-		 * 
-		 * @param protocolVersion protocol version to send hello verify requests
+		 * 4.2.1. Denial-of-Service Countermeasures</a>, that HelloVerifyRequest
+		 * SHOUL be sent using protocol version DTLS 1.0. But that found to be
```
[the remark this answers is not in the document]
boaks: Done.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ClientHandshaker.java
boaks: Done.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/ServerHandshaker.java
boaks: Done.

boaks:
Hope that works for the most modems.

Issue #1345, opened 2020-06-17
Evolving the SessionCache


boaks:
Though some issues around the `SessionCache`have been raised, it may be a good idea to start a discussion about this `SessionCache` and the intended usage.

Some related issues of the past as starting point:
#948 
#1163 
#1343 

From #1343:
The SessionCache seems for me to be not completely "specified". I don't know, if it is assumed to be "consistent" or "eventually consistent". The "removes" in case of a missing session id in the cache seems to point to the first, but intended "remote implementations" would then slow down too much. So, FMPOV, there is even specification work left, which then may change the implementations. That will be "time consuming".

I currently don't use this SessionCache and currently I prefer to improve the connection id usage in a cluster. Therefore my plan to invest time is more to make the "connection id cluster" working, than a clustered session resumption. With that it depends on others (including you :-) ), how much time you want to spend into the SessionCache.

sbernard31:
> I don't know, if it is assumed to be "consistent" or "eventually consistent".

Could you explain what you mean by "consistent" or "eventually consistent" ?

> Therefore my plan to invest time is more to make the "connection id cluster" working

Could you describe what doesn't work for now ?
And what would mean exactly  "connection id cluster" working ?
Are you talking about a share/persistent connection store ?

boaks:
> Could you explain what you mean by "consistent" or "eventually consistent" ?

["eventually consistent"](https://en.wikipedia.org/wiki/Eventual_consistency)

Generally I didn't invest too much time in understanding the `SessionCache`. One of the outcomes of this discussion may be, that it gets clear, which "consistent" model is used and how the connection store then uses the session cache (e.g. no remove in find).  

> Could you describe what doesn't work for now ?

There is simply no adoption of the [DTLS-1.2-connection-ID-based-load-balancer](https://github.com/eclipse/californium/wiki/DTLS-1.2-connection-ID-based-load-balancer). With that, I would like to experiment to forward the messages to the right node within scandium. e.g. check the cid, if it's not the own, forward the message to the right scandium node. Add the original source address to the udp-message, so that i java implementation is possible. That will not be that performant as a "external load balancer solution", but hopefully it works good enough to gain more attraction, more user, and so the chance to get the adoption for a real cid-load-balancer. 

> Are you talking about a share/persistent connection store ?

No, in my opinion the "record sequence numbers" will prevent such a implementation.

sbernard31:
>  `forward the message to the right scandium node`

[...]

PR #1460, opened 2020-11-25
Add parameter to keep session when connection is terminated.

boaks:

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
@@ -1057,15 +1057,41 @@ public class DTLSConnector implements Connector, RecordLayer {
 	 * principals of all connections.
 	 * 
 	 * Note: if {@link SessionCache} is used, it's not possible to remove a
+	 * cache entry, if no related connection is in the connection store. All
+	 * available connections will be removed from that session cache as well.
+	 * 
+	 * @param principalHandler handler to be called within the serial execution
+	 *            of the related connection. If {@code true} is returned, the
+	 *            related connection is terminated and the session is removed
+	 *            from the session cache.
+	 * @return future to cancel or wait for completion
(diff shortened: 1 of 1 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1052,6 +1052,27 @@ public boolean accept(Principal connectionPrincipal) {
...
+	 * @param principalHandler handler to be called within the serial execution
+	 *            of the related connection. If {@code true} is returned, the
+	 *            related connection is terminated and the session is removed
+	 *            from the session cache.
+	 * @return future to cancel or wait for completion
+	 * @see #startTerminateConnectionsForPrincipal(org.eclipse.californium.elements.util.LeastRecentlyUsedCache.Predicate,
+	 *      boolean)
+	 */
+	public Future<Void> startTerminateConnectionsForPrincipal(
+			LeastRecentlyUsedCache.Predicate<Principal> principalHandler) {
```
sbernard31, line 1072: Just some thought about this.
In our case we have a `SessionCache` (RedisSessionCache).
And for some reason we want to remove connection+session for a given DTLS Identity. (e.g. compromised credentials)

As you said in you comment : 
> Note: if **SessionCache** is used, it's not possible to remove a
 cache entry, if no related connection is in the connection store. All
 available connections will be removed from that session cache as well.

My current understanding is in common use case, if you are using SessionCache you will face the situation where connection is not in connection store. Meaning that if you want to remove session you should do it separately.

Does it means that maybe the right default value for this method should be `removeFromCache = false` ?
(Of course we should not change it for 2.x but maybe for 3.x ?)

sbernard31 (approved):
LGTM.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
boaks: Hm, for me these are details for #1345

Issue #315, opened 2017-05-29
Source IP address for response returned by a COAP server created with wildcard IP address


mbaril:
In section 5.3.2 (Request/Response Matching Rules) of ~~RFC7259~~ RFC7252 it is mentioned:
"The source endpoint of the response MUST be the same as the destination endpoint of the original request."

However, from what I have experimented, the responses returned by a CoAP server created using anyLocalAddress do not always set the source IP address of the response with the destination IP address of the request.
(I'm using straight UDP connector without DTLS.)

In the code (org.eclipse.californium.elements.UDPConnector) I see that the same socket is used to send all response so it looks like we let the OS set the source IP address of the IP frames.

boaks:
If your faced issues of such ambiguous source addresses caused by multi interface setups, you may setup several connectors with unique defined interface address. 

`CoapServer.addEndpoint(final Endpoint endpoint)`

I hope this solves your issues.

sophokles73:
I do not know of a way to explicitly set the source IP address and port of a `DatagramPacket` being sent via `DatagramSocket.send()` in Java, maybe you know of one? In any case, IMHO you shouldn't bind to *anyLocalAddress* in a production environment anyway but to an explicit network adapter instead.

mbaril:
@boaks Yes the creation of an endpoint per IP address should solve the problem. However, I was trying to avoid this solution because we have several network interfaces and for each interface several ip adresses (ipv4 + ipv6). The creation of an endpoint per address is a waste of resources (threads, memory, ...) because in my case because I just want to listen on every interfaces.

@sophokles73 I think the real problem is to get the destination IP address from the incoming requests received by the server. I'm surprised, but from what I see, it's looks like in Java this information is not available from the DatagramPacket received. I don't know if somebody know how to retrieve that information?

From my point of view, the current implementation is problematic because it cannot guarantee that a CoapServer having an endpoint created using anyLocal address will return valid CoAP responses (response that will be accepted by the CoapClient).

boaks:
I still think, this is causes by an ambiguous network setup!

But anyway, 
for those, in which environment anyLocal works, they may use it.
for those, in which environment anyLocal is ambiguous, they should use multiple connectors.

Even if RFC7252 requires something, which the java API (as far as I know) doesn't offer, we can't program it using java.

boaks:
@mbaril 
[...]

PR #1463, opened 2020-12-01
Add javadoc about IPv6 and multiaddress interfaces.

boaks:
Issue #315


CiaranWoodward (commented):
Looks good, thanks for putting this together! I would suggest a slight change of wording, but otherwise I think this solves the issue as well as is practical to do right now.

Suggestion:
```
 * Note: using IPv6 interfaces with multiple addresses including permanent and
 * temporary (with potentially several different prefixes) currently causes
 * issues on the server side. The outgoing traffic in response to incoming may
 * select a different src address than the incoming dst address. To overcome this,
 * please ensure that the 'any address' is not used on the server side and a separate
 * CoapEndpoint is created for each address to receive incoming traffic.
```

boaks:
Adapted according your proposal.

CiaranWoodward (approved):
Looks good to me, thanks!

boaks:
Thanks for your feedback!

PR #1480, opened 2020-12-16
Reduce synchronization in blockwise layer.

boaks:
Experimental. There are not too much block1 tests. Adding them will take longer.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -521,10 +521,11 @@ public class BlockwiseLayer extends AbstractLayer {
 	private void handleInboundRequestForNextBlock(final Exchange exchange, final Request request,
 			final KeyUri key, final Block2BlockwiseStatus status) {
 
+		Response block;
 		synchronized (status) {
 
 			BlockOption block2 = request.getOptions().getBlock2();
-			Response block = status.getNextResponseBlock(block2);
+			block = status.getNextResponseBlock(block2);
 			if (status.isComplete()) {
 				// clean up blockwise status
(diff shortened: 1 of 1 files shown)


boaks:
@sbernard31 

I'm not sure, if it is possible for you to test that in your setup. If that is possible, we can create a 2.6.x branch and pick this into that.

sbernard31:
I looked at the PR and that sounds OK at first sight.

> I'm not sure, if it is possible for you to test that in your setup

We have not so much tests on our side too. I can try to run it tomorrow but I'm pretty sure there are not able to detect any race condition issue which could be introduce because of not enough lock.

boaks:
I extended the benchmark in cf-extplugtest-client.
We see, if runs stable. But I guess, it will not show the dead-lock you have reported.
Let's see. If these test doesn't show defects, should I create a 2.6.1-M1 in order that you may test, if that works better for you?

boaks:
From my side:
The unit test have run through the night without trouble.
The benchmark with block1 also run without reporting errors.

That doesn't verify, that this resolves the reported issue.

So, I will merge it to the 3.0.

@sbernard31 
If you also want a 2.6.1 or a 2.6.1-M1, let me know.

sbernard31:
> But I guess, it will not show the dead-lock you have reported.

I agree. You need to block one thread for ever on `onResponse` from block1 request to check this.

>That doesn't verify, that this resolves the reported issue.

I agree but reading the code I'm pretty sure, `onresponse` is not out of sync block which was the main issue I guess. My concern about this PR would be more about removing too much `synchronize`.

>  The unit test have run through the night without trouble.

That's good ! A good hint that you don't remove too much `synchronize`.

> I will merge it to the 3.0.

:+1: 

> If you also want a 2.6.1 or a 2.6.1-M1, let me know.

I think this should be good to backport this in a 2.6.x branch, but there is no urgency for release, let's wait a little bit in case we face more important issue and release for more than this optimization.

PR #1486, opened 2020-12-18
Graceful server restart

boaks:

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
@@ -27,10 +27,17 @@
 package org.eclipse.californium.core;
 
 import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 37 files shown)


boaks:
Still WIP.

The keys are NOT encrypted! So, only for experiments!
The binary format is also not finished. If that changes, a graceful restart will fail :-).

A restart affects also other server state:
- deduplication store
- incoming block1 transfers (post/put on client side)
- outgoing block2 transfers (post/get on server side)

These states are not persisted. The deduplication store will be large, and hits may be rare. FMPOV, that may be an extension, when the dtls-states are handed-over and the old connector is kept running for some seconds. Then it may be possible to forward such deduplication requests on demand to the old running connector.

boaks:
Update:
- restores now also the "last usage"
- encrypts now the states

boaks:
The update of "README" in "demo-apps/cf-extplugtest-server" in this PR contains instructions.

PR #1492, opened 2021-01-04
Remove jobs on cancel from ScheduledThreadPoolExecutor.

boaks:

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/util/ExecutorsUtil.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/util/ExecutorsUtil.java
@@ -21,6 +21,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
@@ -32,6 +33,18 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Utility to create executors.
+ * 
+ * According some web sources, using {@link ScheduledFuture#cancel(boolean)} may
+ * cause resource leaks, because the {@link Runnable} is removed delayed.
+ * Especially "long term" scheduled tasks, which are frequently scheduled and
+ * canceled may be affected of that. To mitigate that, since java 7 the
+ * {@link ScheduledThreadPoolExecutor#setRemoveOnCancelPolicy(boolean)} can be
+ * used. The {@link ScheduledThreadPoolExecutor} will be configured using the
+ * value of environment-variable or java-property
+ * {@code "EXECUTER_REMOVE_ON_CANCEL"} or {@link #DEFAULT_REMOVE_ON_CANCEL}.
+ * 
+ * Test with java 11 didn't show such leaks, even if the task are removed
+ * delayed, they are removed much earlier than their scheduled time.
  * 
  * Note: THE INTERNAL/PRIVATE {@code SplitScheduledThreadPoolExecutor} IS A
  * WORKAROUND! IT MAY BE REPLACED IN THE FUTURE. See issue #690.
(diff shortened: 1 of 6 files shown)


boaks:
@sbernard31 

I'm not sure about this one. The web mentions [leak issues with canceled scheduled tasks](https://medium.com/@nilesh_mahajan/java-concurrency-pitfalls-of-scheduledexecutorservice-d7298b66e59f).
But none of my tests with the 5-minute blockwise cleanup tasks showed such a leak.

Do you have any negative experience with leaks caused by canceled tasks?

(It would be also possible to put this on 2.6.1 with a different default (off/false).)

element-connector/src/main/java/org/eclipse/californium/elements/util/ExecutorsUtil.java
```diff
@@ -169,21 +226,42 @@ public static void shutdownExecutorGracefully(long timeMaxToWaitInMs, ExecutorSe
...
 	 * execution.
 	 * 
-	 * Note: IT'S A WORKAROUND! IT MAY BE REPLACED IN THE FUTURE! See issue #690.
+	 * Note: IT'S A WORKAROUND! IT MAY BE REPLACED IN THE FUTURE! See issue
+	 * #690.
 	 */
 	private static class SplitScheduledThreadPoolExecutor extends ScheduledThreadPoolExecutor {
 
-		private static final long QUEUE_SIZE_DIFF = 20000;
+		private final long SCHEDULE_EXECUTOR_LOGGING_QUEUE_SIZE_DIFF_DEFAULT = 10000;
```
sbernard31, line 264: Reading this quiclky, I see this change which is maybe not intended ?

boaks: It's only for logging, and may now be adapted/changed by EXECUTER_LOGGING_QUEUE_SIZE_DIFF.
That logging helps to see, when the jobs are actually removed. In difference to the web, that logging indicated in my tests, that the postpone of removing jobs is small. At least, if there are always new jobs.

sbernard31:
As far as I remembered I didn't face this kind of issue. 

The [javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledThreadPoolExecutor.html) says : 
> When a submitted task is cancelled before it is run, execution is suppressed. By default, such a cancelled task is not 
> automatically removed from the work queue until its delay elapses. While this enables further inspection and monitoring, it may 
> also cause unbounded retention of cancelled tasks. To avoid this, set setRemoveOnCancelPolicy(boolean) to true, which 
> causes tasks to be immediately removed from the work queue at time of cancellation. 

I'm not sure what means : 
> By default, such a cancelled task is not automatically removed from the work queue until its delay elapses. 

Trying to look at the code too but still not clear.

boaks:
If a task is scheduled to be executed "delayed" for 5 minutes, and that task is canceled, it is kept in the queue until that delay. Basically, AFAIK, the task are kept in a "heap" (see heap-sort). If a task is moved to the head of the "heap", then it gets simply removed, when it was canceled before. The idea seems to be, that this is more performant than removing it on cancel. FMPOV, that idea was always wrong ;-).

With that, blockwise cleanups, which are scheduled with a 5 minute delay for each block and mostly canceled, will stick for the 5 minutes. That may get large job-queues. In my tests using openjdk 11 this wasn't the case, at least according the logging.

It's a little uncomfortable, a situation, where such leaks are "documented", but verifying that fails. Therefore I decided to adapt the executor's configuration.

sbernard31:
I get it.

>  In my tests using openjdk 11 this wasn't the case, at least according the logging.

Maybe you should test with java 7 or 8 ?  (if you are able to test this easily)

Without this fix you just consume more memory ? or you can also reject task which should not be rejected (e.g. if the heap is full ?)
If we reject task do we get some logs when this happens ?

boaks:
> Maybe you should test with java 7 or 8 ?

And with oracle and openjdk :-). The point is, it will take too much time to try to cover all jvms.
Therefore I already mentioned, that I will focus on testing with openjdk 11.

FMPOV, it's easier to backport this to 2.6.1, by default disabled, than to try to check it for other jvms.

boaks:
> Without this fix you just consume more memory ? or you can also reject task which should not be rejected (e.g. if the heap is full ?)
> If we reject task do we get some logs when this happens ?

No experience, though my openjdk 11 seems to behave different.
Just to mention, I also implemented a "blockwise cleanup" using the improved `LeastRecentlyUsedCache` and a general low frequent cleanup task instead of the cancel-task-per-request. But that will anyway only apply to version 3.

sbernard31:
>  And with oracle and openjdk :-). The point is, it will take too much time to try to cover all jvms.
Therefore I already mentioned, that I will focus on testing with openjdk 11.

You don't need to justify yourself if you don't do it. :slightly_smiling_face: 
I just proposed this idea as you seems a bit frustrating to not be able to reproduce the behavior.
Knowing that setRemoveOnCancelPolicy was added in java 1.7 there is maybe more chance to reproduce it with this version of java. 

> FMPOV, it's easier to backport this to 2.6.1, by default disabled, than to try to check it for other jvms.

No objection.

boaks:
> bit frustrating

It's not frustrating. But I prefer not to change a release (2.6) without verify, that this is an improvement/bugfix.

PR #1495, opened 2021-01-04
Reorder message observer in blockwise layer.

boaks:


sbernard31:
Should this be backported in 2.x ?
(Do not hesitate to ping me if you think that maybe a PR should be backported  to 2.x)

boaks:
No, I don't plan that. 
On my side it's to improve the logging in the benchmark client using "blockwise" requests.

PR #1528, opened 2021-02-01
Move certificate supported check to local field.

boaks:
Fixes leshan issue 964.

### DIFF
--- a/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/DefaultCipherSuiteSelector.java
+++ b/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/cipher/DefaultCipherSuiteSelector.java
@@ -36,16 +36,21 @@ public class DefaultCipherSuiteSelector implements CipherSuiteSelector {
 
 	protected static final Logger LOGGER = LoggerFactory.getLogger(DefaultCipherSuiteSelector.class);
 
-	/**
-	 * {@code true} to check supporting certificate based cipher suites,
-	 * {@code false}, if the check already failed.
-	 */
-	private boolean checkCertificateSupport = true;
-
 	@Override
 	public boolean select(CipherSuiteParameters parameters) {
+		boolean certificateSupportFailed = false;
 		for (CipherSuite cipherSuite : parameters.getCipherSuites()) {
-			if (select(cipherSuite, parameters)) {
+			if (cipherSuite.requiresServerCertificateMessage()) {
+				if (!certificateSupportFailed) {
+					if (select(cipherSuite, parameters)) {
+						return true;
+					}
+					// if the check for the certificate support
+					// fails, it fails for all other 
+					// certificate based cipher suites as well
+					certificateSupportFailed = true;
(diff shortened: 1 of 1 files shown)


sbernard31:
I was just going to ask if local field should not be used as this object seems to be shared between several Server handshaker. :-)

boaks:
Yes, the local field approach was very, very bad ;-(.

sbernard31:
I tried to understand the code but I feel I missed something ? 

In : 
```java

	@Override
	public boolean select(CipherSuiteParameters parameters) {
		boolean certificateSupportChecked = false;
		for (CipherSuite cipherSuite : parameters.getCipherSuites()) {
			if (cipherSuite.requiresServerCertificateMessage()) {
				if (!certificateSupportChecked) {
					if (select(cipherSuite, parameters)) {
						return true;
					}
					certificateSupportChecked = true;
				}
			} else if (select(cipherSuite, parameters)) {
				return true;
			}
		}
		return false;
	}
```

I see that if we find a cipher suite which requires Server certificate message.
We try to select it. 
If we failed to select it, we never retry a cipher suite which requires server certificate message.
Maybe this is totally correct, but at first sight tat could sounds like a bug because `select` depends on a given cipher suite and failing for a given cipher will make fail for all others.
(I'm not sure I'm clear :grimacing:)


I maybe something like this could be easier to read ? 
```java
	@Override
	public boolean select(CipherSuiteParameters parameters) {
		boolean certificateSupportChecked = false; 
		boolean certificateSupported = false;
		for (CipherSuite cipherSuite : parameters.getCipherSuites()) {
			if (cipherSuite.requiresServerCertificateMessage()) {
				if (!certificateSupportChecked){
					certificateSupported = isCertificateSupported(parameters);
				}
				if (certificateSupported) {
					if (select(cipherSuite, parameters)) {
						return true;
					}
				}
			} else if (select(cipherSuite, parameters)) {
				return true;
			}
		}
		return false;
	}
```

sbernard31:
(I try to check which Californium version was impacted, I since this bug exist since v2.3.0)

boaks:
The cipher suite within `selectForCertificate` is only used to check
`cipherSuite.getCertificateKeyAlgorithm() == CertificateKeyAlgorithm.EC`. That is always true, until someone extends scandium for other algorithms. Therefore the function mainly checks all (extension) parameters, and they don't vary with the cipher suite.

boaks:
Though I work on a "mismatch-details description", I would prefer just to add more comments for the 2.6.1.

boaks:
The other PR (master, 3.0.x) contains also unit-tests. 

`DTLSConnectorHandshakeTest.testX509HandshakeFailingNoCommonSignatureAlgorithms`

demonstrates, that the old version sticks on the `DefaultCipherSuiteSelector.checkCertificateSupport` failure and causes followup handshakes to fail as well. The bugfix in this PR fixes this bug.

boaks:
[Fix verified also from leshan.](https://github.com/eclipse/leshan/issues/964#issuecomment-771573708)

PR #1531, opened 2021-02-02
Add security policy.

boaks:

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,29 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+Currently, [GitHub security advisories](https://help.github.com/en/articles/managing-security-vulnerabilities-in-your-project) is not activated on [eclipse](https://www.eclipse.org/) projects.
+
+To report a vulnerability, your need to open a [bugzilla ticket](https://bugs.eclipse.org/bugs/enter_bug.cgi?product=Community&component=Vulnerability+Reports&keywords=security&groups=Security_Advisories).
+
+For more details, please look at https://www.eclipse.org/security/.
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 3.0.0-SNAPSHOT (master) | :heavy_check_mark: |
+| 2.6.1   | :heavy_check_mark: |
+| 2.5.0   | :question: |
+| 2.4.1   | :question: |
+| 2.3.1   | :question: |
+| 2.2.3   | :question: |
+| 2.1.0   | :question: |
+| 2.0.0   | :question: |
+| before 2.0.0   | :x: |
+
+:heavy_check_mark: development version / current release - all bugfixes will be applied
+
+:question: the previous (bugfix-)releases - update to the current release is recommended. On exceptions, specific bugfixes may be applied on request. (Create a vulnerability report with the requested vulnerability fix and the (bugfix-)version.)
+
+:x: old releases, milestone releases - usually no bugfixes are applied there.


sbernard31:
Are you sure you want to support all those version ? 

I think you should not list the version itself just the "branch" or maybe I missed something

I guess you should unsupported version too  : 

| Version | Supported          |
| ------- | ------------------ |
| 3.0.0 dev (master) | :white_check_mark: |
| 2.6.x   | :white_check_mark: |
| 2.5.x   | :white_check_mark: |
| 2.4.x   | :white_check_mark: |
| 2.3.x   | :white_check_mark: |
| 2.2.x   | :white_check_mark: |
| 2.1.x   | :white_check_mark: |
| 2.0.x   | :white_check_mark: |
| 1.0.x   | :x: |

OR 

| Version | Supported          |
| ------- | ------------------ |
| 3.0.0 dev (master) | :white_check_mark: |
| 2.6.x   | :white_check_mark: |
| 2.5.x   | :white_check_mark: |
| 2.4.x   | :white_check_mark: |
| 2.3.x   | :white_check_mark: |
| 2.2.x   | :white_check_mark: |
| 2.1.x   | :white_check_mark: |
| 2.0.x   | :white_check_mark: |
| < 2.0.0   | :x: |

boaks:
> Are you sure you want to support all those version ?

No, at least not automatically. On request, with good reasons, maybe.
(I currently pushed also an alternative proposal as second commit. The red cross for version before will be added.)

sbernard31:
About alternative policy.

I think your idea about backporting bug fix on demand could be OK for "classic" bug fixes but for security one I'm not sure.

I mean how does the  "ondemand" security support works ? 
E.g : User opens a ticket and asks for  2.3.1 with all security fixes ? or I ask for backporting a specific bugs ? 

I guess with backward compatibility something like this is maybe enough ?

| Version | Supported          |
| ------- | ------------------ |
| 3.0.0 dev (master) | :white_check_mark: |
| 2.6.x   | :white_check_mark: |
| < 2.6   | :x: |
| 1.0.x   | :x: |

boaks:
Please see the last commit.

Beside the current release and the development version, 
for me a bugfix depends on the amount of work and the user's case.

boaks:
> all security fixes

Ok, I add "specific".

boaks:
> I mean how does the "ondemand" security support works ?

If an user wants/needs to stick to a special previous version (may be not able to update in the next time, for any reason), but feels, that a security bugfix will provide benefit, then it depends on the amount of work to do so.

boaks:
[recent proposal](https://github.com/bosch-io/californium/blob/add_sec_policy/SECURITY.md)

boaks:
> I think you should not list the version itself just the "branch" or maybe I missed something

For me a branch is mainly a github artifact required to prepare for a bugfix release. In the past, it was created, when required.

My feeling is more, that this is also about "grants". But that depends a lot on the specific case. As I wrote, on the amount of time doing so (and being able to spend that time), and the situation of the project. I'm not sure. if hard rules can be applied. Especially, when it's my "time" to do so :-).

sbernard31:
As I said I have concern about this idea of security support on demand but I will not insist.

The other remarks about notation (2.6.x vs 2.6.1) I feel the former does not need update on release and so it's easier to maintain. But OK if you prefer to define the version exactly.

boaks:
> But OK if you prefer to define the version exactly.

My intention was to create awareness on the last bugfix release. Sometimes user may get aware,  that using a available bugfix may solve an issue already. We will see ...

FMPOV, sure, if we see, that this doesn't fit, it's easy to adapt it.

sophokles73:
I am not sure if I understood the support policy correctly. My current understanding is that you want to fix bugs in the latest released version only. In addition, you *might* port individual fixes back to older releases on special request. If this is indeed what you want to do then I would find it easier to understand for users if we simply wrote these sentences explicitly instead of having a table which needs to be updated constantly with every new release ...

boaks:
Basically, that's what the explanation of the "signs" says.

I'm not sure, why the update of the table is considered to be an issue.

For me, that's fast done, and it's the easiest approach to be very clear at the versions.

If we get aware, that changing that table is too much work, then it could be simplified without any trouble-

Or do I oversee something? Is it not possible to change that?

sbernard31:
> Or do I oversee something? Is it not possible to change that?

It's just a file in your repository, so you can change it at anytime I suppose. 

> I'm not sure, why the update of the table is considered to be an issue.

I suppose the divergence comes because we do not feel the current presentation clearer than the one which does not need updates. In this case why choosing the one we need more work.

It seems that you consider the current way clearer and so you consider it deserve the small extra work needed to maintain it.

:-)

boaks:
At least, I think it's worth to collect some experience with it.

To be frank, even the issue template is mostly ignored.
I guess, with a couple of request, which don't consider to read that policy, we can simplify it, as much as we want. I would just try it.

PR #1554, opened 2021-03-01
Complete ticket

boaks:


boaks:
PR #1555 obsoletes the session ticket and removes it.

PR #1556, opened 2021-03-03
Ability to define custom root resource and MessageDeliverer implementation

ashvayka:
added new constructor to CoapServer class & updated access to deliverRequest method in ServerMessageDeliverer class. This is useful when you would like to define more complex MessageDeliverer logic.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapServer.java
@@ -132,7 +132,7 @@ public class CoapServer implements ServerInterface {
 	 * assigned, it will bind to CoAP's default port 5683.
 	 */
 	public CoapServer() {
-		this(NetworkConfig.getStandard());
+		this(NetworkConfig.getStandard(), null ,null);
 	}
 
 	/**
@@ -144,7 +144,7 @@ public class CoapServer implements ServerInterface {
 	 *            will bind to CoAP's default port 5683 on {@link #start()}.
 	 */
 	public CoapServer(final int... ports) {
-		this(NetworkConfig.getStandard(), ports);
+		this(NetworkConfig.getStandard(), null, null, ports);
 	}
 
 	/**
@@ -158,6 +158,24 @@ public class CoapServer implements ServerInterface {
 	 *            will bind to CoAP's default port 5683 on {@link #start()}.
 	 */
 	public CoapServer(final NetworkConfig config, final int... ports) {
+		this(config, null, null, ports);
+	}
+
+	/**
+	 * Constructs a server with the specified configuration that listens to the
+	 * specified ports after method {@link #start()} is called.
+	 *
+	 * @param config the configuration, if {@code null} the configuration
+	 *            returned by {@link NetworkConfig#getStandard()} is used.
+	 * @param ports the ports to bind to. If empty or {@code null} and no
+	 *            endpoints are added with {@link #addEndpoint(Endpoint)}, it
+	 *            will bind to CoAP's default port 5683 on {@link #start()}.
+	 * @param rootResource the Resource to set as root.
+	 * 	 		  If {@code null} will use {@link #createRoot()} method to set root.
+	 * @param messageDeliverer the MessageDeliverer to use.
+	 * 	 		  If {@code null} will create new instance of ServerMessageDeliverer class.
+	 */
+	public CoapServer(final NetworkConfig config, final Resource rootResource, final MessageDeliverer messageDeliverer, final int... ports) {
 		// global configuration that is passed down (can be observed for changes)
 		if (config != null) {
 			this.config = config;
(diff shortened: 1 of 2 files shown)


boaks:
Hi @ashvayka ,

thanks for your interest. I would like to split the extension:

1. If a custom message deliverer should be used, isn't the already available 
`CoapServer.setMessageDeliverer(final MessageDeliverer deliverer)` working? 

2. custom root resource. I'm not sure, if the  idea is only to adapt the text. If so, I would prefer to make that text adaptable using something as "setRootContent(String)". Or do you want to adapt more? If you want to adapt more, please obey, that the root-resource should override the `Resource.getEndpoints()`.  

The error above is caused by a missing ECA, please read [CONTRIBUTING](https://github.com/eclipse/californium/blob/master/CONTRIBUTING.md) for instructions.

sbernard31:
About using a custom root resource, this is something we use in Leshan too.
Currently we do it like this : 
```java
// create coap server
CoapServer coapServer = new CoapServer(coapConfig) {
    @Override
    protected Resource createRoot() {
        return new MyRootResource(this);
    }
};
```

boaks:
Hi @ashvayka ,

is it mainly for hiding the root resource? Should "discover" (".well-known") also be disabled?

Hi @sbernard31 ,

should we add a flag to disable the root resource? Additional also the ".well-known"?

sbernard31:
In Leshan this is not for disable the root resource be really to implement a custom one.
Eg. for Leshan client delete or discover on "/" is possible

ashvayka:
Hi @boaks, I believe the solution provided by @sbernard31 is sufficient for us. We just thought that setting the root resource explicitly in a constructor is a more obvious way for most of the users. What we trying to do is a bit specific. Some of the devices  wrongly construct a COAP URI Path object. The URI e.g. /api/v1/rtddt6576g/telemetry is just one object and COAP server on ThingsBoard expects it to be separate URI Path objects eg /api, /v1, / rtddt6576g, /telemetry. So, we want to create a custom message deliverer that will be "fixing" this. I can close this PR but I would really suggest considering adding at least the ability to set root Resource in the constructor. If you like this, we will submit another PR.

boaks:
> but I would really suggest considering adding at least the ability to set root Resource in the constructor. If you like this, we will submit another PR.

Generally, yes.
One open question would be, how that "Resource.getEndpoints()" stuff will work.
Alternatively, adding the leshan example to the javadoc may also help others.

boaks:
> Some of the devices wrongly construct a COAP URI Path object. The URI e.g. /api/v1/rtddt6576g/telemetry is just one object and COAP server on ThingsBoard expects it to be separate URI Path objects eg /api, /v1, / rtddt6576g, /telemetry.

I'm little mixed up.

What doesn't work is 

```java
CoapServer.add(new CoapResource("api/v1/");
```

What works is 

```java
CoapResource v1 = new CoapResource("v1");
CoapResource api = new CoapResource("api");
api.add(v1);
CoapServer.add(api);
```

ashvayka:
the device firmware has a bug and we can't fix it because we don't control the firmware. According to the [RFC section 5.10.1](https://tools.ietf.org/html/rfc7252#section-5.10.1), the URI Path option should hold only one segment (one segment =
"api", one segment = "v1") but the device sends "api/v1" as one URI Path option.

boaks:
> but the device sends "api/v1"

Ah, OK. Yes, then a custom deliverer helps.

boaks:
I checked the usage of that `CoapResource.getEndpoints()`. AFAIK, it's used for "createClient", which seems to be only used by the `AnnounceResource` in cf-benchmark-observer. There it is easy to get the receiving endpoint from the Exchange. Using an other server endpoint may fail, because that changes the source-address of outgoing messages and may get blocked by NATs.

With that, I think removing `CoapResource.getEndpoints()` is the right way.
And adapt the createClient to use an additional CoapExchange/Endpoint to determine the outgoing endpoint.

So, your PR with a explicit `RootResource` will be welcome.

sbernard31:
@ashvayka, about deliver, if I understand you correctly a possible workaround could be to override : `org.eclipse.californium.core.server.resources.Resource.getChild(String)`

boaks:
See PR #1559 removes the `Resource.getEndpoints()`.
A PR with the root-resource could be based on that.

ShvaykaD:
Hi @boaks  and @sbernard31, after additional investigation, we have noticed that in order to achieve our goal, we can simply create our own message deliverer that will extend `org.eclipse.californium.core.server.ServerMessageDeliverer` class, and override next method:

```
	protected Resource findResource(Exchange exchange) {
		return findResource(exchange.getRequest().getOptions().getUriPath());
	} 
```
with our own implementation that updates UriPath of Exchange before actually calling org.eclipse.californium.core.server.resources.Resource.getChild(String) method.


@sbernard31, idea with overriding method: org.eclipse.californium.core.server.resources.Resource.getChild(String) may work also, but it will basically only fixed the result of org.eclipse.californium.core.server.ServerMessageDeliverer.findResource(final List<String> list) method, and we will still need to do additional logic for parsing URI paths in the resources implementation.

So, probably for the current stage, you can close this pull request.

boaks:
> with our own implementation that updates UriPath of Exchange before actually calling 

That make sense.

Issue #1583, opened 2021-03-29
R8 rules for Californium


Thomas-Vos:
# Please provide basic issue information
* org.eclipse.californium:californium-core:2.6.2
* org.eclipse.californium:scandium:2.6.2
* R8 version 2.1.75 (on Android)

These are the R8 rules I am using for Californium 2.6.2 to get it working on Android with R8 enabled:
```
-keep class org.eclipse.californium.elements.util.StringUtil
-keep class org.slf4j.MDC { *; }
```

# -keep class org.eclipse.californium.elements.util.StringUtil
If this rule is not added you get the following exception. I deobfuscated the stack trace so that could be why it looks a bit strange. I also removed references to my own code.
```
java.lang.ExceptionInInitializerError
    at org.eclipse.californium.elements.util.StringUtil.normalizeLoggingTag(Unknown Source:0)
    at ***.org.eclipse.californium.scandium.config.DtlsConnectorConfig$Builder.build(SourceFile:28)
           setEndpoint
    at ***.<init>(SourceFile:14)
           <init>
    at ***(SourceFile:16)
    at ***.invokeSuspend(SourceFile:7)
    at ***.invoke(SourceFile:2)
           invoke
    at kotlin.time.DurationKt.kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(SourceFile:2)
    at kotlin.time.DurationKt.kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(SourceFile:9)
                              kotlinx.coroutines.BuildersKt.withContext
    at ***(SourceFile:1)
    at ***$suspendImpl(SourceFile:35)
    at ***(SourceFile:22)
       invokeSuspend
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(SourceFile:2)
    at kotlinx.coroutines.DispatchedTask.run(SourceFile:20)
    at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(SourceFile:1)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(SourceFile:11)
                                                               runWorker
                                                               run
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.lang.Package.getImplementationVersion()' on a null object reference
    at org.eclipse.californium.elements.util.StringUtil.<clinit>(SourceFile:6)
[...]

PR #1585, opened 2021-03-29
Update the Android sample and enable R8

Thomas-Vos:
Resolves https://github.com/eclipse/californium/issues/1583

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,7 @@ local.properties
 # Android #
 ###########
 .gradle
-gradle*.properties
+gradle/
 gradlew*
 build/
 
--- a/demo-apps/cf-android/app/build.gradle
+++ b/demo-apps/cf-android/app/build.gradle
@@ -42,7 +42,6 @@ android {
 
 dependencies {
     implementation fileTree(include: ['*.jar'], dir: 'libs')
-    implementation 'com.android.support:appcompat-v7:22.2.1'
     implementation 'org.eclipse.californium:californium-core:3.0.0-SNAPSHOT'
     implementation 'org.eclipse.californium:element-connector:3.0.0-SNAPSHOT'
     implementation 'org.eclipse.californium:scandium:3.0.0-SNAPSHOT'
--- a/demo-apps/cf-android/app/build.gradle
+++ b/demo-apps/cf-android/app/build.gradle
@@ -22,8 +22,8 @@ android {
     }
     buildTypes {
         release {
-            minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            minifyEnabled true
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
 
--- a/demo-apps/cf-android/app/build.gradle
+++ b/demo-apps/cf-android/app/build.gradle
@@ -1,4 +1,6 @@
-apply plugin: 'com.android.application'
+plugins {
+    id 'com.android.application'
+}
 
 repositories {
     mavenLocal()
@@ -8,12 +10,13 @@ repositories {
 }
 
 android {
-    compileSdkVersion 26
+    compileSdkVersion 30
+    buildToolsVersion "30.0.3"
 
     defaultConfig {
         applicationId "org.eclipse.californium.examples.android"
         minSdkVersion 19
-        targetSdkVersion 26
+        targetSdkVersion 30
         versionCode 1
         versionName "1.0"
     }
--- a/demo-apps/cf-android/app/proguard-rules.pro
+++ b/demo-apps/cf-android/app/proguard-rules.pro
@@ -1,26 +1,11 @@
-# Add project specific ProGuard rules here.
-# By default, the flags in this file are appended to flags specified
-# in C:\Tools\Android\sdk/tools/proguard/proguard-android.txt
-# You can edit the include path and order by changing the proguardFiles
-# directive in build.gradle.
-#
-# For more details, see
-#   http://developer.android.com/guide/developing/tools/proguard.html
+# For more details, see http://developer.android.com/guide/developing/tools/proguard.html
 
-# Add any project specific keep options here:
+# Californium should not need any R8 rules for it to work properly.
 
-# If your project uses WebView with JS, uncomment the following
-# and specify the fully qualified class name to the JavaScript interface
-# class:
-#-keepclassmembers class fqcn.of.javascript.interface.for.webview {
-#   public *;
-#}
--keep class org.eclipse.** { *; }
--keep interface org.eclipse.** { *; }
--keep enum org.eclipse.** { *; }
--keep class io.netty.** { *; }
--keep interface io.netty.** { *; }
--keep enum io.netty.** { *; }
--dontwarn org.eclipse.**
--dontwarn org.slf4j.**
--dontwarn io.netty.**
+
+# The rules below are only added here for testing the sample app, you do not need to add them in your project.
+# Keep line numbers for debugging
+-keepattributes SourceFile,LineNumberTable
+-renamesourcefileattribute SourceFile
+# Move all classes to the root package to see what happens with Californium
+-repackageclasses ''
--- a/demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
+++ b/demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
@@ -16,9 +16,9 @@
  ******************************************************************************/
 package org.eclipse.californium.examples;
 
+import android.app.Activity;
 import android.content.Intent;
 import android.os.AsyncTask;
-import android.support.v7.app.AppCompatActivity;
 import android.os.Bundle;
 import android.util.Log;
 import android.view.Menu;
@@ -39,7 +39,7 @@ import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
 
 import java.util.concurrent.TimeUnit;
 
-public class MainActivity extends AppCompatActivity {
+public class MainActivity extends Activity {
 
     public static final String CLIENT_NAME = "client";
 
--- a/demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
+++ b/demo-apps/cf-android/app/src/main/java/org/eclipse/californium/examples/MainActivity.java
@@ -18,8 +18,9 @@ package org.eclipse.californium.examples;
 
 import android.app.Activity;
 import android.content.Intent;
-import android.os.AsyncTask;
 import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
 import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
@@ -29,14 +30,14 @@ import android.widget.TextView;
 
 import org.eclipse.californium.core.CoapClient;
 import org.eclipse.californium.core.CoapResponse;
-import org.eclipse.californium.core.coap.CoAP;
-import org.eclipse.californium.core.coap.Request;
 import org.eclipse.californium.core.network.CoapEndpoint;
 import org.eclipse.californium.core.network.EndpointManager;
 import org.eclipse.californium.elements.UDPConnector;
 import org.eclipse.californium.scandium.DTLSConnector;
 import org.eclipse.californium.scandium.config.DtlsConnectorConfig;
 
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 public class MainActivity extends Activity {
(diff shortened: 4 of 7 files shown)


boaks:
Just to avoid some irritation:

The current release is 2.6.2 (from last week), and a 2.6.3 will come some time in the future. That work is on branch 2.6.x. 

The current development is 3.0.0 (M1 pending, I hope within the next 3 weeks). That work is branch master.

Though the last release review is more than a year ago, we decided, to only release bugfix versions (no API change) for the 2.6 and do that review with the next major release 3.0. At least, until a very strong reason for a 2.7 changes that.

With that, I would prefer to have that work based on 3.0.0-M1 and master.
The issues in the californium code itself, if that doesn't change the API, is usually also back-ported and so also available for the 2.6.3.

boaks:
Ah I see, using 3.0.0-SNAPSHOT requires some changes ahead.

boaks:
Uups, I see the legal misery. Even the old cf-android seems to use "com.android.support:appcompat-v7:22.2.1", the PR now "androidx.appcompat:appcompat:1.2.0". Both would require a CQ (including CQs for all indirect dependencies). I tried to convince Eclipse, that android support libraries are mainly "part of the android OS" and therefore maybe "exempt pre-req", but in vain.

I will try to find some time to ask that again. But to be frank, I don't think, that has changed. So my forecast of the outcome will be, that we even have to remove the simple cf-android ...

boaks:
See PR #1586 for change required to update to 3.0.

(The [certificate failure reported](https://github.com/eclipse/californium/issues/1583#issuecomment-809646490) was even a failure in 2.6, I fixed it for 3.0.)

Anyway, I'm really not sure, what the outcome of the legal stuff will be, including removing the android demo at all.

Thomas-Vos:
What is the min Android API level supposed to be for Californium? In [this commit](https://github.com/eclipse/californium/commit/275dd093ea0b0826ccec2766184aa42079fc4579) I saw you are talking about level 20, but that does not make sense. API level 20 is Android 4.4W (for Android Wear). Maybe it would make more sense to change this to API level 21 (Android 5.0)? Or maybe 19 (Android 4.4), but that is not needed for me.

The Android support libraries (now called AndroidX) are required for backwards compatibility. Furthermore, it is used for consistency across all supported Android API levels for UI themes and views. However, as this sample is just a single screen with a very simple UI and does not really use a lot of AndroidX code, we can probably remove the dependency. I think it would be fine for a demo app like this. The theming may look different on other Android devices, though. Let me know if you want me to do this.

I will rebase this PR on https://github.com/eclipse/californium/pull/1586.

boaks:
Then I would go for 19 (that's my old tablet).

> we can probably remove the dependency.

That would be really great! 

The point using this is much more "boilerplate". Not too many eclipse projects do care about android at all. Therefore it's hard to convince, that android support libraries should be exempt from the eclipse CQ process.

boaks:
Using this PR and the Californium source changes + update to 3.0.0-SNAPSHOT (PR #1586) "works on my machine".

The patch for the package in `StringUtil` is already include in master and I included it as well. (This PR is 1 commit behind the master.)

The demo uses  "com.github.tony19:logback-android:1.1.1-11". With that, it's possible to not keep org.slf4j.MDC. Without I get:  

```sh
2021-03-30 12:07:23.252 8439-8496/org.eclipse.californium.examples.android W/System.err: SLF4J: Failed to load class "org.slf4j.impl.StaticMDCBinder".
2021-03-30 12:07:23.252 8439-8496/org.eclipse.californium.examples.android W/System.err: SLF4J: Defaulting to no-operation MDCAdapter implementation.
2021-03-30 12:07:23.252 8439-8496/org.eclipse.californium.examples.android W/System.err: SLF4J: See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.
```

So, is it possible, that your issue with MDC is caused by the implementation you're using?

boaks:
I got aware, that android seems to report "0.0" as dummy version.
Therefore I added an additional commit to remove the dummy and set it to `null` for "not available".

I also update the the latest commits in this PR.
I'm still not able to see/analyze the MDC issue.
I commented out the "-keep class org.slf4j.MDC", but I can't see an error.
Also commenting out "com.github.tony19:logback-android" show only 

```sh
2021-03-30 13:12:13.512 5007-5007/org.eclipse.californium.examples.android W/System.err: SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
2021-03-30 13:12:13.512 5007-5007/org.eclipse.californium.examples.android W/System.err: SLF4J: Defaulting to no-operation (NOP) logger implementation
2021-03-30 13:12:13.512 5007-5007/org.eclipse.californium.examples.android W/System.err: SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.

2021-03-30 13:12:16.841 5007-5052/org.eclipse.californium.examples.android W/System.err: SLF4J: Failed to load class "org.slf4j.impl.StaticMDCBinder".
2021-03-30 13:12:16.841 5007-5052/org.eclipse.californium.examples.android W/System.err: SLF4J: Defaulting to no-operation MDCAdapter implementation.
2021-03-30 13:12:16.842 5007-5052/org.eclipse.californium.examples.android W/System.err: SLF4J: See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.
```

Maybe you can give me some hints.

Thomas-Vos:
I have not added any implementations for slf4j in my project myself as I do not want logging in release builds. I am not that familiar with slf4j but I assume you do not need to anything to prevent Californium from working.

I am able to reproduce the logged exception in the sample app, however Californium still works (which is not the case in my project).

1. Use this PR and apply this commit: https://github.com/Thomas-Vos/californium/commit/3653d761aa8edbb9703674079284269582cb3099
2. In Android Studio change the "build variant" (at the bottom left of the screen) to "release" (so R8 is active). It will use the keystore and signing config in the commit linked above.
3. Run the app (will run a release APK with R8)
4. In the logcat you should see the following. You may need to change the filter to no filter as this is a non-debuggable build.

```
W/System.err: MDC binding unsuccessful.
W/System.err: Reported exception:
W/System.err: java.lang.NullPointerException
W/System.err:     at org.slf4j.MDC.<clinit>(SourceFile:1)
W/System.err:     at org.eclipse.californium.scandium.DTLSConnector$7.doWork(SourceFile:1)
W/System.err:     at org.eclipse.californium.scandium.DTLSConnector$7.run(SourceFile:6)
```

This is the same exception I get in my real app. However, the demo app is still functional! That is not the case in my real project, that just seems to get "stuck" when doing any calls (hard to debug without debugging enabled).

Adding the keep rule prevents the error above. And in my real project it also causes Californium to start working again.

I decompiled the demo app APK to see where this exception is coming from. See the following (without obfuscation but with optimization):
```
package org.slf4j;

import org.slf4j.helpers.NOPMDCAdapter;
import org.slf4j.helpers.Util;

public abstract class MDC {
    public static NOPMDCAdapter mdcAdapter = new NOPMDCAdapter();

    static {
        try {
            throw null;
        } catch (NoClassDefFoundError e) {
            String message = e.getMessage();
            if (message == null || !message.contains("StaticMDCBinder")) {
                throw e;
            }
            Util.report("Failed to load class \"org.slf4j.impl.StaticMDCBinder\".");
            Util.report("Defaulting to no-operation MDCAdapter implementation.");
            Util.report("See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.");
        } catch (Exception e2) {
            Util.report("MDC binding unsuccessful.", e2);
        }
    }

    public static void put(String str, String str2) {
        if (mdcAdapter == null) {
            throw new IllegalStateException("MDCAdapter cannot be null. See also http://www.slf4j.org/codes.html#null_MDCA");
        }
    }
}
```
It looks like R8 thinks that `bwCompatibleGetMDCAdapterFromBinder` always returns null so it inlined it and that explains the `throw null` code and the error log. However, as you can see the mdcAdapter is always set to a value by R8 so the code should just work fine. R8 also seems to have inlined the other functions like `MDC.clear`.

So this cannot be the issue. I decompiled my real project's APK as well and searched for this file, and it was very similar. I see nothing what could be going wrong. Not sure what is happening and why the rule fixes the issue.

Thomas-Vos:
By the way, as you can see from the latest commits, I removed the AppCompat dependency.

Thomas-Vos:
I think there is something with how Californium calls MDC code that R8 does not like. But I think that it is not something you can fix, I would say it looks more like an R8 issue.

Maybe it is best to add this keep rule to the sample app with a short explanation so other people know that this rule may be needed if Californium does not work for them in release builds.

boaks:
> This is the same exception I get in my real app. However, the demo app is still functional!

I guess, if coaps instead of coap is used, that will change accordingly (it fails).

> I think there is something with how Californium calls MDC

My intention is to execute such a call "ahead". If I get that exception, then scandium will stop to use the MDC.

Thomas-Vos:
> I guess, if coaps instead of coap is used, that will change accordingly (it fails).

Yes, you are right. I can reproduce the issue with the local server and doing a DTLS localhost request. Great, now the sample can reproduce the issue.

Thomas-Vos:
The MainActivity class in the Android sample contains deprecated APIs. That is why you get the warning message. The Android `AsyncTask` class was deprecated recently. See: https://developer.android.com/reference/android/os/AsyncTask

`AsyncTask` should still work fine, the message is just a warning. So we could probably leave it for now. If you want to fix it, the documentation says the following:
```
Deprecated
Use the standard java.util.concurrent or Kotlin concurrency utilities instead.
```

I don't think ignoring this message with `-Xlint:deprecation` is a good idea. You probably want to be reminded that AsyncTask is not the way to build new apps.

boaks:
> I don't think ignoring this message with -Xlint:deprecation is a good idea. 

agreed. My understanding of

> Note: Recompile with -Xlint:deprecation for details.

that this is for details, not to suppress. Anyway, where is that configure in android studio or gradle?

Thomas-Vos:
> Anyway, where is that configure in android studio or gradle?

Add the following code in the project level build.gradle inside the allprojects block:
```groovy
tasks.withType(JavaCompile) {
    options.compilerArgs << "-Xlint:deprecation"
}
```
Then do a Gradle sync, maybe need a clean as well, and run it again.

Thomas-Vos:
I think this PR is ready now as the R8 issues have been fixed. 

It contains one commit of https://github.com/eclipse/californium/pull/1586 because that has not been merged to master yet.

Also replaced AsyncTask with Java concurrent as that does not require any external dependencies.

boaks:
The commit with "Add Gradle files" unfortunately adds files to the repo. That requires licenses and potentially CQ (Contribution Questionnaires in order to check the licenses) . In previous version of android studio, these files were added by that. Do you know, if this is still possible?

Thomas-Vos:
Alright, I removed those files. Android Studio should add them automatically but there could be errors if the Gradle version that is automatically added has breaking changes, but that is highly unlikely for this simple project.

boaks:
Maybe I was wrong.

Using AStudio 4.1.3, I get

> * What went wrong:
> An exception occurred applying plugin request [id: 'com.android.application']
> > Failed to apply plugin [id 'com.android.internal.version-check']
>    > Minimum supported Gradle version is 6.5. Current version is 6.1.1. If using the gradle wrapper, try editing the distributionUrl in /home/achim/repos/californium/demo-apps/cf-android/gradle/wrapper/gradle-wrapper.properties to gradle-6.5-all.zip
> 

when I import the project. I already tried to cleanup the project, removing ".gradle" and ".idea", but for some reasons the android studio seems to stick to gradle 6.1.1 instead of 6.5. Very mystic to me.

boaks:
Adding "gradle/wrapper/gradle-wrapper.jar" and "gradle/wrapper/gradle-wrapper.properties" fixes my broken build again. I'm not longer sure, if the wrapper could be added by astudio or not.

Thomas-Vos:
This is exactly the reason why these files should be checked into version control. Everyone does that as that makes sure builds are consistent for all users who use the project. It would be great if that could be done in the future. (In that case, make sure the gradle folder is included (with the jar and properties files), and two gradlew(.bat) files).

What version of Android Studio are you using? I just tried it with Android Studio 4.1.3 and the Gradle wrapper is not generated. I tried the same project in Android Studio 4.2 beta 6 and the wrapper is generated fine. So it seems something is not working in the stable version of Android Studio 4.1.3. I did not run into this issue before because I was using the beta version.


I can squash the commits into one if that is what you think is best.

Thomas-Vos:
Maybe it is better to keep the commits as is. That would allow you to easily revert changes like the removal of the AppCompat dependency in case you would like to add it again in the future. Please let me know what you would like me to do.

boaks:
> I can squash the commits into one if that is what you think is best.

Thanks!

> This is exactly the reason why these files should be checked into version control.

Sure, as developer, I have no doubts. 
But, if it's in the repo, it's required to check the license (AFAIK gradle wrapper is appache 2.0, so not really a issue.)

boaks:
Ok, then go for the multiple commits.

boaks:
I checked the gradle wrapper in the CQ database. The 5.6 is already approved, so we can include it.

Thomas-Vos:
Gradle 5.6 is a bit old and will probably not work with the current versions used in the project. I updated the target versions and Android plugin to the latest stable versions. Can you use a newer Gradle version? For example, 6.8.3, the latest stable version.

boaks:
5.6 

Sorry that was a typo :-). There are many version approved, e.g. 6.5.1

Thomas-Vos:
Ok great, 6.5.1 should work fine. Will you add the files or do you want me to do that?

boaks:
I've tested it and created a PR.

boaks:
@Thomas-Vos 

thank you very much for your patience and contribution!
Hope, that the update doesn't show too many troubles :-).

PR #1587, opened 2021-03-30
Use MDC optional.

boaks:
Supports environments without org.slf4j.MDC implementation.


Thomas-Vos:
Working great for me in the Android demo app with R8!

PR #1595, opened 2021-04-08
OSCORE: Updates the functionality for securely re-generating contexts

rikard-sics:
This pull request updates the method described in Appendix B.2 of the OSCORE RFC for securely re-generating contexts:

Changes done:
1. For messages exchanged during this procedure, the ID Contexts exchanged on the wire (kid_context) are now CBOR byte strings.
2. Fixed issue related to setting the sequence number to 0 in new contexts for the client. This was causing problems during integration into Leshan.


boaks:
@rikard-sics 

should this be also cherry-picked to 2.6.x?

boaks:
LGTM

rikard-sics:
> should this be also cherry-picked to 2.6.x?

Yes I think that would be a good idea. Hopefully it would not be too complicated as the changes are fairly contained, mainly in one file.

> LGTM

Sounds good, thanks.

Issue #1648, opened 2021-06-23
NetworkConfig - Redesign


boaks:
I didn't want to capture the discussion in the related issue #1601 therefore I created a new one.

For PoC I moved the `NetworkConfig` into the element-connector and added extensions for coap, proxy2 and tcp.
An extension is mainly a set of Keys and a `NetworkConfigDefaultHandler` to setup the defaults. To add such a extension, that default handler is added to a list of such handlers in the base `NetworkConfig`.

With that, a application selects the to used modules and and them prior to use the `NetworkConfig`. Except for the cli-tcp-netty module, that works quite well. With the module, the config depends on the availability of the tcp-jar. Assuming that this may be even added later, I think the right way is to add them always, even it the tcp-jar is not available. That move the tcp-config then also to the element-connector.

Currently it's still the `NetworkConfig` and so still weakly typed. I'm not sure, if providing a typed API really pays-off, I guess that basically depends on how often it is used via the setter API.

boaks:
I pushed a very first version to branch "wip_configuration_redesign".

The new central class is now `org.eclipse.californium.elements.config.Configuration`. In difference to the old solution, it consists now of `Configuration.Definition<T>` and Object and not only String keys and values. A `Configuration.Definition` contains a key (to save properties), a documentation (to amend the properties-format), and a converter from text to object representation.

Modules are collections of such `Configuration.Definition<T>` and they registered using a name and their `DefaultsProvider`.

When a `Configuration` is create, all registered `DefaultsProvider` are called, and they setup the `Configuration.Definition<T>` together with the default values. Though `Configuration.Definition<T>` are typed, the  `T Configuration.get() ` and `Configuration.set(T) ` supports typed access using these generics.

```
config.set(CoapConfig.ACK_TIMEOUT, 2, TimeUnit.SECONDS);
config.set(CoapConfig.MAX_RETRANSMIT, 4);
...

int maxRetransmit = config.get(CoapConfig.MAX_RETRANSMIT);
long ackTimeoutMillis = config.get(CoapConfig.ACK_TIMEOUT, TimeUnit.MILLISECONDS);
```

Properties file:

```
# CoAP acknowledge timeout.
COAP.ACK_TIMEOUT=2[s]
# Maximum number of CoAP retransmissions.
COAP.MAX_RETRANSMIT=4
```

boaks:
PR #1653 was merged to master after the 3.0.0-M3 release.

sbernard31:
[...]

Issue #1601, opened 2021-04-10
could Visitor pattern be usefull for configuration in Cf?


rogierc:
The [Mule Coap connector](https://github.com/teslanet-nl/mule-coap-connector) has a set of configuration classes to define the CoAP servers, clients and CoAP endpoints to use in the Mule application. These classes organise all configuration-parameters in a logical structure to give the application developer a good user-experience.(Mule's development environment enables configuration in it's GUI using java-reflection). The Visitor design pattern is used to separate the configuration classes, that are structured for ease of use, from the proces of building Cf entities such as different types of endpoints and connectors. 

Maybe this approach is usefull for Californium also. To explore this idea, I've put the configuration classes in a Cf-fork and removed Mule particulars. See: [config package](https://github.com/rogierc/californium/tree/visitable_config/californium-core/src/main/java/org/eclipse/californium/core/config)

The package contains an usage-example also. It shows how different visitors can be used to construct CoAP servers with different endpoint types. The visitors sub-package contains also the NetworkConfigGetVisitor and NetworkConfigSetVisitor that convert the new configuration-classes to or from the good old NetworkConfig.  

The solution probably isn't complete but maybe enough to assess feasability of the idea.

boaks:
My idea about a redesign of the configuration was to move common parts to element-connector and to extend that by the specific components (e.g.  DTLS or TCP). Potentially also "network-interface" specific.

I'm not sure, which advantage the visitor pattern brings. Though some upstream projects will depend on that, I would try to get feedback from them as well.

@sbernard31 (leshan) 
@sophokles73 (hono)
@kaikreuzer (openhab) 

WDYT? Short feedback will be welcome.

rogierc:
The main advantage of the Visitor pattern is that the structure of the configuration itself is decoupled from . Different components that have their own view on (a part of) the configuration don't have to 'know' its structure. That is delegated to the visitor which extracts the need configuration items. This way the structure of the configuration can be focused on ease of use (from application developer perspective), without imposing this structure on the internal components of californium / scandium. 

Because of the way Mule SDK works, I had to define configuration classes containing Mule annotions. These classes are scanned by Mule using reflection and are not allowed to depend on Californium classes. So, decoupling using the Vistor pattern was needed there. Could very well be that it isn't worth the effort in other usecases...

sophokles73:
[...]

PR #1653, opened 2021-06-30
Redesign configuration.

boaks:
WIP

First results of the configuration redesign.
See issue #1648


boaks:
Yep.

I currently work on split the config values and adapt the Connector constructors to use `Configuration`, sometimes with additional parameters, see the new Tcp- and TlsClientConnector. There is still so much to type ...

sbernard31:
(with last push --force, It compiles now)

sbernard31:
I looked at this.  (Not so deeply)

And I think it is clearly better :+1: 
We have a better separation of concern and discover API with autocomplete would be probably easier (with to `Definition` + typed setter/getter in `Configuration`)

I'm not sure if this will replace existing Builder or if we will still have both config + builder ?
I refer to my previous comment : 
> d1) NetworkConfig is about everything ? udp, dtls, coap, tls, http, tcp but in a same time there is also a DtlsConnectorConfig.Builder or CoapEndpoint.Builder. The responsibility of those classes sounds not so clear. Why some setting are in some builder and other in NetworkConfig? answering to those responsibility/scope questions could maybe help to define the new design.

(source : https://github.com/eclipse/californium/issues/1601#issuecomment-861328306)

Currently the only part I'm not so comfortable with is the static Configuration.MODULE field and static blocks which registers module.

Looking at PR,  I was thinking that you created a kind of generic **Configuration framework**  (not really linked to Californium or CoAP). I mean it could almost be a separated library. My next thought was : "maybe there is already some library which does this kind of thing ? :thinking: " 
Did you look at this or did you consider this ?

boaks:
> I'm not sure if this will replace existing Builder or if we will still have both config + builder ?

The part of simple parameter maybe repaced by the config. the part of interface implementations not. For DTLS I consider to keep the builder but replace the simple settings by the config.

> My next thought was : "maybe there is already some library which does this kind of thing ? thinking "
Did you look at this or did you consider this ?

It's inspired by [netty.io - AttributeMap](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/AttributeMap.java).

I considered to switch to yaml, but withdrawn it. I don't like external dependencies for close to nothing.
Especially not, if they are used for the core and/or with Android. The implementation is now basically 1 class and 1 interface, and several simple data-definition classes. The rest  is the setup for Californium. And the most work is always to change the available code to use it, regardless what you use.

Just a hint, did you had a look on the created properties files. For me that also a really large improvement to have it sorted and documented.

sbernard31:
> Just a hint, did you had a look on the created properties files? For me that is also a really large improvement to have it sorted and documented.

I just looked at this :  very cleaner :sparkles: . (maybe default value could be added with description in comment)

Any thought about : 
> Currently the only part I'm not so comfortable with is the static Configuration.MODULE field and static blocks which registers module.

boaks:
> Currently the only part I'm not so comfortable with is the static Configuration.MODULE field and static blocks which registers module.

Making that more "dynamic" seems to me to be a difficult idea. Especially the use to generate the initial properties file may get irritating.

sbernard31:
> Making that more "dynamic" seems to me to be a difficult idea. Especially the use to generate the initial properties file may get irritating

I understand the concern but the static one as also some drawback, the main one is : 
A user could have some params in his files which should not be there (which he don't really need)

e.g : I have 2 config(2 different file), one for coap over tcp and one for udp in same application, I will have udp param in my tcp confg and tcp params in my udp config.

boaks:
Yes. 
But I'm not sure, if two files will not create more maintenance (shared coap-config) then provide benefits.
In some examples I use more files, but that's more about different services not about select protocols.

sbernard31:
>  But I'm not sure, if two files will not create more maintenance (shared coap-config) then provide benefits.

It's a case which show that maybe putting everything in the same file is not so good ? 

In my experience static design create issues frequently, I try to think if in this particular case we can imagine some realistic use case where this will be an issue. :thinking:

PR #1694, opened 2021-08-10
Log on client fallback to full handshake.

boaks:


sbernard31 (approved):
:heavy_check_mark: LGTM

PR #1749, opened 2021-09-23
Add rsa related cipherSuites.

### LINKED ISSUE #1693: [Feature request] Ability to define custom cipherSuite.
Our whole certificate chain is RSA and it's kind of impossible to push our client to give us a EDDSA certificate. I searched issues of this repo and I got  #517 . I know there's no point for implementing RSA support in this library, but it would be nice if we can offer our own implementation in our project.

Nick-The-Uncharted:
Fix #1693 

There only one minor issue unresolved: where should I put the "RSA" constant?


Nick-The-Uncharted:
Happened to ran into some minor bugs, I will try to fix them in other mr:
1.  org.eclipse.californium.scandium.dtls.cipher.DefaultCipherSuiteSelector#select check `parameters.getCertificateMismatch()` for select cert. Which leading to mismatch caused previous cipherSuite affect later ones.
2. If we do org.eclipse.californium.scandium.config.DtlsConnectorConfig.Builder#setSupportedCipherSuites without PSK ones but set a custom securityStore, we will get " Advanced PSK store set, but no PSK cipher suite!" Exception.

boaks:
Hi @Nick-The-Uncharted ,

I'm back from vacation, and the very first "large thing", I will work on, is this PR. But first I have to do some smaller things :-) .

So only two questions ahead as comments in the changed files.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SignatureAndHashAlgorithm.java
```diff
@@ -321,16 +322,18 @@ public static SignatureAndHashAlgorithm valueOf(String jcaName) {
...
-				return new SignatureAndHashAlgorithm(hashAlgorithm, signatureAlgorithm);
-			}
+			hashAlgorithm = HashAlgorithm.valueOf(hash);
+			signatureAlgorithm = SignatureAlgorithm.valueOf(signature);
+		} else {
+			hashAlgorithm = HashAlgorithm.INTRINSIC;
+			signatureAlgorithm = SignatureAlgorithm.valueOf(jcaName);
 		}
-		return null;
+		return new SignatureAndHashAlgorithm(hashAlgorithm, signatureAlgorithm);
```
boaks, line 336: For me, the new code now assumes, that all signature- and hash-algorithms are at least known.
I don't think, that is a too good idea. The master branch already contains a similar update. Any reasons to work out your own?

Nick-The-Uncharted: Actully I just copied them from master: https://github.com/eclipse/californium/blob/95f7197750c72d21df1f31b65194671058f1de53/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SignatureAndHashAlgorithm.java#L332
and my IDE remind me that the `if (hashAlgorithm != null && signatureAlgorithm != null) ` check is unnecessary so I removed them.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/x509/KeyManagerCertificateProvider.java
```diff
@@ -363,6 +363,12 @@ private boolean matchCurves(List<SupportedGroup> curves, List<X509Certificate> c
 		for (X509Certificate certificate : chain) {
 			PublicKey certPublicKey = certificate.getPublicKey();
 			if (Asn1DerDecoder.isSupported(certPublicKey.getAlgorithm())) {
+			    // for rsa key we only check whether server support certain curves for key exchange.
+				// As curves is already negotiated based on server config so we always return true
```
boaks, line 367: I have my doubts about the statement of that comment. Sure the common curves have been negotiated at that point, but if you consider "mixed certificate chains" (RSA+EC), then a check for the other (EC) Certificates will be required. With that, I would change the `return` logically in `continue`. I have seen some other places, where something as "isEC" may improve to distinguish between EC based stuff (requires curve check), and non EC based stuff. In the past [SignatureAlgorithm.isEcdsaCompatible()](https://github.com/eclipse/californium/blob/master/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SignatureAndHashAlgorithm.java#L233) was useful, but maybe a more general solution is required.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SignatureAndHashAlgorithm.java
boaks: Interesting ... 
I also read the javadoc for enum `valueOf` this morning and found, that it return an `IllegalArgumentException`. I guess, that requires to be addressed and fixed also in the master ;-).

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/x509/KeyManagerCertificateProvider.java
Nick-The-Uncharted: You are right, I should change return to contine. I failed to consider the "mixed certificate chains" situation.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SignatureAndHashAlgorithm.java
Nick-The-Uncharted: Even value of won't throw, `org.eclipse.californium.scandium.config.DtlsConnectorConfig.Builder#setSupportedSignatureAlgorithms(java.lang.String...)` will still throw IllegalArgumentException if that `valueOf` return null, so I think it probably okay?

boaks: The idea of returning `null` instead of throwing an exception was to support other peer's, which sends certificates with "unsupported signatures and algorithms", if the trusted CA is before that. It requires more time to check, what is really needed to achieve that.

Nick-The-Uncharted: But algorithms are decoded in here (org.eclipse.californium.scandium.dtls.SignatureAlgorithmsExtension#fromExtensionDataReader), it's not using `valueOf` method.

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/x509/KeyManagerCertificateProvider.java
Nick-The-Uncharted: I will fix this when I got home. (I submitted this code from my personal computer, uploading code from company computer requires too many paperwork I simply gave up....). 
Should I amend this commit and perform I force push or just add a normal commit?

scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/SignatureAndHashAlgorithm.java
boaks: > It requires more time to check, what is really needed to achieve that.

But not now. Now I have to focus on the little things :-).

boaks: I checked it, it's only used for configuration. 
I adapted the result according the standard `valueOf` functionallity.
See PR #1752

Nick-The-Uncharted: Should I revert the change of `SignatureAndHashAlgorithm.valueOf` so there won't be conflict? 
And should I amend this commit and perform a force push or just add a normal commit?

Nick-The-Uncharted: And the small problem. Where should I put the "RSA" constant, it's a small thing but get my itchy.

boaks: > Where should I put the "RSA" constant

What's about [Asn1DerDecoder.RSA](https://github.com/eclipse/californium/blob/master/element-connector/src/main/java/org/eclipse/californium/elements/util/Asn1DerDecoder.java#L79)?

boaks: > Should I revert the change of SignatureAndHashAlgorithm.valueOf so there won't be conflict?
> And should I amend this commit and perform a force push or just add a normal commit?

The number of changes rather small. Chose what ever you want to apply.

boaks:
Thanks for your work! I merged in the meantime these changes to the master. 
We will see, if [leshan - PR 1073](https://github.com/eclipse/leshan/pull/1073) gets merged.

Issue #1754, opened 2021-09-30
[coap-observe] deleteting an observed resource


sbernard31:
I'm using californium 3.0.0-M4.

My use case : I observe a resource but at some time the resource is deleted.

The RFC7641 says  :
> A notification SHOULD have a 2.05 (Content) or 2.03 (Valid) response
   code.  **However, in the event that the state of a resource changes in
   a way that would cause a normal GET request at that time to return a
   non-2.xx response (for example, when the resource is deleted), the
   server SHOULD notify the client by sending a notification with an
   appropriate response code (such as 4.04 Not Found) and subsequently
   MUST remove the associated entry from the list of observers of the
   resource.**
   
 In my case  when this happens I get this exception below and the 4.04 response is never sent.  : 
```
BaseCoapStack 2021-09-30 17:45:55,874 [WARN] error send response CON-4.04   MID=   -1, Token=BB7B530AF042FDD5, OptionSet={"Content-Format":"text/plain"}, canceled no payload
org.eclipse.californium.core.network.ExchangeCompleteException: Exchange[R48, complete] is already complete! CON-4.04   MID=   -1, Token=BB7B530AF042FDD5, OptionSet={"Content-Format":"text/plain"}, canceled no payload
	at org.eclipse.californium.core.network.Exchange.assertIncomplete(Exchange.java:1269)
	at org.eclipse.californium.core.network.InMemoryMessageExchangeStore.registerWithMessageId(InMemoryMessageExchangeStore.java:239)
	at org.eclipse.californium.core.network.InMemoryMessageExchangeStore.registerOutboundResponse(InMemoryMessageExchangeStore.java:412)
	at org.eclipse.californium.core.network.UdpMatcher.sendResponse(UdpMatcher.java:160)
	at org.eclipse.californium.core.network.CoapEndpoint$OutboxImpl.sendResponse(CoapEndpoint.java:867)
	at org.eclipse.californium.core.network.stack.BaseCoapStack$StackBottomAdapter.sendResponse(BaseCoapStack.java:236)
	at org.eclipse.californium.core.network.stack.ReliabilityLayer.sendResponse(ReliabilityLayer.java:130)
	at org.eclipse.californium.core.network.stack.BlockwiseLayer.sendResponse(BlockwiseLayer.java:697)
	at org.eclipse.californium.core.network.stack.ObserveLayer.sendResponse(ObserveLayer.java:120)
[...]

PR #1758, opened 2021-10-03
Fix observer 2.6

boaks:
Fixes issue #1754 for 2.6


boaks:
Fixes the issue with `CoapResource.delete()`, but not the issues fixed by 1347359 .

Issue #1078, opened 2019-09-25
Unregistered CoAP-Content Types are defined


jimsch:
In going through a different code based that was originally based on Californium, I noted that the following CoAP Content Types were define there but are not in fact registered by IANA.  These values where were dropped during the development of the original CoAP document should be either removed or moved to a new private value space.  (I intend to just remove them.)

There is an effort going on to get a sweep of all of the known media-types registered, however when that is done different values will be assigned to these content formats.

Text/xml = 1
Text/csv = 2
Text/html = 3
Image/Gif = 21
Image/Jpeg = 22
Image/png = 23
Image/tiff = 24
Audio/Raw = 25
Video/Raw = 26
Application/rdf+xml = 43
Application/soap+xml = 44
Application/atom+xml = 45
Application/fastinfoset = 48
Application/soap+fastinfoset = 49
Application/x-obix-binary = 51

boaks:
Thanks!

[IANA - CoAP Content Formats(https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats)

> There is an effort going on to get a sweep of all of the known media-types registered,

Does this ongoing effort affect that list?

jimsch:
As I said, they will all get new values at that point, probably greater than 255.  You can track the new registrations with the draft https://datatracker.ietf.org/doc/draft-bormann-core-proactive-ct/, but as you can see it has not yet become a working group document.

boaks:
OK, I think 1 milestone deprecate and then remove it would be the best FMOPV.

boaks:
Deprecated.

boaks:
Remove in master.

PR #1814, opened 2021-11-05
Add missing CoAP Content-Formats

sophokles73:
Added the Content-Formats that have been registered with IANA but which
haven't been added to Cf yet.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/MediaTypeRegistry.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/MediaTypeRegistry.java
@@ -71,8 +71,8 @@ public class MediaTypeRegistry {
 	public static final int APPLICATION_SENML_EXI = 114;
 	public static final int APPLICATION_SENSML_EXI = 115;
 	public static final int APPLICATION_COAP_GROUP = 256;
-	public static final int APPLICATION_DOTS = 271;
-	public static final int APPLICATION_MISSING_BLOCKS = 272;
+	public static final int APPLICATION_DOTS_CBOR = 271;
+	public static final int APPLICATION_MISSING_BLOCKS_CBOR_SEQ = 272;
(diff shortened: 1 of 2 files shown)


boaks:
2019 we had issue #1078.

So, should I ask the core-mailing list, if the content types have been consolidated before merging it?

boaks:
[core](https://mailarchive.ietf.org/arch/msg/core/fvqzpz5iKpg4V92hzQeaWhACMxc/)

californium-core/src/main/java/org/eclipse/californium/core/coap/MediaTypeRegistry.java
```diff
@@ -40,17 +40,58 @@
...
+	public static final int APPLICATION_COSE_KEY = 101;
+	public static final int APPLICATION_COSE_KEY_SET = 102;
 	public static final int APPLICATION_SENML_JSON = 110;
+	public static final int APPLICATION_SENSML_JSON = 111;
 	public static final int APPLICATION_SENML_CBOR = 112;
+	public static final int APPLICATION_SENSML_CBOR = 113;
+	public static final int APPLICATION_SENML_EXI = 114;
+	public static final int APPLICATION_SENSML_EXI = 115;
+	public static final int APPLICATION_COAP_GROUP = 256;
+	public static final int APPLICATION_DOTS = 271;
```
EskoDijk, line 74: Above should be APPLICATION_DOTS_CBOR ? (See IANA registry)

californium-core/src/main/java/org/eclipse/californium/core/coap/MediaTypeRegistry.java
```diff
@@ -40,17 +40,58 @@
...
+	public static final int APPLICATION_MISSING_BLOCKS = 272;
```
EskoDijk, line 75: Also here add '_CBOR_SEQ' to name?

PR #1853, opened 2021-12-15
OSCORE: Consider port when storing and retrieving contexts

rikard-sics:
The PR enables functionality to also consider the port for storage and retrieval of OSCORE contexts. This will enable distinguishing between contexts for the same/IP hostname but with different ports.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
@@ -32,6 +32,7 @@ import java.util.regex.Pattern;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.eclipse.californium.core.coap.CoAP;
 import org.eclipse.californium.core.coap.CoAP.ResponseCode;
 import org.eclipse.californium.core.coap.Token;
 import org.eclipse.californium.elements.util.Bytes;
@@ -292,8 +293,10 @@ public class HashMapCtxDB implements OSCoreCtxDB {
 	 */
 	private static String normalizeServerUri(String uri) throws OSException {
 		String normalized = null;
+		int port = -1;
 
 		try {
+			port = (new URI(uri)).getPort();
 			normalized = (new URI(uri)).getHost();
 		} catch (URISyntaxException e) {
 			// workaround for openjdk bug JDK-8199396.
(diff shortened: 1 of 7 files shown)


cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
```diff
@@ -292,8 +293,10 @@ public synchronized void updateSeqByToken(Token token, Integer seq) {
 	 */
 	private static String normalizeServerUri(String uri) throws OSException {
 		String normalized = null;
+		int port = -1;
 
 		try {
+			port = (new URI(uri)).getPort();
```
boaks, line 299: `new URI(String)` isn't that cheap. Maybe you adapt the snippet to create the URI once and reuse it?

cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
```diff
@@ -341,7 +344,12 @@ private static String normalizeServerUri(String uri) throws OSException {
 			LOGGER.error("Error finding host of request URI: " + uri + " message: " + e.getMessage());
 		}
 		if (ipv6Addr instanceof Inet6Address) {
-			normalized = ipv6Addr.getHostAddress();
+			normalized = "[" + ipv6Addr.getHostAddress() + "]";
+		}
+
+		// Consider port, if not default
+		if (port != -1 && port != CoAP.DEFAULT_COAP_PORT) {
```
boaks, line 351: Just as idea, if port is initialized above with `CoAP.DEFAULT_COAP_PORT`, then the `-1` check may be left out.

cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
rikard-sics: Yes, makes sense. I will try to do that.

cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
rikard-sics: Yeah, true I can modify things like that and it should be fine.

rikard-sics: Actually, upon testing this it doesn't work since this line:
`port = (new URI(uri)).getPort();`
Will assign -1 to port if the uri doesn't contain a port. So I still need the -1 check in the end.

cf-oscore/src/main/java/org/eclipse/californium/oscore/HashMapCtxDB.java
rikard-sics: Now done in commit 1f50f113af4dca6c51ac99bdb61a149ed71c10bc

boaks:
LGTM

@rikard-sics are you done? Then I will merge it.

rikard-sics:
> LGTM
> 
> @rikard-sics are you done? Then I will merge it.

Thanks! Yes I am done.

Issue #1909, opened 2022-02-10
Minor Release 3.3.0 - Available


boaks:
See [3.3.0](https://projects.eclipse.org/projects/iot.californium/releases/3.3.0).

boaks:
The 3.3.0 is released, but unfortunately not fully available on [Maven Central](https://search.maven.org/search?q=g:org.eclipse.californium%20a:parent%20v:3.2.0). A ticket is created, and I will report it, if it's available. 

On [Eclipse Repository](https://repo.eclipse.org/content/repositories/californium-releases/org/eclipse/californium) the release 3.3.0 is available. The tools and actinium are available on [Eclipse Repository](https://repo.eclipse.org/content/repositories/californium-releases/org/eclipse/californium) only.

boaks:
The maven central issue is still ongoing.

Please Note:
To use the version 3.3.0 requires currently to use the eclipse repos, see [READEME - current builds](https://github.com/eclipse/californium#current-builds). If the artifacts are available on maven central, that will not longer be required.

boaks:
Unfortunately I found a bug in the dtls graceful restart, introduced in commit  bf4c6b43dc867f4225d8c3b014bb6927d6674980

A 3.3.1 is required.

boaks:
See PR #1929

boaks:
The 3.3.0 is now also available on [3.3.0 - Maven Central](https://search.maven.org/search?q=g:org.eclipse.californium%20a:parent%20v:3.3.0).

Issue #1707, opened 2021-08-22
OSCore observe seconde times occur MAC validation failed


KeithLaiKB:
when I try the case about OSCORE of californium observe 
(my case is in https://github.com/KeithLaiKB/java_test_learn_cf_oscore)

step1
start the server program

step2
start the client program in the same local machine

step2.1
cancel the observer relationship

step2.2
shutdown the client (and then the client program finished)

step3
when I start the client program in the same local machine *again*

then the client console would appear

[main] INFO org.eclipse.californium.core.network.config.NetworkConfig - loading properties from file E:\java_learn_californium_bug_workspace\java_test_learn_cf_oscore\oscore_client_workspace1\com.learn.californium.tryoscore.client\Californium.properties
[main] INFO org.eclipse.californium.core.network.RandomTokenGenerator - using tokens of 8 bytes in length
[main] INFO org.eclipse.californium.core.network.RandomTokenGenerator - using tokens of 8 bytes in length
[main] INFO org.eclipse.californium.ban - Started.
[main] INFO org.eclipse.californium.core.network.CoapEndpoint - coap CoapEndpoint uses udp plain
[main] INFO org.eclipse.californium.core.network.stack.BlockwiseLayer - coap BlockwiseLayer uses MAX_MESSAGE_SIZE=1024, PREFERRED_BLOCK_SIZE=512, BLOCKWISE_STATUS_LIFETIME=300000, MAX_RESOURCE_BODY_SIZE=8192, BLOCKWISE_STRICT_BLOCK2_OPTION=false
[main] INFO org.eclipse.californium.core.network.CoapEndpoint - coap Endpoint [coap://0.0.0.0:0] requires an executor to start, using default single-threaded daemon executor
[main] INFO org.eclipse.californium.elements.UDPConnector - UDPConnector starts up 1 sender threads and 1 receiver threads
[main] INFO org.eclipse.californium.elements.UDPConnector - UDPConnector listening on 0.0.0.0/0.0.0.0:52281, recv buf = 65536, send buf = 65536, recv packet size = 2048
[main] INFO org.eclipse.californium.core.network.CoapEndpoint - coap Started endpoint at coap://0.0.0.0:52281
[main] INFO org.eclipse.californium.core.network.EndpointManager - created implicit endpoint coap://0.0.0.0:52281 for coap
[...]

PR #1910, opened 2022-02-10
OSCORE: Improvements to option handling and replay window

rikard-sics:
This pull request updates the handling of the OSCORE option, including separate option encoder/decoder classes, and now stores the entire OSCORE option in the exchange. This avoids using a separate map to keep track of the sequence number to use for processing incoming or outgoing responses. It also enables improvements to the replay window functionality.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
@@ -75,17 +75,18 @@ public class ObjectSecurityLayer extends AbstractLayer {
 	 * @param ctxDb the OSCore context DB
 	 * @param message the message
 	 * @param ctx the OSCore context
-	 * @param newPartialIV boolean to indicate whether to use a new partial IV or not
-	 * @param outerBlockwise boolean to indicate whether the block-wise options
-	 *            should be encrypted or not
+	 * @param newPartialIV whether to use a new partial IV or not
+	 * @param outerBlockwise whether the block-wise options should be encrypted
+	 * @param requestSequenceNr sequence number (Partial IV) from the request
+	 *            (if encrypting a response)
 	 * 
 	 * @return the encrypted message
 	 * 
 	 * @throws OSException error while encrypting response
 	 */
 	public static Response prepareSend(OSCoreCtxDB ctxDb, Response message, OSCoreCtx ctx, final boolean newPartialIV,
-			boolean outerBlockwise) throws OSException {
-		return ResponseEncryptor.encrypt(ctxDb, message, ctx, newPartialIV, outerBlockwise);
+			boolean outerBlockwise, int requestSequenceNr) throws OSException {
+		return ResponseEncryptor.encrypt(ctxDb, message, ctx, newPartialIV, outerBlockwise, requestSequenceNr);
 	}
 
 	/**
@@ -108,12 +109,16 @@ public class ObjectSecurityLayer extends AbstractLayer {
 	 *
 	 * @param ctxDb the context database used
 	 * @param response the incoming request
+	 * @param requestSequenceNr sequence number (Partial IV) from the request
+	 *            (if decrypting a response)
+	 * 
 	 * @return the decrypted and verified response
 	 * 
 	 * @throws OSException error while decrypting response
 	 */
-	public static Response prepareReceive(OSCoreCtxDB ctxDb, Response response) throws OSException {
-		return ResponseDecryptor.decrypt(ctxDb, response);
+	public static Response prepareReceive(OSCoreCtxDB ctxDb, Response response, int requestSequenceNr)
+			throws OSException {
+		return ResponseDecryptor.decrypt(ctxDb, response, requestSequenceNr);
 	}
 
 	@Override
@@ -161,9 +166,6 @@ public class ObjectSecurityLayer extends AbstractLayer {
 				 */
 				OSCoreEndpointContextInfo.sendingRequest(ctx, exchange);
 
-				exchange.setCryptographicContextID(ctx.getRecipientId());
-				final int seqByToken = ctx.getSenderSeq();
-
 				final Request preparedRequest = prepareSend(ctxDb, request);
 				final OSCoreCtx finalCtx = ctxDb.getContext(uri);
 
@@ -185,11 +187,11 @@ public class ObjectSecurityLayer extends AbstractLayer {
 						}
 
 						ctxDb.addContext(token, finalCtx);
-						ctxDb.addSeqByToken(token, seqByToken);
 					}
 				});
 
 				req = preparedRequest;
+				exchange.setCryptographicContextID(req.getOptions().getOscore());
 
 			} catch (OSException e) {
 				LOGGER.error("Error sending request: " + e.getMessage());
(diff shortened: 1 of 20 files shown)


rikard-sics:
This should resolve the issue reported in #1707, which seems to be caused by a combination of the way the sequence number to use for responses was handled, in combination with the previous code for the replay window.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
```diff
@@ -75,17 +75,18 @@ public static Request prepareSend(OSCoreCtxDB ctxDb, Request message) throws OSE
...
+	 *            (if encrypting a response)
 	 * 
 	 * @return the encrypted message
 	 * 
 	 * @throws OSException error while encrypting response
 	 */
 	public static Response prepareSend(OSCoreCtxDB ctxDb, Response message, OSCoreCtx ctx, final boolean newPartialIV,
-			boolean outerBlockwise) throws OSException {
-		return ResponseEncryptor.encrypt(ctxDb, message, ctx, newPartialIV, outerBlockwise);
+			boolean outerBlockwise, Integer requestSequenceNr) throws OSException {
```
boaks, line 88: Is it intended, that the `Integer requestSequenceNr ` may be `null`? 
If so, would it be possible to add that to the javadoc?
If not, why is `Integer` used instead of `int`?

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
```diff
@@ -0,0 +1,220 @@
...
+
+		// Parsing Partial IV
+		if (n > 0) {
+			partialIV = Arrays.copyOfRange(total, index, index + n);
+			index += n;
+		}
+
+		// Parsing KID Context
+		if (h != 0) {
+			int s = total[index];
```
boaks, line 96: `index++`? (And remove the `+1`s)

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
```diff
@@ -0,0 +1,220 @@
...
+
+	/**
+	 * Retrieve the ID Context
+	 * 
+	 * @return the ID Context (kid context)
+	 */
+	public byte[] getIdContext() {
+		if (!decoded) {
+			decode();
+			decoded = true;
```
boaks, line 123: May consider to move (at least) that line `decode = true;` into `decode()`?

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
```diff
@@ -0,0 +1,220 @@
...
+		}
+
+		if (partialIV == null) {
+			return 0;
+		}
+
+		byte[] partialIvInt = new byte[Integer.BYTES];
+		System.arraycopy(partialIV, 0, partialIvInt, 0, partialIV.length);
+
+		return ByteBuffer.wrap(partialIvInt).getInt();
```
boaks, line 161: Did you also try to just use `ByteBuffer.wrap(partialIv).getInt()` ?

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
```diff
@@ -0,0 +1,220 @@
...
+		if (h != 0) {
+			int s = total[index];
+
+			kidContext = Arrays.copyOfRange(total, index + 1, index + 1 + s);
+
+			index += s + 1;
+		}
+
+		// Parsing KID
+		kid = Arrays.copyOfRange(total, index, total.length);
```
boaks, line 104: The encoding depends on the k-flag in the header-byte.
I'm not sure, where "misformed" options are detected, but I would use the flag to perform some validation here.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
boaks: Seems, that `partialIV` may be too short.

`Integer.BYTES` is not available in java 7 (until now, Californium is still usable with java 7, but not all features.)

I would prefer to go with:


```
byte[] partialIvInt = Arrays.copyOf(partialIV, Integer.SIZE / Byte.SIZE);
```

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
rikard-sics: True, requestSequenceNr does only have to be an Integer in the method Encryptor.encryptAndEncode, but i needlessly did that for other methods too. This has now been fixed in commit b5bc349413ab6d5572d45fff1fcce3c121431a65

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
rikard-sics: Done now in commit 41653450f5ca4444db22eefa050446d0171ce9b4

Would it be even better to put in the beginning of decode something like?:
```
if(!decoded) {
   return;
}
```

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
rikard-sics: Yes unfortunately partialIV is too short. I have updated the code according to your suggestion in commit 25e6209ae03b65ac32de4f6abaf00f4f03db8269

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
boaks:
```
if(!decoded) {
   return;
}
```

Yes, that's what I would do. 
The other variant is also ok for me. Just the `decoded = true;` should really be in the `decode()` function, and not somewhere else.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
rikard-sics: Okay, I updated according to my understanding in commit e284a0587b4aa51110312cf3b30aacd7b10fd8d2

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
```diff
@@ -85,26 +100,32 @@ private void decode() {
...
-		}
+				index += s;
+			}
 
-		// Parsing KID
-		kid = Arrays.copyOfRange(total, index, total.length);
+			// Parsing KID
+			if (k != 0) {
+				kid = Arrays.copyOfRange(total, index, total.length);
+			}
```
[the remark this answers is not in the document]
rikard-sics: Yes, true, that is an extra check that can be done. (Additional checks will also be done later in decryptor.decodeObjectSecurity).

I now throw an Exception if k == 0 but index != total.length, now I just throw a simple exception but perhaps that is fine since it will be caught right below and thrown as CoapOSException with an error message. Updated in commit 210590ce8febf31324d5cd7400a68e6ec94efc01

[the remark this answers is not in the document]
rikard-sics: I updated the error reporting when parsing the OSCORE option to be a bit more fine-grained in commit 51523ef232437132857844087de5c19526163421

rikard-sics:
I was alerted to a potential issue in the code from a colleague. If you could hold off on merging this until Monday so I can test and fix things it would be great. (I know you are also waiting for feedback in issue #1707).

boaks:
I will wait for you OK.

(And I would have, as usually, also asked you, if it's ready to merge from your side).

rikard-sics:
> I will wait for you OK.
> (And I would have, as usually, also asked you, if it's ready to merge from your side).

Thanks! Yeah makes sense :)

boaks:
Any update?
If possible, I would like to include this PR in release 3.3.0 (see #1909).

rikard-sics:
> Any update?
> If possible, I would like to include this PR in release 3.3.0 (see #1909).

I am on a business trip currently, but I will try to resolve this as fast as I can. My goal is to do it latest by Thursday. I hope that works considering the coming release, otherwise I can try to push it earlier.

boaks:
Depends on this week's Thursday or next week's. :-).

It will also be no big thing to postpone it from next week's Thursday to Tuesday the week after (1. March 2022), if you see, it gets required.

rikard-sics:
Alright, I have now managed to fix the identified issue. I will update and finalize the PR during the afternoon.

rikard-sics:
Okay, I have now made a final commit which should fix the remaining issue. Done in commit 6a2b75e4cc7933e30116ccbce1a1873a68a9a7e5

boaks:
Nice, please rebase and squash. I will do a final test run and merge it afterwards.

rikard-sics:
> Nice, please rebase and squash. I will do a final test run and merge it afterwards.

Thanks! I have now rebased and squashed.

rikard-sics:
> LGTM

Great, thanks for the merge!

PR #1914, opened 2022-02-14
Improve JCE startup.

boaks:
Optionally use non-blocking random for BC.


boaks:
@Nick-The-Uncharted

If you have time, you may check, if the BC support still works for you.

boaks:
Thanks!

PR #1918, opened 2022-02-19
OSCORE: Fix Partial IV parsing and add tests sending many requests

rikard-sics:
After the last PR I got feedback from an external collaborator about issues when sending larger numbers of requests. Unfortunately it seems a problem snuck in when parsing the Partial IV values into Sender Sequence Numbers. This PR fixes the parsing of Partial IVs. It also adds further JUnit tests which tests sending large numbers of requests, varying the PIV over its range, and the OSCORE option encoding/decoding.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
@@ -16,10 +16,10 @@
  ******************************************************************************/
 package org.eclipse.californium.oscore;
 
-import java.nio.ByteBuffer;
 import java.util.Arrays;
 
 import org.eclipse.californium.core.coap.CoAP.ResponseCode;
+import org.eclipse.californium.elements.util.DatagramReader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -92,8 +92,8 @@ public class OscoreOptionDecoder {
 		byte flagByte = total[0];
 
 		int n = flagByte & 0x07;
-		int k = flagByte & 0x08;
-		int h = flagByte & 0x10;
+		int k = (flagByte & 0x08) >> 3;
+		int h = (flagByte & 0x10) >> 4;
 
 		byte[] partialIV = null;
 		byte[] kid = null;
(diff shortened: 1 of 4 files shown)


rikard-sics:
I will do one more commit to this PR and things should be solved.

rikard-sics:
> I will do one more commit to this PR and things should be solved.

I did a further commit with additional JUnit tests and further fixes. Now the PIV->SSN conversion should always happen correctly.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OscoreOptionDecoder.java
```diff
@@ -179,7 +179,8 @@ public int getSequenceNumber() {
 			return 0;
 		}
 
-		byte[] partialIvInt = Arrays.copyOf(partialIV, Integer.SIZE / Byte.SIZE);
+		int padding = Integer.SIZE / Byte.SIZE - partialIV.length;
+		byte[] partialIvInt = OSSerializer.leftPaddingZeroes(partialIV, padding);
```
boaks, line 183: Seems, this is somehow more complicated ... 
Have you considered the Californium utility class `DatagramReader` as well? With that it's:

```
return new DatagramReader(partialIV, false).read(partialIV.length * Byte.SIZE);
```

rikard-sics: Thanks for the tip! No I had not considered that one. Let me adjust the code to use that instead, test, and report back.

rikard-sics: Using the DatagramReader for this worked out nicely. I have now changed to using it instead in commit f488416cbb3b3933bde4d8d8453c0aedb86419a7

boaks:
LGTM

If you're ready as well, please rebase and squash.

rikard-sics:
> LGTM
> 
> If you're ready as well, please rebase and squash.

Sounds good, I will rebase and squash shortly.

rikard-sics:
> Sounds good, I will rebase and squash shortly.

Alright, I have now rebased and squashed.

PR #1970, opened 2022-04-07
OSCORE: Add OSCORE support to plugtest server and client

rikard-sics:
This PR adds support for OSCORE to the plugtest server and client. This includes a resource on the plugtest server that can only be accessed using OSCORE (/oscore), and a resource from which the current OSCORE security configuration can be retrieved (/oscoreInfo). In addition, the README file has been expanded with information and settings for OSCORE support in the plugtest server/client.

The OSCORE support in the plugtest server supports the Appendix B.2 context rederivation procedure, and the plugtest client will use it by default. This is because otherwise requests from new clients would be considered replays (as the server's replay window is filled up from earlier clients). Clients that do not support the Appendix B.2 procedure can retrieve the latest security material (including an appropriate Sender Sequence Number) from the resource "/oscoreInfo" using plain CoAP.


rikard-sics:
I went for the solution of enabling the Appendix B.2 mode by default on the client as clients would otherwise just experience replays (unless I disable replay protection).

For now the OSCORE support is always enabled, I could add flags to control this but was a bit hesitant to extend classes such as ClientInitializer and possibly others.

Lastly I enabled TCP support in the OSCoreCoapStackFactory as that seemed to be needed for the default config for the client, and there is no particular reason that OSCORE should not support TCP also. I hope this was not too "radical".

demo-apps/cf-plugtest-client/pom.xml
```diff
@@ -36,6 +36,10 @@
 			<groupId>${project.groupId}</groupId>
 			<artifactId>cf-cli</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.eclipse.californium</groupId>
```
boaks, line 40: Not that important, but may be replaced by

`<groupId>${project.groupId}</groupId>`

demo-apps/cf-plugtest-server/pom.xml
```diff
@@ -36,6 +36,10 @@
 			<groupId>${project.groupId}</groupId>
 			<artifactId>scandium</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.eclipse.californium</groupId>
```
boaks, line 40: Also here

`<groupId>${project.groupId}</groupId>`

boaks:
> Lastly I enabled TCP support in the OSCoreCoapStackFactory as that seemed to be needed for the default config for the client, and there is no particular reason that OSCORE should not support TCP also. I hope that is fine.

If that works, it's sure fine. 

I will need a little time to get my current work done. After that I will then test your PR and provide some feedback.

demo-apps/cf-plugtest-client/pom.xml
rikard-sics: Now updated in commit 705e27c9636e950128fa814e04bb0460dafa2778

demo-apps/cf-plugtest-server/pom.xml
rikard-sics: Now updated in commit 705e27c9636e950128fa814e04bb0460dafa2778

rikard-sics:
> If that works, it's sure fine.

Hmm, on further reflection maybe it was a bit premature before I can implement JUnit tests and such. If it's possible to modify the plugtest server to run with OSCORE not supporting TCP that may be better for now.



> I will need a little time to get my current work done. After that I will then test your PR and provide some feedback.

Sounds good!

boaks:
>  If it's possible to modify the plugtest server to run with OSCORE not supporting TCP that may be better for now.

Generally disable TCP? Or just for the one Resource?
I guess generally. Then I would consider, that we just make a short remark, that for now, using OSCORE over TCP is "experimental". (TCP is anyway still experimental :-) ).

rikard-sics:
> Generally disable TCP? Or just for the one Resource?
> I guess generally. Then I would consider, that we just make a short remark, that for now, using OSCORE over TCP is "experimental". (TCP is anyway still experimental :-) ).

Yeah then either
1. Disabling TCP for the /oscore resource specifically.
2. Or remark that OSCORE over TCP support is experimental.

I am fine with the latter :)

boaks:
> I am fine with the latter :)

Then, please go for the remark.

boaks:
I had a first test run.
Remarks:
I would appreciate, if the log-levels are reconsidered. INFO seems for me to be too much (maybe in a additional PR).
I would also appreciate, if a cli options is added to disable the usage of oscore, see [ping](https://github.com/eclipse/californium/blob/master/demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java#L98-L99) for example.

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
```diff
@@ -107,6 +115,9 @@ public void applyDefinitions(Configuration config) {
 	 */
 	public static void main(String[] args) throws ConnectorException, IOException {
 
+		HashMapCtxDB db = new HashMapCtxDB();
```
boaks, line 118: Please consider to use a cli-option (default enabled) to switch OSCORE on/off.

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
```diff
@@ -145,6 +156,9 @@ public static void main(String[] args) throws ConnectorException, IOException {
 			testCB(clientConfig.uri, context);
 			testCO(clientConfig.uri, context);
 			testCL(clientConfig.uri, context);
+
+			initOscore(clientConfig.uri, db);
```
boaks, line 160: Please use the above cli-option also here.

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
```diff
@@ -705,6 +719,103 @@ public static void testCL(String uri, EndpointContext context) throws ConnectorE
...
+	 * otherwise arise for clients.
+	 * 
+	 * @param uri the URI of the OSCORE resource at the server
+	 * @param db the OSCORE context database
+	 */
+	public static void initOscore(String uri, HashMapCtxDB db) {
+		AlgorithmID alg = AlgorithmID.AES_CCM_16_64_128;
+		AlgorithmID kdf = AlgorithmID.HKDF_HMAC_SHA_256;
+
+		byte[] master_secret = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E,
```
boaks, line 797: Maybe you have a look at [StringUtil.hex2ByteArray(String)](https://github.com/eclipse/californium/blob/master/element-connector/src/main/java/org/eclipse/californium/elements/util/StringUtil.java#L248). At least for initialization, that may be easier.

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
```diff
@@ -705,6 +719,103 @@ public static void testCL(String uri, EndpointContext context) throws ConnectorE
...
+				0x0F, 0x10 };
+		byte[] master_salt = { (byte) 0x9e, (byte) 0x7c, (byte) 0xa9, (byte) 0x22, (byte) 0x23, (byte) 0x78,
+				(byte) 0x63, (byte) 0x40 };
+		byte[] sid = new byte[] { 0x01 };
+		byte[] rid = new byte[] { 0x02 };
+		byte[] id_context = new byte[] { (byte) 0x37, (byte) 0xcb, (byte) 0xf3, (byte) 0x21, (byte) 0x00, (byte) 0x17,
+				(byte) 0xa2, (byte) 0xd3 };
+		int MAX_UNFRAGMENTED_SIZE = 4096;
```
boaks, line 805: I'm not too sure, but `CoapConfig.MAX_RESOURCE_BODY_SIZE` is in an other place. We may also start to add a configuration module for oscore, and add oscore specific values to that.

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
```diff
@@ -269,6 +280,10 @@ public void applyDefinitions(Configuration config) {
 	}
 
 	public static void main(String[] args) {
+		
+		db = new HashMapCtxDB();
```
boaks, line 284: Please move that to [PlugtestServer.init(BaseConfig)](https://github.com/eclipse/californium/blob/master/demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java#L351) and try to use arguments for `db` and `serverRid` instead of `static` fields.

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
```diff
@@ -523,6 +540,40 @@ public static boolean console(ActiveInputReader reader, long timeout) {
...
+		AlgorithmID alg = AlgorithmID.AES_CCM_16_64_128;
+		AlgorithmID kdf = AlgorithmID.HKDF_HMAC_SHA_256;
+
+		byte[] master_secret = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E,
+				0x0F, 0x10 };
+		byte[] master_salt = { (byte) 0x9e, (byte) 0x7c, (byte) 0xa9, (byte) 0x22, (byte) 0x23, (byte) 0x78,
+				(byte) 0x63, (byte) 0x40 };
+		byte[] sid = new byte[] { 0x02 };
+		byte[] rid = new byte[] { 0x01 };
+		serverRid = rid;
```
boaks, line 560: That may be returned instead of written to a `static` field.

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
```diff
@@ -361,6 +376,8 @@ public static Configuration init(BaseConfig config) {
 		// create server
 		try {
 			List<Protocol> protocols = config.getProtocols();
+			protocols.remove(Protocol.DTLS);
```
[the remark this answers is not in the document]
boaks: I've tested it with DTLS. Seems to work.
I'm not sure, why it should be removed.

boaks:
TCP support:

In order to use CoAPoverTCP, the CoapStack uses some different layers. Therefore the default stack-factory distinguish, which stack to use, by the protocol, see [default stack-factory](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java#L1740-L1747) .

if OSCORE is to be used with TCP, a tcp-oscore stack , similar to CoapTcpStack](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapTcpStack.java) is required, and the oscore stack-factory should then also use the protocol specific stack.

For now, it may be also ok, if the oscore stack-factory just use the CoapTcpStack without supporting oscore.
If the client is the adapted to disable oscore by cli-option, I think, the client may the also just warn, that oscore is currently not supported for TCP instead of sending the oscore-test-request.

README.md
```diff
@@ -180,6 +180,17 @@ A test server is running at <a href="coap://californium.eclipseprojects.io:5683/
 It is an instance of the [cf-plugtest-server](https://repo.eclipse.org/content/repositories/californium-releases/org/eclipse/californium/cf-plugtest-server/3.4.0/cf-plugtest-server-3.4.0.jar) from the demo-apps.
 The root resource responds with its current version.
 
+The server has a resource only accessible using OSCORE under "/oscore". It is configured with the following security material:
```
boaks, line 183: Maybe add "(in preparation)", though it will take perhaps a couple of times, to deploy it to the sandbox after merging this PR.

rikard-sics:
> I would appreciate, if the log-levels are reconsidered. INFO seems for me to be too much (maybe in a additional PR).

You mean in general in the OSCORE code? Yeah I could make a follow-up PR to downgrade the logging levels, especially INFO.

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
rikard-sics: Done in commit e0ea3913838eecdd4908b73034ba5aa8a41858de

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
```diff
@@ -523,6 +540,40 @@ public static boolean console(ActiveInputReader reader, long timeout) {
...
+	 * configuration and adds it to the OSCORE context database. The created
+	 * context will support the Appendix B.2. context rederivation procedure.
+	 * 
+	 * @param db the OSCORE context database
+	 */
+	public static void initOscore(HashMapCtxDB db) {
+		AlgorithmID alg = AlgorithmID.AES_CCM_16_64_128;
+		AlgorithmID kdf = AlgorithmID.HKDF_HMAC_SHA_256;
+
+		byte[] master_secret = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E,
```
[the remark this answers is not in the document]
rikard-sics: Done in commit e0ea3913838eecdd4908b73034ba5aa8a41858de

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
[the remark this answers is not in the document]
rikard-sics: Yeah seems to be just a mistake/leftovers from testing. Reverted in commit 082725ddf6e06e789b9535e79a3734af66f2116f

What about TLS? Perhaps I should delete also the following?
https://github.com/eclipse/californium/blob/082725ddf6e06e789b9535e79a3734af66f2116f/demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java#L379

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
rikard-sics: Done in commit 95a165891b5fec3c4bb8eaacdc1e14da386bd6b8

rikard-sics:
> CP support:
> 
> In order to use CoAPoverTCP, the CoapStack uses some different layers. Therefore the default stack-factory distinguish, which stack to use, by the protocol, see [default stack-factory](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java#L1740-L1747) .
> 
> if OSCORE is to be used with TCP, a tcp-oscore stack , similar to [CoapTcpStack](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/stack/CoapTcpStack.java) is required, and the oscore stack-factory should then also use the protocol specific stack.
> 
> For now, it may be also ok, if the oscore stack-factory just use the CoapTcpStack without supporting oscore. If the client is the adapted to disable oscore by cli-option, I think, the client may the also just warn, that oscore is currently not supported for TCP instead of sending the oscore-test-request.

I will explore a bit and think about it. Just saying that OSCORE doesn't support TCP for now may be fine.

rikard-sics:
I made a commit 3700d5ca6a5963cb164a3ad1ff8985596f125b6c to remove some of the information in the /oscoreInfo resource. After internal discussions we determined that it is better not to send the master secret etc. over the network in plaintext like that as it may give incorrect assumptions on how OSCORE normally functions. That information can be gotten from the code or the README file.

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
boaks: FMPOV, better than compiled in values.

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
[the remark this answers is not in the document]
boaks: Yes, if the stack-factory gets fixed.

boaks:
By the way, though I'm not in the OSCORE spec, maybe you also consider to add a README.md with some instructions/tutorial how to use it in the future? Does and No-Does?

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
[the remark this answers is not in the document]
boaks: Sorry, I may have got it wrong.

No, don't remove neither TCP nor TLS, that is not required!
It is required to add TCP/TLS support to the [OSCoreCoapStackFactory](https://github.com/eclipse/californium/blob/master/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCoapStackFactory.java#L41), either by using the available [CoapTcpStack](https://github.com/eclipse/californium/blob/master/californium-core/src/main/java/org/eclipse/californium/core/network/CoapEndpoint.java#L1742-L1744) or try to implement a  `OSCoreTcpStack` using the TCP variants of the layers.

And for the client, once the usage of oscore is out in a cli-option, also check the protocol flavor. If TCP is not supported for OSCORE, then just print a warning instead of executing the tests.

But the very most important thing is:
Adapt the  OSCoreCoapStackFactory to support the TCP flavor, even without OSCORE support.

boaks:
> I will explore a bit and think about it. Just saying that OSCORE doesn't support TCP for now may be fine.

That OSCORE doesn't support TCP is one thing, the real pain is, it disables/prevent the usage of TCP then at all.
That could be easily fixed by adapting the OSCoreCoapStackFactory. Please go for that.

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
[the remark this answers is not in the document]
rikard-sics: Removal of TLS has been deleted now in commit 04e3600364132a1e37a76fa60d5773628254ae1e

rikard-sics:
> That OSCORE doesn't support TCP is one thing, the real pain is, it disables/prevent the usage of TCP then at all.
> That could be easily fixed by adapting the OSCoreCoapStackFactory. Please go for that.

An update on this: I worked on creating a TCP-enabled stack for OSCORE so that OSCORE can support TCP just as it does UDP. 

I could not get the plugtest client to work with TCP but could adapt the TcpThroughputClient for my testing. After testing with it I believe OSCORE over TCP should now work also.

This work was done in commit 04e3600364132a1e37a76fa60d5773628254ae1e

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
rikard-sics: Now retrieving values from the standard configuration, modified in commit ee60caa4c866d635e36ee0e499433017246d7423

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
rikard-sics: I struggle a bit in how to do this in the best way. The command line parameters are parsed in `ClientInitializer.init` and it is also in there the Endpoint is created.

However I need to know whether to do `OSCoreCoapStackFactory.useAsDefault(db);` before the endpoint is created, which I can't know since I don't know the results of the command line argument parsing.

For alternatives I came up with three options

- Adding the cf-oscore as a dependency to cf-cli, creating the HashMapCtxDB and setting it with `OSCoreCoapStackFactory.useAsDefault(db);` inside there (using `setCustomCoapStackArgument(customStackArgument)` & `setCoapStackFactory` should work too).
- Moving out the Endpoint creation from `ClientInitializer.init` into the code of the plugtest client.
- Manually pre-parsing the command line options (for OSCORE only). If OSCORE is used I can add to the clientConfig the OSCORE context DB as customStackArgument and the OSCoreCoapStackFactory as coapStackFactory. Then within `ClientInitializer.init` if OSCORE is used those values are read from the config and used when creating the Endpoint.

Not sure what you prefer or if there are better options.

demo-apps/cf-plugtest-server/src/main/java/org/eclipse/californium/plugtests/PlugtestServer.java
rikard-sics: I tried to restructure things accordingly in commit c9dff1a0e4c982531367ca05e9c287a174bf7f48

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCoapStackFactory.java
```diff
@@ -38,9 +37,7 @@
 	@Override
 	public CoapStack createCoapStack(String protocol, String tag, Configuration config,
 			EndpointContextMatcher matchingStrategy, Outbox outbox, Object customStackArgument) {
-		if (CoAP.isTcpProtocol(protocol)) {
```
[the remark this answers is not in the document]
rikard-sics: Should be resolved in commit 04e3600364132a1e37a76fa60d5773628254ae1e. See also my comment later on in the thread.

README.md
rikard-sics: Done in commit ad70aba45b8a0006a67118fd0b8696d9211c0b1b

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreTcpStack.java
```diff
@@ -0,0 +1,78 @@
...
+	 * @param config The configuration values to use.
+	 * @param outbox The adapter for submitting outbound messages to the
+	 *            transport.
+	 * @param ctxDb context DB.
+	 * @deprecated use
+	 *             {@link #OSCoreTcpStack(String, Configuration, EndpointContextMatcher, Outbox, OSCoreCtxDB)}
+	 *             instead.
+	 * @since 3.0 (logging tag added and changed parameter to Configuration)
+	 */
+	public OSCoreTcpStack(String tag, Configuration config, Outbox outbox, OSCoreCtxDB ctxDb) {
```
boaks, line 75: For a new `Stack`, just don't provide the deprecated constructor.
(for UDP this was done for backwards compatibility.)

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreUdpStack.java
```diff
@@ -60,19 +62,19 @@ public OSCoreStack(String tag, Configuration config, EndpointContextMatcher matc
...
 	 *            transport.
 	 * @param ctxDb context DB.
 	 * @deprecated use
-	 *             {@link #OSCoreStack(String, Configuration, EndpointContextMatcher, Outbox, OSCoreCtxDB)}
+	 *             {@link #OSCoreUdpStack(String, Configuration, EndpointContextMatcher, Outbox, OSCoreCtxDB)}
 	 *             instead
 	 * @since 3.0 (logging tag added and changed parameter to Configuration)
 	 */
-	public OSCoreStack(String tag, Configuration config, Outbox outbox, OSCoreCtxDB ctxDb) {
+	public OSCoreUdpStack(String tag, Configuration config, Outbox outbox, OSCoreCtxDB ctxDb) {
```
boaks, line 77: Now it's a new class, so we don't need the deprecated constructor.
With the new name, it's anyway required to adapt the using code. With that, the backwards compatibility is broken, and so wen don't benefit from providing the deprecated variant.

boaks:
> I could not get the plugtest client to work with TCP

That requires a `mvn clean install` in the parent folder (californium's root folder). After that, the cf-plugtest-client target folder will contain both, the `cf-plugtest-client-3.5.0-SNAPSHOT.jar` and the `cf-cli-tcp-netty-3.5.0-SNAPSHOT.jar`. If the plugtest-client is then started from the command line with 

```
java -jar target/cf-plugtest-client-3.5.0-SNAPSHOT.jar coap+tcp://localhost
```

it works. Great, thanks for the efforts to make it working with the TCP flavor.

boaks:
About the cli-option, I will have a look at it and report, if I found a solution.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCoapStackFactory.java
```diff
@@ -27,8 +27,8 @@
 import org.eclipse.californium.elements.config.Configuration;
```
boaks, line 27: Please remove the

```
import org.eclipse.californium.core.network.CoapStackFactory;
```

above. It's not longer required and reported with a "deprecated warning".

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
boaks: With PR #1972 it will be

```
		ClientInitializer.init(args, clientConfig, false);
		HashMapCtxDB db = null;
		if (clientConfig.oscore) {
			db = new HashMapCtxDB();
			OSCoreCoapStackFactory.useAsDefault(db);
		}
		ClientInitializer.registerEndpoint(clientConfig, null);
```

boaks: PR #1972 is merged. If you rebase, you should be able to apply to above snippet.

boaks:
From my side:

Removing `import org.eclipse.californium.core.network.CoapStackFactory;` and the cli-option is the only left thing. If you're done with that, please "squash" (rebase is required anyway for the cli-option).

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
rikard-sics: Thanks!  Now I have added cli-option for the client in commit 20aa762ebf9b41e73f67f4b9fd28afdf2f08ebd0

demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
rikard-sics: Done in commit 20aa762ebf9b41e73f67f4b9fd28afdf2f08ebd0

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreTcpStack.java
rikard-sics: Done in commit 2861fb22ba8d37687159909b017663b15631fded

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreUdpStack.java
rikard-sics: Done in commit 2861fb22ba8d37687159909b017663b15631fded

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreCoapStackFactory.java
rikard-sics: Done in commit 2861fb22ba8d37687159909b017663b15631fded

rikard-sics:
> By the way, though I'm not in the OSCORE spec, maybe you also consider to add a README.md in the future with some instructions/tutorial how to use it? Does and No-Does?

Yeah, I think that can definitely make sense. Like with information on how to get started and other relevant things.

rikard-sics:
> Yes.
> 
> FMPOV, though it's about encryption, also:
> 
>     * which credentials are involved? Scope (e.g. peers, set of peers)? Provisioning?
> 
>     * are credentials derived? Scope? Timespan?
> 
> 
> What ever is useful to understand how to use it.

True, it could go into details about those aspects also. I will start thinking about this for a future PR.

rikard-sics:
> I would appreciate, if the log-levels are reconsidered. INFO seems for me to be too much (maybe in a additional PR).

Taking care of the excessive logging levels can be the next PR I go for.

boaks:
After merging, I tested it a little more.
I think, the POST is not really implemented.

boaks:
I found 

```
LOGGER.info("Request: " + exchange.getRequest().toString());
```

if possible, when revisiting the logging levels, please cleanup such stuff as well.

```
LOGGER.info("Request: {}", exchange.getRequest());
```

Especially, if the logging level is disabled, this has a "large" benefit, though in the current version, the `toString()` of `Request` is always called and the resulting string is also always created. In the later, both are only done, when the logging level is enabled.

rikard-sics:
> After merging, I tested it a little more. I think, the POST is not really implemented.

Yeah, I see what you mean. Both OSCORE 03 and OSCORE 04 is using newPut(). That's my bad, I missed that completely.

If it is fine I can fix this also as part of the log-level PR? (Or I make a separate PR now for it)

boaks:
> If it is fine I can fix this also as part of the log-level PR? (Or I make a separate PR now for it)

I would prefer, to have first the log-level stuff.
And later you may add the POST in a separate PR.

PR #1977, opened 2022-04-11
Add OSCORE to plugtest-checker.

boaks:
Also transfer MID to outer request.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
@@ -186,6 +186,10 @@ public class ObjectSecurityLayer extends AbstractLayer {
 							request.setToken(token);
 						}
 
+						if (!request.hasMID() && preparedRequest.hasMID()) {
+							request.setMID(preparedRequest.getMID());
+						}
+
 						ctxDb.addContext(token, finalCtx);
 					}
 				});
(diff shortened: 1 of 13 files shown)


boaks:
@rikard-sics 

Just in the case you can spend some time in the changes of [ObjectSecurityLaye](https://github.com/eclipse/californium/pull/1977/files#diff-e1ba1103f30f6728d503e453033a9c26f2dd9dbb8b630160adfe603d57c40d73)

and the changed unit tests.

rikard-sics:
> Just in the case you can spend some time in the changes of [ObjectSecurityLaye](https://github.com/eclipse/californium/pull/1977/files#diff-e1ba1103f30f6728d503e453033a9c26f2dd9dbb8b630160adfe603d57c40d73)
> 
> and the changed unit tests.

Yes, let me have a look during the day.

rikard-sics:
Yeah, it looks good to me.

I had one thought though which is that perhaps some of the logic of transferring information from the message before protection to the message after protection (like the code about duplicate and nanoTimestamp) could be done in the following methods instead?
https://github.com/eclipse/californium/blob/d3be27d7b48b21e466bd72cbc5d04529059a9db0/cf-oscore/src/main/java/org/eclipse/californium/oscore/OptionJuggle.java#L256
or
https://github.com/eclipse/californium/blob/d3be27d7b48b21e466bd72cbc5d04529059a9db0/cf-oscore/src/main/java/org/eclipse/californium/oscore/OptionJuggle.java#L289

There is where a new request/response is built from an old one.

PR #1978, opened 2022-04-12
OSCORE: Reduce logging levels and improve logging statements

rikard-sics:
This PR reduces logging levels (especially info) and restructures the logging statements to avoid building string output when the related logging level is turned off.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
@@ -287,7 +287,7 @@ public class ContextRederivation {
 
 		// Check if context re-derivation is enabled for this context
 		if (ctx.getContextRederivationEnabled() == false) {
-			LOGGER.debug("Context re-derivation not initiated due to it being disabled for this context");
+			LOGGER.debug("Context re-derivation not considered due to it being disabled for this context");
 			return ctx;
 		 }
 
@@ -537,6 +537,11 @@ public class ContextRederivation {
 	 * @param ctx the OSCORE context in use
 	 */
 	private static void printStateLogging(OSCoreCtx ctx) {
+
+		if (!LOGGER.isDebugEnabled()) {
+			return;
+		}
+
 		PHASE currentPhase = ctx.getContextRederivationPhase();
 
 		String supplemental = "";
@@ -573,11 +578,10 @@ public class ContextRederivation {
 			break;
 		}
 
-		String output = "Context re-derivation phase: " + currentPhase + " (" + supplemental + ")";
 		if (currentPhase == PHASE.INACTIVE) {
-			LOGGER.debug(output);
+			LOGGER.trace("Context re-derivation phase: {} ({})", currentPhase, supplemental);
 		} else {
-			LOGGER.info(output);
+			LOGGER.debug("Context re-derivation phase: {} ({})", currentPhase, supplemental);
 		}
 	}
 
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/Decryptor.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/Decryptor.java
@@ -140,8 +140,9 @@ public abstract class Decryptor {
 			plaintext = enc.decrypt(key);
 
 		} catch (CoseException e) {
-			LOGGER.error(ErrorDescriptions.DECRYPTION_FAILED + " " + e.getMessage());
-			throw new OSException(ErrorDescriptions.DECRYPTION_FAILED + " " + e.getMessage());
+			String details = ErrorDescriptions.DECRYPTION_FAILED + " " + e.getMessage();
+			LOGGER.error(details);
+			throw new OSException(details);
 		}
 
 		return plaintext;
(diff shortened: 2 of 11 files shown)


cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
```diff
@@ -575,9 +575,9 @@ private static void printStateLogging(OSCoreCtx ctx) {
 
 		String output = "Context re-derivation phase: " + currentPhase + " (" + supplemental + ")";
```
boaks, line 576: If the output is "assembled" with in the "debug" or "trace", the string is not created in vain. If you consider that as "code duplicate", you may also use a `LOGGER.isInfoEnabled()` to guard this lines.

cf-oscore/src/main/java/org/eclipse/californium/oscore/Decryptor.java
```diff
@@ -140,7 +140,7 @@ protected static byte[] decryptAndDecode(Encrypt0Message enc, Message message, O
 			plaintext = enc.decrypt(key);
 
 		} catch (CoseException e) {
-			LOGGER.error(ErrorDescriptions.DECRYPTION_FAILED + " " + e.getMessage());
+			LOGGER.error(ErrorDescriptions.DECRYPTION_FAILED + " {}", e.getMessage());
```
boaks, line 143: The (theoretical) downside of this approach is:
It still creates a temporary string, the `ErrorDescriptions.DECRYPTION_FAILED + " {}"`.
And, if `ErrorDescriptions.DECRYPTION_FAILED` would contain some meta (`{}`), the it will fail either. Consider to use:

```
LOGGER.error("{} {}", ErrorDescriptions.DECRYPTION_FAILED, e.getMessage());
```

cf-oscore/src/main/java/org/eclipse/californium/oscore/Decryptor.java
```diff
@@ -140,7 +140,7 @@ protected static byte[] decryptAndDecode(Encrypt0Message enc, Message message, O
...
 			throw new OSException(ErrorDescriptions.DECRYPTION_FAILED + " " + e.getMessage());
```
boaks, line 144: In such a case, where the same message is used for logging and the Exception, I would got for:

```
String details = ErrorDescriptions.DECRYPTION_FAILED + " " + e.getMessage();
LOGGER.error("{}", details);
throw new OSException(details);
```

boaks:
I miss `CoapOSExceptionHandler`. No that much, gut I hope, fast done.

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
rikard-sics: Hmm, what about something like in commit 4a06fd47ed6343e508b3b3b06b31de08e20c10e9 ?

cf-oscore/src/main/java/org/eclipse/californium/oscore/Decryptor.java
rikard-sics: Fixed in commit d75f91db7d1c68682787a92248a0e8ec91ad6ccf

cf-oscore/src/main/java/org/eclipse/californium/oscore/Decryptor.java
rikard-sics: Yes I realize searching around that this pattern happens very many times in the current code.

Is this good to do in general (also for fixed strings) or only when part of the message is gotten from some object?

Like 
```
LOGGER.error("{} {}", ErrorDescriptions.DECRYPTION_FAILED, e.getMessage());
throw new OSException(ErrorDescriptions.DECRYPTION_FAILED + " " + e.getMessage());
```
vs.
```
LOGGER.error(ErrorDescriptions.TOKEN_NULL);
throw new NullPointerException(ErrorDescriptions.TOKEN_NULL);
```

I could certainly go through and change this for all instances if that is better.

rikard-sics:
> I miss `CoapOSExceptionHandler`. No that much, gut I hope, fast done.

Now covering this class also in commit 258c0c9976cfb7d852085382d8caa4729232ee45

cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
```diff
@@ -338,7 +338,7 @@ public void receiveResponse(Exchange exchange, Response response) {
 			if (!isProtected(response) && expectProtectedResponse) {
 				LOGGER.warn("Incoming response is NOT OSCORE protected!");
```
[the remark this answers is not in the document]
rikard-sics: Yeah, this may be worth "info" as it is quite important information. Changed in commit 7c9644408e3bf4d5f6caaac3f98cae612a8e4d70

cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
boaks: The approach in 4a06fd4 works for me

rikard-sics: > The approach in [4a06fd4](https://github.com/eclipse/californium/commit/4a06fd47ed6343e508b3b3b06b31de08e20c10e9) works for me

Great! Then I resolve this convo

boaks:
LGTM

(Please squash.)

rikard-sics:
> LGTM
> 
> (Please squash.)

Great! The commits have now been squashed.

PR #1991, opened 2022-04-21
Add patch request for CoapClient.

Ernest-su:


boaks:
Thanks for your contribution.
Would it be possible, that you add a ` @since 3.5` tag to your new methods?

Ernest-su:
added.

Issue #2031, opened 2022-05-26
DTLS enhanced cluster support


kiemlicz:
Hello,

In my company we're interested in connection sharing mechanism
and it _looks like_ we will have some time to invest

Some past issue to bring back some discussions: https://github.com/eclipse/californium/issues/1484

#### Reasoning
The assumption
```
I'm not sure, how many vm/node crashes you have been affected. In my experience, these are very rare, once or twice a year per system
```
is simply wrong. The DTLS state loss happens more often and this is a problem for us.
The graceful shutdown is a nice feature (thank you) however we need more reliability

#### (Very) Roughly we would like to introduce:
1. Storage to keep the CID to NODE_ID mapping (which node handles given CID), populated on client first contact
2. The in-memory state cache, periodically (not with every DTLS packet) flushed to persistent store (only the state change is flushed)
3. Given the node failure occurs (the packet is handled by the node not listed as 'designated' by given CID-NODE_ID mapping),
transfer the state to new node, update CID-NODE_IP mapping

Before starting the work, we would like to have some information regarding:
- are there any particular requirements that we should fulfill in order to get such PR merged into main branch?
- is the JDK7 requirement still valid?

Kind regards

boaks:
> In my experience, these are very rare, once or twice a year per system

> is simply wrong. The DTLS state loss happens more often and this is a problem for us.

What is wrong? My experience?
Obviously not !!!! I run about 4 servers on my own, I had not downtime in the last 1,5 years. There was one move of the californium sandbox to a new vm, but except of that, no I had no downtimes. 

So, how often does your VM crash?
Or are there any other reasons for lose the DTLS state?

> Storage to keep the CID to NODE_ID mapping (which node handles given CID), populated on client first contact

So, you don't want to encode the NODE_ID in the CID? OK then use and maintain a map. The `NodeConnectionIdGenerator`should enable you to do so.

> The in-memory state cache, periodically (not with every DTLS packet) flushed to persistent store (only the state change is flushed)

"(not every DTLS packet)", so you deal with the record sequence number on your own? OK, that record sequence number (write and read) is the only data changing after the handshake. You get callbacks for these changes in `ConnectionListener.onConnectionUpdatesSequenceNumbers(Connection connection, boolean writeSequenceNumber)`, that should do it. Just mark changes (read/write) and then flush these changes, when ever you think it's the right time.

> is the JDK7 requirement still valid?

Until now, JDK7 was working, and so we haven't been forced to update that.
 
[...]

PR #2038, opened 2022-07-05
Optional callback based action before retrieving connection

Pinioo:
Discussion ref: #2031

We'd like to introduce an optional, configurable action to be performed during datagram processing before the connection store is accessed. Injected actions may execute on separate threads, so datagram processing continues after calling the callback parameter.

This mechanism is necessary to be able to implement any logic which looks for cached CID-NODE mappings and performs relevant actions.


boaks:
Thanks for your contribution.

General:
To contribute to an eclipse project, you need to "sign" an [ECA](http://www.eclipse.org/legal/ECA.php).
For more information about contributing, please read [CONTRIBUTING](#CONTRIBUTING.md). 

About the changes:

I'm not sure, maybe I missed it, but there was not really a discussion.
If you don't want an real information exchange, OK.
 
AFAIK, you want to have a "(maybe remote) lookup". As I wrote in that issue #2031, I prefer changes, which are outside the implementation core. That ensures, that other users are not affected.

What is currently offered is 

```
protected void processDatagram(DatagramPacket packet, InetSocketAddress router)
```

I guess, your extension could be moved into a new class derived from `DTLSConnector`, similar to `DtlsClusterConnector`, or? 

If so, I would prefer that as solution.

boaks:
Just as additional hint/proposal:

Though your callback uses `ConnectionId cid` and `InetSocketAddress addr`

[DtlsClusterConnector](https://github.com/eclipse/californium/blob/main/scandium-core/src/main/java/org/eclipse/californium/scandium/DtlsClusterConnector.java#L467-L476)

demonstrates, how to read the cid without parsing all the records. It uses

```
ConnectionId cid = Record.readConnectionIdFromReader(reader, connectionIdGenerator); 
```

for that.

boaks:
Alternatively, we can add

```
protected void processRecords(List<Record> records, InetSocketAddress peerAddress, InetSocketAddress router)
```

to `DTLSConnector` and the new class overrides that.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1799,7 +1802,39 @@ protected void processDatagram(DatagramPacket packet, InetSocketAddress router)
...
+		final ConnectionId connectionId = firstRecord.getConnectionId();
 
+		if (beforeConnectionRetrievalAction == null) {
+			processRecords(records, peerAddress, connectionId, router);
+		} else {
+			final Thread usedReceiverThread = Thread.currentThread();
+			final Runnable processingPart = new Runnable() {
+				@Override
+				public void run() {
+					processRecords(records, peerAddress, connectionId, router);
```
boaks, line 1814: There is no need to execute the `processRecords` from the receiver thread or executor.
`executeInbound` will execute the job anyway using the executor. The code executed by the calling thread of `processRecords` is very small and simple.

boaks:
I scheduled the release 3.6.0 for next Thursday, 14. July.

I'm not sure, how to proceed with this PR.

Please, check with your employer, if there is interest in contributing.

If you prefer, that we just add `protected void processRecords(List<Record> records, InetSocketAddress peerAddress, InetSocketAddress router)` in order to enable you to add additional stuff on your own, I can change that for you.

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1740,7 +1740,7 @@ public void processDatagram(DatagramPacket datagram) {
 	 * @param router router address, {@code null}, if no router is used.
 	 * @since 2.5
 	 */
-	protected void processDatagram(DatagramPacket packet, InetSocketAddress router) {
+	protected void processDatagram(DatagramPacket packet, final InetSocketAddress router) {
```
boaks, line 1743: Please remove the added `final`s. I don't see, its required, or?

Pinioo:
Thank you for the response.

I pushed the corrected version with a new connector extending `DTLSConnector` with a simple usage example (a show of the concept) written in Javadoc. It is now isolated from core implementations.

I followed your suggestion to introduce `protected void processRecords(List<Record> records, InetSocketAddress peerAddress, InetSocketAddress router)`, it makes this new implementation much easier than it would be with overriding `processDatagram`.

scandium-core/src/main/java/org/eclipse/californium/scandium/DtlsBindingPreprocessingConnector.java
```diff
@@ -0,0 +1,145 @@
...
+ * 				}
+ * 			}
+ * 			callback.run();
+ * 		});
+ * 	} else {
+ * 		callback.run();
+ * 	}
+ * });
+ * </pre>
+ */
```
boaks, line 62: Please add a "@since 3.6"

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1798,8 +1798,13 @@ protected void processDatagram(DatagramPacket packet, InetSocketAddress router)
 			return;
 		}
 
-		final Record firstRecord = records.get(0);
+		processRecords(records, peerAddress, router);
+	}
 
+	protected void processRecords(final List<Record> records,
```
boaks, line 1804: Please add javadoc and

```
@since 3.6
```

scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
Pinioo: You're right, I left them accidentally

boaks:
If you can do the "polish" (it's not that much), I will continue with some tests. If these are passing (I don't see any reason, why not), then it will be merged and included in the version 3.6.

Just as hint:
In order to protect against DoS attack, I would consider to "limit" such pending jobs. I introduced

```
protected boolean executeInbound(Executor executor, InetSocketAddress peer, Runnable job);
```

for that. Not sure, if that is also useful for you, but a similar approach may help.

Pinioo:
> Just as hint:
> In order to protect against DoS attack, I would consider to "limit" such pending jobs.

That's a good point. Currently, this implementation doesn't introduce any async boundary explicitly so I would keep it as simple as it is but I will mention it in some Javadoc.

scandium-core/src/main/java/org/eclipse/californium/scandium/DtlsBindingPreprocessingConnector.java
```diff
@@ -0,0 +1,156 @@
...
+ * is accessed.
+ * </p>
+ * <p>
+ * Example of usage; simple reading and restoring connection from shared storage.
+ * This code uses a separate ForkJoinPool to execute its action in order to release connector's
+ * receiver threads in case of some complex operations (reading from map is here to
+ * just show the concept).
+ * </p>
+ * <pre>
+ * final Set<ConnectionId> knownCids = new HashSet<>();
```
boaks, line 42: I like it, but javadoc complains about `<` and `>` ;-(.
Please replace them by "&lt;" and "&lt;"

(And please add a hint, that `->` lamdas require to use at least java 8.)

boaks:
Beside of the `<>` in the javadoc, the changes have passed my tests.
So, please do a final "fix", then I set my LGTM and will merge it the day after the LGTM.

Pinioo:
Instead of replacing signs, I wrapped the whole code sample in "code" block, it also makes formatting prettier, at least inside IntelliJ.

I hope it does work with your javadoc version.

PR #2078, opened 2022-09-28
Update web-site.

boaks:

### DIFF
--- a/site/content/_global/nav.md
+++ b/site/content/_global/nav.md
@@ -1,13 +1,13 @@
 +++
 fragment = "nav"
 #disabled = true
-date = "2019-10-29"
+date = "2022-09-27"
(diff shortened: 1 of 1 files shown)


sophokles73:
IMHO we should not change the base URL until a conclusion has been reached in https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1991

boaks:
OK, I wasn't aware, that this is in discussion. 
So, wait for the outcome. 
The old github link will be forwarded, so that's also not that urgent.

boaks:
I removed the adaption of the web-site URL, if that comes, I will add it again.
So it's only the update to the new github repo.

sophokles73 (approved):
LGTM

Issue #2085, opened 2022-11-08
Adding DTLS sequence number to endpoint context


adamsero:
Hi, in this issue I wanted to request adding a small feature to the existing endpoint context code. I'm currently working on a service for monitoring the quality of network traffic that reaches Leshan server. 

The service would benefit greatly if it could make use of the DTLS sequence number field in COAP message's endpoint context - it can be used to assure continuous traffic and easy detection of missing packets. 

Since Leshan uses Californium, I had to make changes to the source code here and I'm hoping you could integrate it. The changes can be seen [here](https://github.com/JaroslawLegierski/californium/compare/main...opl/dtls_sequence_number) - they're really small and hopefully not a big problem to integrate.

boaks:
A PR will be welcome! 
There is already a valid ECA for your e-mail address, so a PR should be easy, or?

A first look:

Did you consider [KEY_PREFIX_NONE_CRITICAL](https://github.com/eclipse-californium/californium/blob/main/element-connector/src/main/java/org/eclipse/californium/elements/MapBasedEndpointContext.java#L64-L68)?

For the CID, we distinguish between read and write, so maybe

```
KEY_PREFIX_NONE_CRITICAL + "DTLS_READ_SEQUENCE_NUMBER"
```

boaks:
Just to mention:

There is also a [DatagramFilter](https://github.com/eclipse-californium/californium/blob/main/scandium-core/src/main/java/org/eclipse/californium/scandium/DatagramFilter.java), which may be also useful for analysis.

adamsero:
> Did you consider [KEY_PREFIX_NONE_CRITICAL](https://github.com/eclipse-californium/californium/blob/main/element-connector/src/main/java/org/eclipse/californium/elements/MapBasedEndpointContext.java#L64-L68)?

Yes, I added that, thanks for noticing.

Created PR #2086 for this.

boaks:
PR #2086 has been merged.

PR #2086, opened 2022-11-09
Add DTLS sequence number to endpoint context

adamsero:


element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
```diff
@@ -56,6 +56,12 @@ public class DtlsEndpointContext extends MapBasedEndpointContext {
 	 */
 	public static final Definition<Long> KEY_HANDSHAKE_TIMESTAMP = new Definition<>("DTLS_HANDSHAKE_TIMESTAMP",
 			Long.class, ATTRIBUTE_DEFINITIONS);
+	/**
+	 * The name of the attribute that contains the dtls packet sequence number
+	 * during the DTLS session as {@link Number}.
+	 */
+	public static final Definition<Long> DTLS_SEQUENCE_NUMBER = new Definition<>(
+			KEY_PREFIX_NONE_CRITICAL + "DTLS_SEQUENCE_NUMBER", Long.class, ATTRIBUTE_DEFINITIONS);
```
boaks, line 64: As I wrote in my comment on issue #2085, there are two different sequence number in a session. One for reading and one for writing. I would prefer to reflect that by naming it `DTLS_READ_SEQUENCE_NUMBER`.

WDYT?

boaks:
Beside of the name, LGTM.

element-connector/src/main/java/org/eclipse/californium/elements/DtlsEndpointContext.java
adamsero: Sorry, I focused on `KEY_PREFIX_NONE_CRITICAL` and completely overlooked that you also wanted to add `READ` in the name too, my bad 😅

boaks: Thanks for your contribution! You're welcome.

I don't think, that the "DTLS_WRITE_SEQUENCE_NUMBER" will be added, but if, then it's easier to use a specific name.

PR #2088, opened 2022-11-17
Add configuration to disable reuse of tokens on blockwise transfer.

boaks:
Improve protection from delay attacks, if no other means, maybe on application level, are available.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/config/CoapConfig.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/config/CoapConfig.java
@@ -611,12 +611,27 @@ public final class CoapConfig {
 	 * adapt the block size when receiving a 4.13 Entity too large response
 	 * code.
 	 * <p>
-	 * See https://tools.ietf.org/html/rfc7959#section-2.9.3 for more details.
+	 * @see <a href="https://tools.ietf.org/html/rfc7959#section-2.9.3" target="_blank">RFC7959, 2.9.3. - 4.13 Request Entity Too Large</a>
 	 */
 	public static final BooleanDefinition BLOCKWISE_ENTITY_TOO_LARGE_AUTO_FAILOVER = new BooleanDefinition(
 			MODULE + "BLOCKWISE_ENTITY_TOO_LARGE_AUTO_FAILOVER",
 			"Enable automatic failover on \"entity too large\" response.",
 			DEFAULT_BLOCKWISE_ENTITY_TOO_LARGE_AUTO_FAILOVER);
+	/**
+	 * Property to indicate that blockwise follow-up requests are reusing the
+	 * same token for traceability.
+	 * <p>
+	 * <b>Note:</b> reusing tokens may introduce a vulnerability, if
+	 * requests/response are captured and sent later without protecting the
+	 * integrity of the payload by other means.
+	 * </p>
+	 * 
+	 * @see <a href="https://github.com/core-wg/attacks-on-coap" target="_blank">attacks-on-coap</a>
+	 * @since 3.8
+	 */
+	public static final BooleanDefinition BLOCKWISE_REUSE_TOKEN = new BooleanDefinition(
+			MODULE + "BLOCKWISE_REUSE_TOKEN",
+			"Reuse token for blockwise requests. Ease traceability but may introduce vulnerability.", false);
 
 	/**
 	 * Time interval for a coap-server to check the client's interest in further
@@ -794,6 +809,7 @@ public final class CoapConfig {
 			config.set(BLOCKWISE_STATUS_INTERVAL, DEFAULT_BLOCKWISE_STATUS_INTERVAL_IN_SECONDS, TimeUnit.SECONDS);
 			config.set(BLOCKWISE_STRICT_BLOCK2_OPTION, DEFAULT_BLOCKWISE_STRICT_BLOCK2_OPTION);
 			config.set(BLOCKWISE_ENTITY_TOO_LARGE_AUTO_FAILOVER, DEFAULT_BLOCKWISE_ENTITY_TOO_LARGE_AUTO_FAILOVER);
+			config.set(BLOCKWISE_REUSE_TOKEN, false);
 			// BERT enabled, when > 1
 			config.set(TCP_NUMBER_OF_BULK_BLOCKS, 4);
 
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/Block1BlockwiseStatus.java
@@ -33,6 +33,13 @@ public final class Block1BlockwiseStatus extends BlockwiseStatus {
 	private static final Logger LOGGER = LoggerFactory.getLogger(Block1BlockwiseStatus.class);
 
 	/**
+	 * Current pending block wise request.
+	 * 
+	 * @since 3.8
+	 */
+	private Request current;
+
+	/**
 	 * Create block1wise status.
 	 * 
 	 * @param keyUri key uri of the blockwise transfer
@@ -47,6 +54,7 @@ public final class Block1BlockwiseStatus extends BlockwiseStatus {
 	private Block1BlockwiseStatus(KeyUri keyUri, RemoveHandler removeHandler, Exchange exchange, Request request,
 			int maxSize, int maxTcpBertBulkBlocks) {
 		super(keyUri, removeHandler, exchange, request, maxSize, maxTcpBertBulkBlocks);
+		current = request;
 	}
 
 	/**
@@ -189,6 +197,7 @@ public final class Block1BlockwiseStatus extends BlockwiseStatus {
 		block.getOptions().setBlock1(blockSzx, m, num);
 
 		setComplete(!m);
+		current = block;
 		return block;
 	}
 
@@ -214,13 +223,15 @@ public final class Block1BlockwiseStatus extends BlockwiseStatus {
 	}
 
 	/**
-	 * Checks whether a response has the same token as the request that
-	 * initiated the block1 transfer that this is the tracker for.
+	 * Checks whether a response has the same token as the current request of
+	 * this tracker.
 	 * 
 	 * @param response The response to check.
 	 * @return {@code true} if the tokens match.
+	 * @since 3.8 use the current request instead of the initial request to
+	 *        support blockwise transfer with changing tokens.
 	 */
-	public boolean hasMatchingToken(final Response response) {
-		return response.getToken().equals(firstMessage.getToken());
+	public synchronized boolean hasMatchingToken(final Response response) {
+		return response.getToken().equals(current.getToken());
 	}
 }
--- a/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/network/stack/BlockwiseLayer.java
@@ -218,6 +218,7 @@ public class BlockwiseLayer extends AbstractLayer {
 	private volatile boolean enableStatus;
 	private ScheduledFuture<?> statusLogger;
 	private ScheduledFuture<?> cleanup;
+	private final long healthStatusInterval;
 	private final int maxTcpBertBulkBlocks;
 	private final int maxMessageSize;
 	private final int preferredBlockSzx;
@@ -226,12 +227,23 @@ public class BlockwiseLayer extends AbstractLayer {
 	private final int maxResourceBodySize;
 	private final boolean strictBlock1Option;
 	private final boolean strictBlock2Option;
-	private final long healthStatusInterval;
+	/**
+	 * Reuse tokens for follow-up requests.
+	 * <p>
+	 * <b>Note:</b> reusing tokens may introduce a vulnerability, if
+	 * requests/response are captured and sent later without protecting the
+	 * integrity of the payload by other means.
+	 * </p>
+	 * 
+	 * @see <a href="https://github.com/core-wg/attacks-on-coap" target="_blank">attacks-on-coap</a>
+	 * @since 3.8
+	 */
+	private final boolean reuseToken;
 	/* @since 2.4 */
 	private final boolean enableAutoFailoverOn413;
 
 	private final EndpointContextMatcher matchingStrategy;
-	
+
 	/**
 	 * Creates a new blockwise layer for a configuration.
 	 * <p>
@@ -372,7 +384,7 @@ public class BlockwiseLayer extends AbstractLayer {
 		});
 		strictBlock1Option = config.get(CoapConfig.BLOCKWISE_STRICT_BLOCK1_OPTION);
 		strictBlock2Option = config.get(CoapConfig.BLOCKWISE_STRICT_BLOCK2_OPTION);
-
+		reuseToken = config.get(CoapConfig.BLOCKWISE_REUSE_TOKEN);
 		healthStatusInterval = config.get(SystemConfig.HEALTH_STATUS_INTERVAL, TimeUnit.MILLISECONDS);
 
 		enableAutoFailoverOn413 = config.get(CoapConfig.BLOCKWISE_ENTITY_TOO_LARGE_AUTO_FAILOVER);
@@ -761,6 +773,46 @@ public class BlockwiseLayer extends AbstractLayer {
 	}
 
 	/**
+	 * Get outer response to pass to application.
+	 * 
+	 * The outer response matches to initial application request.
+	 * 
+	 * @param exchange exchange
+	 * @param response actual response
+	 * @return outer application response
+	 * @since 3.8
+	 */
+	private Response getOuterResponse(Exchange exchange, Response response) {
+		// check, if response is for original request
+		if (exchange.getRequest() != exchange.getCurrentRequest()) {
+			// prepare the response as response to the original request
+			Response outerResponse = new Response(response.getCode());
+			// adjust the token using the original request
+			outerResponse.setToken(exchange.getRequest().getToken());
+			if (exchange.getRequest().getType() == Type.CON) {
+				outerResponse.setType(Type.ACK);
+				// adjust MID also
+				outerResponse.setMID(exchange.getRequest().getMID());
+			} else {
+				outerResponse.setType(Type.NON);
+			}
+			outerResponse.setSourceContext(response.getSourceContext());
+			outerResponse.setPayload(response.getPayload());
+			outerResponse.setOptions(response.getOptions());
+			outerResponse.setApplicationRttNanos(exchange.calculateApplicationRtt());
+			Long rtt = response.getTransmissionRttNanos();
+			if (rtt != null) {
+				outerResponse.setTransmissionRttNanos(rtt);
+			}
+			exchange.setResponse(outerResponse);
+			return outerResponse;
+		} else {
+			exchange.setResponse(response);
+			return response;
+		}
+	}
+
+	/**
 	 * Invoked when a response has been received from a peer.
 	 * <p>
 	 * Checks whether the response
@@ -800,32 +852,7 @@ public class BlockwiseLayer extends AbstractLayer {
 				default:
 				}
 
-				// check, if response is for original request
-				if (exchange.getRequest() != exchange.getCurrentRequest()) {
-					// prepare the response as response to the original request
-					Response resp = new Response(response.getCode());
-					// adjust the token using the original request
-					resp.setToken(exchange.getRequest().getToken());
-					if (exchange.getRequest().getType() == Type.CON) {
-						resp.setType(Type.ACK);
-						// adjust MID also
-						resp.setMID(exchange.getRequest().getMID());
-					} else {
-						resp.setType(Type.NON);
-					}
-					resp.setSourceContext(response.getSourceContext());
-					resp.setPayload(response.getPayload());
-					resp.setOptions(response.getOptions());
-					resp.setApplicationRttNanos(exchange.calculateApplicationRtt());
-					Long rtt = response.getTransmissionRttNanos();
-					if (rtt != null) {
-						resp.setTransmissionRttNanos(rtt);
-					}
-					exchange.setResponse(resp);
-					upper().receiveResponse(exchange, resp);
-				} else {
-					upper().receiveResponse(exchange, response);
-				}
+				upper().receiveResponse(exchange, getOuterResponse(exchange, response));
 				return;
 			}
 
@@ -1039,11 +1066,9 @@ public class BlockwiseLayer extends AbstractLayer {
 				LOGGER.debug("{}Block1 followed by Block2 transfer", tag);
 			} else {
 				// All request blocks have been acknowledged and we have
-				// received a
-				// response that does not need blockwise transfer. Thus, deliver
-				// it.
-				exchange.setResponse(response);
-				upper().receiveResponse(exchange, response);
+				// received a response that does not need blockwise transfer.
+				// Thus, deliver it.
+				upper().receiveResponse(exchange, getOuterResponse(exchange, response));
 			}
 		}
 	}
@@ -1058,8 +1083,10 @@ public class BlockwiseLayer extends AbstractLayer {
 				int blockSzx = Math.min(response.getOptions().getBlock1().getSzx(), preferredBlockSzx);
 				nextBlock = status.getNextRequestBlock(blockSzx);
 
-				// we use the same token to ease traceability
-				nextBlock.setToken(response.getToken());
+				if (reuseToken) {
+					// we use the same token to ease traceability
+					nextBlock.setToken(response.getToken());
+				}
 				nextBlock.setDestinationContext(status.getFollowUpEndpointContext(response.getSourceContext()));
 
 				LOGGER.debug("{}sending (next) Block1 [num={}]: {}", tag, nextBlock.getOptions().getBlock1().getNum(),
@@ -1152,8 +1179,7 @@ public class BlockwiseLayer extends AbstractLayer {
 
 			if (response.isNotification()) {
 				// We have received a notification for an observed resource that
-				// the
-				// application layer is no longer interested in.
+				// the application layer is no longer interested in.
 				// Let upper layers decide what to do with the notification.
 				upper().receiveResponse(exchange, response);
 			}
@@ -1253,7 +1279,8 @@ public class BlockwiseLayer extends AbstractLayer {
 			 * a different KeyToken in exchangesByToken, which is cleaned up
 			 * with the CleanupMessageObserver above.
 			 */
-			if (!response.isNotification()) {
+			if (reuseToken && !response.isNotification()) {
+				// we use the same token to ease traceability
 				block.setToken(response.getToken());
 			}
 
--- a/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
+++ b/californium-core/src/test/java/org/eclipse/californium/core/test/lockstep/BlockwiseClientSideTest.java
@@ -844,7 +844,7 @@ public class BlockwiseClientSideTest {
 		server.sendResponse(ACK, CONTINUE).loadBoth("B").block1(1, true, 128).go();
 
 		server.expectRequest(CON, PUT, path).storeBoth("B").block1(2, false, 128).payload(reqtPayload, 256, 300).go();
-		server.sendResponse(ACK, CHANGED).loadBoth("B").go();
+		server.sendResponse(ACK, CHANGED).loadBoth("B").block1(2, false, 128).go();
 
 		Response response = concurrentRequest.waitForResponse(ERROR_TIMEOUT_IN_MS);
 		assertThat(response, is(notNullValue()));
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityContextLayer.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityContextLayer.java
@@ -233,9 +233,9 @@ public class ObjectSecurityContextLayer extends AbstractLayer {
 		// Handle incoming OSCORE responses that have been re-assembled by the
 		// block-wise layer (for outer block-wise). If a response was not
 		// processed by OSCORE in the ObjectSecurityLayer it will happen here.
-		boolean outerBlockwise = exchange.getCurrentResponse() != null
-				&& exchange.getCurrentResponse().getOptions().hasBlock2()
-				&& ctxDb.getContextByToken(exchange.getCurrentResponse().getToken()) != null;
+		Response rawResponse =  exchange.getCurrentResponse();
+		boolean outerBlockwise = rawResponse.getOptions().hasBlock2()
+				&& ctxDb.getContextByToken(rawResponse.getToken()) != null;
 		if (outerBlockwise) {
 
 			LOGGER.debug("Incoming OSCORE response uses outer block-wise");
@@ -270,8 +270,6 @@ public class ObjectSecurityContextLayer extends AbstractLayer {
 			if (exchange.getRequest().isObserveCancel()) {
 				ctxDb.removeToken(response.getToken());
 			}
-
-			super.receiveResponse(exchange, response);
 		}
 
 		super.receiveResponse(exchange, response);
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ObjectSecurityLayer.java
@@ -129,11 +129,19 @@ public class ObjectSecurityLayer extends AbstractLayer {
 				// Handle outgoing requests for more data from a responder that
 				// is responding with outer block-wise. These requests should
 				// not be processed with OSCORE.
-				boolean outerBlockwise = request.getOptions().hasBlock2() && exchange.getCurrentResponse() != null
-						&& ctxDb.getContextByToken(exchange.getCurrentResponse().getToken()) != null;
-				if (outerBlockwise) {
-					super.sendRequest(exchange, req);
-					return;
+				if (request.getOptions().hasBlock2() && exchange.getCurrentResponse() != null) {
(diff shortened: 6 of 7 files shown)


boaks:
This PR is changing the behavior of Californium within RFC7959.

Californium was using the same token for blockwise transfers in order to ease traceability.

[attacks-on-coap](https://github.com/core-wg/attacks-on-coap) shows the downside of this. Even if for ACKs (piggy-backed-responses) Californium also uses the MID, using different tokens is the most common practice.

If someone detects trouble with it, it's possible to switch it back using 

"COAP.BLOCKWISE_REUSE_TOKEN=true"

SeppoTakalo:
I am not an CoAP expert, but I think this change has introduced problems, without fixing the potential repply attack.

## In summary:
I claim that this change broke Blockwise transfers that might require you to maintain a state. I think this would be the  case if the targeted resource might change between queries. If token is not kept the same, you cannot ensure integrity of the content as you cannot distinct between queries. Etag option would ensure that you have received a same representation of the content, but you cannot really use that in the GET query as it is generated by the server.

## Resoning
I believe this change is meant to fix potential [Response Delay and Mismatch Attack](https://core-wg.github.io/attacks-on-coap/draft-ietf-core-attacks-on-coap.html#name-the-response-delay-and-mism) described in Attacks on CoAP page.

However, I think the approach to fix the issue is wrong. As explained in the link above, the attack requires that attacker can guess the re-use of the token, in order to repply. And to fix the issue, you should use better random source to generate tokens, which I believe is not an issue here. But this change did not do anything with token generations. So if there was issue, it is still there. (I'm not suggesting there was a problem).

In blockwise transfer, if the server is implemented in stateless manner, this approach here is working. However, each query of a next block, have now no relationship to a previous query. They are completely independent. So even if server is implemented to maintain a state it cannot work anymore. I'll explain why this is a problem in a bit..

In [RFC7959: Section 3.4](https://datatracker.ietf.org/doc/html/rfc7959#section-3.4) there are few examples of usage of tokens. If you follow the example after "Retrieval of remaining blocks" note in the first sequence diagram, you notice that GET queries of remaining block keeps using the same token. It is not very clear there in RFC, but my assumption here is, that as long as you are requesting pieces of the same block, you should retain the same token.

If you use the same token for the whole blockwise transmission, it does not danger you to the reply attack, because that token exist only once. You only query one block once using the same token. Then next block uses the same token, but different block number. If your token algorithm works correctly, you are not going to use same token ever when requesting the same block for the same resource. You you do repply a packet during the blockwise transaction, the requesting client will just deal that block as a dublicate.

I have an example in my mind why this token rotation is problematic during the blockwise. Lets imagine that we have a resource called "/camera". Every time you do a GET request for it, it gives you a still image. But every request give a new still image from a live feed. So how can you do a blockwise transfer of a image, if you cannot tell which requests belong together? Every block would be a block from a new image, not from the same that started the transaction. So, this token-rotation only works with stateles servers. I breaks things if there is state involved.

The problem I'm facing, is LwM2M related. I'm using Leshan server which uses this library and after an update, things started to break. In LwM2M, you could send GET requests that might result more that one resource to be packet into a payload. If, for example TLS, SenML JSON/CBOR are used, the payload might contain lots of resources. This means that when first request arrives, the payload is formed, and kept in the memory until all is send. Now if we  do this stateless, it means that on every request we form a new paylod, split it, and send one block of it. How can you ensure that all the blocks are from similar payload? For example, if one of the resource is a "timestamp" that changes value on every second (or ms). Then every time you form a payload, it will be different than last time.

Last problem, which clearly broke Leshan, is that after this change, LwM2M Composite-Read does not work anymore. In composite-read, the payload of GET request contains a list of resources that should be put into the response payload. In next query, Leshan is not sending the paylod anymore, just the indicator to get Block N=1. But if your token changes, there is no indication that where your request should be targeted?
Here is the example of the first request:
Screenshot from 2023-08-25 13-32-30
Now here is the second request:
Screenshot from 2023-08-25 13-33-32
As you can see, there is no payload, and the token is different. How am I supposed to tell that this request is supposed to continue the previous request?

boaks:
Maybe just as very first information:

- this PR make the reuse of token in a blockwise transfer configurable. You may configure the old behavior or the new, see [BLOCKWISE_REUSE_TOKEN](https://github.com/eclipse-californium/californium/blob/main/californium-core/src/main/java/org/eclipse/californium/core/config/CoapConfig.java#L621-L635).  The PR changes the default behavior, but it's possible to overwrite that default application specific, if that's required.

- this PR doesn't fix the vulnerabilities of plain CoAP. If there are vulnerabilities left for encrypted CoAP is then an question, which may get different answers. It may be that implementing RFC 9175 improves that, but for now, I don't see that soon happens. 

I will try to provide an answer  for the points in your issue, but I will need some time. 
In the past, several implementations try to use "protocol artifacts" to make things possible, which are just excluded by RFC7959, including using etags or reuse tokens. 
For more general discussion about CoAP (RFC 7252, 7641, 7959 ...) our [Wiki](https://github.com/eclipse-californium/californium/wiki/Links-to-CoAP-or-DTLS-1.2-research-information) contains a couple of useful links, here the [IETF core mailing-list](https://mailarchive.ietf.org/arch/browse/core/) and [IETF Constrained Application Protocol (CoAP): Corrections and Clarifications](https://github.com/core-wg/corrclar) will be a good choice.

boaks:
> I claim that this change broke Blockwise transfers that might require you to maintain a state. I think this would be the case if the targeted resource might change between queries. If token is not kept the same, you cannot ensure integrity of the content as you cannot distinct between queries. Etag option would ensure that you have received a same representation of the content, but you cannot really use that in the GET query as it is generated by the server.

OK, that beams back ... as I already wrote, with the years a lot of approaches have been implemented, but they change the trade-offs chosen in [RFC7959](https://datatracker.ietf.org/doc/rfc7959/). Same approaches have also be applied to [RFC7641](https://datatracker.ietf.org/doc/rfc7641/), when some want to use CON notifies to achieve a "value stream without gaps". In my sum: that may be all very nice ideas, but the place to discus is the [IETF core mailing-list](https://mailarchive.ietf.org/arch/browse/core/). Alternatively, we added such stuff with a configuration flag, e.g. [BLOCKWISE_STRICT_BLOCK1_OPTION and BLOCKWISE_STRICT_BLOCK2_OPTION](https://github.com/eclipse-californium/californium/blob/main/californium-core/src/main/java/org/eclipse/californium/core/config/CoapConfig.java#L565-L602). 

To your claim:

[RFC7959 - 2.4 Using the Block2 Option](https://www.rfc-editor.org/rfc/rfc7959.html#page-13) bottom of the page:

> The Block2 Option provides no way for a single endpoint to perform
>    multiple concurrently proceeding block-wise response payload transfer
>    (e.g., GET) operations to the same resource.  This is rarely a
>    requirement, but as a workaround, a client may vary the cache key
>    (e.g., by using one of several URIs accessing resources with the same
>    semantics, or by varying a proxy-safe elective option).

Basically, that means, if the resource is changing, the etag is changing. Old transfers are detected by the changed etag and are canceled and the new presentation of the resource gets available for download. That affects also RFC7641. Alternative definitions would cause the server to allocate (much?) more memory. If I remember well, it's now a coupe of years, when Californium (miss)used the etag for "multiple concurrently". We stopped with that (miss)used. There is no configuration value available to enable that again.

boaks:
> In [RFC7959: Section 3.4](https://datatracker.ietf.org/doc/html/rfc7959#section-3.4) there are few examples of usage of tokens. If you follow the example after "Retrieval of remaining blocks" note in the first sequence diagram, you notice that GET queries of remaining block keeps using the same token. It is not very clear there in RFC, but my assumption here is, that as long as you are requesting pieces of the same block, you should retain the same token.

[RFC7959: Section 3.4](https://www.rfc-editor.org/rfc/rfc7959.html#page-27)

This shows not a GET block2 transfer, instead it shows a observer & get.

The GET uses token 0xfb, and so the first response (etag 6f00f38e). The next response/notify (etag 6f00f392) also uses that token 0xfb, as defined by RFC7641. But the follow-up request for that notify are then using a different token 0xfc. Finally the RFC mentions:

>  (Note that the choice of token 0xfc in this example is arbitrary;
>    tokens are just shown in this example to illustrate that the requests
>    for additional blocks cannot make use of the token of the Observation
>    relationship.  As a general comment on tokens, there is no other
>    mention of tokens in this document, as block-wise transfers handle
>    tokens like any other CoAP exchange.  As usual, the client is free to
>    choose tokens for each exchange as it likes.)    

If the client is free to choose tokens, a server can't bind state to it!

boaks:
> The problem I'm facing, is LwM2M related. I'm using Leshan server which uses this library and after an update, things started to break.

Just to ensure, the token is the root-cause, could you please test that with `COAP.BLOCKWISE_REUSE_TOKEN=true`?

Did you already open an issue in Leshan (I haven't found that), maybe changing the default value specific for the lwm2m application helps. 

> But if your token changes, there is no indication that where your request should be targeted?

That's not defined by the token, that's defined by the the clients-identity and the  coap-options, mainly the uri and uri-query. There is a second PR #2161, which adds the message code also to that.

jvermillard:
@SeppoTakalo I confirm this change in behavior breaks Zephyr LWM2M clients and Leshan, I solved it 5 months ago by configuring BLOCKWISE_REUSE_TOKEN to true on my servers. I should have opened a zephyr ticket or sent a discord message

boaks:
@jvermillard 

I'm wondering. As I wrote in the leshan issue:
If californium is the CoAP client, then this may have caused the failure.
If californium is the CoAP server, then I don't get it, why that should have caused an issue.

Do you remember the case, when you observed the error?

sbernard31:
I just take a quick look at this, so I could have missed a lot.

But concerning : 
> Last problem, which clearly broke Leshan, is that after this change, LwM2M Composite-Read does not work anymore. In composite-read, the payload of GET request contains a list of resources that should be put into the response payload. In next query, Leshan is not sending the paylod anymore, just the indicator to get Block N=1. But if your token changes, there is no indication that where your request should be targeted?

Note that Composite-Read is a FETCH on `/` (I mean all read-composite request target same URI which is `/`)
"What should be returned in the response' is driven by the FETCH payload. (which is always the case with FETCH not specific to LWM2M)

So there is maybe a potential issue here.   
Not even sure to know what should be the right behavior, I mean : 
- does a FETCH using block2 should send payload for each block request ?
- If yes what happened when you use block1 and block 2 with FETCH request.

And for more fun, FETCH can also be used with observe. (see : https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/528)

boaks:
Yeep, some of those questions are already pending on [IETF Constrained Application Protocol (CoAP): Corrections and Clarifications](https://github.com/core-wg/corrclar).

If you like, I will create a PR for leshan to change the default for the token reuse in order to keep the old behavior until the IETF Core clarifies the usage of the token.

jvermillard:
FYI openthread is not affected: https://github.com/openthread/openthread/issues/7976

sbernard31:
My guess :  we found some issue mainly with **FETCH and blockwise which could take long time before to be solve.**
So this is not only a LWM2M / Leshan issue, this is also a CoAP/Californium issue.

So I guess the question are : 

1) does `BLOCKWISE_REUSE_TOKEN=false` really help to solve [Response Delay and Mismatch Attack](https://core-wg.github.io/attacks-on-coap/draft-ietf-core-attacks-on-coap.html#name-the-response-delay-and-mism) ?
(Personnally, I don't know but I see that at least @SeppoTakalo seems not to be sure of that)

2) **IF** we think it does not OR we are not very sure, then maybe better to also move Californium to `BLOCKWISE_REUSE_TOKEN=true` as default. 

3) **ELSE IF** we think it helps to solve this attacks, let's keep `BLOCKWISE_REUSE_TOKEN=false` as default behavior but in this case, all users which are using FETCH with blockwise should use `BLOCKWISE_REUSE_TOKEN=false` and so there will be vulnerable to [Response Delay and Mismatch Attack](https://core-wg.github.io/attacks-on-coap/draft-ietf-core-attacks-on-coap.html#name-the-response-delay-and-mism), so that sounds not so good too.

(changing or not the default behavior in Leshan depends a lot of answers to questions  above :point_up:, so we can see that after :slightly_smiling_face: )

sbernard31:
About fetch and blockwise : 
- https://github.com/core-wg/corrclar/issues/27#issuecomment-1646655627
and maybe : 
- https://github.com/core-wg/corrclar/issues/28
- https://github.com/core-wg/corrclar/issues/18

boaks:
I'm aware of those corrclar issues.

FMPOV, if the client is defined to be free to choose the token, then no server state could be related to.

If some new RFC seems to be undefined, I would rather wait on their clarification before manifest, the choose a token is required in order to have some of the new stuff working. I think, it doesn't help, if that is done in advance of the answer. I will try to contact Jon from libcoap in order to get his opinion, maybe also Olaf (libcoap).

cabo:
I have been alerted to the existence of this discussion.
If I understand the problem correctly (correlating requests), the solution is RFC 9175 Request-Tag.
I don't think the approach "we don't know when that will be implemented, so instead we invent our own protocol" (giving the token semantics it does not have) will help.
But maybe I don't fully understand the discussion yet.
Instead of everybody scrambling to implement the unspecified deviant protocol, wouldn't it be easier if everyone implemented RFC 9175 Request-Tag?

boaks:
If RFC 9175 solves the issue, I think LwM2M needs to consider that.

boaks:
> wouldn't it be easier if everyone implemented RFC 9175 Request-Tag?

FMPOV, the question is, who needs it, and who will implement it.

The misuse of the token, caused by the behavior of Californium before 3.8 is already implemented.
That makes the idea of document the misuse as Interim solution attractive.

sbernard31:
> a client is free to do so. If a coap-server implementation breaks using different tokens it's not a compliant implementation.

I understand that too.

> it breaks only not compliant devices.

FMPOV, This is what we need to clarify, if this is true. There is nothing to change in Californium OR Leshan.

> To keep not compliant devices running, would makes it hard to change anything. 

I agree

> What I usually try is to add a configuration, to switch back to the old behavior, as here.

Some thought about it, from Leshan Project : https://github.com/eclipse-leshan/leshan/wiki/How-Leshan-should-behave-with-Non-Compliant-Implementations-%3F

cabo:
On 28. Aug 2023, at 11:14, Simon ***@***.***> wrote:
> 
> It seems some implementation decide to use that token to identify the "block exchange”.

The problem is that this deviant behavior is not only not interoperable, it creates an incentive for other implementations to cater to this behavior (which is essentially the discussion we are having here for Californium), effectively forking the CoAP protocol.
There are good reasons to limit token re-use (see Section 4.2 of RFC 9175), so clients need to be free to implement appropriate strategies; nurturing implementations that get in the way of that is not a good idea.

To the contrary, to avoid turning CoAP into soup, we should be deliberately *not* cater for such implementations, to create pressure that the deviant behavior gets fixed there.

Grüße, Carsten

sbernard31:
> I would prefer to move the discussion to a separate issue to have it more transparent for others.

I don't know, if following comments better fit here or at #2168.
If feel this is still about @SeppoTakalo issue so I decide to put it here. (feel free to move it or some part of it to a better place)

About rfc : 

1. a blockwise exchange is identify by "foreign peer identity" + requested resource. Is that correct ? 
2. this imply blockwise doesn't really support concurrent transfer for same peer/resource.
 >   The Block1 Option provides no way for a single endpoint to perform
   multiple concurrently proceeding block-wise request payload transfer
   (e.g., PUT or POST) operations to the same resource

 >   The Block2 Option provides no way for a single endpoint to perform
   multiple concurrently proceeding block-wise response payload transfer
   (e.g., GET) operations to the same resource.  This is rarely a
   requirement, but as a workaround, a client may vary the cache key
   (e.g., by using one of several URIs accessing resources with the same
   semantics, or by varying a proxy-safe elective option).


About reporting problem : 

1. For GET,  there is only a problem for concurrent access to same resource (same URI), right  ? 
The solution for that is using [etag-request](https://www.rfc-editor.org/rfc/rfc9175.html#name-the-request-tag-option) ? 
(This sounds like a new feature request not a bug/regression)

2. For FETCH, 
   2.1. this is pretty much the same, except that maybe "same resource" is more ambiguous.. Does it means same "targeted URI" or same "targeted URI + payload".  Currently Californium [doesn't send payload](https://github.com/eclipse-californium/californium/issues/2168). I'm not sure this is a killer feature to be able to do several FETCH on same URI + different payload. I feel doing concurrent request on same peer is already not a  so good idea.
  2.2. a stateless server implementation also need the payload in each FETCH block. maybe the real blocking issue ?

(more details about drawback or unclear details at : https://github.com/eclipse-californium/californium/issues/2168)

@cabo,  I think we need RFC clarification about FETCH+blockwise.   

@SeppoTakalo, Is it well resume or is there missing point ?

sbernard31:
> The problem is that this deviant behavior is not only not interoperable, it creates an incentive for other implementations to cater to this behavior

I totally understand that's what I try to support more generally at : https://github.com/eclipse-leshan/leshan/wiki/How-Leshan-should-behave-with-Non-Compliant-Implementations-%3F

boaks:
> I don't know, if following comments better fit here or

My point with that is, this is somehow "closed" and not easy visible to others.
Therefore a new, related issue, makes it easier for others to get aware.

sbernard31:
Thinking more about it and It's hard to me to imagine it's possible to implement a stateless blockwise support for LWM2M Read-composite operation. Because you need to generate all payload before to be able to send it by block. 

So about, 
> Last problem, which clearly broke Leshan, is that after this change, LwM2M Composite-Read does not work anymore. In composite-read, the payload of GET request contains a list of resources that should be put into the response payload. In next query, Leshan is not sending the paylod anymore, just the indicator to get Block N=1. But if your token changes, there is no indication that where your request should be targeted?

I understand this is only an issue if you want to support several concurrent ReadComposite request from same peer. Not sure this is critical use case.

Or I miss something ?

(Nevertheless, maybe it exists other use cases where FETCH could be implemented in a stateless way :shrug:)

boaks:
> support several concurrent ReadComposite request

If that is a block2 blockwise transfer, the RFC7959 already states, that this is not supported.

As I cited above:

[RFC7959 - 2.4 Using the Block2 Option](https://www.rfc-editor.org/rfc/rfc7959.html#page-13) bottom of the page:

> The Block2 Option provides no way for a single endpoint to perform
>    multiple concurrently proceeding block-wise response payload transfer
>    (e.g., GET) operations to the same resource.

So, if the ReadComposites are addressing the same resource, then  concurrent operation is not supported. The term `resource` reflects here in my interpretation mainly the resource path.

sbernard31:
> The term resource reflects here in my interpretation mainly the resource path.

I have some doubt that maybe this could be interpreted as URI + payload.

So reading [RFC8132§2. FETCH Method](https://www.rfc-editor.org/rfc/rfc8132.html#section-2)  a bit more , I guess that same resource probably also means same URI (but not crystal clear to me), e.g. : 
> The CoAP FETCH method is used to obtain a representation of a
   resource, specified by a number of request parameters.  Unlike the
   CoAP GET method, which requests that a server return a representation
   of the resource identified by the effective request URI (as defined
   by [[RFC7252](https://www.rfc-editor.org/rfc/rfc7252)]), the FETCH method is used by a client to ask the server
   to produce a representation as described by the request parameters
   (including the request options and the payload) based on the resource
   specified by the effective request URI.  The payload returned in
   response to a FETCH cannot be assumed to be a complete representation
   of the resource identified by the effective request URI, i.e., it
   cannot be used by a cache as a payload to be returned by a GET
   request.

boaks:
The term is RFC7959 and that's the base for the implementation decision.

RFC8132 was later and never really considered. We added mainly just the message codes and some methods to override the fetch operation. FMPOV, it's still in discussion. Once that discussion concluded, the implementation will be adapted.

boaks:
PFC 2252 = RFC 7252?

mrdeep1:
> PFC 2252 = RFC 7252?
Oops - corrected

boaks:
Usage of RFC9175:
The pain is not only CoAP related, the issue caused an impact in LwM2M.
If the solution would be RFC 9175, then this would require to update LwM2M as well.
I'm not sure, who is still working on IETF/core and LwM2M. Hannes has changed the job, so I'm afraid, he will not be able to care.

@SeppoTakalo 

Are you still interested? Could you provide the infos about the zephyr implementation?

sbernard31:
Usage of RFC9175 : 

It solves the concurrent access issue.
But this doesn't solve the stateless implementation support because using Request-Tag means that you need to store a state at server side, right ?

cabo:
> Usage of RFC9175 :
> 
> It solves the concurrent access issue. But this doesn't solve the stateless implementation support because using Request-Tag means that you need to store a state at server side, right ?

Indeed, if we want to enable clients to do a FETCH without sending the entire request body again for every block, the server needs to keep state, corrclar 27/28.  (This gets ridiculous when Block1 is used with Block2, but still is an issue for Block2 only.)

Blockwise tries to be open both to stateless servers and servers that want to keep state; so if the client behavior needs to depend on which of these is the case, we'd need additional signaling.

mrdeep1:
> Just to be sure, you mean this is ridiculous to re-send all payload using block1 for each block2 request, right ?

For stateful servers, subsequent FETCH requests for the next Block2 do not need the FETCH data as the server can build a cache-key that contains the cachable options that points to the appropriate data that needs a Block2 slice returned. BUT, if there is a chance that there are multiple concurrent FETCH to the same resource, but with different FETCH data, then the request will need a Request-Tag per different FETCH data so that the server can include this in the cache-key to differentiate which data-set should be used for getting the appropriate slice.

For non-stateful servers, the entire FETCH request needs to be repeated so the server can re-generate the data and then send back the appropriate slice of the data based on the block size and requested block number.  This includes any Block1 set of transfers handling the FETCH data.  The server will need to maintain some sort of state during the Block1 transfers to be able to assemble the entire FETCH data (note that if there are Block1 transfers and there is any chance of concurrent FETCH, then you must use Request-Tag to differentiate between the discrete FETCHes so server can correctly assemble the Block1s during the transitional state assemble phase).

Note that if Observe is being used, the server will need to maintain something to be able to generate any unsolicited responses, and that if the client wants to explicitly de-register the Observe, this has to be the original register FETCH request (including data), with just the original Observe option updated to de-register.

mrdeep1:
>> Blockwise tries to be open both to stateless servers and servers that want to keep state; so if the client behavior needs to depend on which of these is the case, we'd need additional signaling.

> This additional signaling should ideally be part of [RFC8132](https://www.rfc-editor.org/rfc/rfc8132), right ? (I mean we should not need additional RFC ?)

Agreed. the client has no way of knowing that the server is stateless or not unless there is some sort of out of band knowledge, or the server signals (new/bis RFC) its capabilities.

chrysn:
I don't want to dogpile on here (as cabo already made the good points, and I hope that this option stays off by default, so that breakage is not silent), but I'd like to weigh in on the Zephyr side, providing more evidence points that their implementation *does* break when interacting with other implementations. I failed to find a reference to Zephyr's issue tracker, where the reliance on tokens in block-wise is discussed. Could you give me a pointer?

mrdeep1:
In the same way that Extended Tokens ([RFC8974 Section 2.2 Discovering Support](https://www.rfc-editor.org/rfc/rfc8974.html#section-2.2)) and Q-Block ([RFC9177 Section 4.1 Properties of the Q-Block1 and Q-Block2 Options](https://datatracker.ietf.org/doc/html/rfc9177#section-4.1)) test for functionality support in the server, I guess we could suggest something like (for stateless FETCH support) on the client
````
-> FETCH + Block2 option (num = 0, szx = 0) + data against resource URI client is going to do FETCH against
<- check response - is resource valid for FETCH etc.  If OK continue.
-> FETCH + Block2 option (num = 1, szx = 0)  with no data against same resource URI
<- check response  - ok - supports stateful, failure is stateless (or perhaps something else?)
````

**Comment by sbernard31** [comment]
>  I guess we could suggest something like (for stateless FETCH support) on the client

That sounds a good possible way.

**Comment by chrysn** [comment]
As Request-Tag has been thrown around here as a solution to what apparently was done with token matching before: A Request-Tag will only ever *distinguish* two requests. If two requests don't match in their block key (eg. have a different Uri-Path), the request tags will not magically make them match.

**Comment by boaks** [comment]
> magically

Yep. 
For me the outcome of this discussion is:

- I don't know, what is specified in LwM2M nor what is exactly implemented in the Zephyr lwm2m client.
- without that, the discussion is more theoretical, then it could help here with this specific "unsharp" issue.
- the feature this PR added, stop reusing the token on a blockwise transfer, is in line with the specification. It harms mainly non-compliant implementations. For short term work-around someone may configure Californium the reuse the token again, but for long- and midterm that kind of token usage **MUST** be omitted.

Everything else, details of the [RFC 8132](https://www.rfc-editor.org/rfc/rfc8132) or contribution of [RFC9175](https://www.rfc-editor.org/rfc/rfc9175) to a solution of this issue, is for me out of the scope of discussing the re-use of tokens in a blockwise transfer. If that stuff requires attention, then I think, a separate issue (as #2168) makes that work easier. 

Thanks to all, who contributed their knowledge, experience and opinions.

**Comment by sbernard31** [comment]
> I don't know, what is specified in LwM2M nor what is exactly implemented in the Zephyr lwm2m client.
> without that, without that, the discussion is more theoretical, ...

@boaks, about LWM2M (not Zephir), I tried to answer over the various discussions we had about it but I can try to summarize here :
 - FETCH issues raised here seems not specific to LWM2M FETCH usage. 
 - Etag seems not to be mentionned in LWM2M (https://github.com/eclipse-leshan/leshan/issues/1499#issuecomment-1693635749)
 - Knowing how looks like Read-Composite payload, it's hard to me to imagine this can be done in a stateless way.
 
 If you want to know more, there is not too much to read about it in LWM2M specification : 
 - [LWM2M-v1.1.1@core§6.3.8. Read-Composite Operation](https://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#6-3-8-0-638-Read-Composite-Operation)
 - [LWM2M-v1.1.1@transport§Table: 6.4.4.-1 Operation to Method and URI Mapping (Device Management & Service Enablement Interface)](https://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Transport-V1_1_1-20190617-A.html#Table-644-1-Operation-to-Method-and-URI-Mapping-Device-Management-and-Service-Enablement-Interface)
 
IF you don't know so much about LWM2M, you should at least try to understand : 
- LWM2M object tree, maybe reading [this](https://github.com/eclipse-leshan/leshan/wiki/Adding-new-objects) OR [this](https://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#7-0-7-Identifiers-and-Resources) can help you)
- and also that for FETCH : a LWM2M client acts as CoAP server and vice versa.

For people we know a bit about LWM2M, this looks like : 

You send a **FETCH** on `/` with a payload containing LWM2M nodes you want to read, e.g using ct=SENML-JSON: 
```
[{"n":"/3/0/0"},
{"n":"/3/0/9"},
{"n":"/1/0/1"}]
```
Then you can get an answer containing LWM2M nodes value in the payload, e,g. using ct=SENML-JSON : 
```
{"n":"/3/0/0", "vs":"Open Mobile Alliance"},
{"n" :"/3/0/9", "v":95},
{"n":"/1/0/1", "v":86400}]
```
In this example, we only ask for LWM2M Single resource but you can ask for object, object instance, any kind of resource, resource instance or even the whole LWM2M object tree using `"n":"/")

**About Request-Tag**, I just see that since LWM2M v1.1.x, the specification says : 
> The CoAP Request-Tag Option [CoAP_ERT] SHOULD be used to detect interchange of blocks between different blockwise requests to the same resource over unreliable transport.

boaks:
> I tried to answer over the various discussions 

Sure, my point is more:
The use of the token in CoAP seems to be clear now. So at least the discussion for this PR is done for me.
If there is more to discus, that should be done under a better title, but not here in this PR. this PR is about
being able to not reuse the token in a blockwise transfer.
And that's done for me.
And because this turned into a long discussion even without more comments by the original reporter, I don't plan to spend here more time in.

RomainPelletant:
For information, the related Zephyr PR is [here](https://github.com/zephyrproject-rtos/zephyr/pull/61858)
This discussion was so instructive to me so thanks for your time.

SeppoTakalo:
I have been away for a week, so I did not find time to comment. Sorry about that.

Thank you for all the valuable information here. It definitely looks like this falls in between not so properly defined behaviour of CoAP.
As pointed out, the Request-Tag seem to fix the FETCH/PATCH queries.

However, I still don't think its use is properly defined, and I'm mostly considering LwM2M usecase here, when the response payload might be generated, and each time it is generated it might be different. So on most cases, I need to maintain a state so I can ensure the integrity of the payload.
The RFC 9175 says:
> The Request-Tag option MUST NOT be present in response messages.

So clearly whoever initiates a GET query must already know that response might be a block-wise transfer and append the Request-Tag option. Or (worst case and against recommendation) append a Request-Tag into all queries. Otherwise, if you send a normal GET and the server splits it into multiple blocks and sends BLOCK N=0, then what do you do for the next block query? Do you generate a Request-Tag or not? If you generate a Request-Tag, this is actually a new transfer and you need to start block block N=0.

When does Californium add the Request-Tag into queries? Or is it implemented?

As I'm one of the contributor to Zephyr's LwM2M client, I can alter the behavior here. But going through this long discussion have not really helped. You seem to have concluded that re-using the token is wrong, but what should I then use to match the following GET query to a previous one, if I need to maintain a state?

boaks:
I'm still not sure, why it is considered, that a lwm2m-server uses concurrent GET/FETCH operations (Composite-Read). From my point of view, there is not that much benefit in that. If the server finishes the blockwise transfer before starting the next, the current RFC 7959 works pretty well. Using ETAG also ensures, that if the resource is changing during the transfer, the client gets noticed about that. 
It may be a topic, how the lwm2m-server "serializes" the requests, but I would not move then that question into concurrent blockwise operations.
(By the way PR #2161 enables to have concurrent FETCH and GET operations, even if I would serialize them also.)

sbernard31:
My understanding, 

>  You seem to have concluded that re-using the token is wrong, but what should I then use to match the following GET query to a previous one, if I need to maintain a state?

If no `request-tag` used, you can not have concurrent block transfer for same resource/same peer.
So you should match by peer identity / resource URI. 
If you face concurrent transfer, cancel previous one (I think this is the californium choice)  or reject second one.
[As explained](https://www.rfc-editor.org/rfc/rfc9175.html#section-3.3)  : 
_"As it has always been, a server that can only serve a limited number of block-wise operations at the same time can delay the start of the operation by replying with 5.03 (Service Unavailable) and a Max-Age indicating how long it expects the existing operation to go on, or it can forget about the state established with the older operation and respond with 4.08 (Request Entity Incomplete) to later blocks on the first operation."_

If `request-tag` is used you can match by peer Identity / resource URI / request-tag and you can support concurrent block transfer for same resource of same peer. 

(In all case, FETCH doesn't work with stateless implementation with current RFC states)

So unless I missed something request-tag will not solve your issue unless you really want to support concurrent block transfer. 

> When does Californium add the Request-Tag into queries? Or is it implemented?

Not implemented and It seems to be not planned : https://github.com/eclipse-californium/californium/issues/2174

> Or (worst case and against recommendation) append a Request-Tag into all queries. 

This seems to be the simple choice and the [choice made by libcoap](https://github.com/open-coap/java-coap/issues/65#issuecomment-1708653579) :

_"For libcoap, I took the decision that by default, Request-Tag is sent with every request (even if Block1 is not being used and Block2 was not defined) "just in case" there is a Block2 sized response. The client application can however disable the CoAP stack doing this, only sending (done by CoAP stack) Request-Tag if Block1 or Block2 were defined."_

But I agree this sounds not recommended by the RFC : 
- _"The Request-Tag option is only used in the request messages of block-wise operations."_
- _"... this means sending messages without Request-Tag options whenever possible "_

Even if that should work because : 
_"Note that Request-Tag options can be present in request messages that carry no Block options (for example, because a proxy unaware of Request-Tag reassembled them)."_

> Otherwise, if you send a normal GET and the server splits it into multiple blocks and sends BLOCK N=0, then what do you do for the next block query? Do you generate a Request-Tag or not? If you generate a Request-Tag, this is actually a new transfer and you need to start block block N=0.

I guess this is another working possible option, but I can understand that at first sight this could be considered as not ideal.

boaks:
> When does Californium add the Request-Tag into queries? Or is it implemented?

It's not implemented. For me concurrent blockwise transfers have no general benefit.
There maybe a benefit assuming that the blockwise resources are changing fast, but that is not too frequently the case. And also there you need to decide, if you really want to spend the bandwidth into transfer an old resource representation.

At least for now, Californium is unfortunately not a AI, which implements the stuff by importing RFCs on itself ;-).

So it requires some contribution.

In the years ago I had a paid job for doing so, since a year this is not longer the case. I currently try to limit the invested free time in open source to 4h a week. That's eaten up by answering questions and providing bugfixes.
I'm not sure, if that changes.At least, I don't see that. 
There is even one more consequence of that time shortage: 
Contribution will require also to do some "quality work", because this would otherwise also fall into that 4h.

chrysn:
> However, I still don't think its use is properly defined, and I'm mostly considering LwM2M usecase here, when the response payload might be generated, and each time it is generated it might be different. So on most cases, I need to maintain a state so I can ensure the integrity of the payload.

(Trying to only add what has not been said)

All the CoAP server can do to protect the integrity of the payload is to set an ETag (eg. a hash of the full body). It is the CoAP client that must then verify that all the ETags match.

Unless there are concurrent requests (in which case the client will set a Request-Tag), the server that receives a GET for FETCH request for Block2:1/-/.. (i.e. a non-initial one) does not know which "instance" of the larger request this comes for. But it doesn't have to: It will just pick the latest one it has (for the given client and the given set of non-block options), trusting that the client really doesn't do concurrent requests.

How long the server keeps that "instance" around will depend on its capabilities, but at any rate, the lookup criterion are the client's address and the relevant options. (7959 had no precise definition of "relevant" options; 9175 calls the criterion being "matchable").

PR #2097, opened 2022-12-19
Update license_header_template.

boaks:
Add SPDX-License-Identifier: (EPL-2.0 AND BSD-3-Clause)

### DIFF
--- a/license_header_template.txt
+++ b/license_header_template.txt
@@ -1,17 +1,16 @@
-/*******************************************************************************
- * Copyright (c) <year (e.g. 2020)> <(company) name> and others.
+/********************************************************************************
+ * Copyright (c) {date} Contributors to the Eclipse Foundation
  * 
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v2.0
- * and Eclipse Distribution License v1.0 which accompany this distribution.
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
  * 
- * The Eclipse Public License is available at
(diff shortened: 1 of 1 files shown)


waynebeaton:
Think of the SPDX expression from the consumer/adopter's perspective.

I believe in this case that the idea is that one can consume Eclipse Californium content under either the EPL-2.0 **OR** the EPL-1.0 (which is an instance of the BSD-3-Clause). 

So the expression should be `EPL-2.0 OR BSD-3-Clause`. The parentheses are not required (they're not wrong).

FYI, the ["Documentation" link](https://www.eclipse.org/projects/tools/documentation.php?id=iot.californium#copyright) on the PMI provides some example copyright headers.

boaks:
After reading the provided link, this requires to collect the current "copyrights" into the Notice file.
so, I guess, we better keep the old style and just add the SPDX line.

waynebeaton:
> After reading the provided link, this requires to collect the current "copyrights" into the Notice file. 

How so? The first option more-or-less exactly matches what you're proposing. 

Only the second option--which is consistently presented as as _alternative_--requires tracking of specific copyright holders via a NOTICE file. FWIW, _NOTICE file_ is intended in the abstract to mean some indication of notices; AFAIAC, the commit record service serves this purpose.

You may have noticed that none of our examples include the text "All rights reserved". This term is meaningless. My recommendation is that you remove it from your template. It's not necessarily wrong for it to be there, but it doesn't actually add anything.

boaks:
>  AFAIAC, the commit record service serves this purpose.

OK, that sounds like adding a "refer to git reflog" in Californium's NOTICE.md will do it.
I will revise the PR.

Issue #2099, opened 2022-12-21
Data regarding dropped records in DTLSConnector


Wisien999:
Hello,
I use Scandium as the DTLS server and I would like to know what exact record/datagram was dropped in `DTLSConnector` (or at least its source address/ConnectionID). Is there any existing mechanism which allows me to do so?

In case it currently is impossible to do this could we introduce such a functionality? Also, how would you inject it into the current implementation? I can add the PR with my suggested changes if needed.

boaks:
At least for messages from "valid peer addresses/connection ids",
[DatagramFilter.onMacError(Record record, Connection connection)](https://github.com/eclipse-californium/californium/blob/main/scandium-core/src/main/java/org/eclipse/californium/scandium/DatagramFilter.java#L50) reports that.
If more is required, e.g. dropped messages from unknown sources, then the `DatagramFilter` must be extended. Usually, adding just a method/function fails, revapi will report that as semantic versioning error. We usually go then for something as `AdvancedDatagramFilter`and merge the interfaces again on the next major release.

Wisien999:
Thank you for your response. Unfortunately, `DatagramFilter.onMacError(Record record, Connection connection)` won't work. I need to know every dropped datagram/record. So, more or less, in every place where `DtlsHealth.receivingRecord(true)` is called. Is something like this possible?

boaks:
>  Is something like this possible?

Sure. I would prefer to have the `DatagramFilter` extended as described above.
Please be aware, that using resources (CPU/RAM) for "unverified peers" are pretty much in danger of DoS. So be very careful.

Wisien999:
Thank you for your valuable advice and resources. 
What I can see records can be dropped in a lot of places. For example:
```java
DTLSConnector.processRecords(...) {
	...
	if (connection == null) {
		if (health != null) {
			health.receivingRecord(true);
		}
		...
	}
	...
}
```
or
```java
DTLSConnector.processRecords(...) {
	...
	if (record.getConnectionId() == null && !connection.equalsPeerAddress(record.getPeerAddress())) {
		if (health != null) {
			health.receivingRecord(true);
		}
		return;
	}
	...
}
```
or more.
[...]

PR #2100, opened 2022-12-26
Datagram drop callback

Wisien999:
Discussion ref: #2099 

I'd like to introduce injectable action on datagram/record drop with access to dropped object. As discussed in the issue two methods were prepared to be added to `DatagramFilter`.

### DIFF
--- a/element-connector/src/main/java/org/eclipse/californium/elements/util/LimitedRunnable.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/util/LimitedRunnable.java
@@ -94,7 +94,7 @@ public abstract class LimitedRunnable implements Runnable {
 	 * Handles {@code RejectedExecutionException}
 	 * @param ex the thrown exception
 	 */
-	public abstract void onError(RejectedExecutionException ex);
+	public void onError(RejectedExecutionException ex) {};
 
 	/**
 	 * Execute this job.
--- a/element-connector/src/main/java/org/eclipse/californium/elements/util/LimitedRunnable.java
+++ b/element-connector/src/main/java/org/eclipse/californium/elements/util/LimitedRunnable.java
@@ -91,6 +91,12 @@ public abstract class LimitedRunnable implements Runnable {
 	}
 
 	/**
+	 * Handles {@code RejectedExecutionException}
+	 * @param ex the thrown exception
+	 */
+	public abstract void onError(RejectedExecutionException ex);
+
+	/**
 	 * Execute this job.
 	 * 
 	 * @param executor executor to execute jobs.
@@ -101,6 +107,7 @@ public abstract class LimitedRunnable implements Runnable {
 			executor.execute(this);
 		} catch (RejectedExecutionException ex) {
 			onDequeueing();
+			onError(ex);
 			throw ex;
 		}
 	}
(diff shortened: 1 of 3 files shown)


scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java
```diff
@@ -1969,7 +2004,7 @@ public void run() {
 	 * @since 3.5
 	 */
 	@NoPublicAPI
-	protected boolean executeInbound(Executor executor, InetSocketAddress peer, LimitedRunnable job) {
+	protected boolean executeInbound(Executor executor, InetSocketAddress peer, LimitedRunnable job, Runnable onError) {
```
boaks, line 2007: Maybe, extending `LimitedRunnable`with an `onError(RejectedExecutionException ex)` callback is an alternative to a second `Runnable`. 
Or pass in the `Record` instead of the  second `Runnable`

scandium-core/src/main/java/org/eclipse/californium/scandium/DatagramFilterExtended.java
```diff
@@ -0,0 +1,40 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Bosch.IO GmbH and others.
```
boaks, line 2: Are you an employee of Bosch.IO? 

If not, please edit this according [license_header_template.txt](https://github.com/eclipse-californium/californium/blob/main/license_header_template.txt). if this is your own work not belonging to an employer, then just use your name. 

There is also a pending PR #2097 . If you like, you may chose also that. I plan to replace all header when start to work in the 4.0.

boaks:
Thanks a lot!

I would like to spend some time in testing it.
Reviewing, I also recognized, that there is the one or other "gap" related to the "running".
Please correct the license header in the `DatagramFilterExtended`.

boaks:
I see two more candidates, `dropReceivedRecord(Record record)` and `discardRecord(final Record record, final Throwable cause)`.

element-connector/src/main/java/org/eclipse/californium/elements/util/LimitedRunnable.java
```diff
@@ -90,6 +90,12 @@ public boolean isOverflown() {
 		return overflow;
 	}
 
+	/**
+	 * Handles {@code RejectedExecutionException}
+	 * @param ex the thrown exception
+	 */
```
boaks, line 96: Some polishing, please add

@since 3.8

scandium-core/src/main/java/org/eclipse/californium/scandium/DatagramFilterExtended.java
```diff
@@ -0,0 +1,39 @@
...
+package org.eclipse.californium.scandium;
+
+import org.eclipse.californium.elements.util.PublicAPIExtension;
+import org.eclipse.californium.scandium.dtls.Record;
+
+import java.net.DatagramPacket;
+
+/**
+ * Extension of DatagramFilter
+ */
```
boaks, line 25: Some polishing, please add a "." at the end of the first sentence and a

```
@since 3.8
```

Wisien999:
Thank you for noticing `dropReceivedRecord` and `discardRecord` I overlooked. I also added the callback to the end of `processApplicationDataRecord(final Record record, final Connection connection)` where `health.receivingRecord(true)` is not called but record is definitely dropped. I you think it should not be here I can remove it.

boaks:
LGTM

boaks:
I missed to ask for squashing.
I will just do this on my own ;-).

PR #2107, opened 2023-01-20
Explain GitHub security advisories

mrybczyn:
Add direct link to the form to report a new security issue in GitHub security advisories and explain how to access that form.

The goal is to make reporting even easier.

Signed-off-by: Marta Rybczynska <marta.rybczynska@eclipse-foundation.org>


boaks:
LGTM

PR #2113, opened 2023-02-01
coap-over-tcp: cancel all observation relations on send errors.

boaks:
Add caching of principal and session-id to TLS.
Update netty.io to 4.1.87.Final.


boaks:
See PR #2114

Issue #2109, opened 2023-01-24
Debug log for mismatching PSK Secret


Warmek:
Currently an unsuccessful handshake caused by mismatched PSK secrets isn't loged. Would it be possible to log such an event in symilar maner to how "unknown psk identity" is handled?

```java
AlertMessage alert = new AlertMessage(AlertLevel.FATAL, AlertDescription.UNKNOWN_PSK_IDENTITY);
	if (hostName != null) {
		throw new HandshakeException(
				String.format("No pre-shared key found for [virtual host: %s, identity: %s]", hostName,
						pskIdentity),
				alert);
	} else {
		throw new HandshakeException(
				String.format("No pre-shared key found for [identity: %s]", pskIdentity), alert);
	}
```
Handshaker.java (line 1164) in scandium

boaks:
> Currently an unsuccessful handshake caused by mismatched PSK secrets isn't loged. 

The point is, that this mismatch isn't detected on its own, its detected causing an MAC error.
PSK isn't like "username/password", which is usually sent to an already authenticated server in order to authenticate the client. Instead PSK just exchanges the identity and use the secret to generate the "master secret" as base for the session keys. A mismatch is then detected with a MAC failure of the Finish message.

There have been also a general discussion about security leaks in PSK [how to handle unknown identity and bad psk ?](https://mailarchive.ietf.org/arch/msg/tls/3sXyPNowGI1zn3qwTQr7ZWhIBMI/).

What may be possible, is add a note, when PSK is used and the received `Finish` causes a MAC error (server side, the client usually reach the timeout).

Generally, I would like to postpone this after release 3.8 (hope in the next 2-4 weeks) in order to put all such "diagnose messages/hints" onto an separate logger. Such details in the wrong eyes may harm more.

boaks:
Please check PR #2116, if that works for you.

PR #2116, opened 2023-02-14
Add hint about wrong PSK secret.

boaks:
Add logging message, if the handshake failure may be caused by a wrong PSK secret.

Issue: #2109


Warmek (approved):
That works for us. 
Thank you!

boaks:
I adapted it a little and reset the log level to debug.

PR #2145, opened 2023-06-17
Update bouncy castle.

boaks:
Use 1.74 for java 8 and newer.
Use 1.73 for java 7.


boaks:
See [Bouncy Castle, issue 1429](https://github.com/bcgit/bc-java/issues/1429)

boaks:
Java 7 issues are fixed in bc 1.75. Update dependency to that.

PR #2153, opened 2023-06-28
Experimentally relax the constraints of the exchange executor.

boaks:
Don't fail, when a null executor is passed in.
Still not supported and may fail later, including hard to find race conditions.


JimmyBaize (approved):
Thank you very much. Very thoughtful. So that I can just pass null in executor.
After the PR is merged, I will try to upgrade the version for verification.

boaks:
> After the PR is merged, I will try to upgrade the version for verification.

"upgrade the version" may be not that easy. See [MIGRATION_HINTS](https://github.com/eclipse-californium/californium/blob/main/MIGRATION_HINTS.md) for a list of details.

PR #2158, opened 2023-07-12
Minor tcp changes

boaks:


element-connector/src/main/java/org/eclipse/californium/elements/config/TcpConfig.java
```diff
@@ -97,7 +97,10 @@ public String getModule() {
 
 		@Override
 		public void applyDefinitions(Configuration config) {
-			config.set(TCP_WORKER_THREADS, 1);
+			final int CORES = Runtime.getRuntime().availableProcessors();
+			final int THREADS = CORES > 3 ? 2 : 1;
```
jvermillard, line 101: In californium-core/src/main/java/org/eclipse/californium/core/config/CoapConfig.java
We use `config.set(PROTOCOL_STAGE_THREAD_COUNT, CORES);`
Why having this if more than 3 => 2 or 1 if less logic?
In general a comment with an explanation of magic number is useful

jvermillard: Reusing those numbers is fine, but it's different from what the Java NIO TCP consensus thinks it's optimal.

Usually, the  consensus on TCP worker thread count (which reads the incoming buffer and decode the message, nothing more) is, at minimum, the number of core:

Apache mina uses: number of cores +1
https://github.com/apache/mina/blob/trunk/core/src/main/java/org/apache/mina/transport/nio/NioTcpServer.java#L82-L83 

Netty uses something more aggressive: 2* number of cores.
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/MultithreadEventLoopGroup.java#L40-L41

I would use the netty default (0, which lets netty decide) since it was agreed upon following Netty's generic TCP benchmarking and research, except if we have specific knowledge or requirements regarding CoAP TCP.

boaks: Thanks!
I will try, if switching to 0 as default works.

boaks: See #2162 

AFAIK, no one did recently such benchmarks and has provided the results. Without that it's simply hard to see, what the best number will be. The old value of 1 was for the most cases not that good.

PR #2165, opened 2023-07-24
Fix deprecation of LeastRecentlyUsedCache.Predicate.

boaks:
Introduce separate Filter instead and deprecated
LeastRecentlyUpdatedCache.Predicate.
Cleanup internal usage of deprecated config API.


sbernard31:
I didn't test it and only review it quickly. It seems to be a good short term solution waiting  for 4.0.0 clean-up

PR #2187, opened 2023-09-27
OSCORE: Link target attribute

rikard-sics:
This PR adds the link target attribute 'osc' for resources which are only accessible using OSCORE.

It also corrects the behavior of the _OSCoreResource_ when it is set to only be accessible using OSCORE, but receives a non-OSCORE request.

### DIFF
--- a/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
@@ -51,6 +51,7 @@ public class LinkFormat {
 	public static final String TITLE = "title";
 	public static final String OBSERVABLE = "obs";
 	public static final String LINK = "href";
+	public static final String OSCOREONLY = "osc"; // RFC 8613
 
 	// for Resource Directory
 	public static final String LIFE_TIME = "lt";
--- a/californium-core/src/main/java/org/eclipse/californium/core/server/resources/ResourceAttributes.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/server/resources/ResourceAttributes.java
@@ -252,6 +252,34 @@ public class ResourceAttributes {
 	}
 
 	/**
+	 * Marks the resource as only accessible using OSCORE.
+	 *
+	 * @since 3.10
+	 */
+	public void setOscoreOnly() {
+		findAttributeValues(LinkFormat.OSCOREONLY).setOnly("");
+	}
+
+	/**
+	 * Checks if the resource is only accessible using OSCORE.
+	 *
+	 * @return {@code true}, if only accessible using OSCORE
+	 * @since 3.10
+	 */
+	public boolean hasOscoreOnly() {
+		return hasAttribute(LinkFormat.OSCOREONLY);
+	}
+
+	/**
+	 * Marks the resource as not only accessible using OSCORE.
+	 *
+	 * @since 3.10
+	 */
+	public void clearOscoreOnly() {
+		attributes.remove(LinkFormat.OSCOREONLY);
+	}
+
+	/**
 	 * Replaces the value for the specified attribute with the specified value.
 	 * If another value has been set for the attribute name, it will be removed.
 	 * 
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreResource.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreResource.java
@@ -12,13 +12,13 @@
  * 
  * Contributors:
  *    Tobias Andersson (RISE SICS)
+ *    Rikard Höglund (RISE)
  *    
  ******************************************************************************/
 package org.eclipse.californium.oscore;
 
 import org.eclipse.californium.core.CoapResource;
 import org.eclipse.californium.core.coap.CoAP.ResponseCode;
-import org.eclipse.californium.core.coap.CoAP.Type;
 import org.eclipse.californium.core.network.Exchange;
 import org.eclipse.californium.core.coap.OptionSet;
 import org.eclipse.californium.core.coap.Response;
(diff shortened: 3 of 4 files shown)


rikard-sics:
This was another point brought up during the recent interop tests.

This functionality is in accordance with the OSCORE RFC Section 9:
https://datatracker.ietf.org/doc/html/rfc8613#section-9

Not sure if JUnit tests are also desired, if so I can add them.

californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
```diff
@@ -51,6 +51,7 @@ public class LinkFormat {
 	public static final String TITLE = "title";
 	public static final String OBSERVABLE = "obs";
 	public static final String LINK = "href";
+	public static final String OSCOREONLY = "osc";
```
boaks, line 54: Please add 

```
// RFC 8613
```

californium-core/src/main/java/org/eclipse/californium/core/server/resources/ResourceAttributes.java
```diff
@@ -251,6 +251,31 @@ public void clearObservable() {
 		attributes.remove(LinkFormat.OBSERVABLE);
 	}
 
+	/**
+	 * Marks the resource as only accessible using OSCORE.
+	 */
```
boaks, line 256: Please add:

```
@since 3.10
```

californium-core/src/main/java/org/eclipse/californium/core/server/resources/ResourceAttributes.java
```diff
@@ -251,6 +251,31 @@ public void clearObservable() {
...
+	public void setOscoreOnly() {
+		findAttributeValues(LinkFormat.OSCOREONLY).setOnly("");
+	}
+
+	/**
+	 * Checks if the resource is only accessible using OSCORE.
+	 *
+	 * @return true, if only accessible using OSCORE
```
boaks, line 264:
```
{@code true}
```

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreResource.java
```diff
@@ -34,6 +35,7 @@ public class OSCoreResource extends CoapResource {
 
 	public OSCoreResource(String name, final boolean isProtected) {
 		super(name);
+		getAttributes().setOscoreOnly();
```
boaks, line 38: Is there a relation betwenn "isProtected" and "Oscore Only"?
I searched for references to is Protected but could not find something.

rikard-sics:
Thanks for the feedback, I will take care of the points you mentioned.

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreResource.java
rikard-sics: Yes, great point, I missed that. At a minimum _setOscoreOnly()_ should only be invoked when _isProtected_ is set to true. Since _isProtected_ is the flag that is actually controlling if the resource is only accessible with OSCORE, it's not enough that the resource type is OSCoreResource.

I will take care of this.

californium-core/src/main/java/org/eclipse/californium/core/coap/LinkFormat.java
rikard-sics: Added in commit 4281708a39346675a1d6ee0f88f26bab0c902280

californium-core/src/main/java/org/eclipse/californium/core/server/resources/ResourceAttributes.java
rikard-sics: Done in commit e41bd07e2f4956c067eb4774a2edcee45e7e7a3a

californium-core/src/main/java/org/eclipse/californium/core/server/resources/ResourceAttributes.java
rikard-sics: Added both in commit e41bd07e2f4956c067eb4774a2edcee45e7e7a3a

cf-oscore/src/main/java/org/eclipse/californium/oscore/OSCoreResource.java
rikard-sics: Took care of only setting the OSCORE link target attribute if _isProtected_ is true in commit b9996e21d07b0c0bbe1ce45962ad54bf5cb03aff

rikard-sics:
I also took care of the issue of responding with RST to non-OSCORE requests to OSCORE-only resources in commit c1fb46aed9b8d684ffed895ee8da84bffe2924d5

If that looks good to you, we could keep it in this PR.

californium-core/src/main/java/org/eclipse/californium/core/server/resources/ResourceAttributes.java
```diff
@@ -251,6 +251,34 @@ public void clearObservable() {
...
+	 * @since 3.10
+	 */
+	public void setOscoreOnly() {
+		findAttributeValues(LinkFormat.OSCOREONLY).setOnly("");
+	}
+
+	/**
+	 * Checks if the resource is only accessible using OSCORE.
+	 *
+	 * @since 3.10
```
boaks, line 266: please exchange the order of since and return.

boaks:
Beside of the order of since and return, LGTM.
Please fix the order and squash the commits.

californium-core/src/main/java/org/eclipse/californium/core/server/resources/ResourceAttributes.java
rikard-sics: I have fixed this also now

rikard-sics:
> Beside of the order of since and return, LGTM.
> Please fix the order and squash the commits.

Sounds great. I have now fixed the order and squashed the commits.

PR #2188, opened 2023-09-28
Block malformed RST.

boaks:


boaks:
This PR will block sending malicious RST messages. Please fix such code before using an update with this PRR.
On receiving, `CoapConfig.STRICT_EMPTY_MESSAGE_FORMAT` may be used temporary to continue to process malicious RST messages, while the clients are fixing that issue.

PR #2197, opened 2023-11-30
Deprecate exposed logger and replace them by local instances.

boaks:
Redesign FilteredLogger to use names instead of exposing logger.


boaks:
In order to prepare for decouple californium from the slf4j version, all exposed loggers are now deprecated. The most will change their scope to `private` on the next major release. And a few will be removed, because they are not used at all.

PR #2212, opened 2024-01-26
Add Exchange as a parameter to the ObserveResource::getNotificationSequenceNumber

ashvayka:


boaks:
Just to be clear: did you read RFC7641?
I guess, you considerations about the grant of the "sequence numbers" doesn't match that.

PR #2213, opened 2024-01-26
Add Exchange as a parameter to the ObserveResource::getNotificationSequenceNumber

ShvaykaD:

### DIFF
--- a/californium-core/api-changes.json
+++ b/californium-core/api-changes.json
@@ -109,7 +109,7 @@
 			}
 		}
 	],
-	"3.10.0": [
+	"3.11.0": [
 		{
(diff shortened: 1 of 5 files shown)


boaks:
Let me try it again:
I assume, you don't comply with RFC7641.
If so, I would prefer to have such stuff clearly marked as "experimental/not compliant".
As different implementation idea:
If "handleGET/???" sets already a "observe option" in the response, "ObserveRelation" won't overwrite that for "success" responses.

That's a pretty small "behavior" change. It's the common way done for other stuff. The stack sets the things, if it's not already set.

Would that help you?

PR #2214, opened 2024-01-26
Added parameter to the ObserveResource::getNotificationSequenceNumber

ShvaykaD:

### DIFF
--- a/californium-core/api-changes.json
+++ b/californium-core/api-changes.json
@@ -108,6 +108,27 @@
 				]
 			}
 		}
+	],
+	"3.11.0": [
+		{
+			"extension": "revapi.differences",
+			"configuration": {
+				"ignore": true,
+				"differences": [
+					{
+						"code": "java.method.numberOfParametersChanged",
+						"old": "method int org.eclipse.californium.core.server.resources.ObservableResource::getNotificationSequenceNumber()",
+						"new": "method int org.eclipse.californium.core.server.resources.ObservableResource::getNotificationSequenceNumber(org.eclipse.californium.core.network.Exchange)",
+						"justification": "Redesign to support separate sequence numbers for different observe tokens"
+					},
+					{
+						"code": "java.method.numberOfParametersChanged",
+						"old": "method int org.eclipse.californium.core.CoapResource::getNotificationSequenceNumber()",
+						"new": "method int org.eclipse.californium.core.CoapResource::getNotificationSequenceNumber(org.eclipse.californium.core.network.Exchange)",
+						"justification": "Redesign to support separate sequence numbers for different observe tokens"
+					}
+				]
+			}
+		}
 	]
-	
 }
--- a/californium-core/src/main/java/org/eclipse/californium/core/CoapResource.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/CoapResource.java
@@ -376,7 +376,7 @@ public class CoapResource implements Resource, ObservableResource {
 	}
 
 	@Override
-	public int getNotificationSequenceNumber() {
+	public int getNotificationSequenceNumber(Exchange exchange) {
 		return notificationOrderer.getCurrent();
 	}
 
--- a/californium-core/src/main/java/org/eclipse/californium/core/observe/ObserveRelation.java
+++ b/californium-core/src/main/java/org/eclipse/californium/core/observe/ObserveRelation.java
@@ -403,7 +403,7 @@ public class ObserveRelation {
 	 * The first response will {@link #setEstablished()} the relation and
 	 * {@link ObservableResource#addObserveRelation(ObserveRelation)} it. If not
 	 * {@link #isCanceled()}, and the response {@link Response#isSuccess()},
-	 * {@link ObservableResource#getNotificationSequenceNumber()} will be set.
+	 * {@link ObservableResource#getNotificationSequenceNumber(Exchange)} will be set.
 	 * 
 	 * @param response response
 	 * @return current relation state.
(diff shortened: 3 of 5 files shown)


ShvaykaD:
Hi @boaks, thank you for highlighting RFC7641. We acknowledge that our approach to "sequence numbers" may not fully align with it. However, due to the specific requirements of our multi-tenant environment, we need to apply our changes to ensure distinct observe sequences for each device. We're keen to explore how we can achieve this within the framework of RFC7641 and welcome any suggestions you may have.

ShvaykaD:
@boaks, regarding your proposed solution involving the "observe option" in the response and the behavior of "ObserveRelation".

> Let me try it again:
I assume, you don't comply with RFC7641.
If so, I would prefer to have such stuff clearly marked as "experimental/not compliant".
As different implementation idea:
If "handleGET/???" sets already a "observe option" in the response, "ObserveRelation" won't overwrite that for "success" responses.

>That's a pretty small "behavior" change. It's the common way done for other stuff. The stack sets the things, if it's not already set.

>Would that help you?

We will analyze and test this method in our environment to see if it aligns with our requirements and constraints.

boaks:
```
if (!response.isNotification()) {
	response.getOptions().setObserve(resource.getNotificationSequenceNumber());
}
```

Similar as that. `setObserve` is used twice, please apply it to both.

boaks:
From many other examples, I don't think, that this makes any sense!
You didn't really provide the information, why that should be required.
I personally would prefer, if you document in your system, what you do and why.
In other cases, e.g. lwm2m "blockwise/token" it finally ends up in a mess.

boaks:
> In a multi-tenant environment (ThingsBoard), we use the same resource path (e.g.,/api/v1/device/attributes) that multiple devices use.

Even in a single-tenant environment, multiple devices may observe the same resource.

I guess, the multi-tenant requirement here is more, that this are "different resources" (per tenant) under the same path.
If that assumption is right, I guess, you split also the content of that tenant specific resource and therefore it should be easy possible to add there also the specific sequence number/observe option.

A more general alternative implementation would be to keep different `CoapResource` per tenant and use a customized `ServerMessageDeliverer.

In sum: if it's mainly about "different resources for different tenant under the same path", then I would recommend to not grant more about the sequence-number, except, that the "multi-tenant-function" enables to use tenant specific resources under a common path.

ashvayka:
Hi @boaks, The PR is related to the problem we encountered while trying to upgrade our code base from 3.0.0 to the latest release (3.10).  I will explain the situation at the end of this message. 

**The use case**

Let's go ahead and start with the use case in detail so you may understand our needs better. There are two aspects.
1. Let's assume we have 1M devices with different credentials that use plain CoAP (no DTLS).  We identify devices based on our credentials that we call access tokens (not coap observe tokens). So, to get updates from the server, the device needs to send observe request to the following resource:
```
coap://host/api/v1/$DEVICE_ACCESS_TOKEN/updates
```
2.  In a similar case, but for DTLS, we identify the device via certificate, and we don't have a device access token in the observed URI path:
```
coaps://host/api/v1/updates
```

Since we have a lot of devices and only one URI `/api/v1/updates`, we decided to create a "super" resource. See the [CoapTransportResource](https://github.com/thingsboard/thingsboard/blob/master/common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/CoapTransportResource.java) and add it to [CoapTransportService](https://github.com/thingsboard/thingsboard/blob/master/common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/CoapTransportService.java#L60). Please note that we handle all requests in this "super" resource instead of creating a lot of sub-resources because in DTLS, we don't have separate URI per each client device. 

**The blocker**

Previously, we have overwritten the [CoapResource::checkObserveRelation](https://github.com/thingsboard/thingsboard/blob/7776bdb0e89c37ee365b262b8d2d1b91c154b12a/common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/CoapTransportResource.java#L91C17-L91C37) to use a separate notification counter per each device+token pair (client IP + client port + observe token. see [getTokenFromRequest](https://github.com/thingsboard/thingsboard/blob/master/common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/CoapTransportResource.java#L369)

You have deprecated this method in the latest code, and we cannot overwrite and add our desired logic. That is why we submitted this PR, and we hope you can merge it so we can migrate to the newer version without forking the repo. 

P.S. The RFC 7641 does not explicitly forbid or disallow this practice. Please provide your comments if you disagree. I would like to avoid creating separate resources per each device since this may impact performance.

boaks:
> P.S. The RFC 7641 does not explicitly forbid or disallow this practice. 

As I wrote, I consider that you don't use the "same resource" but instead "different resources under the same path". It's not important how you implement that, the point is that the logic is that of "different resources".

>  I would like to avoid creating separate resources per each device since this may impact performance.

That are different things, if you really use different "CoapResource"s or implement that logic in a different way.

```
    @Override
    public void checkObserveRelation(Exchange exchange, Response response) {
        String token = getTokenFromRequest(exchange.getRequest());
        final ObserveRelation relation = exchange.getRelation();
        if (relation == null || relation.isCanceled()) {
            return; // because request did not try to establish a relation
        }
        if (response.getCode().isSuccess()) {
            if (!relation.isEstablished()) {
                relation.setEstablished();
                addObserveRelation(relation);
            }
            AtomicInteger state = clients.getNotificationCounterByToken(token);
            if (state != null) {
                response.getOptions().setObserve(state.getAndIncrement());
            } else {
                response.getOptions().removeObserve();
            }
        } // ObserveLayer takes care of the else case
    }
```

I haven't tested it, but I guess it will work straight forward to merge the "observe option" handling into the "processHandleGet" method, once the ObserveRelation keeps already provided observe options instead of overwriting it.

I currently prepare a PR with that function.

boaks:
Please check PR #2215 , if that works for you

It would require to adapt your implementation to set the:

``` 
AtomicInteger state = clients.getNotificationCounterByToken(token);
```
ahead.

ashvayka:
From the first look, this is exactly what we need. But I will confirm once we do additional tests. Thank you so much!

ShvaykaD:
Pr was closed since the [PR: Adapt ObserveRelation to keep provided observe options](https://github.com/eclipse-californium/californium/pull/2215) with an alternative solution satisfies our goals.

PR #2215, opened 2024-01-26
Adapt ObserveRelation to keep provided observe options.

boaks:


ashvayka:
From the first look, this is exactly what we need. But I will confirm once we do additional tests. Thank you so much!

boaks:
Your welcome.

ashvayka:
Hi @boaks, Thank you for the proposed changes. I confirm that those changes will help us implement our case.

boaks:
Does it require a minor release, 3.11 for you?

Usually I need to start and pass a "progress review" (once a year) in order to do new releases. So I guess, it will take at least 2 weeks.

ashvayka:
We plan to release ThingsBoard 3.7 within the next 3-4 months. So there is no rush, but we just need to know which version of the Californium we should use and when to expect it.

Issue #2252, opened 2024-06-20
Always a 5.02 response in ProxyHttpClientResource if http response contains no content-type header


basimons:
Hello,

After updating from californium 2.7.0 -> 3.12, I ran into a small issue.

I have a resource like this:

```java
@Slf4j
public class HttpClientResourceForwarder extends ProxyHttpClientResource {

    public HttpClientResourceForwarder (HttpProxyProperties props, MeterRegistry meterRegistry) {
        super(props.getCoapProxyVersionPath(), props.getCoapProxyVisible(), true,
            new Coap2ForwarderHttpTranslator(props, meterRegistry), props.getHttpScheme());
    }

    @Override
    public Resource getChild(String name) {
        // This ensures that all paths after the version path (/v0/**) end up at this Resource
        return this;
    }

    @Override
    public void handleRequest(final Exchange exchange) {
        log.debug("Received COAP proxy request: {}", exchange.getRequest());
        super.handleRequest(exchange);
    }
}

```

This worked fine in 2.7.0, but after upgrading I sometimes encounter an issue. This is caused when the server that the `ProxyHttpClientResource` does a request to, returns a response WITHOUT a content-type header, the response will always be a 5.02, with the message being "content type must not be null", which is set on org.eclipse.californium.proxy2.http.ContentTypedEntity#49.

IMO this is a bug, because getting a response without a Content-Type header is fine as long as the Content-Length is also 0.

Kind regards,

Bram

boaks:
I guess the culprit is

[CrossProtocolTranslator.setCoapPayload()](https://github.com/eclipse-californium/californium/blob/main/californium-proxy2/src/main/java/org/eclipse/californium/proxy2/http/CrossProtocolTranslator.java#L408-L409) .

If the "empty http content" is an empty array, then the check must be

```
if (payload != null && payload.length > 0)
```

Not sure, are you able to test that, if it fixes your issue?

boaks:
Maybe it's [ContentTypedEntity](https://github.com/eclipse-californium/californium/blob/main/californium-proxy2/src/main/java/org/eclipse/californium/proxy2/http/ContentTypedEntity.java#L48-L52), at least the error message indicates that.

Then a fix may be:

```
	public ContentTypedEntity(ContentType contentType, byte[] payload) {
		this.payload = payload != null && payload.length > 0 ? payload : null;
		if (contentType == null && this.payload != null) {
			throw new NullPointerException("content type must not be null, if payload is provided!");
[...]

PR #2253, opened 2024-06-20
Coap-Http-Cross-Prox: don't fail on missing content type, if no payload

boaks:
is provided as well.

Fix issue #2252


boaks:
See PR #2257

PR #2273, opened 2024-08-22
OSCORE: Fix incorrect CoAP method in plugtest client

rikard-sics:
Fixes a small issue where the incorrect CoAP method was used for one of the plugtests for OSCORE.

### DIFF
--- a/demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
+++ b/demo-apps/cf-plugtest-client/src/main/java/org/eclipse/californium/plugtests/PlugtestClient.java
@@ -779,7 +779,7 @@ public class PlugtestClient {
 
 		System.out.println("===============\nOSCORE 04");
 		System.out.println("---------------\nPOST /oscore\n---------------");
-		r = Request.newPut();
+		r = Request.newPost();
 		r.setPayload("non-empty");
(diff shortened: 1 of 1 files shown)


rikard-sics:
> LGTM

Great, yes it ready from my side too. Just a small fix.

Just to let you know, I plan to check the problem with OSCORE outer block-wise and the JUnit test that had to be ignored.

boaks:
Maybe we've discussed that in e-mails, PR #2161 required to disable [OSCoreOuterBlockwiseTest.testOuterBlockwiseGetProxyClientBW](https://github.com/eclipse-californium/californium/blob/main/cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreOuterBlockwiseTest.java#L309-L321), because the implementation used the original outer method for the follow up blockwise requests. Just in the case, your "on the fix".

rikard-sics:
> Maybe we've discussed that in e-mails, PR #2161 required to disable [OSCoreOuterBlockwiseTest.testOuterBlockwiseGetProxyClientBW](https://github.com/eclipse-californium/californium/blob/main/cf-oscore/src/test/java/org/eclipse/californium/oscore/OSCoreOuterBlockwiseTest.java#L309-L321), because the implementation used the original outer method for the follow up blockwise requests. Just in the case, your "on the fix".

Thanks, yes I recall we discussed this. Yes I will work to try to fix this issue, hopefully some time in the coming weeks.

PR #2284, opened 2024-09-29
Fix potential memory leak in InMemoryReadWriteLockConnectionStore

zakersimonjack:
During our stress test, I noticed that memory usage kept going up and didn't decrease significantly after GC.

f38a-4841-a4f1-7c6e0347faf8

Then I checked the classes in memory. The `Connection` class in connectionsByPrincipal never seems to get cleaned up. Finally I found `removeByPrincipal` seems to be missed in `onEviction` of `EvictionListener`.


boaks:
Thanks for reporting!

boaks:
My current plan is to release 3.13 on Wednesday, 9. October. (I will do the orga work for 3.13 and schedule tomorrow.)
Just in the case your stress test uncovers more, let me know.

Issue #2294, opened 2024-10-13
Duplicate requests not handled during OSCORE context rederivation


burrrrrr:
With the plugtest server app, during OSCORE context rederivation (Appendix B.2), the server doesn't appear to handle duplicate requests from the client (same token and MID). This is an issue in low availability areas where context rederivation will need to be restarted if a response is lost. If I'm not wrong, it is expected the server can handle multiple context rederivations from replayed requests in parallel.

boaks:
I'm not common with OSCORE.

Duplicate requests (same peer, same MID within EXCHANGE_LIFETIME) are handled according [RFC 7252 4.5 Message Deduplication](https://www.rfc-editor.org/rfc/rfc7252#section-4.5). With that I don't see, that there is something special to do.

@rikard-sics 

Does OSCORE change that RFC 7252 4.5?

boaks:
I got aware, that the `OSCoreUdpStack` puts the `ObjectSecurityLayer` below the `ReliabilityLayer`, which then prevents that from deduplicating requests in `ReliabilityLayer.receiveRequest`.  
Not sure, what is the intended behavior or best to fix that.
It may be possible, to change the order of the layer. 
Or to add a `if (request.isDuplicate())` check to the `ObjectSecurityLayer.receiveRequest`.

boaks:
I very first test showed, that exchanging the layers may work.

```
new BlockwiseLayer(tag, false, config, matchingStrategy),
new ObjectSecurityLayer(ctxDb),
CongestionControlLayer.newImplementation(tag, config)};
```

@burrrrrr 

would it be possible for you to test with that approach until we get a feedback from Rikard?

rikard-sics:
> would it be possible for you to test with that approach until we get a feedback from Rikard?

Yes my spontaneous comment without checking deeper is that swapping the orders of the layers makes sense

burrrrrr:
> I very first test showed, that exchanging the layers may work.
> 
> ```
> new BlockwiseLayer(tag, false, config, matchingStrategy),
> new ObjectSecurityLayer(ctxDb),
> CongestionControlLayer.newImplementation(tag, config)};
> ```
> 
> @burrrrrr
> 
> would it be possible for you to test with that approach until we get a feedback from Rikard?

@rikard-sics 

Great, that fixes the retransmission issue for me.

[...]

PR #2295, opened 2024-10-14
OSCORE: fix HMAC validation during context rederivation

burrrrrr:
This is related to #2294 but does not directly address it. 

I found that the server was always accepting duplicate requests as valid for request 2 during context rederivation. It looks like the expected HMAC output is compared against itself, so the received context ID from the request is not validated.

### DIFF
--- a/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
+++ b/cf-oscore/src/main/java/org/eclipse/californium/oscore/ContextRederivation.java
@@ -317,15 +317,16 @@ public class ContextRederivation {
 			// Generate HMAC output using S2
 			byte[] hmacOutput = performHMAC(ctx.getContextRederivationKey(), contextS2);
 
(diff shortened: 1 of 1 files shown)


boaks:
Thanks for your contribution.
Please read the [CONTRIBUTION](../../CONTRIBUTION.md) instructions and sign the ECA.
In general that's required to "grant" others the usage of your contribution.

boaks:
As I wrote in your issue, I'm not common to OSCORE.

@rikard-sics 
Could you please have a look this?

rikard-sics:
> Could you please have a look this?

Yes, I will have a look at this PR and the related issue. My initial impression is that the way to go is to stop the OSCORE layer from accepting requests from the client (with same token & MID) in the first place.

boaks:
Please see alternative PR #2296

boaks:
PR #2296 got merged.

PR #2296, opened 2024-10-18
Fix deduplication for OSCORE.

boaks:
Swap OSCORE and CongestionControl layers.
Add logback configuration for unittests.

Issue #2294


rikard-sics:
Thanks! I will have a look and test through some OSCORE-related things

rikard-sics:
Thanks for the PR! 

I have now done some testing and I have not found any issues introduced by this swap of layers. So I'd say we should go ahead and do it.

boaks:
Do you know, if the OSCORE RFC refers somehow to the CoAP MID? That may be need to consider.

boaks:
@rikard-sics 

Any news? I would like to go for 4.0.0-M1 next week.

PR #2320, opened 2025-01-09
Remove Actinium from web-site.

boaks:
Discontinued with Californium v4.0.


boaks:
This change is already ["online"](https://eclipse.dev/californium/).

PR #2392, opened 2026-02-23
OSCORE: Allow only select response codes for unprotected responses

rikard-sics:
When a response comes to an OSCORE-protected request, only specific CoAP response codes should be expected in that response. These are specified in the OSCORE RFC and extended with further codes in the case that a CoAP proxy was used for the associated request. If the response is not using one of these response codes it will be dropped.


erik-borgstrom:
I will try to find time tomorrow to take a look.

PR #2410, opened 2026-09-19
Fix LinkFormat token and quoted extension parsing

shrutikbalwan:
Preserve valid punctuation in unquoted values and split quoted lists only for list-valued attributes. Add regressions for token characters, following links, extension strings and preserved list/flag behavior.

Assisted-by: OpenAI Codex


boaks:
Thanks for your contribution.

For a Eclipse project it is required to sign a [ECA](https://github.com/eclipse-californium/californium/blob/main/CONTRIBUTING.md#eclipse-contributor-agreement).

Is that possible for you?

